PR #4, opened 2016-10-05
Use single argument completion handler

lightsprint09:
If a request finishes, it is either a success with a response or it is an error. But bot can be nil at the same time. Two completion handler for success and error solve this problem.

### DIFF
--- a/Apollo/ApolloClient.swift
+++ b/Apollo/ApolloClient.swift
@@ -29,7 +29,7 @@ public class ApolloClient {
     self.init(networkTransport: HTTPNetworkTransport(url: url))
   }
   
-  public func fetch<Query: GraphQLQuery>(query: Query, completionHandler: @escaping (_ result: GraphQLResult<Query.Data>?, _ error: Error?) -> Void) {
-    networkTransport.send(query: query, completionHandler: completionHandler)
+  public func fetch<Query: GraphQLQuery>(query: Query, onCompletion: @escaping (GraphQLResult<Query.Data>) -> Void, onError: @escaping ( _ error: Error) -> Void) {
+    networkTransport.send(query: query, onCompletion: onCompletion, onError: onError)
   }
 }
--- a/Apollo/NetworkTransport.swift
+++ b/Apollo/NetworkTransport.swift
@@ -19,7 +19,7 @@
 // SOFTWARE.
 
 public protocol NetworkTransport {
-  func send<Query: GraphQLQuery>(query: Query, completionHandler: @escaping (_ result: GraphQLResult<Query.Data>?, _ error: Error?) -> Void)
+    func send<Query: GraphQLQuery>(query: Query, onCompletion: @escaping (GraphQLResult<Query.Data>) -> Void, onError: @escaping ( _ error: Error) -> Void)
 }
 
 struct GraphQLResponseError: Error, LocalizedError {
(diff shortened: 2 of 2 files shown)


martijnwalraven:
I definitely see where you are coming from, and I'm not sure what the right decision is here, but having two separate callbacks feels less than ideal to me.

One of the reasons for sticking to a single completion handler would be that it fits better with the rest of the Cocoa API, and I think it is important that the API feels familiar to most iOS developers. Although to be fair, it makes sense to have Swift establish its own patterns.

Another benefit of having a single callback is that it allows you to use trailing closure syntax.

My person ideal would be to get a `Result` or `Promise` type in the language, with the ability to use standard `try` error handling with it (like `async`/`await` does in JavaScript). 

In the meantime, we could decide to use our own `Result` type, which [Alamofire](https://github.com/Alamofire/Alamofire) has also done for instance. But I'm not sure if that wouldn't overcomplicate matters.

ephemer:
Yes, to be honest I've wrapped our use of apollo iOS in a promise library,
so this is kind of a moot point for us.

In general I am all for the trailing closure syntax, so that's a good
point. I haven't used this API enough to worry about conciseness, but it
may be the most important argument to make about this change.

As a philosophical point, I'm not sure adhering to Cocoa is the way
forward. Since swift's inception, the battle has pretty much always been to
make cocoa more swifty, rather than the other way around. As a random side
note, I wouldn't be surprised if Apple was already well underway with a
Swift-only successor to Cocoa.

The main advantage I still see with two callbacks is reducing boilerplate
and separation of both the error and error-free codepaths (mostly optional
unwrapping). As is, it feels a bit like dealing with untyped JSON (but
thankfully not quite that extreme), which is the entire thing we're trying
to avoid
Martijn Walraven <notifications@github.com> schrieb am Do. 6. Okt. 2016 um
17:44:

ephemer:
But re: promises in the language. I'd be all for it. Would you consider
proposing it on swift-evolution?
Geordie Jay <geojay@gmail.com> schrieb am Fr. 7. Okt. 2016 um 20:30:

mgwidmann:
Any update on this? I'm noticing for a response like the following that I'm not receiving the error correctly in my completion handler:

```javascript
{
  "errors": [
    {
      "user_message": "No user found for given email.",
      "message": "In field \"signInUser\": Email not found error",
      "locations": [
        {
          "line": 2,
          "column": 0
        }
      ]
    }
  ],
  "data": {}
}
```

I'm seeing both `result` is `nil` and `error` is `Optional(Apollo.GraphQLResultError(path: ["signInUser"], underlying: Apollo.JSONDecodingError.missingValue))`

I cannot get at the `message` field within the error because of the missing data in the response. Would this PR fix this issue or should I open another issue?

martijnwalraven:
@mgwidmann: You should open a new issue for this, as this PR doesn't really address your problem.

What server-side technology are you using? Your response seems invalid to me, because `data` is an empty object rather than `null`, as it should be according to the spec.

justinmakaila:
I mean, any opposition to adding `antitypical/Result` as a dependency?

I'm fairly certain the Swift timeline doesn't include a `Result` type, and that's pretty much the standard when it comes to implementing one.

```
apollo.fetch(MyQuery()) { result in
  switch result {
    case let .success(value):
      successHandler(value)
    case let .failure(error):
      errorHandler(error)
  }
}
```

mgwidmann:
@justinmakaila What you described is better known as a `Maybe` monad. I would prefer it as its a much cleaner approach than the nodejs style 2 parameter result & error. Maybe Apollo should provide it to distinguish between logical errors and catastrophic failures if Swift doesn't plan to support it any time soon.

PR #5, opened 2016-10-13
Allow using booleans in the generated graphql schema 

smowden:
using booleans currently leads to compile time errors, because they cant be converted into swifts types


martijnwalraven:
Thanks for the pull request, merged!

PR #8, opened 2016-10-18
Added SPM support

dfirsht:


martijnwalraven:
This looks great! Sorry the Travis CI is a bit slow, there seems to be a queue for iOS builds. But the iOS 10 tests have already passed, so the iOS 9 ones should follow soon hopefully.

Have you already started using Apollo iOS with SPM?

dfirsht:
Thanks!

I am just starting to play around with it. I was able to get a fetch to the Frontpage-server using Apollo iOS to run successfully using the SPM.

martijnwalraven:
Good to hear! Let me know how things go, or if you have any questions or suggestions. Also curious to hear about your use cases. Are you planning on using Apollo iOS in non-Cocoa environments?

dfirsht:
Thanks for the quick merge! Right now I am just playing around with it to see what it can do. I was able to integrate Apollo iOS with the Kitura web framework so Kitura can make GraphQL requests on behalf of the client.

Btw SPM requires repos to be tagged with version numbers to be pulled so you'll need to release a new version. It also doesn't recognize the `v` in front so it will have to be something like `0.3.1`

martijnwalraven:
Ah, good to know! I changed the versioning script to remove the `v` from the tag and pushed a new version `0.3.1`. Let me know if that doesn't work.

PR #9, opened 2016-10-20
add cancelling to apollo requests

jamiepinkham:
In the current implementation, both the `fetch` and `perform` operations of `ApolloClient` return a `Void` type. However, it's important, especially in fetching data, to be able to cancel an outbound request/response if, for instance, the user leaves the screen, or any other number of reasons.

### DIFF
--- a/Sources/ApolloClient.swift
+++ b/Sources/ApolloClient.swift
@@ -1,5 +1,9 @@
 import Foundation
 
+public protocol Cancellable {
+    func cancel()
+}
+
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks for the pull request, that looks great! Nice work extracting the `Cancellable` protocol to abstract from `URLSessionTask`.

There seems to be a queue for Mac builds on Travis CI, but I'll merge it as soon as the tests pass.

PR #31, opened 2016-12-09
Make GraphQLError.message publicly accessible

timbotnik:
@martijnwalraven - I assume this would need to be public in order to access the message returned inside the GraphQLError?

### DIFF
--- a/Sources/GraphQL.swift
+++ b/Sources/GraphQL.swift
@@ -29,7 +29,7 @@ public struct GraphQLResult<Data> {
 }
 
 public struct GraphQLError: Error {
-  let message: String
+  public let message: String
 }
 
 extension GraphQLError: GraphQLMappable {


martijnwalraven:
@timbotnik: Good catch!

benrom:
This is an important fix, without that it's not possible to access error messages...

Any idea when a release containing this fix will be available?

martijnwalraven:
@benrom: I was waiting for normalized caching support to stabilize before doing a new release, but I just made a patch 0.4.3 release with this commit in it. Hopefully that solves your issue for now. Thanks for bringing this to my attention!

(The new release will also have some changes to `GraphQLError` to allow access to more fields, see https://github.com/apollostack/apollo-ios/issues/39.)

benrom:
Great, I will test it now. Thanks!!

PR #32, opened 2016-12-09
Make GraphQLResponse initialiser public

JanGorman:


martijnwalraven:
Out of curiosity, what is your use case for constructing your own `GraphQLResponse`? Are you creating a custom `NetworkTransport`? Or implementing some kind of caching layer?

JanGorman:
@martijnwalraven Yes, it's to do with a custom `NetworkTransport`. Need to pass the `operation.variables` in a different format

martijnwalraven:
@JanGorman: Great, let me know how things go and if you think the API could be improved in any other way. It is pretty minimalistic right now, and we still have to figure out how people will be using it.

aschober:
I too was looking to make a custom NetworkTransport to handle including an authorization header with tokens that get refreshed ~30mins. My custom NetworkTransport will get the current token from auth logic on every request.

Glad to see this already merged in.

aschober:
Another thought is to make `GraphQLResponseError` public. On the other hand, maybe a custom NetworkTransport should also have its own custom error.

PR #46, opened 2017-02-01
Make parseResult public again

fwal:
We currently use Apollo as de-/serialiser for our graphql requests within our pre-existing network stack and it works really well. 

We used this method for achieving this but it was recently made internal.

### DIFF
--- a/Sources/GraphQLResponse.swift
+++ b/Sources/GraphQLResponse.swift
@@ -7,7 +7,7 @@ public final class GraphQLResponse<Operation: GraphQLOperation> {
     self.body = body
   }
 
-  func parseResult(cacheKeyForObject: CacheKeyForObject? = nil) throws -> (GraphQLResult<Operation.Data>, RecordSet?)  {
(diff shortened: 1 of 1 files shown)


martijnwalraven:
I'd like to better understand your use case. Which part of your app calls `parseResult`? To integrate with a pre-existing network stack, wouldn't it be enough to implement your own `NetworkTransport`?

Maybe you could also chime in on https://github.com/apollographql/apollo-ios/pull/6?

martijnwalraven:
I'm a little hesitant to make `parseResult` public because its current implementation includes result normalization, and if you only want to deserialize a result, that is wasted effort that could be avoided by implementing result parsing yourself (which should only take a few lines of code).

What do you think?

martijnwalraven:
@MrAlek: Let's make it public for now then, with the understanding that the API may change again in the future. I also still think it's valuable to continue the conversation in https://github.com/apollographql/apollo-ios/issues/6 to see if there isn't a better way to do this.

MrAlek:
@martijnwalraven Thanks! And yes, of course, I'm working on a second integration from our end currently (using ApolloClient) and will get back to you with some thoughts based on that.

martijnwalraven:
Great! I'm also in the process of writing some API documentation, so that may also help.

PR #68, opened 2017-03-08
Fixed issue where a field that was removed from a Record would still exist after a merge.

mckaskle:
This could happen when a value changed to `null`. This fixes a bug in the following scenario:

1. Make a query where field `x` has value `y`. The result is cached.
2. Make the same query but bypass the cache. This time the field `x` has value `null`.
3. Make the same query again but allow fetching from the cache.

*Expected Result*: field `x` has value `null`.
*Actual Result*: field `x` still has value `y`.

The fix was to ensure that keys/fields that do not exist in a new record are removed from the old record during a merge.

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


mckaskle:
Closing. Not only did this not pass tests but it looks like the issue I was trying to address was fixed w/ #63.

martijnwalraven:
@mckaskle: Good to hear the issue was fixed!

PR #79, opened 2017-04-11
SQLite persistent normalized cache

kompfner:
This PR does the following:
* Introduces `SqliteNormalizedCache`, which persists records in a SQLite database.
* Introduces the `SqliteNormalizedCache` subspec, which is necessary for two reasons:
    * It allows consumers of the Apollo pod to decide whether or not they need SQLite persistence for their application - if they don’t need it, they don’t need to take a dependency on `SQLite.swift`.
    * It disables `SqliteNormalizedCache` usage on tvOS projects, which don’t have access to the file system.
* Shares existing unit tests to test `InMemoryNormalizedCache` and `SqliteNormalizedCache`.
* Adds a unit test verifying that cached data is persisted in `SqliteNormalizedCache`.

Known remaining work before this PR can be merged:

- [x] Properly handle versioning both Apollo.xcodeproj and ApolloSQLite.xcodeproj
- [x] Naming: "Sql" --> "SQL"
- [x] In `loadRecords(forKeys:)`, implement correct expected behavior of returning `nil` for keys that don't have corresponding records (instead of the misunderstood expected behavior described in the comment)

Next steps:

- [ ] Clear cached records when Apollo version changes.


justinmakaila:
Hey @kompfner, I'm not certain that Apollo should include a specific SQLite cache. I think the goal is to provide protocols for end-users to plug their own dependencies or solutions in to. A lot of the work here looks like it would be relevant to a project I just (released)[https://github.com/justinmakaila/Lunar]. It provides a cache that uses CoreData as a persistence layer, and it specifies Apollo as a dependency rather than adding it into Apollo.

martijnwalraven:
The idea is definitely that normalized caches are pluggable, so anyone can write their own. But I would also like to include sensible defaults, and SQLite seems like the best option for a lightweight persistent cache. So I think it makes sense to include it in this repo, but in a separate framework, so people can import it if they want but don't need to take the dependency if they don't need it.

One thing I've been talking about with @kompfner is that I would prefer if we could keep unit tests for caches separate. Ideally, we'd have a set of unit tests that represent the contract any `NormalizedCache` should follow, and we can run these for any cache. So Lunar should also be able to run these tests to verify that the cache implementation works as expected.

justinmakaila:
@martijnwalraven I'm not necessarily convinced. I think that persistence should be on behalf of the end user, and dependencies plugged in as they need them. 

In order for Carthage to build such a thing, it would need to check out SQLite, regardless of whether you use it or not, and build all of those symbols into the Apollo framework. Then you'd be forced to include it in your project (please correct me if I'm wrong). It sounds wonderful for CocoaPods becuase of subspecs, but it's plain nasty with Carthage. Currently on Moya, we have 3 flavors Moya, RxMoya, and ReactiveMoya. Carthage users need to download and build ALL of the dependencies to build all 3 flavors and then selectively import the ones they want. It's a huge pain when managing dependencies (you better not try to use a different version of `ReactiveSwift` in your project, otherwise `ReactiveMoya` won't build).

I don't think it's an un-sensible default to only provide an in-memory cache. The tools are there for the community or the developer to build the cache that best suits their needs. I think by including one as a "sensible default" will lead to others building cases on why including their specific caching use-case is sensible, given their specific architecture.

justinmakaila:
I don't like bringing in a dependency unless it's absolutely critical to the function of the library. In this case, the SQLite cache can exist in it's on repo that depends on Apollo, and we can recommend to just use that dependency if you want persistence out of the box in the README:
```
To use Apollo with Carthage, you can use
github "apollographql/apollo-ios"

Here's a list of community supported caches to suit your needs:
Lunar (CoreData cache for Apollo):
github "justinmakaila/lunar"

SQLite Apollo:
github "kompfner/sqlite-apollo-cace"
```

justinmakaila:
Huge fan of the unit tests idea, mostly because I had to write my own and it wasn't awesome.

martijnwalraven:
I was hoping we could keep the SQLite cache in a separate Xcode project, with a separate framework target. I definitely don't want Apollo framework to have a dependency on SQLite. It sounds like you have a lot more experience attempting the same thing with Moya, so if we can't selectively build frameworks with Carthage a separate repository might be a better solution. @kompfner: What do you think?

justinmakaila:
@martijnwalraven Yeah so in the case of a github dependency, (`github "apollographql/apollo-ios" "master"`) it will clone the repo from that github URL, point it to whatever version or branch is specified (in this case, master) and then find the `xcodeproj`s or `xcworkspace` in the directory and build all of the shared schemes as specified by the target configuration.

I think if we were already headed down the path of "separate project" we might as well finish it off with the "separate repo" logic. I mean, it may lead to more maintenance down the line, but it allows for cache maintainers to point to a specific version of Apollo, and it also helps solidify what Apollo is: A normalized interface for interacting with your remote. Apollo doesn't, and should not care about what you use as a `NetworkTransport` or a `NormalizedCache`, but rather that one is provided and implemented.

kompfner:
@justinmakaila, @martijnwalraven, you bring up some really good points here. I agree that it’s generally desirable to have projects with a single narrow focus that do an amazing job at one thing and one thing only. The `NormalizedCache` protocol is a great example of using a contract to enable enhancements to a project while keeping them separate from the core. We’re 100% in agreement on our principles, I think.

What I disagree with is the following:

> I don't think it's an un-sensible default to only provide an in-memory cache.

I would argue that a persistent cache should *not* be considered an enhancement - it’s essential for a native Apollo client. While an in-memory cache has been sufficient for web pages so far (though there is talk of introducing a persistent cache for `apollo-client`), there are very different expectations about how a native mobile client should work. If your app can be killed in the background at any time, and you’re much more likely to open it while offline or in poor network conditions, cache persistence is critical. I might go so far as to claim that a persistent cache is a more sensible default than an in-memory cache - it’s at least definitely a need we have at my company. And if we’re going to provide a default persistent cache, we should go with the most ubiquitous, lowest-common denominator form of storage - SQLite. `apollo-android` reached the same conclusion [last month](https://github.com/apollographql/apollo-android/issues/286).

With that in mind, then, I think it’s important that we optimize for the persistent cache case, in terms of developer ergonomics for both consumers of `apollo-ios` as well as its maintainers. From the consumer side, it allows us to provide a better experience for our Cocoapods users (who I think there will be many of), letting them “opt out” by specifying a subspec. From the maintainer side, it allows us to more easily work on all the cache implementations and the rest of the `apollo-ios` code. This is especially important now, when `apollo-ios` is still very much undergoing active development, and when it may very well have to evolve and change based on what we learn from introducing cache persistence. Having to worry about different repos at this point would likely slow us down.

I do admit that there is a notable trade-off: this presents some challenges for those using Carthage. It’s still possible to avoid the SQLite dependency, but it could involve more headache for the consumer.

I’m not advocating that we close the door indefinitely on separating the SQLite cache from the `apollo-ios` repo indefinitely. Eventually we may learn some things: that our consumers are overwhelmingly preferring the in-memory cache, or are far and away Carthage users. The project may stabilize, such that occasionally working across repos wouldn’t present much of a headache. When these things happen, I’d be totally willing to rethink this, but right now it seems clear to me that the pros of bundling the SQLite cache like this outweigh the cons.

justinmakaila:
Ok so my biggest gripes: you're introducing a new dependency for opt-in features, and treating Carthage users as second class citizens in favor of a featureset only provided by CocoaPods.

I'm 50/50 on it, given the explanation. I don't think it's unreasonable to say "Apollo supports in-memory caches out of the box, and custom caching strategies are provided by the community, here's a list: ..." and I'm also a big fan of opt-in.

As a Carthage user, you've essentially said that you're fine with forcing me to download and build SQLite. We're not talking about a system dependency on SQLite, we're talking 3rd party, and I have almost the same grounds to say that we should support CoreData out of the box strictly because it's not dependent on a third party library.

In my experience, handling things this way, like in the case of Moya, results in more PRs for people who design for their specific use case and say "well I really need this, so here's a PR for it" when they could have easily handled that specific use case by creating their own dependency ON Moya rather than building the dependency INTO Moya.

justinmakaila:
I'm not saying the SQLite cache is wrong, or shouldn't exist or anything. I'm saying it should exist as a third-party dependency that users can literally opt-in to by adding it to their `Podfile` or `Cartfile` rather than building in specific subspec support that favors `CocoaPods` users over `Carthage` users.

martijnwalraven:
I'm a bit torn here. I agree having a default persistent cache makes sense, and we want to make using it as easy as possible. But I also feel the current setup complicates the project structure, and keeping SQLite support in a separate repo may set a better example for future extensions.

@kompfner: Couldn't we have the subspec depend on a separate `ApolloSQLite` pod? That way, CocoaPods users would get the same default behavior, but we could keep `ApolloSQLite` in a separate repo.

kompfner:
Many good comments! Here are a few I'd like to address:

> Couldn't we have the subspec depend on a separate `ApolloSQLite` pod?

The subspec convenience is only part of the argument for including the SQLite cache in this repo (and not even the biggest part). Since we should be thinking of persistence/offline as a first-class concern, it means we'll be actively developing the SQLite cache in tandem with everything else in the library. *`apollo-ios` is still a work in progress. Let's not make it more difficult for ourselves by fragmenting across repos.*

> In my experience, handling things this way, like in the case of Moya, results in more PRs for people who design for their specific use case and say "well I really need this, so here's a PR for it"

I don't think we need to invoke a slippery-slope argument in this case. If we agree that it makes sense to provide exactly one default out-of-the-box persistent cache, then we can clearly articulate/argue for this position. In fact, I'll go ahead and update the `podspec` with a commented explanation for the existence of this subspec.

> As a Carthage user, you've essentially said that you're fine with forcing me to download and build SQLite.

I agree that this is less than ideal. Fortunately, though, Carthage lets you opt out of actually including ApolloSQLite/SQLite.swift in your app if you don't want it. Also fortunately, SQLite.swift is relatively thin. It would've been great to link to the system sqlite C library directly, but it didn't seem worth the tradeoff of re-inventing the complexity of making it ready for Swift consumption (it's not a framework).

One more point I wanted to emphasize again: this is the route `apollo-android` has taken. I think there's something to be said about aligning approaches across platforms.

justinmakaila:
I think the slipper-slope argument is very relevant to this conversation. You're saying that we should add a new dependency because of a specific use-case where you need persistence between sessions. 

I can, in the same light, make the case that Lunar, the CoreData cache I'm currently working on (heavily borrowing from your `SQLiteNormalizedCache`, thanks!) is a better fit for the persistence layer in `Apollo`. It doesn't require `Carthage` users to download an extra dependency, and it wouldn't even require a subspec (I think? Correct me if I'm wrong, I haven't used CocoaPods in years). I'm not entirely sure of the specifics around this, but I'm sure we can optionally link the `CoreData` framework so that people who don't use it don't even have to know that those classes exist, and can go on with their lives and their custom caching strategies or having the app state live in ephemeral memory.

In a perfect world, I don't think `Apollo` would provide an implementation of `NormalizedCache`. The `ApolloClient` would require initialization with a `NetworkTransport` and an `ApolloStore`, requiring a custom implementation of `NormalizedCache`.

However, if we were to not consider the pain points of our end-users, we would never get any traction. 

Providing an in-memory store allows users to get up and running without us making too many assumptions, and it doesn't require any third-party dependencies. It's dependency manager agnostic (no special privileges are provided to certain consumers).

> Fortunately, though, Carthage lets you opt out of actually including ApolloSQLite/SQLite.swift in your app if you don't want it.

AFAIK, Carthage clones your repo, and finds ALL shared schemes and builds them. If this is the case, we would still need to manage independent Xcode projects based on your flavor of Apollo (apollo-ios, apollo-ios-sql, etc).

We're already expanding Apollo's responsibilities by including these caches. I don't feel comfortable adding in a cache for persistence that literally forces a dependency on some consumers of the library. It might not get used in the project, but every time they go to update their dependencies, time will be spent building something which isn't used.

I'd almost be more comfortable for building a `NormalizedCache` implementation that keeps everything in memory and then writes the store to a `.txt` file or something for the sake of the simplest common denominator for storage options.

Let me also mention, in reference to `apollo-android`, that we use `Realm` for our android app, and for our implementation, are already planning on building out a `RealmNormalizedCache` for both iOS and Android.

I don't mean to be a brat, but I'm going to invoke the slippery-slope argument again: `Realm` is the least-common denominator between iOS and Android for persistence in terms of using the same platform. Why don't we include `RealmNormalizedCache` in this repository?

martijnwalraven:
> In a perfect world, I don't think Apollo would provide an implementation of `NormalizedCache`. The `ApolloClient` would require initialization with a `NetworkTransport` and an `ApolloStore`, requiring a custom implementation of `NormalizedCache`.

I don't think I agree with this. We definitely want `NormalizedCache` to be an extension point people can use to plug in alternative cache implementations, but I also believe we should have sensible defaults built-in. I'm a firm believer in convention over configuration.

The question then becomes what persistent cache implementation we want to include by default. Although I can see why people want to integrate with Core Data or Realm if they have an existing model layer, and I'm excited to see Lunar, SQLite seems like a more lightweight option that can stand on its own. I think it will be easier to experiment with some of the caching semantics we want to implement when we have full control over the schema and data manipulation. That experimentation is also the reason I'd like to keep things in one repository for now.

Having said that, I'm not happy with Carthage users having to download and potentially build a dependency they don't need. They won't have to include the `ApolloSQLite` and `SQLite` frameworks in their projects, but it still feels like unnecessary overhead. Unfortunately, Carthage isn't very flexible in this regard, and it doesn't allow you to depend on an individual project or target.

I have high hopes for Swift Package Manager, so we may be able to implement a better solution when a new version of Xcode with built-in support for it is released. These decisions would be a lot easier with more fine grained modularity and sane dependency management, but right now I think we should focus on getting things off the ground.

justinmakaila:
But SQLite isn't lightweight. It requires a new dependency, new xcodeproj, and adds more overhead if you're not a CocoaPods user. Lunar doesn't. It literally just links CoreData. Not to mention it pretty much is this SQLiteNormalizedCache, just with DB transactions expressed in CoreData logic, not SQLite.

If you want to use Apollo and Lunar, you add this to your `Cartfile`:
```
github "justinmakaila/Lunar"
```
It will clone Lunar, clone Apollo, and compile both to `.framework`s for you. You then can link both of them as necessary. Sure, when you initialize your `ApolloStore`, you have to pass it a `LunarCache`, but that's not even painful. It literally removes the black magic of Apollo just using some "Magic Cache" that you don't have access to.

(Side note: I initially brought this up and started development on Lunar because I wanted to use our already existing model framework. After exploring the code shared here, I realized that I don't need that dependency anymore, because Lunar now has a single entity, LunarRecord, which provides two properties: "key" and "record".)

I'm not happy, as a contributor, a maintainer of my own cache (that literally does the same thing without the dependency), and more specifically, as a Carthage user, that we're trending towards including something because it was PR'd first and is easy to maintain if you use CocoaPods.

My dependency build time is going to go up because I now have to build Apollo, Lunar, and SQLite (which I don't use), even though I cut my local dependencies out (no more model or sync frameworks).

I share the general sentiment here that we should be able to get people up and running with persistence quickly, and that's why I'm still proposing that we just provide install instructions with how to include whatever specific community-maintained cache. It's really not beyond our audience to have to make some choices on their own backend and maybe, if there's no community support, help support the community and write some code.

justinmakaila:
Apollo's beauty isn't even in it's caching, nor should it concern itself with persistence. Apollo's beauty is that it was a lightweight interface that could live on top of your network and persistence layer without being opinionated about either.

This takes a step away from that, and starts to add weight and overhead, absolutely.

justinmakaila:
At this point, I've said literally all there is to say. If you guys are set on merging this in and forcing SQLite on Carthage users, than there's nothing I can say. I'm just trying to share what I've learned in working with Moya and what I'm trying to use Apollo for. SQLite has nothing to do with Apollo, IMO, and should not be included in the repository.

martijnwalraven:
@kompfner: I cleaned up some code related to promises and the fields serialization. We no longer need to special case `NS` types now. I tried to clarify typing by adding some typealiases to `Record`, but I’m not sure if this actually makes things more or less readable. Definitely something we should also document. Anyway, the rest of the code looks pretty simple, and all tests seem to pass, so I think we’re ready to merge (squash)!

ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -15,31 +15,19 @@ public final class SQLiteNormalizedCache: NormalizedCache {
...
   public func merge(records: RecordSet) -> Promise<Set<CacheKey>> {
-    return Promise<Set<CacheKey>> { fulfill, reject in
-      do {
-        fulfill(try mergeRecords(records: records))
-      }
-      catch {
-        reject(error)
-      }
-    }
+    return Promise { try mergeRecords(records: records) }
```
kompfner: Oh, sweet, your `Promise` implementation automatically turns thrown errors into rejections. 👍

Apollo/JSONStandardTypeConversions.swift
```diff
@@ -117,14 +117,13 @@ extension Dictionary: JSONEncodable {
   public var jsonObject: JSONObject {
     var jsonObject = JSONObject(minimumCapacity: count)
     for (key, value) in self {
+      print("key: \(key), \(type(of: key))")
```
kompfner: Should these be removed?

ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -83,60 +83,56 @@ public final class SQLiteNormalizedCache: NormalizedCache {
...
-    let recordJSON = try SQLiteSerialization.deserialize(data: recordData)
-    return Record(key: row[key], recordJSON)
+    let fields = try SQLiteSerialization.deserialize(data: recordData)
+    return Record(key: row[key], fields)
   }
 }
 
 private let serializedReferenceKey = "reference"
 
 final class SQLiteSerialization {
```
kompfner: Hm, I think this should probably be private.

ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -83,60 +83,56 @@ public final class SQLiteNormalizedCache: NormalizedCache {
...
-      deserializedObject[key] = try deserialize(valueJSON: value)
+      fields[key] = try deserialize(jsonValue: value)
     }
-    return deserializedObject
+    return fields
   }
 
-  private static func deserialize(valueJSON: Any) throws -> Any {
-    switch valueJSON {
+  private static func deserialize(jsonValue: JSONValue) throws -> Record.Value {
```
kompfner: How about this for the naming of the `serialize` and `deserialize` method parameters, for even greater clarity?
- `serialize(fieldValue:)`
- `serialize(fieldJSON:)`

ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -83,60 +83,56 @@ public final class SQLiteNormalizedCache: NormalizedCache {
...
 private let serializedReferenceKey = "reference"
 
 final class SQLiteSerialization {
-  static func serialize(fields: JSONObject) throws -> Data {
+  static func serialize(fields: Record.Fields) throws -> Data {
     var objectToSerialize = JSONObject()
     for (key, value) in fields {
       objectToSerialize[key] = try serialize(value: value)
     }
-    return try JSONSerializationFormat.serialize(value: objectToSerialize)
```
kompfner: The cleanup in this commit is great! Perhaps `JSONSerializationFormat` should have a more descriptive name, to make its purpose more distinct than the system-provided `JSONSerialization`?

ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -88,7 +88,7 @@ public final class SQLiteNormalizedCache: NormalizedCache {
   }
 }
 
-private let serializedReferenceKey = "reference"
+private let serializedReferenceKey = "$reference"
```
kompfner: Is this really necessary? What distinguishes a reference field from any other field is that it's a dictionary (of the form `[reference: <key>]`), whereas other fields are primitive types.

kompfner:
@martijnwalraven I'm going to do a tiny bit more cleanup based on the comments I left on your last few commits. Please let me know if you agree with those. Those commits were really nice improvements, btw!

ApolloSQLite/SQLiteNormalizedCache.swift
martijnwalraven: Yeah, you're right, I wasn't thinking clearly, there are no nested objects in records of course. Although we may want to leave open the possibility of storing JSON in a field if something like that becomes part of the GraphQL spec? May be too forward looking.

kompfner:
I think the reason the unit tests were failing was due to race conditions around multiple tests attempting to open the same file on disk. My fix was to have each test open a SQLite file at a different file URL.

kompfner:
Disable iOS 9.3 unit tests, since that simulator doesn't seem to be an available option on Travis right now.

PR #82, opened 2017-04-13
Feature/public interface updates

justinmakaila:
Renamed and exposed the `InMemoryCache`. It's much easier to set up mock `ApolloClient` instances for testing without duplicating, and also could be useful for feeding in data.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,23 @@
 language: objective-c
 
-osx_image: xcode8.2
+osx_image: xcode8.3
 
 xcode_project: Apollo.xcodeproj
 
 matrix:
   include:
   - xcode_scheme: Apollo
-    xcode_sdk: iphonesimulator10.2
-    env: DESTINATION="platform=iOS Simulator,OS=10.1,name=iPhone 7"
+    xcode_sdk: iphonesimulator10.3
+    env: DESTINATION="platform=iOS Simulator,OS=10.3,name=iPhone 7"
   - xcode_scheme: Apollo
-    xcode_sdk: iphonesimulator10.2
+    xcode_sdk: iphonesimulator10.3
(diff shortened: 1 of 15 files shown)


martijnwalraven:
I realize Objective-C has predisposed me to long descriptive names, so `InMemoryNormalizedCache` sounds perfectly fine to me :) @kompfner also mentioned he thought the name could be shortened, so maybe I'm the minority here.

My concern is that we may want to add other types of caches in the future. So Apollo Android has a response cache in addition to a normalized cache for example. So they do use the `NormalizedCache` suffix.

justinmakaila:
@martijnwalraven I'm into descriptive names as well, I just figured with the existence of the `NormalizedCache` protocol, and the conformance of `InMemory[Normalized]Cache` to the protocol, that dropping the `Normalized` part would be cool.

Could you explain what the response cache is used for? I was under the impression that a `NormalizedCache` was just a protocol to be implemented, and that any "new" caches would be implemented as dependencies (i.e. how I built Lunar to cache records in CoreData).

martijnwalraven:
@justinmakaila: The idea of a response cache is to cache individual responses as opposed to normalized records. So you don't get any of the data sharing and consistency benefits, but it might be more efficient to cache complete responses without normalizing them. On Android, a use case that came up is downloading large blobs of data in the background for example.

I think the `NormalizedCache` should definitely be the focal point of our efforts, and that is where the interesting cache implementations like SQLite and Core Data come in. A response cache is more likely to just write to disk.

designatednerd:
I'm going to close this based on a combination of @martijnwalraven's response and the lack of activity here for quite a while. We've definitely got some thinking to do about how we're handling caching, I've got a few ideas but nothing I want to put out quite yet.

PR #83, opened 2017-04-25
Update `check-and-run-apollo-codegen.sh` to install `apollo-codegen`

kompfner:
With a team the size of mine, I wouldn't be comfortable adding an Xcode Build Phase that requires everyone to manually install the appropriate version of `apollo-codegen`. This PR addresses that burden.

Additionally, this would allow Apollo-dependent Xcode projects to build in CI environments (provided they have `npm` installed).

An even better solution would be to somehow distribute `apollo-codegen` with `apollo-ios`, so that the appropriate `apollo-codegen` package would never have to be downloaded at build time. This would avoid build failures on CI machines when they're having network issues. It would also avoid the occasional super-frustrating failure when a developer attempts to build for the first time after a `git pull` in a place with no WiFi.

This seemed like an easy first step in the right direction.

### DIFF
--- a/scripts/check-and-run-apollo-codegen.sh
+++ b/scripts/check-and-run-apollo-codegen.sh
@@ -1,9 +1,30 @@
+# Only major and minor version should be specified here
 REQUIRED_APOLLO_CODEGEN_VERSION=0.11
+
+REQUIRED_APOLLO_CODEGEN_SEMVER="$REQUIRED_APOLLO_CODEGEN_VERSION.x"
 
 # Part of this code has been adapted from https://github.com/facebook/react-native/blob/master/packager/react-native-xcode.sh
 
 # This script is supposed to be invoked as part of the Xcode build process
 # and relies on environment variables set by Xcode
+
+install_apollo_codegen() {
+  npm install -g apollo-codegen@"$REQUIRED_APOLLO_CODEGEN_SEMVER"
+}
+
+# We consider versions to be compatible if the major and minor versions match
+are_versions_compatible() {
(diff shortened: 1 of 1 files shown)


scripts/check-and-run-apollo-codegen.sh
```diff
@@ -1,10 +1,31 @@
+# Only major and minor version should be specified here
 REQUIRED_APOLLO_CODEGEN_VERSION=0.11
 
+REQUIRED_APOLLO_CODEGEN_SEMVER="$REQUIRED_APOLLO_CODEGEN_VERSION.x"
```
martijnwalraven: I don't think this is actually needed. You can run `npm install` with just the major and minor versions, and it will automatically install the latest patch version.

martijnwalraven:
This definitely seems useful. Agreed that somehow avoiding a separate install would be ideal, but that would complicate our build setup even further so we may want to wait until things settle down a bit before we tackle that.

kompfner:
Good call on the semantic version specification not requiring ".x".

As for the setup complication, is the concern the introduction into `apollo-ios` of a dependency on `npm`? It seems to me like we already have that dependency implicitly (or explicitly if you count the user-directed message that is produced when the appropriate version of `apollo-codegen` is not installed). The way I've been thinking about it, this PR simply takes the small step of making automatic what the user would otherwise have to do manually.

kompfner:
Oh maybe I misread your comment. Correct me if I'm wrong here: you're in favor of the step taken in this PR, but are proposing that we hold off on the possible next step of bundling `apollo-codegen` with `apollo-ios`. If that's the case, then we're in agreement.

martijnwalraven:
Yep, that's what I meant!

PR #89, opened 2017-05-09
Imperative store API (WIP)

martijnwalraven:
This is work in progress to add read/write methods to `ApolloStore`, inspired by [Apollo JavaScript's new imperative store API](https://dev-blog.apollodata.com/apollo-clients-new-imperative-store-api-6cb69318a1e3).

`ApolloStore` exposes these read/write methods on transaction objects accessed through`withinReadTransaction` and `withinReadWriteTransaction`. 

```swift
store.withinReadTransaction { transaction in
  // We can load queries
  let data = try transaction.read(query: HeroAndFriendsNamesQuery())
  data.hero?.name // "R2-D2"
  data.hero?.friends?.flatMap { $0?.name } // ["Luke Skywalker", "Han Solo", "Leia Organa"]

  // But also individual fragments
  let r2d2 = try transaction.readFragment(ofType: HeroDetails.self, withKey: "2001")
  r2d2.name // "R2-D2"
  r2d2.asDroid?.primaryFunction // "Protocol"
}
```

In contrast to the JavaScript client, Apollo iOS exposes typed read/write methods. To accommodate that, the properties of the generated model structs are now mutable (they still obey value semantics of course).

```swift
store.withinReadWriteTransaction { transaction in
  try transaction.update(query: HeroAndFriendsNamesQuery()) { data in
    // We can construct models of the right type (using the shorthand `.init` here)
    data.hero?.friends?.append(.init(__typename: "Droid", name: "C-3PO"))
    // Or convert other models to the right type
    let c3po = try transaction.readFragment(ofType: HeroDetails.self, withKey: "2000")
    data.hero?.friends?.append(try .init(c3po))
  }
}
```

This PR relies on `apollo-codegen` 0.12.x, which can be installed through `npm install -g apollo-codegen@next`.


stubailo:
Awesome!

Sources/Apollo/ApolloStore.swift
```diff
@@ -141,17 +141,17 @@ public final class ApolloStore {
       self.cacheKeyForObject = cacheKeyForObject
     }
     
-    public func readObject<Query: GraphQLQuery>(forQuery query: Query) throws -> JSONObject {
-      return try readObject(forSelectionSet: Query.selectionSet, withKey: rootKey(forOperation: query), variables: query.variables)
+    public func read<Query: GraphQLQuery>(query: Query) throws -> Query.Data {
```
martijnwalraven: In Swift, argument labels are part of the method signature, so `read(query:)` is more conventional. It is in line with the standard library, and Swift actually has some pretty good [API design guidelines](https://swift.org/documentation/api-design-guidelines/).

stubailo: Sure, but there's also `readFragment`, right? Shouldn't that be `read(fragment: )` then?

martijnwalraven: Oh, sorry, totally missed your reponse. But the answer is that `readFragment` doesn't take a fragments as an argument, but a fragment type. So the signature is `readFragment(ofType:)`.

Sources/Apollo/ApolloStore.swift
```diff
@@ -188,17 +188,39 @@ public final class ApolloStore {
...
-    public func write<Fragment: GraphQLFragment>(object: JSONObject, forFragment fragment: Fragment.Type, withKey key: CacheKey, variables: GraphQLMap? = nil) throws {
-      try write(object: object, forSelectionSet: fragment.selectionSet, withKey: key, variables: variables)
+    public func updateFragment<Fragment: GraphQLFragment>(ofType fragmentType: Fragment.Type, withKey key: CacheKey, variables: GraphQLMap? = nil, _ body: (inout Fragment) throws -> Void) throws {
+      var fragment = try readFragment(ofType: fragmentType, withKey: key, variables: variables)
+      try body(&fragment)
+      try write(fragment: fragment, withKey: key, variables: variables)
     }
     
-    func write(object: JSONObject, forSelectionSet selectionSet: [Selection], withKey key: CacheKey, variables: GraphQLMap?) throws {
+    public func updateSelectionSet<SelectionSet: GraphQLSelectionSet>(ofType selectionSetType: SelectionSet.Type, withKey key: CacheKey, variables: GraphQLMap? = nil, _ body: (inout SelectionSet) throws -> Void) throws {
```
kompfner: When would a consumer update the store with a selection set that is neither a query or a fragment?

Sources/Apollo/Utilities.swift
```diff
@@ -21,7 +21,7 @@ public protocol Matchable {
   static func ~=(pattern: Self, value: Base) -> Bool
 }
 
-func isNil(_ value: Any) -> Bool {
+func isNil(_ value: Any, nested: Bool = false) -> Bool {
```
kompfner: is this function used anywhere anymore?

Tests/ApolloCacheDependentTests/ReadWriteFromStoreTests.swift
```diff
@@ -0,0 +1,265 @@
...
+
+      let result = try await(store.load(query: HeroAndFriendsNamesQuery()))
+      guard let data = result.data else { XCTFail(); return }
+
+      XCTAssertEqual(data.hero?.name, "R2-D2")
+      let friendsNames = data.hero?.friends?.flatMap { $0?.name }
+      XCTAssertEqual(friendsNames, ["Luke Skywalker", "Han Solo", "Leia Organa", "C-3PO"])
+    }
+  }
+}
```
kompfner: Nice tests! Some other scenarios that could be useful to exercise here are:
* updating a query with variables
* reading/updating a fragment with variables
* reading/updating a selection set (unless it's not in fact necessary to expose that functionality publicly, a question my other comment poses)

kompfner:
Good stuff! This yielded a really nice API.

Sources/Apollo/ApolloStore.swift
martijnwalraven: Yeah, I'm not a 100% sure if this is needed. But the idea is that you don't always want to define a fragment just to read or update data. So you can update a particular `AllPostsQuery.Data.Post` directly for instance.

Sources/Apollo/Utilities.swift
martijnwalraven: Good catch! Not right now I think, so we could probably delete it.

Tests/ApolloCacheDependentTests/ReadWriteFromStoreTests.swift
martijnwalraven: Yeah, I definitely need to add tests with variables, and make them pass :) Because there are some remaining issues with that.

martijnwalraven:
@kompfner: Thanks for taking a look at this!

PR #105, opened 2017-07-05
Option to use operation ids rather than operation text

kompfner:
This PR introduces to `HTTPNetworkTransport` the option to use `operationIdentifier`s rather than `requestString`s when making requests.

The operation identifier is sent as the "id" field of the request's JSON payload, just like it is in the [network layer](https://github.com/apollographql/persistgraphql/blob/master/src/network_interface/ApolloNetworkInterface.ts) provided by `persistgraphql`. An important deviation from `persistgraphql`'s approach is that here we are not sending an "operationName" field. While the inclusion of this field may help debugging in some cases, I suspect most development will happen with `sendOperationIdentifiers = false`, and that most developers will have access to a mapping between operation identifiers and their corresponding text, allowing for debugging anyway. Furthermore, inclusion of "operationName" makes requests a bit bigger and a bit easier to understand if intercepted.

Note that I'm using the argument name `sendOperationIdentifiers` rather than `persistgraphql`'s `enablePersistedQueries`. The Apollo client is concerned only with client behavior, and `sendOperationIdentifiers` seems to me to be a clearer description of how the flag alters client behavior than `enablePersistedQueries`.

### DIFF
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -1,6 +1,7 @@
 public protocol GraphQLOperation: class {
   static var operationString: String { get }
   static var requestString: String { get }
+  static var operationIdentifier: String { get }
   
   var variables: GraphQLMap? { get }
   
@@ -11,7 +12,11 @@ public extension GraphQLOperation {
   static var requestString: String {
     return operationString
   }
-  
+
+  static var operationIdentifier: String {
+    return ""
+  }
+
   var variables: GraphQLMap? {
     return nil
   }
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -1,7 +1,7 @@
 public protocol GraphQLOperation: class {
   static var operationString: String { get }
   static var requestString: String { get }
-  static var operationIdentifier: String { get }
+  static var operationIdentifier: String? { get }
   
   var variables: GraphQLMap? { get }
   
@@ -13,8 +13,8 @@ public extension GraphQLOperation {
     return operationString
   }
 
-  static var operationIdentifier: String {
-    return ""
+  static var operationIdentifier: String? {
+    return nil
   }
 
   var variables: GraphQLMap? {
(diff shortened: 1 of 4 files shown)


Sources/Apollo/GraphQLOperation.swift
```diff
@@ -1,6 +1,7 @@
 public protocol GraphQLOperation: class {
   static var operationString: String { get }
   static var requestString: String { get }
+  static var operationIdentifier: String { get }
```
martijnwalraven: Should we make this optional instead?

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -11,7 +12,11 @@ public extension GraphQLOperation {
   static var requestString: String {
     return operationString
   }
-  
+
+  static var operationIdentifier: String {
+    return ""
```
martijnwalraven: And then this would return `nil`.

Sources/Apollo/GraphQLOperation.swift
kompfner: The problem is that the generated `operationIdentifier`s would then have to be `String?`s as well, which doesn't make much sense from the standpoint of the generated code.

kompfner: One thought I had is to actually make `operationIdentifier`s in `apollo-codegen` *not* be controlled by the command-line `operation-ids-path` argument, and just *always* be generated. Then no default implementation would have to be provided. What do you think?

martijnwalraven: What do you mean by:
> The problem is that the generated `operationIdentifier`s would then have to be `String?`s as well, which doesn't make much sense from the standpoint of the generated code.

kompfner: I'll go ahead and update `apollo-codegen` to produce `String?`s, in keeping with our treatment of `variables`.

kompfner:
Hmm, I'm not sure I can explain why we're getting the exact error messaging we're getting in CI, but it seems like the failing tests are the ones that are taking more than 5s to execute...

These failures also coincided with a period of Travis being extraordinarily slow.

I'm opening another identical branch to kick off another CI build and check whether these failures have to do with some sort of timeout.

kompfner:
Hmm, ok, Travis seemed to execute much faster today and the tests passed. I'm going to merge this since I've discussed this with @martijnwalraven. Might not be the worst idea to address our timeout-related failures in another PR, so that we can survive Travis slow days.

Issue #37, opened 2017-01-05
Provide a pattern for dynamic reconfiguration of network transport / auth headers


timbotnik:
It's not clear how to use the API to implement login / logout, as it would require us to reconfigure headers.  We’ll need a way to dynamically configure / reconfigure the client instance to provide different auth headers, perhaps other transport options as well.

xiekevin:
For what it's worth, the solution @idris brought up is pretty much what we're doing as well

fruitcoder:
How do you handle queries that fail because of invalid/expired tokens? 
Do you catch a 401 in the completion handler of the apollo client and refresh the token? Then the request would still have failed and just the next request might go through (if token refresh was successful). 
I would prefer to have a delegate method with an incoming url request and a closure that takes another url request. Here I would refresh my token, update the header for the request and call the closure with the now authenticated header request which is consumed by the apollo client.

MrAlek:
I'd prefer this to not be in the network transport as a delegate but rather maybe an authentication delegate on the client. Currently, the network transport is the only customization point (see discussion in #6) and if you'd implement your own transport, you would then have to re-implement authentication. 

Cases the authentication mechanism should be able to handle (long-term):
* Inject means of authentication to the operation before it's sent to the network transport
* Blocking an outgoing operation from being sent to the network transport (if unauthorized)
* Re-authenticating before an operation is sent (if token, etc needs to refresh)
* (optionally) Handle unauthorized responses by either failing or re-authenticating. (could be handled by the application)

martijnwalraven:
@MrAlek: Thanks for writing these cases down! The more I think about this, the more I realize this will take some effort to get right. And we should probably think about this in conjunction with #6.

One idea would be to implement something akin to [RequestAdapter](https://github.com/Alamofire/Alamofire/blob/master/Documentation/Alamofire%204.0%20Migration%20Guide.md#request-adapter) and [RequestRetrier](https://github.com/Alamofire/Alamofire/blob/master/Documentation/Alamofire%204.0%20Migration%20Guide.md#request-retrier) in Alamofire. These are fairly low-level callbacks, but they do offer quite a bit of flexibility. You could then perform the authentication yourself, or plug in something like [p2/OAuth](https://github.com/p2/OAuth2/wiki/Alamofire-4). (Or perform exponential backoff on connection failures, etc.)

[...]

PR #109, opened 2017-07-12
Enable to add additional headers after creating a client

weapon3042:
Right now, additional headers can only be specified when creating a client. This Pull Request provides a solution to add additional headers for each request after creating a client.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -65,17 +65,17 @@ public class ApolloClient {
   ///   - result: The result of the fetched query, or `nil` if an error occurred.
   ///   - error: An error that indicates why the fetch failed, or `nil` if the fetch was succesful.
   /// - Returns: An object that can be used to cancel an in progress fetch.
-  @discardableResult public func fetch<Query: GraphQLQuery>(query: Query, cachePolicy: CachePolicy = .returnCacheDataElseFetch, queue: DispatchQueue = DispatchQueue.main, resultHandler: OperationResultHandler<Query>? = nil) -> Cancellable {
-    return _fetch(query: query, cachePolicy: cachePolicy, queue: queue, resultHandler: resultHandler)
+  @discardableResult public func fetch<Query: GraphQLQuery>(query: Query, cachePolicy: CachePolicy = .returnCacheDataElseFetch, additionalHeaders: [String: String]?, queue: DispatchQueue = DispatchQueue.main, resultHandler: OperationResultHandler<Query>? = nil) -> Cancellable {
+    return _fetch(query: query, cachePolicy: cachePolicy, additionalHeaders: additionalHeaders, queue: queue, resultHandler: resultHandler)
   }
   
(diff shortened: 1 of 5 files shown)


martijnwalraven:
Thanks for submitting this PR. While I agree we need a way to add additional headers after creating a client, I don't think we should specify headers as part of the fetch/mutate methods, both because it leaks HTTP specific details and because I'd like to avoid having to force every caller to keep track of header information.

Instead, I think we'll probably want this to be limited to a specific`NetworkTransport` implementation, like `HTTPNetworkTransport`. We've had [some discussions about this](https://github.com/apollographql/apollo-ios/issues/37) before, and one idea was to use a delegate protocol that allows you to modify a `URLRequest` before it gets sent to the server, similar to [`RequestAdapter`](https://github.com/Alamofire/Alamofire#requestadapter) in Alamofire.

Maybe something like:

```
public protocol HTTPNetworkTransportDelegate: class {
  func networkTransport<Operation: GraphQLOperation>(_ networkTransport: HTTPNetworkTransport, willSend request: inout URLRequest, forOperation operation: Operation)
}
```

Do you think that would help solve your use case?

weapon3042:
@martijnwalraven Thanks for your quick response, I agree disclosing header information is not an idea approach. However, my use case requires to fetch the authorization headers from server asynchronously for every request, which required to hold on sending the request until the authorization header fields is fetched. With that said, I doubt the delegate protocol approach will work for this special case and currently I put the authorization header reaching logic inside the application layer, and it will fetch/mutate only when the header fields are successfully fetched. 

Please let me know if you have a better approach to handle setting the header fields  asynchronously, thanks.

martijnwalraven:
Hmmm, do you mean you need fresh credentials for every request? Out of curiosity, what is the server setup that requires that? It seems really inefficient, because you're doing an extra roundtrip per request!

That being said, I think we could still make that work with the delegate approach if we made the method asynchronous.

weapon3042:
@martijnwalraven I agree, actually I have tried to make the delegate method asynchronous, but the side effect is the some functionalites can be broken without a bigger change to HTTPNetworkTransport. Part of the issues is that the URLRequest is added to the operation queue and headers will be fetched and added to the request simultaneously, which can not guarantee the headers are added to the request before the request is executed in the operation queue.  Otherwise, if we have to hold off the request until the headers are properly fetched, a bigger changes in the HTTPNetworkTransport need to by applied.

martijnwalraven:
I'm not sure I understand the problem you're describing. The idea is that we would be modifying the `URLRequest` before starting the session task, so it would block until the credentials have been fetched.

That being said, I still question the sanity of doing this per request :)  So I'm curious to hear the reasons for this requirement.

weapon3042:
@martijnwalraven I have to rephrase it's not making a new request to fetch the headers per request, it's simply a sync API to return the correct HHTP heathers for the current user, so it may or may not make a networking call, most likely making a call only for the first time. As we discussed above, I will adjust this Pull Request to use delegate approach and make it asynchronous. 

Does this sound a good approach to you?

martijnwalraven:
@weapon3042: Yes, that sounds like a good approach, but we may want to spend some time on designing the API first and also consider retrying requests when credentials have expired. I think we could look at [Alamofire](https://github.com/Alamofire/Alamofire#adapting-and-retrying-requests) for inspiration.

weapon3042:
@martijnwalraven Sounds good, thanks for your information!

weapon3042:
Closed for now.

PR #111, opened 2017-07-18
Remove development team from Apollo framework codesigning settings

mochimoshi:
Similar to pull request #50 , this removes the Apollo development team from the Apollo framework code-signing settings. This is needed in order to develop on device - otherwise, Xcode will complain with the following error:

```
None of your accounts are a member of '5TL...':  An unexpected error occurred. Xcode cannot find a team matching '5TL...'.
No signing certificate "iOS Development" found:  No "iOS Development" signing certificate matching team ID "5TL..." with a private key was found.
```

There should be no negative impact with this PR.


martijnwalraven:
Thanks, that makes sense!

Issue #128, opened 2017-08-08
Add CachePolicy to return cache data and then fetch


jasonsilberman:
Currently, I cannot find a way to hit the cache for data first, and then also always hit the network. Sometimes I want to show cached data first, but then fetch from the network to get updated data as well.

If we added a cache policy like: `returnCacheDataAndFetch` then I could do this.

Without this I often times get left with stale data in the cache that continues to be returned, but if I were to use the `fetchIgnoringCacheData` then it would not hit the cache when I am offline. Adding `returnCacheDataAndFetch` would allow developers to have both effective use of the cache and updated data from the network.

Additionally, this would increase parity with Apollo Client because it has the `FetchPolicy` option of [`cache-and-network`](http://dev.apollodata.com/react/api-queries.html#graphql-config-options-fetchPolicy) which allows developers to achieve the same thing.

martijnwalraven:
Yep, that makes sense, and I like `returnCacheDataAndFetch`.

jasonsilberman:
@martijnwalraven okay I think I am going to make a pull request for this, it seems like [`FetchQueryOperation.start()`](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloClient.swift#L168) is the right place to add this?

martijnwalraven:
@jasonsilberman: That would be great! I think it should be fairly straightforward to add to `FetchQueryOperation.start()`.

PR #129, opened 2017-08-08
Add new cache policy: returnCacheDataAndFetch

jasonsilberman:
This pull request adds the CachePolicy: `.returnCacheDataAndFetch`, as discussed in #128.

This new cache policy allows the option to fetch from the cache and then always fetch from the network, similar to the [`cache-and-network`](http://dev.apollodata.com/react/api-queries.html#graphql-config-options-fetchPolicy) option from Apollo Client.


martijnwalraven:
Thanks, this looks good! Could you add a test to `FetchQueryTests.swift` maybe?

jasonsilberman:
@martijnwalraven I added the test and cleaned up the indentation of the first change to match the rest of the file.

Tests/ApolloCacheDependentTests/FetchQueryTests.swift
```diff
@@ -43,6 +43,48 @@ class FetchQueryTests: XCTestCase {
...
+            "__typename": "Human"
+          ]
+        ]
+        ])
+      
+      let client = ApolloClient(networkTransport: networkTransport, store: store)
+      
+      let expectation = self.expectation(description: "Fetching query")
+      
+      client.fetch(query: query, cachePolicy: .fetchIgnoringCacheData) { (result, error) in
```
martijnwalraven: It seems this isn't testing what you intended to test, because the cache policy is `.fetchIgnoringCacheData`.

jasonsilberman: You are correct. Should be fixed now.

PR #131, opened 2017-08-18
Add the ability to set custom http headers for HTTPNetworkTransport

gabrielar:
Sometimes it is necessary to provide custom http headers, eg. to pass requests through a gateway. Thus this pull request adds functionality that does just that by adding an extra parameter to the HTTPNetworkTransport init method. This parameter is a block which is run for each request and must return a dictionary of type [String: String] which will be used to set up the HTTPRequest headers.


designatednerd:
Hey @gabrielar - Thank you for doing this, and sorry for the long radio silence on our end!

Issue #77, opened 2017-04-06
How to utilise Apollo with background session configuration


attheodo:
Hello, I have an operation that I want to continue running in the background. 
I am creating the Apollo client using `let sessionConfig = URLSessionConfiguration.background(withIdentifier: Bundle.main.bundleIdentifier! + ".apolloBackground")` passed in the `HTTPNetworkTransport` object. 

However when I am trying to test how the request behaves in the background, I am getting:
```
2017-04-06 12:10:25.885 Waldo Dev[14754:6295257] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Completion handler blocks are not supported in background sessions. Use a delegate instead.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001074d8b0b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x0000000106bd4141 objc_exception_throw + 48
	2   CFNetwork                           0x0000000109f37e45 -[__NSURLBackgroundSession validateSerializabilityForRequest:completion:] + 172
	3   CFNetwork                           0x0000000109f384fe -[__NSURLBackgroundSession _onqueue_dataTaskForRequest:completion:] + 43
	4   CFNetwork                           0x0000000109f3841d __58-[__NSURLBackgroundSession dataTaskForRequest:completion:]_block_invoke + 34
	5   CFNetwork                           0x0000000109f381ec __68-[__NSURLBackgroundSession performBlockOnQueueAndRethrowExceptions:]_block_invoke + 67
	6   libdispatch.dylib                   0x0000000109a2c05c _dispatch_client_callout + 8
	7   libdispatch.dylib                   0x0000000109a09de6 _dispatch_barrier_sync_f_invoke + 346
	8   CFNetwork                           0x0000000109f38155 -[__NSURLBackgroundSession performBlockOnQueueAndRethrowExceptions:] + 166
	9   CFNetwork                           0x0000000109f3838c -[__NSURLBackgroundSession dataTaskForRequest:completion:] + 211
	10  Apollo                              0x0000000103468626 _TFC6Apollo20HTTPNetworkTransport4senduRxS_16GraphQLOperationrfT9operationx17completionHandlerFTGSqGCS_15GraphQLResponsex__GSqPs5Error___T__PS_11Cancellable_ + 1126
[...]

PR #141, opened 2017-09-12
Make parseResult public

MrAlek:
This ties in with #77. For background URL sessions, you as an app developer isn't in control of neither the app's nor the session's lifecycle. This currently makes you unable to use `ApolloClient` for doing background uploads because you can't really resume an operation from a previous launch.

By making `parseResult` public on `GraphQLResponse` you can manually parse and extract the data once a background session launches your app upon task completion. An alternative would be to have some way to report in a response directly to an `ApolloClient` but I feel that's a lot of more work, especially in finding a good API design.

### DIFF
--- a/Sources/Apollo/GraphQLResponse.swift
+++ b/Sources/Apollo/GraphQLResponse.swift
@@ -8,7 +8,7 @@ public final class GraphQLResponse<Operation: GraphQLOperation> {
     self.body = body
   }
 
-  func parseResult(cacheKeyForObject: CacheKeyForObject? = nil) throws -> Promise<(GraphQLResult<Operation.Data>, RecordSet?)>  {
+  public func parseResult(cacheKeyForObject: CacheKeyForObject? = nil) throws -> Promise<(GraphQLResult<Operation.Data>, RecordSet?)>  {
     let errors: [GraphQLError]?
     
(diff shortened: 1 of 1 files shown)


martijnwalraven:
Sorry for the late response.

I feel making `parseResult` public wouldn't really solve your issue, because the results won't be stored in the normalized cache or used to update other query results. I also don't want to commit to `parseResult` as part of the public API.

I like your alternative better, and it seems that wouldn't have to be that complicated. What do you think of adding this to `ApolloClient`:

```swift
  @discardableResult public func process<Operation: GraphQLOperation>(response: GraphQLResponse<Operation>) throws -> GraphQLResult<Operation.Data>
```

martijnwalraven:
@MrAlek Do you think adding the `process(response:)` method I mentioned above would solve your issue?

MrAlek:
@martijnwalraven Yeah I think that looks great, sorry for the late reply. I'm actually gonna work on revamping our background uploads this week, would be great to synchronize, maybe I can contribute.

martijnwalraven:
@MrAlek Yeah, collaborating on this would be great! Would you have time for a short call to sync up maybe?

MrAlek:
@martijnwalraven Sure, let's talk on Slack.

PR #156, opened 2017-09-27
Add watchOS platform to podspec

### LINKED ISSUE #153: Add watchOS support to podspec
With the introduction of the Apple Watch Series 3, watchOS has become a somewhat mature platform with a full network stack. Since watchOS 2, WatchKit applications have enjoyed on-device execution and separation of concerns from iOS apps.

It would be nice to use Apollo in WatchKit App Extensions. Seeing as Apollo is already fully usable in iOS extensions (like Siri Intent and Today Extensions), it appears that it does not use extension-unsafe features (such as UIApplication.shared).

Have maintainers of this repository tried using Apollo in WatchKit apps in the past? From what I can see at a brief glance, it may work out of the box.

Currently, `Apollo.podspec` does not specify a watchOS version, and thus it cannot be installed using Cocoapods in watchOS apps whatsoever.

ethansinjin:
Resolves #153

### DIFF
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -16,6 +16,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '9.0'
   s.osx.deployment_target = '10.10'
   s.tvos.deployment_target = '9.1'
+  s.watchos.deployment_target = '3.0'
 
   s.subspec 'Core' do |ss|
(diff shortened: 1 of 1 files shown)


martijnwalraven:
Good to hear it's working for you on watchOS, thanks!

PR #165, opened 2017-10-17
Add docs & param names to NetworkTransport & OperationResultHandler

MrAlek:
Noticed this while looking at making the network transport throwing.

Before:
skarmavbild 2017-10-17 kl 18 35 57

After:
skarmavbild 2017-10-17 kl 18 35 43


martijnwalraven:
Thanks! I have some vague recollection of struggling with this a while ago, I think there was an issue with specifying parameters on type aliases before that may have been fixed in a more recent Xcode.

PR #172, opened 2017-10-26
Aynchronous request modification

DarkDust:
This is a proposal to solve #37 by adding a closure that may be executed asynchronously to modify the request before it's actually sent.


DarkDust:
have signed the Meteor Contributor Agreement.

martijnwalraven:
Thanks for the PR! I'm a bit preoccupied with GraphQL Summit right now, so unfortunately I won't have time to give it the attention it deserves until next week.

My thought is that it would be great to see if we can make this more generic with a design similar to Apollo Link. The [concept of links](https://www.apollographql.com/docs/link/overview.html) is pretty powerful, and I've seen some great uses of it at the conference today, so I would really like to open up that flexibility on iOS as well. 

I would also want to involve @MrAlek, because I know he has some use cases of his own that he would need support for.

designatednerd:
Hey @DarkDust  - Thank you for doing this, and sorry for the long radio silence on our end!

Issue #170, opened 2017-10-25
Expose whether a particular result was served from cache or fetched from the server


AnandWalvekar:
**What**
iOS client want to have custom TTL(Time To Live) policy for caching

**Current Behavior**
When cachePolicy = .returnCacheDataAndFetch, client.fetch API gives 2 callbacks. The client has don't have information regarding the source of data in the closure callback

**Expected Behavior**
It would be nice if the closure give 3 params (result, dataStore, error). 
dataStore = ( .cache or .server)

martijnwalraven:
I understand your expected behavior, but I'm not sure how that relates to custom TTL. As far as I can see, that would only tell you whether data came from the cache or the network. So really curious about your reasoning and use case.

You may also be interested in the recent [Apollo Cache Control](https://github.com/apollographql/apollo-tracing) format, which it would be great to take advantage of on the client as well.

AnandWalvekar:
Apollo Cache Control => Gives performance metrix in response. This part is more of a server teams responsibility which iOS client to worry about(unless it is job is to check API performance)

**Use case**
Client maintains a cache time (say 10 sec) for a particular query. After fetching from server for the first time it would update **lastFetchedTime** . Client keep querying for the data from cache till cache time expires. Once that time expires it would make server request with .returnCacheDataAndFetch policy and updates **lastFetchedTime** again

**Concern**
The client should update **lastFetchedTime** only upon successful server response.

martijnwalraven:
Ah, I'm sorry, I included the wrong link for  [Apollo Cache Control](https://github.com/apollographql/apollo-cache-control) (since corrected)! The idea there is that servers can specify which part of a response can be cached, and for how long (TTL). That would be useful for server-controlled cache expiration.

Let's keep this issue focused on the idea of wanting to know whether a particular result was served from cache or fetched from the server. That could also include how long it has been in the cache. 

Instead of adding a third parameter to the closure, I'd prefer to make that part of `GraphQLResult`. What do you think?

AnandWalvekar:
That make sense too.

AnandWalvekar:
@martijnwalraven 
Could you please confirm when can I expect for it?

martijnwalraven:
While feature requests are always welcome, and this sounds like a useful addition, I can't give you any guarantees when it will be implemented. As with any open source project, the quickest way to get your feature in is to open a PR yourself.

AnandWalvekar:
Understood.

[...]

PR #176, opened 2017-10-31
Graph ql result update

AnandWalvekar:
This commit is done as per discussion in https://github.com/apollographql/apollo-ios/issues/170

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -98,7 +98,7 @@ public final class ApolloStore {
 
       return try transaction.execute(selections: Query.Data.selections, onObjectWithKey: Query.rootCacheKey, variables: query.variables, accumulator: zip(mapper, dependencyTracker))
     }.map { (data: Query.Data, dependentKeys: Set<CacheKey>) in
-      GraphQLResult(data: data, errors: nil, dependentKeys: dependentKeys)
+      GraphQLResult(data: data, errors: nil, dataSource:.cache, dependentKeys: dependentKeys)
     }
   }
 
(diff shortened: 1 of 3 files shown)


AnandWalvekar:
Hi @martijnwalraven 
Can you please review the PR

Sources/Apollo/ApolloStore.swift
```diff
@@ -98,7 +98,7 @@ public final class ApolloStore {
 
       return try transaction.execute(selections: Query.Data.selections, onObjectWithKey: Query.rootCacheKey, variables: query.variables, accumulator: zip(mapper, dependencyTracker))
     }.map { (data: Query.Data, dependentKeys: Set<CacheKey>) in
-      GraphQLResult(data: data, errors: nil, dependentKeys: dependentKeys)
+      GraphQLResult(data: data, errors: nil, dataSource:.cache, dependentKeys: dependentKeys)
```
martijnwalraven: I prefer `source` over `dataSource` here (and everywhere else), with a corresponding change of `GraphQLDataSource` to `GraphQLResult.Source`.

Sources/Apollo/GraphQLResult.swift
```diff
@@ -1,15 +1,24 @@
+/// Represents source of data in GraphQLResult
+public enum GraphQLDataSource {
```
martijnwalraven: I think `GraphQLResult.Source` (so nesting the `Source` enum in `GraphQLResult`) would work better here.

Sources/Apollo/GraphQLResult.swift
```diff
@@ -1,15 +1,24 @@
...
+
 /// Represents the result of a GraphQL operation.
 public struct GraphQLResult<Data> {
   /// The typed result data, or `nil` if an error was encountered that prevented a valid response.
   public let data: Data?
   /// A list of errors, or `nil` if the operation completed without encountering any errors.
   public let errors: [GraphQLError]?
-  
+  /// Source of data, or `nil` if there are any errors
+  public let dataSource:GraphQLDataSource?
```
martijnwalraven: I don't think this should be optional. Even if there are errors, these are still part of the result and come from either the network or the cache.

AnandWalvekar:
Hi @martijnwalraven 
Thank you for reviewing. I have made changes as commented

martijnwalraven:
Thanks, and congrats on your first contribution!

PR #180, opened 2017-11-14
Update fetching-queries.md

Gregoirevda:
Log is missing `result`
Note that the link to edit this file is not correct.


abernix:
Thanks so much, @Gregoirevda.  I've fixed the broken "Edit on GitHub" link via https://github.com/apollographql/apollo-ios/pull/258.  So sorry this took so long to land — this looks right to me!

Issue #173, opened 2017-10-27
Discussion: What would an Apollo Link implementation look like?


nubbel:
Hey!

Modifying/intercepting the way the Client communicates with the network is a popular request by the community. Solutions have been proposed in #131 and #172, but have not been accepted. Instead a more generic design akin to [Apollo Link](https://github.com/apollographql/apollo-link) was suggested.

I'm interested in having a stab at implementing Apollo Link for apollo-ios.
Before I get down to coding, I'd like to discuss a few points:

1. The Javascript version of Apollo Link heavily relies on Observables. Currently, there is no notion of Observables in Swift nor in apollo-ios. I understand that Observables are useful for things like automatic retrying etc., but would you consider that a strict requirement or would an implementation based on Promises also be acceptable for you?
2. In case a design based on Observables is a strict requirement, would you prefer an implementation of Observables for Swift from scratch or could we use an already existing solution (similar to how the Javascript version depends on zen-observables)? I'm not sure if that would cause any licensing issues though. Implementing Observables from scratch is doable, but definitely not trivial. You can check out my first experiments in this [gist](https://gist.github.com/nubbel/a71943f7d4a3bee7fb09133c4edd2eca) (neglecting all concurrency). The most popular implementation of Observables in Swift seems to be [RxSwift](https://github.com/ReactiveX/RxSwift). It is extremely feature-rich and well-tested, but I would prefer a more lightweight solution if there is any.
3. The Javascript version uses a context object for link-to-link communication. Could anyone think of a more Swifty and typesafe solution?
4. Should Apollo Link be a separate Pod (just like Apollo/SQLite) or should it be integrated into Core?

Let me know what you think!

@martijnwalraven, @MrAlek, @DarkDust

martijnwalraven:
Really happy you're opening this issue, and I think you're asking exactly the right questions. Would be great to bring more people together to figure this out. I'm traveling right now, but maybe we could set up a meeting for next week.

Let me give you some preliminary answers, but we should probably schedule a call to talk about this in more depth and discuss next steps:
- I don't think our Link implementation necessarily has to use Observables, and I agree we don't want a dependency on something like RxSwift. If we do, I'd prefer a lightweight built-in implementation as well, similar to what we've done with Promises.
[...]

PR #186, opened 2017-11-22
[WIP] Add Apollo Link

nubbel:
Also see the discussion in #173.

# Terminating vs. Non-Terminating Links

ApolloLink has a concept of terminating and non-terminating links. In the Javascript implementation, the link type is determined at runtime.
I thought we could de better and determine it at compile time by using two distinct types TerminatingLink and NonTerminatingLink. This eliminates some compile time checks. For example, concat is only defined on NonTerminatingLink and returns a terminating or non-terminating link, depending on the type of the second link.
However, it makes it difficult to define a split of a terminating and a non-terminating link or vice-versa, because what should the return type be? SchrödingersLink?

# Cancellation

Requests need to be cancellable, that means links need to be cancellable. I considered three options to achieve that:

Using Observables. Observables are inherently cancellable, but non-trivial to implement.
Adding a Cancelled state to Promise.
Using a cancellable object, that us to register onCancel handlers. This was inspired by an experimental Javascript feature, AbortController.
I chose the first third option and created an CancelController and CancelSignal similar to AbortController/AbortSignal in Javascript. The signal is passed in the context.
It seemed to be the less invasive solution and I actually really like the design.

# Caveats

The current design has a few caveats. Most notably, links can only return one result, which renders them unsuitable for subscriptions.

# Next steps

As a next step, I would like to implement an HTTPLink that pulls configuration from the context and makes a network request.
Then, I'd like to see how we could implement a RetryLink.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -28,7 +28,7 @@ public typealias OperationResultHandler<Operation: GraphQLOperation> = (_ result
 
 /// The `ApolloClient` class provides the core API for Apollo. This API provides methods to fetch and watch queries, and to perform mutations.
 public class ApolloClient {
-  let networkTransport: NetworkTransport
+  let link: TerminatingLink
   let store: ApolloStore
   public var cacheKeyForObject: CacheKeyForObject? {
     get {
@@ -43,17 +43,27 @@ public class ApolloClient {
   private let queue: DispatchQueue
   private let operationQueue: OperationQueue
   
+  
+  /// Creates a client with the specified link and store.
+  ///
+  /// - Parameters:
+  ///   - link: The {@link ApolloLink} over which GraphQL documents will be resolved into a response.
+  ///   - store: A store used as a local cache. Defaults to an empty store backed by an in memory cache.
+  public init(link: TerminatingLink, store: ApolloStore = ApolloStore(cache: InMemoryNormalizedCache())) {
+    self.link = link
+    self.store = store
+    
+    queue = DispatchQueue(label: "com.apollographql.ApolloClient", attributes: .concurrent)
+    operationQueue = OperationQueue()
+  }
+  
   /// Creates a client with the specified network transport and store.
   ///
   /// - Parameters:
   ///   - networkTransport: A network transport used to send operations to a server.
   ///   - store: A store used as a local cache. Defaults to an empty store backed by an in memory cache.
-  public init(networkTransport: NetworkTransport, store: ApolloStore = ApolloStore(cache: InMemoryNormalizedCache())) {
-    self.networkTransport = networkTransport
-    self.store = store
-    
-    queue = DispatchQueue(label: "com.apollographql.ApolloClient", attributes: .concurrent)
-    operationQueue = OperationQueue()
+  public convenience init(networkTransport: NetworkTransport, store: ApolloStore = ApolloStore(cache: InMemoryNormalizedCache())) {
+    self.init(link: NetworkTransportLink(networkTransport: networkTransport), store: store)
   }
   
   /// Creates a client with an HTTP network transport connecting to the specified URL.
@@ -125,26 +135,28 @@ public class ApolloClient {
       }
     }
     
-    return networkTransport.send(operation: operation) { (response, error) in
-      guard let response = response else {
-        notifyResultHandler(result: nil, error: error)
-        return
-      }
+    let controller = CancelController()
+    
+    var linkContext = LinkContext()
+    linkContext.signal = controller.signal
+    
+    link.request(operation: operation, context: linkContext).flatMap { response in
+      try response.parseResult(cacheKeyForObject: self.store.cacheKeyForObject)
+    }.andThen { (result, records) in
+      notifyResultHandler(result: result, error: nil)
       
-      firstly {
-        try response.parseResult(cacheKeyForObject: self.store.cacheKeyForObject)
-      }.andThen { (result, records) in
-        notifyResultHandler(result: result, error: nil)
-        
-        if let records = records {
-          self.store.publish(records: records, context: context).catch { error in
-            preconditionFailure(String(describing: error))
-          }
+      if let records = records {
+        self.store.publish(records: records, context: context).catch { error in
+          preconditionFailure(String(describing: error))
         }
-      }.catch { error in
-        notifyResultHandler(result: nil, error: error)
       }
+    }.catch { error in
+      notifyResultHandler(result: nil, error: error)
     }
+    
+        
+    
+    return controller
   }
 }
 
--- a/Sources/Apollo/CancelController.swift
+++ b/Sources/Apollo/CancelController.swift
@@ -1,0 +1,37 @@
+public final class CancelController: Cancellable {
+  public let signal = CancelSignal()
+  
+  public func cancel() {
+    signal.cancel()
+  }
+}
+
+public final class CancelSignal {
+  private let promise: Promise<Void>
+  private let fulfill: (() -> Void)
+  
+  fileprivate init() {
+    var _fulfill: (() -> Void)!
+    promise = Promise { (fulfill, _) in
+      _fulfill = fulfill
+    }
+    
+    fulfill = _fulfill
+  }
+  
+  fileprivate func cancel() {
+    fulfill()
+  }
+  
+  public var isCancelled: Bool {
+    return !promise.isPending
+  }
+  
+  public func onCancel(_ whenCancelled: @escaping () -> Void) {
+    promise.andThen(whenCancelled)
+  }
+}
+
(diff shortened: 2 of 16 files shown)


Tests/ApolloTests/LinkTests.swift
```diff
@@ -0,0 +1,433 @@
...
+    
+    waitForExpectations(timeout: 0.1)
+  }
+  
+  func testSplitNonTerminatingLinkFirst() {
+    let queryLink = SetContextLink.incrementCounter(by: 1)
+    let mutationLink = SetContextLink.incrementCounter(by: -1)
+    
+    let splitLink = passthroughLink.split(first: queryLink, second: mutationLink) { op, _  in
+      type(of: op).rootCacheKey == "QUERY_ROOT"
```
martijnwalraven: We shouldn't rely on `rootCacheKey` to get the operation type (and probably not expose it). Instead, maybe we could have a `GraphQLOperationType` enum and `operationType` property.

martijnwalraven: I would also prefer a way to avoid having to go through `type(of: op)`, making the properties non-static. (They can still be static in the generated code, but maybe we can define instance properties that forward to the static ones in a protocol extension.)

Sources/Apollo/ConcatLink.swift
```diff
@@ -0,0 +1,41 @@
...
+fileprivate final class TerminatingConcatLink: TerminatingLink {
+  private var first: NonTerminatingLink
+  private var second: TerminatingLink
+  
+  init(first: NonTerminatingLink, second: TerminatingLink) {
+    self.first = first
+    self.second = second
+  }
+  
+  func request<Operation>(operation: Operation, context: LinkContext) -> Promise<GraphQLResponse<Operation>> {
```
martijnwalraven: We really can't rely on `Promise` here, because links need the ability to return multiple responses. That's important for subscriptions support, but also for something like a polling link. I'd rather not implement `Observable` (or rely on RxSwift), but maybe a response handler callback would work? On the other hand, promises make the link code a lot easier to write, and so would observables, so maybe it is worth implementing something minimal.

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -1,12 +1,14 @@
-public protocol GraphQLOperation: class {
+public protocol GraphQLOperationBase {
```
martijnwalraven: We may want to keep the `: class` because it makes code using it [slightly more efficient](https://github.com/apple/swift/blob/master/docs/OptimizationTips.rst#advice-mark-protocols-that-are-only-satisfied-by-classes-as-class-protocols), and there is no reason to have structs conform to this protocol.

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -1,12 +1,14 @@
-public protocol GraphQLOperation: class {
```
martijnwalraven: Hmmm, actually, I don't think links ever need knowledge of the associated type. Here is an alternative idea: 

We keep `GraphQLOperation` as is (so without `GraphQLOperationBase`), but we define a new `GraphQLRequest` protocol and make it the basis of links. So instead of `request(operation:context:)` it would now be `perform(request:)`. 

That means `GraphQLRequest` also contains the context, basically everything needed to perform a request. Something like:

```swift
protocol GraphQLRequest {
  var context: GraphQLRequestContext { get }
  var operationType: GraphQLOperationType { get }
  var requestString: String { get }
  var variables: GraphQLMap? { get }
}
```

 I like the idea that it makes links `GraphQLRequest -> GraphQLResponse`, and it cleanly differentiates between operations (reusable across requests) and requests (containing everything that is needed fora single request, including context)

We could then have a `GraphQLOperationRequest` class that initialises itself based on a `GraphQLOperation`. So similar to your `LinkOperationDescriptor`, but with instance properties and also containing the context.

An added benefit is that it also allows you to use links without generated code, or even without the rest of Apollo iOS by implementing something like a `GraphQLDynamicRequest` class that isn't initialised from a `GraphQLOperation`.

We do need to make `GraphQLResponse` depend on `GraphQLRequest` instead of `GraphQLOperation` as well, but that might be a good idea anyway because it no longer has to be generic and that way we can more easily store responses.

What do you think?

Sources/Apollo/ApolloClient.swift
```diff
@@ -28,7 +28,7 @@ public typealias OperationResultHandler<Operation: GraphQLOperation> = (_ result
 
 /// The `ApolloClient` class provides the core API for Apollo. This API provides methods to fetch and watch queries, and to perform mutations.
 public class ApolloClient {
-  let networkTransport: NetworkTransport
+  let link: TerminatingLink
```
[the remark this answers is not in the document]
nubbel: I agree, it could be confusing, but actually it makes quite a lot of sense.
The confusion comes from the belief that there is a chain of links, but there isn't.
The only way to combine links is by composition, which produces a new link.

Consider concatenation:
`concat(NT, T) -> T` and `concat(NT, NT) -> NT` where `NT` is a non-terminating link and `T` is a terminating link.

And split:
`split(NT, NT) -> NT` and `split(T, T) -> T`

Then there is also forced termination:
`makeTerminating(NT) -> T`

As you can see, no matter how many links go into the composition the end result is always a single link.

It makes sense to give the client a terminating link, because it is what produces a result in the end.
Note that this very terminating link may be a single link (e.g. `HTTPLink`) or a link that was composed from (multiple) other links (e.g. `concat(AuthLink, split(HTTPLink, WebSocketLink))`).

Does that make sense to you?

Sources/Apollo/GraphQLOperation.swift
nubbel: Sound great!

Sources/Apollo/GraphQLOperation.swift
nubbel: Interesting

Sources/Apollo/ConcatLink.swift
nubbel: Yeah, I know :(. 
Observables would be nice to have and it would probably be fun to implement, but difficult to get right.
Here's another idea: Similar to the cancel signal, we could add a `onNextResponse` callback to the context which is called whenever a link produces a new response. What do you think?

Sources/Apollo/ApolloClient.swift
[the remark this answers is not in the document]
martijnwalraven: Ah, you're right. I was thinking of this as the first link in the chain, but it's function composition. If we can get this working reliably that would be great, but I don't think it's absolutely necessary to make termination part of the type system. 

> However, it makes it difficult to define a split of a terminating and a non-terminating link or vice-versa, because what should the return type be? SchrödingersLink?

Have you thought more about this yet?

And what does `makeTerminating` do?

Sources/Apollo/ConcatLink.swift
martijnwalraven: Can you give an example of how `onNextResponse` would work? Would this replace returning a promise?

dangthaison91:
How is this going on @nubbel @martijnwalraven ? This is really helpful!

Sources/Apollo/CancelController.swift
```diff
@@ -0,0 +1,37 @@
+public final class CancelController: Cancellable {
```
wtrocki: @ziccardi :)

koenpunt:
Hi there, what's the status of this PR?

LeeWong:
What's the status of this PR?

designatednerd:
@nubbel Do you have any interest in continuing this PR? If not, I can close this out for you.

wtrocki:
@designatednerd I'm so excited, but I first want to make sure. By closing, you mean resolving and merging this PR?

This PR is a really awesome way to enable people to build advanced customization on top of Apollo IOS.
There is a lot of work done already. It may require very little effort now to get it merged, but IMHO it will be the game changer that may really help the community to customize their clients without pushing some very custom cases to the core.

designatednerd:
@wtrocki Closing would mean closing without merging.  

We'd like to add this functionality in general, but this PR is pretty old and has a lot of outstanding comments, and I want to determine if this is something I should take the time to review here, or if that time would be better used looking into a new solution.

wtrocki:
@designatednerd well understood thank you.

designatednerd:
We're definitely still interested in making request and response handling more composable but at this point, this PR is super out of date and I haven't heard back from the original author for a week about whether they want to proceed. I'll close this one out but definitely keeping #173 open until I'm able to get a better line on this. 

Thanks very much for your work and sorry this wound up getting neglected!

nubbel:
I'm sorry I haven't been able to finish this.

@wtrocki If you intend to work on this issue and anything in this PR is useful, please feel free to reuse as much as you want.

PR #187, opened 2017-11-30
Trigger watches when write transactions update cache.

rhishikeshj:
Context
If the cache is updated from an read-write transaction, watch queries are not triggered even though the data is updated.

Solution 
Added a new member to the ReadWriteTransaction object to get the ApolloStore object.
Triggered the subscriber.store functions to make sure watch queries are triggered.

This is required because the subscribers and subscriber.store function both are dependent upon the current ApolloStore object on which the transaction is executing.
If this breaks the abstraction of the ReadWriteTransaction in any way, let me know and also if you could suggest a better way to do this.

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -77,7 +77,9 @@ public final class ApolloStore {
       self.queue.async(flags: .barrier) {
         self.cacheLock.lockForWriting()
 
-        fulfill(ReadWriteTransaction(cache: self.cache, cacheKeyForObject: self.cacheKeyForObject))
+        fulfill(ReadWriteTransaction(cache: self.cache,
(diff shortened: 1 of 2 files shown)


martijnwalraven:
Thanks, this looks great! Happy you included a test. 

Maybe it would be cleaner to extract subscriber notification from `publish()` into its own `fileprivate` `didChangeKeys(_ changedKeys:context:)` function on `ApolloStore`. That avoids code duplication and respects encapsulation a little better. We could then pass that function into the transaction (as `self. didChangeKeys`) instead of the `store` itself.

rhishikeshj:
@martijnwalraven Is this closer to what you would be comfortable with ?

martijnwalraven:
Looks good to me, thanks!

PR #202, opened 2018-01-04
Use soft unwrapping of error dictionary keys to avoid crashes

MrAlek:
This prevents the client from crashing trying to print out malformed errors received from the GraphQL server.

### DIFF
--- a/Sources/Apollo/GraphQLError.swift
+++ b/Sources/Apollo/GraphQLError.swift
@@ -20,13 +20,13 @@ public struct GraphQLError: Error {
   }
   
   /// A description of the error.
-  public var message: String {
-    return self["message"] as! String
+  public var message: String? {
+    return self["message"] as? String
   }
   
   /// A list of locations in the requested GraphQL document associated with the error.
   public var locations: [Location]? {
-    return (self["locations"] as? [JSONObject])?.map(Location.init)
+    return (self["locations"] as? [JSONObject])?.flatMap(Location.init)
   }
   
   /// Represents a location in a GraphQL document.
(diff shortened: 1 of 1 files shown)


martijnwalraven:
According to [the spec](http://facebook.github.io/graphql/October2016/#sec-Errors), errors must contain a `message` key, and if they contain `locations`, each location must have a `line` and `column`. So I'd really like to keep these non-optional to reflect that.

Have you encountered issues with non-compliant GraphQL servers? I agree we should avoid crashes in these cases, but maybe we can return something like `<missing error message>` for `message`, or `-1` for `line`/`column` instead?

MrAlek:
You have a point. What about keeping it required and doing all the JSON parsing in a failable initializer and on
https://github.com/apollographql/apollo-ios/blob/d93b763226c97eaf9128f643fb56c756e2ba7560/Sources/Apollo/GraphQLResponse.swift#L15

do `errors = errorsEntry.flatMap(GraphQLError.init)`

MrAlek:
Bumping this one again :) @martijnwalraven

martijnwalraven:
A failable initializer seems like a great solution. My only concern is that using `errorsEntry.flatMap` would hide malformed errors received from the GraphQL server, so maybe we still want to log these somehow.

MrAlek:
@martijnwalraven Do we have equivalent logging for response objects that fail to parse somewhere?

martijnwalraven:
Nope, but we probably should :)

MrAlek:
Could we first do the flatMap and at least prevent the client from crashing and then think about logging as a separate project? There are many things to consider and I'd rather not just put in stray `print`s. Unit tests should suffice for verifying functionality.

MrAlek:
But to catch errors later on, we can make the error initializer throwing.

martijnwalraven:
Yep, that makes sense. I don't think we should block this on logging.

Out of curiosity, is this something you've encountered with a non-spec compliant GraphQL server?

MrAlek:
Can't remember the scenario, might have been some API facade before the GraphQL server that responded too.

tonythebully:
Just had the same issue where the app was pretty much crashing on launch due to this.
While understandably, the graphql servers should follow the conventions, client probably shouldn't have to pay for their sins so to speak.

martijnwalraven:
@tonythebully Thanks for bringing this to my attention again, I agree we should be lenient here.

PR #229, opened 2018-03-18
Adds support for running codegen from node modules if available

### LINKED ISSUE #219: Use a local reference for exec-ing apollo-codegen in build script first
Hey all, I had [some problems](https://github.com/rnystrom/GitHawk/pull/1508#issuecomment-364670552) with compiling [GitHawk](https://github.com/rnystrom/GitHawk) recently. What happened is that GitHawk would lock Apollo codegen to [0.17-a13](https://github.com/rnystrom/GitHawk/blob/359773a427a2d4b8a029b8b0fc2efbb16d773e4d/package.json#L3) and then in the app, it would use the version of apollo-codegen that is installed globally.

I somehow have a later version of apollo codegen installed ( 0.17.2 ) and so the code generated between the two packages doesn't compile.

Would it make sense to have something like:

```sh
if [[ -s "./node_modules/apollo-codegen/cli/index.js" ]]; then
  exec ./node_modules/apollo-codegen/cli/index.js "$@"
else
  exec apollo-codegen "$@"
fi
```

at the end of  [check-and-run-apollo-codegen.sh](https://github.com/apollographql/apollo-ios/blob/master/scripts/check-and-run-apollo-codegen.sh) to prioritise the exact version of apollo-codegen would have installed?

/cc  @rnystrom

orta:
This fixes #219 by looking in `node_modules` in the same folder as the Xcode project for the app running the script.

### DIFF
--- a/scripts/check-and-run-apollo-codegen.sh
+++ b/scripts/check-and-run-apollo-codegen.sh
@@ -1,7 +1,8 @@
 # Only major and minor version should be specified here
 REQUIRED_APOLLO_CODEGEN_VERSION=0.19
 
-# Part of this code has been adapted from https://github.com/facebook/react-native/blob/master/packager/react-native-xcode.sh
+# Part of this code has been adapted from
+# https://github.com/facebook/react-native/blob/master/scripts/react-native-xcode.sh
 
 # This script is supposed to be invoked as part of the Xcode build process
 # and relies on environment variables set by Xcode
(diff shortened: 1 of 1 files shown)


martijnwalraven:
Thanks!

scripts/check-and-run-apollo-codegen.sh
```diff
@@ -46,20 +47,28 @@ if [[ -x "$HOME/.nodenv/bin/nodenv" ]]; then
...
-if ! type "apollo-codegen" >/dev/null 2>&1; then
-  echo "Can't find apollo-codegen. Installing..."
-  install_apollo_codegen
-fi
+if [[ -s "$SRCROOT/node_modules/.bin/apollo-codegen" ]]; then
+  # If it's installed locally, use version build instead
+  set -x
 
-INSTALLED_APOLLO_CODEGEN_VERSION="$(get_installed_version)"
+  exec "$SRCROOT/node_modules/.bin/apollo-codegen" "$@"
```
cysp: Do we not want to test the version of `apollo-codegen` here?
It could still be an incompatible version, couldn’t it?

PR #233, opened 2018-03-23
Fix an issue that watcher doesn't notify specific change

mitsuyoshi-yamazaki:
# Agenda

Fix an issue that watcher doesn't notify specific change.

# Detail

`GraphQLQueryWatcher.dependentKeys` doesn't contain specific kind of key.
In a following query and `apolloClient.cacheKeyForObject = { $0["id"] }`, `<user_id>.company` is not contained in `GraphQLQueryWatcher.dependentKeys`.

```graphql
query GetUser($id: ID!) {
  getUser(id: $id) {
    id
    name
    company {
      id
      name
    }
  }
}
```

### Side effects

I don't think it has side effects, since `GraphQLQueryWatcher.dependentKeys` is `Set` so keys can't be duplicated.

### DIFF
--- a/Sources/Apollo/GraphQLDependencyTracker.swift
+++ b/Sources/Apollo/GraphQLDependencyTracker.swift
@@ -14,6 +14,7 @@ final class GraphQLDependencyTracker: GraphQLResultAccumulator {
   }
   
   func accept(fieldEntry: Void, info: GraphQLResolveInfo) {
+    dependentKeys.insert(joined(path: info.cachePath))
   }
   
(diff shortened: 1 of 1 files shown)


martijnwalraven:
@mitsuyoshi-yamazaki Thanks! I know this sounds like overkill for a single line change, but I still need to ask you to sign the CLA. Otherwise this seems ready to merge!

mitsuyoshi-yamazaki:
Oh, I didn't "submit" that.
Now it should be fine.

PR #246, opened 2018-04-05
Subscription support: Fixed bug when re-connecting and multiple active subscriptions

knutaa:


SRozhina:
@knutaa @martijnwalraven Hello.
I could not find any information about subscriptions usage in the documentation. But as I understood from this PR you've added a new init for ApolloClient which has "networkTransport" parameter. 
And in tests, you create an instance ApolloWebSocket and give it to ApolloClient. Great. But I could not create ApolloWebSocket in my project (Apollo version 0.9.5). What should I do to create ApolloWebSocket or to make subscription work?
Thanks :)

Issue #228, opened 2018-03-16
Support Certificate Pinning


tsorencraig:
Public key pinning is a great step for apps to reduce MITM vulnerabilities. The current instantiation of the Apollo Client allows a parameter for the URLSession configuration. In order to implement SSL pinning, however, an entire URLSession object should be able to be passed in. This would allows the URLSession object to register a delegate that handles the pinning by implementing

```
optional public func urlSession(_ session: URLSession, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Swift.Void)
``` 

This can be done simply by adding another constructor to the HttpNetworkTransport object that accepts the URLSession object instead of just it's configuration. 

After that, injecting the session object with the appropriate delegate is simple.

```
let configuration = URLSessionConfiguration.default
        
let networkTransport = HTTPNetworkTransport(url: "https://some-gql-endpoint/graphql",
                                            session: URLSession(configuration: configuration,
                                                                                delegate: NSURLSessionPinningDelegate(),
                                                                                delegateQueue: nil))

let client = ApolloClient(networkTransport: networkTransport)
```

martijnwalraven:
I don't think we should give out control of `URLSession` completely, because `HTTPNetworkTransport` should have the ability to implement `URLSessionDataDelegate` methods. But maybe we could add a `HTTPNetworkTransportDelegate` protocol that would pass through the authentication challenge.

This is also one of the possible solutions to handling token authentication. See [this issue](https://github.com/apollographql/apollo-ios/issues/37#issuecomment-278016619) for a long discussion about this :) We should really make some decisions here and solve these issues. @MrAlek, what do you think?

MrAlek:
Though adding a `HTTPNetworkTransportDelegate` would be one way to open up the API, I think people will find more and more cases for customization. Maybe `HTTPNetworkTransport` should be opened up for subclassing instead, so you can override delegate methods, inspect the `HTTPURLResponse` & modify outgoing `URLRequest`s? That could be a good intermediary solution before links are in place.

johntmcintosh:
I'd second the idea of opening `HTTPNetworkTransport` for subclassing. 

In our app, we've needed to re-implement `HTTPNetworkTransport` to have some hooks into both outgoing requests and incoming responses. Our use cases are:

- Injecting logging of the request and response objects 
- Automatically triggering de-authentication if a 403 response is received
[...]

PR #265, opened 2018-04-25
Add HTTPNetworkTransport.init to take URLSession instance to allow …

chrislconover:
Add HTTPNetworkTransport.init to take URLSession instance to allow caller to handle exceptional cases by way of URLSessionDelegate.  The problem was that HTTPNetworkTransport was hiding and subsequently blocking access to existing functionality on URLSession.  

This change is to add an additional initializing that takes an URLSession instance from the caller, allowing said caller to first register a delegate handler.  By contrast, Alamofire handles this in arguably a more elegant method, by exposing optional closure methods to handle each of the three URLSessionDelegate methods.

The approach taken here is simpler to implement, but is less elegant for calling code.  It is more conventional though and will suffice until any higher level solution decision is warranted.

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -53,9 +53,20 @@ public class HTTPNetworkTransport: NetworkTransport {
   ///   - url: The URL of a GraphQL server to connect to.
   ///   - configuration: A session configuration used to configure the session. Defaults to `URLSessionConfiguration.default`.
   ///   - sendOperationIdentifiers: Whether to send operation identifiers rather than full operation text, for use with servers that support query persistence. Defaults to false.
-  public init(url: URL, configuration: URLSessionConfiguration = URLSessionConfiguration.default, sendOperationIdentifiers: Bool = false) {
+  public convenience init(url: URL, configuration: URLSessionConfiguration = URLSessionConfiguration.default, sendOperationIdentifiers: Bool = false) {
+    self.init(url: url, session: URLSession(configuration: configuration),
+              sendOperationIdentifiers: sendOperationIdentifiers)
+  }
+
+  /// Creates a network transport with the specified server URL and session.
+  ///
+  /// - Parameters:
+  ///   - url: The URL of a GraphQL server to connect to.
+  ///   - session: An URLSession instance to be used for ensuing operations.
+  ///   - sendOperationIdentifiers: Whether to send operation identifiers rather than full operation text, for use with servers that support query persistence. Defaults to false.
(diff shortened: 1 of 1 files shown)


martijnwalraven:
I don't think we can allow passing in an existing session, because we want `HTTPNetworkTransport` to be able to set itself as the delegate. What use cases do you have in mind for this? Hopefully we could address these as part of the `HTTPNetworkTransportDelegate` API in https://github.com/apollographql/apollo-ios/pull/257.

tsorencraig:
https://github.com/apollographql/apollo-ios/issues/228

designatednerd:
Hey @chrislconover - Thank you for doing this, and sorry for the long radio silence on our end! Can you please expand on the use cases here? Certificate setup seems to clearly be one, but I want to see if there's other stuff that isn't covered by the changes in #602.

designatednerd:
@chrislconover Any thoughts on this? If I don't hear back within another week I'll close this PR out in favor of the changes that have shipped with 0.11.0.

chrislconover:
I don't have the calling code in front of me, but effectively you are breaking the existing interface, by blocking access to session configuration.  I think I was having to deal with authentication challenge issue that are handled through the session and session configuration.  It's been a while now, but in my collection, you broke the interface by wrapping with with an opaque layer, preventing one from accessing the existing URL session functionality.  

It is generally unsafe and unwise to assume that nobody will need the existing flexibility and then wrap it with an opaque layer - this is generally considered poor API design.

designatednerd:
@chrislconover OK, it's helpful to understand the use case. 

We're still working through a lot of the decisions that were made when the library was first created (and wasn't used as widely for things which weren't Apollo products). There was a time when we were trying to avoid passing in the session since we wanted to have the option of being the session's delegate ourselves, but I need to talk through with some colleagues the implications of removing that ability before we allow that.

designatednerd:
@chrislconover OK after some discussion we are changing direction and letting this in.

In practice, we've never been using the delegate properties of the session, and not allowing the user to pass in their own session leads to all kinds of problems, including making background sessions really hard to use, making mocking a total pain in the butt, and making it impossible to either certificate pin or use a self-signed certificate. 

Can you please fix the merge conflicts and then we can (finally) get this shipped?

designatednerd:
@chrislconover If I don't hear from you by tomorrow, I'll close this out and start a separate PR with this functionality. Thanks!

curiousapplications:
Please do whatever is easiest for you.  I don't need or want credit or copyright, and hereby revoke any claims in perpetuity to any and all of my changes to the Apollo codebase.  Go nuts!

-c

PR #266, opened 2018-04-28
Added GraphQLContext - and used for response header parameters

knutaa:
Proposed implementation for #223 based on the suggested use of Context. 

Context is added to GraphQLResponse and carried over to GraphQLResult and there available to the client. HTTPNetworkTransport insert the context directly when creating the GraphQLResponse entity.

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -100,7 +100,7 @@ public class HTTPNetworkTransport: NetworkTransport {
         guard let body =  try self.serializationFormat.deserialize(data: data) as? JSONObject else {
           throw GraphQLHTTPResponseError(body: data, response: httpResponse, kind: .invalidResponse)
         }
-        let response = GraphQLResponse(operation: operation, body: body)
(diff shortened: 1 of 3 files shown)


Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -100,7 +100,7 @@ public class HTTPNetworkTransport: NetworkTransport {
         guard let body =  try self.serializationFormat.deserialize(data: data) as? JSONObject else {
           throw GraphQLHTTPResponseError(body: data, response: httpResponse, kind: .invalidResponse)
         }
-        let response = GraphQLResponse(operation: operation, body: body)
+        let response = GraphQLResponse(operation: operation, body: body, context: httpResponse.allHeaderFields)
```
martijnwalraven: I don't think we should use the headers as the context directly. We still need the ability to pass `context` from callers (as an optional argument to `fetch`) as well. So at a minimum this should be nested under `headers`. 

But I still like the ability of a delegate method to set/augment the context better, because it allows you to abstract from transport-specific notions like headers, and you can centralize interpreting these headers.

PR #297, opened 2018-06-26
Add HTTPNetworkTransportDelegate and HTTPNetworkTransportInterception

brennantaylor:
Add HTTPNetworkTransportDelegate and HTTPNetworkTransportInterception

So that mutating headers and other inspection can be done by the delegate. This
means being able to change headers without recreating clients or logging
request/response timing at a high level by instrumenting the calls to the
interception. Functionality this should give callers adequate control over the
http pipeline.

By using an `Interception` to encapsulate state it would be easy to implement a
timer for a request/response pair. By flattening it into the delegate itself
that information would be lost without using a unique id to tie them together.

Alternatively one could implement http chaining like seen in okhttp which would
give a cleaner contract in terms of showing that only one call to
`Chain.proceed` can perform an interception but is ultimately more complicated.

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -41,11 +41,24 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
   }
 }
 
+/// Used to intercept the http requests/errors/responses. This allows modifying url requests dynamically without recreating transports
+public protocol HTTPNetworkTransportDelegate: class {
+  func intercept() -> HTTPNetworkTransportInterception
+}
+
+/// An encapsulation of state for a single request (response/error) pair. If any method returns non-null it will be used in the http pipeline
+public protocol HTTPNetworkTransportInterception {
+  func willSend(request: URLRequest) -> URLRequest?
+  func didRecieve(error: Error) -> Error?
+  func didRecieve(response: HTTPURLResponse) -> HTTPURLResponse?
+}
+
 /// A network transport that uses HTTP POST requests to send GraphQL operations to a server, and that uses `URLSession` as the networking implementation.
 public class HTTPNetworkTransport: NetworkTransport {
   let url: URL
   let session: URLSession
   let serializationFormat = JSONSerializationFormat.self
+  public weak var delegate: HTTPNetworkTransportDelegate?
   
   /// Creates a network transport with the specified server URL and session configuration.
   ///
@@ -75,15 +88,25 @@ public class HTTPNetworkTransport: NetworkTransport {
 
     let body = requestBody(for: operation)
     request.httpBody = try! serializationFormat.serialize(value: body)
-    
+    let interception = delegate?.intercept()
+    if let interceptedRequest = interception?.willSend(request: request) {
+        request = interceptedRequest
+    }
     let task = session.dataTask(with: request) { (data: Data?, response: URLResponse?, error: Error?) in
-      if error != nil {
+      if var error = error {
+        if let interceptedError = interception?.didRecieve(error: error) {
+          error = interceptedError
+        }
         completionHandler(nil, error)
         return
       }
       
-      guard let httpResponse = response as? HTTPURLResponse else {
+      guard var httpResponse = response as? HTTPURLResponse else {
         fatalError("Response should be an HTTPURLResponse")
+      }
+
+      if let interceptedResponse = interception?.didRecieve(response: httpResponse) {
+        httpResponse = interceptedResponse
       }
       
       if (!httpResponse.isSuccessful) {


MarkPerkins:
WOW!. this is awesome

martijnwalraven:
@brennantaylor I would really like us to get a solution for this over the finish line. Thanks for opening this PR. One thing missing from this solution is support for asynchronous callbacks. I think that is a hard requirement to implement use cases like retries on errors and token refresh (see the [Alamofire RequestRetrier](https://github.com/Alamofire/Alamofire/blob/master/Documentation/Alamofire%204.0%20Migration%20Guide.md#request-retrier) for an example of this).

There is [an existing PR](https://github.com/apollographql/apollo-ios/pull/257) that deals with this. Maybe we can continue the discussion there to keep everyone involved.

Definitely open to exploring an interception mechanism as an alternative to a delegate as well. We have something similar in JavaScript ([Apollo Link](https://www.apollographql.com/docs/link/)), and that has proven to be a very fruitful extension mechanism. There is [an existing PR](https://github.com/apollographql/apollo-ios/pull/186) that tried to explore this.

RyanCodes:
@brennantaylor I would love for this to make it into master soon! This is exactly what i need :)

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -41,11 +41,24 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
   }
 }
 
+/// Used to intercept the http requests/errors/responses. This allows modifying url requests dynamically without recreating transports
+public protocol HTTPNetworkTransportDelegate: class {
```
wtrocki: I'm using this directly on the application level to verify if this works for various use cases.
Question: what is the reason for this protocol - it looks like the only responsibility is to create interception for each request.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -41,11 +41,24 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
...
+  func intercept() -> HTTPNetworkTransportInterception
+}
+
+/// An encapsulation of state for a single request (response/error) pair. If any method returns non-null it will be used in the http pipeline
+public protocol HTTPNetworkTransportInterception {
+  func willSend(request: URLRequest) -> URLRequest?
```
wtrocki: When using willSend I can access request fields but I cannot add headers (request is immutable object)

wtrocki:
I have done some investigations with this by trying to write Alamofire version of the NetworkTransport. 
From my point of view users should be able to use their own network framework if they need - which is possible for long time on Android by supplying okhttp implementation with all interceptors etc.

What is currently blocking users from IOS is that essential parts that are required to write your own NetworkTransport are not available. Some elements used in apollo network transport are internal - means that when writing your own version of it it will not be possible to reuse them. 

IMHO easy fix will be to make  change `private func requestBody`[1] to protected (fields and classes used inside are internal and cannot be used outside library. When this is done everyone will be free to write their own network transport and community can catch up with Alamofire NetworkTransport etc.

This PR provides some level of flexibility, but allowing users to write their own NetworkTransport will be ultimate fix that will support all use cases.

[1] https://github.com/apollographql/apollo-ios/blob/2bd859d483fe8b6fa01a602f0315f5f09b0894a9/Sources/Apollo/HTTPNetworkTransport.swift#L140-L149

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -41,11 +41,24 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
...
+  func didRecieve(error: Error) -> Error?
+  func didRecieve(response: HTTPURLResponse) -> HTTPURLResponse?
+}
+
 /// A network transport that uses HTTP POST requests to send GraphQL operations to a server, and that uses `URLSession` as the networking implementation.
 public class HTTPNetworkTransport: NetworkTransport {
   let url: URL
   let session: URLSession
   let serializationFormat = JSONSerializationFormat.self
+  public weak var delegate: HTTPNetworkTransportDelegate?
```
wtrocki: I might miss the usage for this, but when testing this I needed to drop `weak` keyword to make everything work. 
This keyword may be not needed here (although I might miss some context that author had when doing this)

naknut:
Hi! Whats the status on this request? I would really need this feature for my use case. Whats blocking this from getting approved?

wtrocki:
@naknut There are couple PR's currently open with similar approaches. 
This will work if you bring this files into your project.

Sources/Apollo/HTTPNetworkTransport.swift
davidwellsmith: Delegates should be weak to avoid reference cycles. As recommended in the SwiftLint [rules](https://github.com/realm/SwiftLint/blob/83a2ca4be7fc2c1d94ae99baa9409ad9f91e84c1/Rules.md#weak-delegate).

tsorencraig:
Any update on this? @wtrocki @brennantaylor @naknut Lots of open issues and PRs for similar functionality, so would be great to get something merged in

wtrocki:
@theadactyl There seem to be no interest in core maintainers to get the solution for this knowing that there is more than couple PR's solving the same issue. 
Community build fully functional https://github.com/maxdesiatov/ApolloAlamofire 

This just does the job and adding it to documentation will be quite handy as many people asking for it.

designatednerd:
Hey @brennantaylor - Thank you for doing this, and sorry for the long radio silence on our end!

PR #299, opened 2018-06-26
Adding public init to GraphQLHTTPResponseError to allow to easier use in other projects

liamjdouglas:
Currently it's not possible to create an GraphQLHTTPResponseError outside the Apollo framework meaning testing for expected responses from Apollo is not possible


liamjdouglas:
Updated

liamjdouglas:
@martijnwalraven CI looks to be failing with bootstrapping errors. We've had those in the past mainly when a framework is being used but not imported where it's used. Not sure that's much help though

PR #304, opened 2018-06-28
Use Apollo CLI for codegen

shadaj:
Switches to the Apollo CLI to generate Swift types for queries.


martijnwalraven:
Great, thanks!

PR #306, opened 2018-06-29
Fix folder case name

js:
The following folders exist on disk:
```
Sources/ApolloWebSocket/ # capital S
Sources/ApolloWebsocket/  # lowercase s
```
and is reflected in the [Github UI](https://github.com/apollographql/apollo-ios/tree/7e90a9640dda767e464f5e86b771fc731303a05c/Sources)

Since macOS by default has a case-insensitive file system this confuses tools like Git and Xcode.


martijnwalraven:
Thanks!

PR #309, opened 2018-07-03
fix #308

### LINKED ISSUE #308: Can't subscribe via WebSocket after server send 'ka' message right after 'connection_ack'
If after ‘connection_ack’ server sends me ‘ka’ message, my messages can’t be sent (-> subscription can't be made), because in method `write(_ str: String, force forced: Bool = false, id : Int? = nil)` field ‘acked’ of class `WebSocketTransport` will never be ‘true’. My though is:  ‘ka’ message shouldn’t reset `acked` property.

In `WebSocketTransport` class:
So here is why `acked` field resets to "false" after "ka" message arrived 
````
 case OperationMessage.Types.CONNECTION_ACK.rawValue,
             OperationMessage.Types.CONNECTION_KEEP_ALIVE.rawValue:
          
          acked = (type == OperationMessage.Types.CONNECTION_ACK.rawValue)
````

And then we never can't make `websocket.write()` here

````
private func write(_ str: String, force forced: Bool = false, id : Int? = nil) {
    print("WebSocket: write: forced \(forced) acked \(acked) \n\(str) ")
    if let websocket = websocket {
      if websocket.isConnected && (acked || forced) {
        websocket.write(string: str)
      } else {
        // using sequence number to make sure that the queue is processed correctly
        // either using the earlier assigned id or with the next higher key
        if let id = id {
          queue[id] = str
        } else if let id = queue.keys.max() {
          queue[id+1] = str
        } else {
          queue[1] = str
        }
      }
    }
  }
````

My thoughts is doing like so instead of first code chunk:

````
        case OperationMessage.Types.CONNECTION_ACK.rawValue:
          acked = true
          writeQueue()

        case OperationMessage.Types.CONNECTION_KEEP_ALIVE.rawValue:
          writeQueue()
````

- [ x] has-reproduction
- [ ] feature
- [ ] docs
- [x] blocking
- [ ] good first issue

psi-gh:
fix #308

- [ ] feature
- [ x] blocking
- [ ] docs

### DIFF
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -116,12 +116,13 @@ public class WebSocketTransport: NetworkTransport, WebSocketDelegate {
             notifyErrorAllHandlers(WebSocketError(payload: payload, error: error, kind: .unprocessedMessage(text)))
           }
           
-        case OperationMessage.Types.CONNECTION_ACK.rawValue,
-             OperationMessage.Types.CONNECTION_KEEP_ALIVE.rawValue:
-          
-          acked = (type == OperationMessage.Types.CONNECTION_ACK.rawValue)
+        case OperationMessage.Types.CONNECTION_ACK.rawValue:
+          acked = true
           writeQueue()
-          
+
+        case OperationMessage.Types.CONNECTION_KEEP_ALIVE.rawValue:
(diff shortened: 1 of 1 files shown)


knutaa:
This issue is also handled in #307. As #307 is addressing additional issues, it may be relevant to proceed with #307 and close this PR.

PR #310, opened 2018-07-05
ApolloWebSockets code style, 

js:
Since #307 is a bit unwieldy here's one with the stylistic and initializer changes, so they can be bikeshedded separately.

Stylistic changes:
 - Consistent use of spacing around variable type declarations
 - Switch on enum cases instead of their raw value
 - Avoid heavy indentation by using guard

I found the Optionality of the websocket property really odd, when would we ever want a nil websocket on a websocket transport class?
I removed the `connect(request:)` since a better API would be to supply a new WebSocketTransport if you need to call another endpoint, instead of mutating the underlying request potentially mid-flight. 

One thing it did provide was the option to "initialize now, but connect later" and I briefly considered not calling `websocket.connect()` in the initializer, in favour of requiring the caller to issue a `connect()` later, but since 99% will probably either call connect() immediately after initialising or forgetting it all-together I decided to just always immediately connect, side-effects be damned

Header configurations (confusingly named `params` in the removed initializers) has been removed, since the caller should instead pass in an URLRequest configured to their needs. That's not our job.

### DIFF
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -2,7 +2,7 @@ import Apollo
 import Starscream
 
 // To allow for alternative implementations supporting the same WebSocketClient protocol
-public class ApolloWebSocket : WebSocket, ApolloWebSocketClient {
+public class ApolloWebSocket: WebSocket, ApolloWebSocketClient {
   required public convenience init(request: URLRequest, protocols: [String]? = nil) {
     self.init(request: request, protocols: protocols, stream: FoundationStream())
   }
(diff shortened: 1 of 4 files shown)


knutaa:
The original intention was to allow for initialisation like the following

let httpTransport = HTTPNetworkTransport(url: ...)
let websocketTransport = WebSocketTransport()
let networkTransport = SplitNetworkTransport(httpNetworkTransport: httpTransport,
                                             webSocketNetworkTransport: websocketTransport)
var apollo = ApolloClient(networkTransport: networkTransport)
// at this point, the websocket part of the split network transport is not connected
// some operation using the http transport
// setting the request based on received information
let request = URLRequest(...)
// and complete the websocket set-up
websocketTransport.connect(request: request)

Suggest that this capability is retained.

js:
Why not just make a new ApolloClient instead of mutating the WebSocketTransport? Which is the pattern the HTTPNetworkTransport uses, since it leads to more resilient and predictable code since it can't exist without having the url (request) properly initialized.

knutaa:
That can be done assuming the store is passed between the ApolloClients.

knutaa:
@martijnwalraven Suggest to accept, it will break any implementations using delayed connect, but is possible to handle with the available features.

PR #315, opened 2018-07-13
Made clear where and how to use the commands

MrMatten:
I was unaware of the apollo-cli package and as the text didn't state it was a command line prompt, I used a little time figuring out how and where to use them


martijnwalraven:
Thanks!

PR #319, opened 2018-07-17
Add operation id in header

gridNA:
/label feature

In our team we have a use case, when we need to send operationID in header with every query request. Because  we can set up headers only during the `ApolloClient` init, we can't send operationID header dynamically. With my pr I've added operationID (if exists) in header for every request. The same logic is currently implemented in Android client.


designatednerd:
@gridNA does the work in #602 to allow header modifications cover this use case?

designatednerd:
This has been implemented in #658 (with SHOUTY HEADER NAMES to match Android) and shipped in 0.13.0. Gonna close this one out, thanks for your work!

PR #341, opened 2018-08-04
 Fixed "Ignoring import" warning with Cocoapods

Camji55:


cocojoe:
@martijnwalraven worth adding a `pod lint` test phase in CI? (Although this can be rather slow)

martijnwalraven:
@cocojoe Yeah, I'd be afraid of slowing builds down even more. And we'd still run `pod lint` as part of publishing a new release (we've been running it with `--allow-warnings` so far, but fortunately can stop doing that now).

PR #350, opened 2018-08-15
Allow spaces in Project name

bretep:
Fixes npm install with a project prefix that contains spaces.


bretep:
The build failed for unrelated reasons.

PR #361, opened 2018-08-31
Invalidate items in cache

ashiemke:
Define a delete cache function for ApolloCache and implement it for InMemoryNormalizedCache and SQLiteNormalizedCache, along with tests to ensure watchers are fired and items are removed from the cache and all places where they are referenced.

### DIFF
--- a/ApolloSQLite.xcodeproj/xcshareddata/xcschemes/ApolloSQLite.xcscheme
+++ b/ApolloSQLite.xcodeproj/xcshareddata/xcschemes/ApolloSQLite.xcscheme
@@ -26,7 +26,6 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      disableMainThreadChecker = "YES"
       shouldUseLaunchSchemeArgsEnv = "NO">
       <Testables>
          <TestableReference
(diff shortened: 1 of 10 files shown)


ashiemke:
Addresses #362

larryonoff:
It would be great adding method `deleteAll`

designatednerd:
@ashiemke Thanks for doing this - I'm not sure if we updated Starscream or if you did, but if you can get that to the same point as the main fork, this looks like it'd be quite helpful. 

I agree with @larryonoff that a `deleteAll` method would also be great, but I'm fine doing that on my end once this is merged.

designatednerd:
@ashiemke Are you interested in continuing to work on this PR?

designatednerd:
Since I haven't heard back in a couple weeks on this one, I'm going to close it out. @ashiemke if you change your mind and decide to work on this again, please feel free to reopen. 

Cache invalidation is always a pain in the ass, and honestly I have to deal with some other stuff before I get to it, so any help we can get on this is most appreciated!

mdarnall:
@designatednerd Hi, curious if this PR is something I can re-open as it's functionality we need going forward. I believe the "deleteAll" is there now as `clear`.

designatednerd:
@mdarnall Again any help we can get here would be great. This approach may be a bit more complex than is actually necessary, but if you think it's a helpful starting point, go right ahead.

mdarnall:
@designatednerd I agree - in talking with our Android team it sounds like apollo-android doesn't go to the trouble of removing the references to the deleted cacheKey, but that also when a subsequent query is executed, rows that have invalid references are ignored as opposed to what I see on iOS where an `Apollo.JSONDecodingError.missingValue` is returned. 

If that seems reasonable to follow that same pattern I can look into matching that behavior.

PR #363, opened 2018-09-04
Make operationDefinition on GraphQLOperation static.

bclymer:
# What

This makes `operationDefinition` on `GraphQLOperation` `static`. 

# Why

Since it is already a constant, there are a couple reasons why making it `static` is nice.

1. It's not allocated each time the class is instantiated.
1. It allows accessing the string outside of the context of an instance.

The latter is the main reason I'm making this PR. We were previously using `requestString` from the older `apollo-codegen` to detect changes in the query in order to invalidate local caches. Ignoring the fact that we could have a better way of determining that, I think this property _should_ be `static` anyway. 

# Problems

This requires a change in apollo-cli as well. Here's my PR -- https://github.com/apollographql/apollo-cli/pull/579

### DIFF
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -7,7 +7,7 @@ public enum GraphQLOperationType {
 public protocol GraphQLOperation: class {
   var operationType: GraphQLOperationType { get }
   
-  var operationDefinition: String { get }
+  static var operationDefinition: String { get }
   var operationIdentifier: String? { get }
   
   var queryDocument: String { get }
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -19,7 +19,7 @@ public protocol GraphQLOperation: class {
 
 public extension GraphQLOperation {
   var queryDocument: String {
-    return operationDefinition
+    return type(of: self).operationDefinition
   }
 
   var operationIdentifier: String? {


bclymer:
Tests are failing as expected since I'm not pointing to my CLI PR. My only thought is changing [this line](https://github.com/apollographql/apollo-ios/blob/master/.travis.yml#L56) to `npm install -g bclymer/apollo-cli`, but given that's a mono-repo I don't think that will work. Any ideas on how to get CI to test this correctly?

martijnwalraven:
Thanks for bringing this up. `operationString` and `requestString` actually used to be `static`, but with the rename to `operationDefinition` and `queryDocument` I made them instance properties to keep open the possibility of using dynamic operation instances (the idea then was that we'd want to be able to make parts of your app written in React Native use the same cache, returning JSON instead of query-specific generated types).

I'm not against reversing that change, but it doesn't really make sense to only make `operationDefinition` `static` since that doesn't contain the fragments it may depend on. (Presumably it doesn't serve your use case either, because changes in fragment definitions can change the query without the operation definition changing.) 

If we make this change, I think it should apply to all operation-level properties, so `operationType`, `operationDefinition`, `queryDocument`, and `operationIdentifier`. You'd also have to change all the uses of these properties.

Curious to hear what you mean by 'using `requestString` from the older `apollo-codegen` to detect changes in the query in order to invalidate local caches' however. If you're talking about the built-in normalized cache, there should be no need for that. And if you're using your own cache I'd be interested in the reasoning behind that.

bclymer:
That's a good point, depending on where this discussion leads I'll go and make `queryDocument` and `operationIdentifier` `static` as well. 

I think we could have just used the `SQLite` subspec in order to achieve our caching needs, but here's our use case and the logic behind using our own cache:

* We have a long list of user specific items that don't change often (outside of new items being added).
* We wanted to be able to load the most recent 20 items instantly on a new app launch.

Normally we'd cache recent content in Core Data (our app supports a pretty heavy amount of offline mutations, so Apollo's cache didn't work great for most use cases), however given the overhead of making new Core Data model versions, and the expected rate of us changing the requested fields from this specific endpoint, we decided to store the `snapshot` (now `resultMap`) from the first page response. The next time the user launches the app, if the `requestString` is identical to the old `requestString`, we will populate that first page immediately and then make a network request to load the up-to-date first page.

If there is future work planned around these being instance properties like you mentioned, the workaround for us is not bad; just create an empty instance of the Operation (using empty strings, empty arrays, etc) in order to access the definition. I don't foresee us using this caching pattern much in the future either.

designatednerd:
@bclymer Are you interested in continuing the work in this PR, or should I close it out?

bclymer:
@designatednerd feel free to close it out, along with the associated PR here - https://github.com/apollographql/apollo-tooling/pull/579

I can always come back it to later and open a fresh PR.

designatednerd:
Thanks! I'll close both.

PR #365, opened 2018-09-11
Cache load errors cause deadlock in write transaction

ashiemke:
Reject batch loader loads if a loader fails to allow exceptions to cascade through promise chain. Previous behavior (run test without code) causes the load to never complete, and more importantly, causes the store to hold a read lock indefinitely, causing a deadlock when a write is attempted later. In practice, this is every time a fetch occurs with default cache policy.

The root of the issue is a non-normalized dictionary which is stored in the cache. In our schema, we have some JSON blobs that come from an external service which are represented as a JSON custom scalar. Storing these in the cache worked fine, but fetching these types caused attempts to read from cache to never return, and other strange behavior due to writes also failing.

Solution was pretty simple, just needed to pass the failures in the batch loader to the individual load promises so they can be picked up and forwarded on the chain. Since these are wrapped in a `whenAll`, only the first `reject` matters.

### DIFF
--- a/Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
+++ b/Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
@@ -210,6 +210,37 @@ class LoadQueryFromStoreTests: XCTestCase {
     }
   }
   
+  
+  func testLoadingWithBadCacheSerialization() throws {
+    let initialRecords: RecordSet = [
+      "QUERY_ROOT": ["hero": Reference(key: "2001")],
+      "2001": [
+        "name": "R2-D2",
+        "__typename": "Droid",
+        "friends": [
+          Reference(key: "1000"),
+          Reference(key: "1002"),
+          Reference(key: "1003")
+        ]
+      ],
+      "1000": ["__typename": "Human", "name": "Luke Skywalker", "badField":
+        ["dictionary": "badValues", "nested bad val": ["subdictionary": "some value"] ]
+      ],
+      "1002": ["__typename": "Human", "name": "Han Solo"],
+      "1003": ["__typename": "Human", "name": "Leia Organa"],
+      ]
+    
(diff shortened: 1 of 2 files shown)


ashiemke:
Looks like the star scream dependency is failing for some reason, not sure what the root issue is.

ashiemke:
One travis failure, which passes on my local. Unsure why it's breaking. Can someone kick it? I suspect it's a timing-related failure.

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -210,6 +210,36 @@ class LoadQueryFromStoreTests: XCTestCase {
...
+      "1003": ["__typename": "Human", "name": "Leia Organa"],
+      ]
+    
+    withCache(initialRecords: initialRecords) { (cache) in
+      store = ApolloStore(cache: cache)
+      
+      let query = HeroAndFriendsNamesQuery()
+      load(query: query) { (result, error) in
+        XCTAssertNil(result)
+        XCTAssertNotNil(error)
```
paulmelnikow: Updating this to 

```swift
XCTAssertTrue(error is GraphQLResultError)
XCTAssertTrue((error as! GraphQLResultError).underlying is JSONDecodingError)
```

provides a bit more confidence about what is happening here.

designatednerd: I'd use `guard let` instead of force unwrapping because force unwrapping crashes the whole test suite instead of just failing the test, but I agree here

designatednerd:
@ashiemke are you interested in continuing this PR?

designatednerd:
@ashiemke Since I haven't heard back from ya in a while, I'm going to take this over so we can try to get it merged.

designatednerd:
Closing this one out, please follow #763 for an attempted resolution here.

PR #370, opened 2018-09-21
Update dataloader dispatch to propagate errors

dshusta-coursera:
Seeing some SQLiteNormalizedCacheErrors that get swallowed in DataLoader and no-op without recovery. Add handling for rejected promises so they get propagated up to the ApolloStore and can fall back to network fetch.


dshusta-coursera:
Oh, maybe going to defer to https://github.com/apollographql/apollo-ios/pull/365

PR #381, opened 2018-10-05
Adding swift_version to the podspec

sirghi:
When trying to package a library that's dependent on Apollo using cocoapods-packager, the script fails saying it doesn't specify a SWIFT_VERSION. I reckon it's proper behaviour in general to specify the version in the podspec.


sirghi:
Anyway feedback on this? I see the CI failed because of a stale simulator not booting up.

Issue #225, opened 2018-03-13
Code-generation fails with spaces in target name


eliotfowler:
**Installation method:** Cocoapods
**Version:** 0.7.0

After installing the Apollo pod, I built my project against the iPhone 7 simulator. The project built successfully. I then added the "Run Script" build phase pasted from the [documentation](https://www.apollographql.com/docs/ios/installation.html#adding-build-step). The project fails on this build step every time now.

**Expected Result**: Successful build

**Observed result**: Failed build

**Details:**

Here is the error output from the find command:
```find: -: unknown primary or operator```

One thing of note (and my best guess of the issue): my $FRAMEWORK_SEARCH_PATHS is both large *and* has directories with spaces in it. **Notably: my target is "JAM Debug - Staging" which shows up in various directory names.**

Here is a dump of the variable. Important to note that I'm simply using the inherited FRAMEWORK_SEARCH_PATHS from Cocoapods. Nothing custom:

```
[...]

PR #387, opened 2018-10-17
Fix Apollo Build Step script because it fails when Schemes have space in their names

Valpertui:
When there is spaces in any path of the FRAMEWORK_SEARCH_PATHS, the `find` command fails.
Having a scheme with a space in its name will cause FRAMEWORK_SEARCH_PATHS to have spaces in each path.

This edit changes the `find` separation character to `\n` from ` ` through `IFS` and fixes the FRAMEWORK_SEARCH_PATHS to include double quotes on its first path (which is without double quotes even if it has a space in the path). Not adding the double quotes also breaks `find` when it reads a path with a space if `IFS` is set

- [x] blocking
- [x] docs


Valpertui:
It's also a fix for this issue https://github.com/apollographql/apollo-ios/issues/225

ricsantos:
Would be good to get this merged, @Valpertui thanks for making it, can you sign the agreement so the Apollo team can merge it?

Valpertui:
@ricsantos I did sign it back in october, but I signed it again today.

designatednerd:
I'm going to close this one out in favor of #610 - there was some stuff that also needed to be fixed in the docs and with CocoaPods, but your work on that command to get everything working fine with spaces in project and scheme names was super-helpful with that. Thanks so much!

PR #390, opened 2018-10-18
export createRequest method for extending the transport

koenpunt:
To make even the most trivial adjustment to the request, like being able to dynamically specify the headers, you have to copy the full implementation of the transport class.

By exposing the `createRequest` method, this would be possible by simply extending the existing transport, and only override the `createRequest` method.


designatednerd:
Hey @koenpunt  - Thank you for doing this, and sorry for the long radio silence on our end!

Issue #357, opened 2018-08-24
How to perform cache manipulation after mutations / subscriptions?


Nickersoft:
As a longtime user of Apollo's React / JS-based client, I know that a pretty common practice when performing mutations or subscriptions to update Apollo's internal cache so any updates can propagate throughout the rest of the app. This practice is heavily described in the docs on [optimistic UI responses](https://www.apollographql.com/docs/react/features/optimistic-ui.html) and [subscriptions](https://www.apollographql.com/docs/react/advanced/subscriptions.html#subscribe-to-more). For example, in JS this might look like:

```javascript
            update: (proxy, { data: { submitComment } }) => {
              // Read the data from our cache for this query.
              const data = proxy.readQuery({ query: CommentAppQuery });
              // Add our comment from the mutation to the end.
              data.comments.push(submitComment);
              // Write our data back to the cache.
              proxy.writeQuery({ query: CommentAppQuery, data });
            }
``` 

I know questions about the cache have been asked here before, but there do not seem to be any examples, recommendations, docs, etc. on how you might be able to achieve this kind of manipulation.

I use ReSwift in my app and have considered passing in a custom store object, and am wondering if this might be the only way. Without ReSwift, any modification I make to an external cache object won't affect the client cache because Apollo [copies the init parameter to an internal variable](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloClient.swift#L53). 

This seems like a pretty instrumental concept in the Apollo ecosystem... am I missing something?

ashiemke:
You can use ApolloStore's WithinReadWriteTansaction stuff to modify the contents of the cache. There is not good documentation right now, but you can probably find what you need by reading tests. See ReadWriteStoreTests.swift.

Nickersoft:
[...]

PR #394, opened 2018-10-22
Fix writing a query with variables to the store.

kevbuchanan:
This issue is also discussed in https://github.com/apollographql/apollo-ios/issues/357

The ReadWriteTransaction's [resolver](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloStore.swift#L148) expects the cache key to be in the object data. When writing to the store, the data will not already have this. The existing tests for this didn't include a query with variables, which would cause the lookup to fail.

I'm not sure if the below is the best way to fix this. Seems like either the data needs to have the cache key added, or the write needs to use a different resolver that just uses the field name.

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -143,9 +143,10 @@ public final class ApolloStore {
 
     fileprivate lazy var loader: DataLoader<CacheKey, Record?> = DataLoader(self.cache.loadRecords)
 
-    fileprivate func makeExecutor() -> GraphQLExecutor {
+    fileprivate func makeExecutor(keySelector: @escaping (GraphQLResolveInfo) -> String = { $0.cacheKeyForField }) -> GraphQLExecutor {
       let executor = GraphQLExecutor { object, info in
-        let value = object[info.cacheKeyForField]
+        let key = keySelector(info)
+        let value = object[key]
         return self.complete(value: value)
       }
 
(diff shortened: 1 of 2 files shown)


kevbuchanan:
On further testing of this, it looks like this is an issue at any level of the response, so just updating the top level keys won't work. Using a `responseKeyForField` on writes seems to work. Not sure how to best write a unit test for this case though, I don't see any obvious queries in `StarWarsAPI` to use.

kevbuchanan:
@Nickersoft Not sure what else needs to be done here. As far as making the `store` variable public from `ApolloClient`, I'll leave that for others to implement. I don't have any strong opinions on that as you can currently create and reference the store prior to creating the `ApolloClient`.

Nickersoft:
I suppose my concern there @kevbuchanan is just the fact I wasn't sure that if you defined a store, passed it into the ApolloClient constructor, then called a write-to-cache method on the store variable, if it would actually update the Apollo cache. From what I saw [the constructor copies to the passed store variable](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloClient.swift#L53) to a member variable by-value, which would mean you'd have to call the write cache method on the class member directly. That is, unless I'm mistaken somewhere. I am relatively new to Swift, and am not 100% on how the cache is handled and propagated internally.

rad182:
please merge this 👍

Nickersoft:
Can anyone make sense of why the build might be failing? I can't find a solid error message

Nickersoft:
@kevbuchanan @rad182 Created a separate PR that exposes the methods needed for writing to the store and features this fix: #413. This will ideally let me fix any build issues as they arise and hopefully push this feature through.

kevbuchanan:
Closing in favor of https://github.com/apollographql/apollo-ios/pull/413

Issue #262, opened 2018-04-13
Trying to parse Long-values, getting error: "Could not convert to String"


sfla:
Using Apollo 0.8.0.
The API I'm querying has a few values represented as Long. Since "there's no such thing" in Swift, I was expecting it to be parsed into `Int64` or simply `Int`. However, when running the request, the query returns with a `GraphQLResultError` saying this:
`GraphQLResultError(path: ["path", "to", "value"], underlying: Apollo.JSONDecodingError.couldNotConvert(value: 123, to: Swift.String))`
It looks like it attempts to parse the Long-value as a String instead of an Int or any other number.
This seems quite severe to me, but I'm pretty new to the entire GraphQL-thing. Is Long not supposed to be a value? Could it be that the creators of this service has injected some values for their own purpose?
If this is the case, is there any way for me to add certain "exceptions" or "improvements" on my own?

martijnwalraven:
`Long` is what is known as a custom scalar, there is no such type built-in to GraphQL. Unfortunately, GraphQL currently doesn't have a way of specifying how these custom scalars should be encoded/decoded. Without further configuration, Apollo iOS assumes `String`. 

You can use the `--passthrough-custom-scalars` option to `apollo-codegen` to leave custom scalars as is. That means you will need to define something like `typealias Long = Int64` to avoid compiler errors.

sfla:
@martijnwalraven Thank you! It solved my issue. When using `--passthrough-custom-scalars` I also noticed I needed a few other typealiases to make it work. For example, the data contained two types for date-handling. One `Date` and one `DateTime`. Simply adding a `typealias DateTime = Date` solved it, though I now decode them both equally. It's not a problem for my implementation, but strictly speaking, they probably should have separate `jsonValue`-implementations, because they're now also encoded identically.
What's strange is that before using `--passthrough-custom-scalars`, I didn't have to do this with `Date` and `DateTime`, it was handled automatically.

Issue #218, opened 2018-03-02
GraphQLResultError(path: ["idp", "authenticate", "userId"], underlying: Apollo.JSONDecodingError.couldNotConvert(value: 27, to: Swift.String))


duydtdev:
I got an error when call a request to get data below
```swift 
GraphQLResultError(path: ["idp", "authenticate", "userId"], underlying: Apollo.JSONDecodingError.couldNotConvert(value: 27, to: Swift.String))
```

This is my model and query:
```swift 
query Authenticate ($username: String!, $method: String!, $credentials: [String!]!) {
  idp {
    authenticate(username: $username, method: $method, credentials: $credentials) {
      userId
      organizations { id name }
      featureLevel
      token
    }
  }
}
```

```swift
type Authenticate {
    token: String
    userId: Long
    role: Role!
    organizations: [OrganizationInfo!]
    featureLevel: Int
    data: [ CustomerCredentials ]
}

type CustomerCredentials {
    organization: String!
    token: String!
}
```
Could you help tell me why? Many thanks

martijnwalraven:
`userId` is defined as `Long`, which is not a built-in type but a custom scalar. So it is treated as `String` by default. See [this issue](https://github.com/apollographql/apollo-ios/issues/23) for a discussion and solution.

designatednerd:
This is fixed by #402 and will ship with `0.14.0`

designatednerd:
Shipped with `0.14.0`!

Issue #94, opened 2017-06-06
How to define the rules for type convert


jzhw0130:
@martijnwalraven 

Server define a type 'Date', the format is '2017-06-06' or 1496736244000, (string or int).
Generate the scheme.json by apollo-codegen download-schema, the type is 'Date', too

Generate the API.swift by apollo-codegen generate , the type convert to 'String' automatically. Then data decode will throw exception when server format is int(1496736244000).  Because it can not convert 'Int' to 'String'..

Now I have to change the API.swift manually, from 'String' to 'Int'. But it is not good,  I have to do it many times when scheme changes...

Could you tell me the best way to solve the problem?

Thanks a lot

Jing

cerupcat:
Did you find a solution for this? I'm also looking how to automatically create Date types during apollo codegen.

martijnwalraven:
@cerupcat @jzhw0130: You should be able to pass `--passthrough-custom-scalars` to `apollo-codegen` to avoid generating a `typealias` to `String` for custom scalars. You can then add your own alias and conversion code, see [here](https://github.com/apollographql/apollo-ios/issues/23#issuecomment-261697349) for an example.

cerupcat:
Awesome, thanks! Will take a look.

jzhw0130:
@cerupcat  I just modify the scheme.json manually to solve the problem

cerupcat:
Thanks @jzhw0130. I don't think that would work for us since we share the schema file across platforms and it gets auto generated when we make changes. Will be giving the other solution a try when we get back to that part.

shehang:
@jzhw0130 & @cerupcat Do you guys found a way to solve this?
I tried modifying the scheme.json file but it doesn't work.
Having trouble Decoding & Encoding DateTime type to Date in typealias.

Any help is appreciated.

designatednerd:
I've (finally) added some actual documentation and an example for this with #1474 - would appreciate any feedback.

Issue #23, opened 2016-11-18
Incorrect code generation: String instead of Int


nikola-mladenovic:
Steps to reproduce:

- Use the schema from: https://github.com/commercetools/commercetools-sunrise-ios/blob/issues/%2394-product-graphql/Sunrise/schema.json
- The only query used was: https://github.com/commercetools/commercetools-sunrise-ios/blob/issues/%2394-product-graphql/Sunrise/Models/Queries/ProductOverview.graphql

The resulting `API.swift` that was generated using `apollo-ios` version 0.4 and `apollo-codegen` version 0.9 contains: `public let centAmount: String` (`MoneyDetails` struct property). The actual type from the schema:
```
{
"name": "centAmount",
"type": {
  "kind": "NON_NULL",
  "name": null,
  "ofType": {
    "kind": "SCALAR",
    "name": "Long",
    "ofType": null
  }
}
```

martijnwalraven:
According to the [GraphQL spec](http://facebook.github.io/graphql/#sec-Scalars), `Long` is not a built-in scalar, so it is treated as a custom scalar. Right now in Apollo iOS, all custom scalars are mapped to string (which is also how they are encoded in JSON).

I would definitely like to add better support for custom scalars, which should give you the opportunity to define your own mapping.

But in this case, do you really need `Long` here, or would `Int` suffice?

martijnwalraven:
> thanks @martijnwalraven! judging by the nature of `centAmount` field, I think the value can be pretty high in some cases
> the current workaround could be manually altering the local schema file; is there anything better you’d suggest?

I hope to add better support for custom scalars soon, but in the meantime there is a temporary `apollo-codegen` option `--passthrough-custom-scalars` that doesn't attempt to map custom scalars to string but keeps them as is. That means you're responsible for defining the type and making sure it is `JSONDecodable` and `JSONEncodable`.

I haven't tested it, but something like this should work:
```swift
typealias Long = Int64

extension Int64: JSONDecodable, JSONEncodable {
  public init(jsonValue value: JSONValue) throws {
    guard let string = value as? String else {
      throw JSONDecodingError.couldNotConvert(value: value, to: String.self)
    }
    guard let number = Int64(string) else {
      throw JSONDecodingError.couldNotConvert(value: value, to: Int64.self)
    }
    
    self = number
  }
  
  public var jsonValue: JSONValue {
    return String(self)
  }
}
```

AnandWalvekar:
[...]

PR #402, opened 2018-11-06
Cast Int as String if the jsonValue is of type Int. (Custom Scalars better compatibility)

cadizjavier:
The support of Apollo for custom scalars is quite simple.
Referring to this [issue](https://github.com/apollographql/apollo-ios/issues/23#issuecomment-261624608) all custom scalars are mapped to String.
The recommended option for this cases is try to adjust the GraphQL backend for using the `built-in` scalars or just pass the `--passthroughCustomScalars` flag at the `codegen` tool.
Passing the flag is a really good option but it is super annoying when the Backend defines dozens of custom types that potentially conflicts with the app types.
The solution is to use the `built-in` scalars and just expect that everything custom returns as a `String` and do the parsing in the Data layer.

The main issue with this is that the Apollo parser breaks when the custom scalar is of type `Int` since it fails to do the casting to `String`.
This happens for example when a custom type called `Timestamp` represents the value as `1540993726`. Then the Apollo parser treats it as `String` but fails to do the casting since isn't a `String` really, it comes in the `JSON` as `Int`.

This PR solves this. I adjusted all the tests to pass but they need some check if I did something weird there, specially the `testGetScalarListWithWrongType` one since I don't know if I used the proper approach there.
This can be potentially used for Bools too if necessary.

As a side note this kinda helps to fix the following issues without using the `--passthroughCustomScalars` flag  and then parse the data to the proper type in the Model layer or something.
- [23](https://github.com/apollographql/apollo-ios/issues/23) Open
- [262](https://github.com/apollographql/apollo-ios/issues/262) Closed
- [218](https://github.com/apollographql/apollo-ios/issues/218) Open
- [94](https://github.com/apollographql/apollo-ios/issues/94) Open. Kinda handles this one too since Date is a custom scalar and regarding the returned value it will parse everything to String.


almassapargali:
Hey @martijnwalraven, can you take a look at this PR, we're having this issue as well, and looks like it's pretty common.

Sources/Apollo/JSONStandardTypeConversions.swift
```diff
@@ -2,10 +2,14 @@ import Foundation
 
 extension String: JSONDecodable, JSONEncodable {
   public init(jsonValue value: JSONValue) throws {
-    guard let string = value as? String else {
-      throw JSONDecodingError.couldNotConvert(value: value, to: String.self)
+    switch value {
+    case let string as String:
+        self = string
+    case let int as Int:
```
designatednerd: Is it the `case let` that allows this to only need `as` instead of `as?`?

Tests/ApolloTests/ReadFieldValueTests.swift
```diff
@@ -105,19 +99,13 @@ class ReadFieldValueTests: XCTestCase {
...
-    XCTAssertThrowsError(try readFieldValue(field, from: object)) { (error) in
-      if let error = error as? GraphQLResultError, case JSONDecodingError.couldNotConvert(let value, let expectedType) = error.underlying {
-        XCTAssertEqual(error.path, ["name"])
-        XCTAssertEqual(value as? Int, 10)
-        XCTAssertTrue(expectedType == String.self)
-      } else {
-        XCTFail("Unexpected error: \(error)")
-      }
-    }
+    let value = try readFieldValue(field, from: object) as! String
```
designatednerd: I would strongly recommend against force-unwrapping here as instead of failing the test, it crashes the whole test suite. Usually I do something like 

```swift
guard let foo = bar as? String else {
   XCTFail("Wrong type!")
   return 
}

XCTAssertEqual(foo, "expectedFoo")
```

Tests/ApolloTests/ParseQueryResponseTests.swift
```diff
@@ -46,15 +46,9 @@ class ParseQueryResponseTests: XCTestCase {
...
-        XCTAssertEqual(error.path, ["hero", "name"])
-        XCTAssertEqual(value as? Int, 10)
-        XCTAssertTrue(expectedType == String.self)
-      } else {
-        XCTFail("Unexpected error: \(error)")
-      }
-    }
+    let (result, _) = try response.parseResult().await()
+
+    XCTAssertEqual(result.data?.hero?.name, "10")
```
designatednerd: Why shouldn't this throw an error when it's not a string? `name` isn't using a custom scalar type

designatednerd:
@cadizjavier are you interested in continuing this PR?

Sources/Apollo/JSONStandardTypeConversions.swift
cadizjavier: Yes, instead of throw if the value isn't string I want to have better granularity when managing the type. If the value is a `String` keep it as it, if is an `Int` cast it to `String` because is a custom scalar and we want to parse it properly in our app (specific case of the Timestamp described in the PR description). This of course could be extended to other basic types, like Doubles, Bool, etc.

Tests/ApolloTests/ParseQueryResponseTests.swift
cadizjavier: You are absolutely right. I will check this test case again and come with a better solution.

Tests/ApolloTests/ReadFieldValueTests.swift
cadizjavier: Totally agree. Force unwrapping here is a non go. Your solution sounds very clean to me. I will update this code too.

cadizjavier:
Yes, sorry for the long delay. I provided feedback and context to the comments and will update the PR this week. Thanks for taking over this amazing project !!

Sources/Apollo/JSONStandardTypeConversions.swift
designatednerd: Yeah i think for now let's keep it to `String` and `Int` - my recollection is that we recommend explicitly against doing that for Bool scalars. In practice `Int` is really the only one I've seen people requesting.

designatednerd:
Hooray! Thank YOU!

cadizjavier:
Updated PR with your feedback and some improvements in the testing side.
Test cases now have a differentiation between a different type and a wrong type. 
A `testGetScalarWithDifferentType` test, for example, handle the cases where I pass an `Int` and the parser returns a `String` as expected and a `testGetScalarWithWrongType` test just works as expected the old way (a type that can't be cast to the specified type).

There is actually a kinda confusing test which you commented previously regarding
`XCTAssertEqual(result.data?.hero?.name, "10")` where you mention why we don't throw if isn't a `String`.
I'm not exactly sure what to do here. It occurs to me that we can check the original type and see if it isn't a custom scalar type and act accordingly.
I don't know if we have this kind of introspection in the defined types or we can't do better at the implementation level.
Technically speaking I think it is correct to cast as String something that comes as Int with the change made in the PR but I can see pros and cons of both arguments.
What do you think about it ?

designatednerd:
Yeah - after having spent the last week diving into some codegen and typesystem stuff, we can punt on handling the "Well is this a custom scalar or should it have been a string in the first place?" thing until codegen is in swift. I'll make a follow-up issue for it.

Tests/ApolloTests/ParseQueryResponseTests.swift
designatednerd: Punting on this until later - we'll figure it out at some point.

Tests/ApolloTests/ReadFieldValueTests.swift
```diff
@@ -105,14 +119,28 @@ class ReadFieldValueTests: XCTestCase {
     XCTAssertNil(value)
   }
   
-  func testGetOptionalScalarWithWrongType() throws {
+  func testGetOptionalScalarWithDifferentType() throws {
     let object: JSONObject = ["name": 10]
     let field = GraphQLField("name", type: .scalar(String.self))
-    
+
+    XCTAssertNoThrow(try readFieldValue(field, from: object))
```
designatednerd: I think this is redundant since the test `throws` - if this were to throw here or where you're getting the value, the whole test would fail

Tests/ApolloTests/ReadFieldValueTests.swift
```diff
@@ -59,14 +59,28 @@ class ReadFieldValueTests: XCTestCase {
     }
   }
   
-  func testGetScalarWithWrongType() throws {
+  func testGetScalarWithDifferentType() throws {
     let object: JSONObject = ["name": 10]
     let field = GraphQLField("name", type: .nonNull(.scalar(String.self)))
-    
+
+    XCTAssertNoThrow(try readFieldValue(field, from: object))
```
designatednerd: Samesies on the redundancy of no throw

Tests/ApolloTests/ParseQueryResponseTests.swift
```diff
@@ -46,10 +46,26 @@ class ParseQueryResponseTests: XCTestCase {
       ]
     ])
 
+    XCTAssertNoThrow(try response.parseResult().await())
```
designatednerd: Samesies on the redundancy

designatednerd:
I think just fixing the test redundancies and this should be 👍

designatednerd:
Eh, I'll remove the redundant checks myself, I'd like to get this out with `0.14.0` 😃

cadizjavier:
Wohooo, excited that this gets approved. Thanks for the feedback and being supportive.

PR #422, opened 2018-12-12
Exposed a method to ping

delannoyk:
Our server-side Websocket implementation forces us to ping every 30seconds for the connection to stay alive.

This change just exposes a ping method on a `WebSocketTransport` instance.


designatednerd:
Thanks @delannoyk!

PR #424, opened 2018-12-17
Make `field` and `alias` on *GraphQLField* public

CalQL8ed-K-OS:
I'd like to utilize these fields in order to facilitate automatic conversion back to an auto-generated structs JSON representation

### DIFF
--- a/Sources/Apollo/GraphQLSelectionSet.swift
+++ b/Sources/Apollo/GraphQLSelectionSet.swift
@@ -31,8 +31,8 @@ public protocol GraphQLSelection {
 }
 
 public struct GraphQLField: GraphQLSelection {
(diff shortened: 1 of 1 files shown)


designatednerd:
Hi @CalQL8ed-K-OS - I'm not entirely sure I follow on the use case - can you explain in a bit more detail? There might be some stuff we've already got that could help solve this without making these public, but I'm not sure without more detail.

CalQL8ed-K-OS:
This was for an old job, so I don't much care for this PR any longer personally 😛
I'll close it since it's no longer my responsibility 🤷🏽‍♂️

designatednerd:
Haha, fair enough! Thanks for closing it out.

PR #427, opened 2018-12-29
Update for compatibility with Swift5

cysp:
- Remove an unused subscript implementation on Dictionary that triggers a compiler crash
- Tweak the type of an iterator


JaviSoto:
👍 

Can we get this merged? Xcode 10.2 Beta is out with Swift 5 now.

PR #428, opened 2019-01-08
Add file uploads [RFC] [WIP]

JustinDSN:
### Add Upload Support to Apollo-iOS
* [GraphQL Multipart Request Spec](https://github.com/jaydenseric/graphql-multipart-request-spec) by @jaydenseric.

### Depends on:
- https://github.com/apollographql/starwars-server/pull/9

### Summary of Changes:
- Adds a new protocol `GraphQLUpload` and `typealias` for `Upload`.
- Adds new `struct`s for `DataGraphQLUpload`, `FileGraphQLUpload`, `InputStreamGraphQLUpload` to allow clients flexibility to supply an uploads.
- Adds a new protocol `GraphQLUploadOperation` which inherits from `GraphQLOperation`.
- Adds a new protocol `GraphQLUploadMutation` which inherits from `GraphQLUploadOperation`.
- Adds a public function to `ApolloClient` to `perform` `GraphQLUploadMutation`s
-  *Changes* generic constraints on `Operation` type on `NetworkTransport` protocol.
- Adds a new requirement to `send<Operation: GraphQLUploadOperation>` on `NetworkTransport` protocol.
- Added implementation for `send<Operation: GraphQLUploadOperation>` in `HTTPNetworkTransport`.
- Adds AlamoFire's `MultipartForm`, `AFError`, and `HTTPHeaders` to provide robust MultipartFrom encoding.

### TODO:
- [] Change apollo-cli to generate `GraphQLUploadMutations` correctly.
- [] Implement multipart encoding memory thresholds to either write to disk or stream from memory.


Sources/Apollo/GraphQLUpload.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public protocol GraphQLUpload: JSONEncodable {
```
MrAlek: Maybe rename this to `GraphQLFile`? Since conceptually, one "upload" can contain multiple files?

Sources/Apollo/GraphQLUpload.swift
```diff
@@ -0,0 +1,26 @@
...
+  var data: Data { get }
```
MrAlek: This property should preferably be a stream instead of actual bytes. For large upload, you don't want to load the entire file into memory at once.

You mentioned you have had a look at https://github.com/Alamofire/Alamofire/blob/master/Source/MultipartFormData.swift ?

Sources/Apollo/GraphQLUpload.swift
jaydenseric: I don't really understand the Swift implementation, but regarding naming I recommend maintaining consistency with other file upload implementations. The scalar is called `GraphQLUpload` because technically it is not a file, but a promise of a file being uploaded that might, or might not, finish uploading.

Sources/Apollo/AFError.swift
```diff
@@ -0,0 +1,222 @@
...
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+/// `AFError` is the error type returned by Alamofire. It encompasses a few different types of errors, each with
+/// their own associated reasons.
+///
+/// - multipartEncodingFailed:     Returned when some step in the multipart encoding process fails.
+public enum AFError: Error {
```
martijnwalraven: Hmmm, exposing `AFError` seems a bit confusing because we're not depending on AlamoFire. Maybe we could rework this to expose a `MultipartEncodingError`?

Sources/Apollo/ApolloClient.swift
```diff
@@ -123,6 +123,21 @@ public class ApolloClient {
...
   }
 
+  /// Performs a mutation by sending it to the server.
+  ///
+  /// - Parameters:
+  ///   - mutation: The mutation to perform.
+  ///   - queue: A dispatch queue on which the result handler will be called. Defaults to the main queue.
+  ///   - resultHandler: An optional closure that is called when mutation results are available or when an error occurs.
+  /// - Returns: An object that can be used to cancel an in progress mutation.
+  @discardableResult public func perform<Mutation: GraphQLUploadMutation>(mutation: Mutation, queue: DispatchQueue = DispatchQueue.main, resultHandler: OperationResultHandler<Mutation>? = nil) -> Cancellable {
```
martijnwalraven: I think we should be able to avoid duplicating `perform` and related methods by removing `GraphQLUploadMutation` and relying on protocol composition instead.

Sources/Apollo/ApolloClient.swift
```diff
@@ -165,6 +179,37 @@ public class ApolloClient {
       }
     }
   }
+
+  fileprivate func send<Operation: GraphQLUploadOperation>(operation: Operation, context: UnsafeMutableRawPointer?, handlerQueue: DispatchQueue, resultHandler: OperationResultHandler<Operation>?) -> Cancellable {
```
martijnwalraven: Same here, I don't think we need to duplicate this.

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -41,6 +57,11 @@ public extension GraphQLMutation {
   var operationType: GraphQLOperationType { return .mutation }
 }
 
+public protocol GraphQLUploadMutation: GraphQLUploadOperation {}
```
martijnwalraven: I don't think there is a need for `GraphQLUploadOperation`.

victormihaita:
Any updates when this PR will be available for merging. It seems that the community, as well as me, has a big interest into this :)

victormihaita:
Hi guys. Do you have by any chance any updates for this PR?

Sources/Apollo/AFError.swift
JustinDSN: Hi @martijnwalraven I renamed the error and other references to Alamofire.  Although, I've kept the original copyright message from the files.

JustinDSN:
Hi @victormihaita, I'm sorry for the delay, but this PR took a back burner for a little while.  

I've addressed the comments left by @martijnwalraven, but it's been a while since we've collaborated on this, so we'll have to see if we can get this off the shelves.

@martijnwalraven from my perspective heres the checklist:
- [] Review and merge https://github.com/apollographql/starwars-server/pull/10, this will allow the unit-tests I've written in this PR to pass.
- [] Change apollo-cli to generate `GraphQLUploadMutations` correctly as a separate PR.  Can someone from the Apollo team help with this?
- [] Implement multipart encoding memory thresholds to either write to disk or stream from memory as a separate PR.

What do you think?  Do you think we should approach it this way or some other way.  I'm also very curious how many members of the community are looking for this?

victormihaita:
This feature is something that we are really looking forward to have it. Right now, we had to create a separate REST Endpoint in our API just for uploading images (while the rest of it is based on GraphQL), and this was made just for the iOS client... I consider it as a really important feature for a networking service such as Apollo... it would be great if we can see any progress in this direction :) @martijnwalraven  @JustinDSN

designatednerd:
Hey @JustinDSN - thanks for your hard work on this, and apologies this was left fallow for so long. Do you have any thoughts on rebasing this vs moving forward with the simpler implementation in #626?

designatednerd:
I've gone ahead and merged #626 - I'm going to close this one out, but please feel free to open a new PR if there's anything that's missing from the implementation there!

Issue #423, opened 2018-12-13
After updating to Apollo 0.9.4, CircleCI can't find ApolloCLI


adamirowe:
I recently updated to Apollo 0.9.4 via Cocoa Pods, which works great locally, but when building project via fastlane on CircleCI I'm getting output that Apollo iOS can't find 1.9.x of Apollo CLI.

<img width="1065" alt="screen shot 2018-12-13 at 11 40 59 am" src="https://user-images.githubusercontent.com/43765296/49953368-fe1e1280-fecb-11e8-9e6d-d9eb2d999a6e.png">

paulmelnikow:
I'm running into this warning locally as well.

stNamco:
Did you set an output file path? if you set that, generating script is executed before compile source.
https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/XcodeBuildSystem/200-Build_Phases/bs_build_phases.html#//apple_ref/doc/uid/TP40002690-BBCCFEDJ

paulmelnikow:
> I'm running into this warning locally as well.

I should clarify that I see this warning all the time though my build works fine.

structAnkit:
I just updated from Apollo 0.8.x to 0.9.5 yesterday and started running into this issue.

When I run `fastlane` and don't have Apollo CLI installed (when I run my build phases on a fresh environment such as Bitrise CI), the `npm install` step doesn't run synchronously so the build phase is basically skipped. Since the Apollo CLI is not being installed, the codegen isn't running and the generated file doesn't exist.

designatednerd:
Hey @adamirowe - were you ever able to resolve this issue? I know some folks had success running a pre-build step that used `npm install - g` with our CLI client on Circle.

adamirowe:
> Hey @adamirowe - were you ever able to resolve this issue? I know some folks had success running a pre-build step that used `npm install - g` with our CLI client on Circle.

Yes but I do not remember how honestly. Let me know if you need anything else from me.

Issue #383, opened 2018-10-16
graphql apollo fails to run script


mkamal13:
i followed the tutorial provided by apollo and each time i build to run the script that generate the api file ,i get the following massage:

iphonesimulator/Apollo/Apollo.framework/check-and-run-apollo-cli.sh: No such file or directory
Command PhaseScriptExecution failed with a nonzero exit code
this is the project:https://github.com/mkamal13/graphqlapollo can any body help?

mkamal13:
no

azablotskiy:
pod update should do the trick ;)

moger777:
I get same issue with 0.9.4 but only when archiving.

stNamco:
Did you set an output file path? if you set that, generating script is executed before compile source.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/XcodeBuildSystem/200-Build_Phases/bs_build_phases.html#//apple_ref/doc/uid/TP40002690-BBCCFEDJ

designatednerd:
Please see the updated codegen script setup instructions [here](https://www.apollographql.com/docs/ios/installation/#adding-a-code-generation-build-step) - these have been updated in 0.11.0 and 0.11.1. Gonna close this one out!

PR #431, opened 2019-01-11
add output files path to sample project script

stNamco:
# Agenda
add output files path to sample project script. it is needed for executing a script before compile sources.
because Xcode10 new build system is Parallelize build.

# Detail
another library fixed this issue like this.
https://github.com/mac-cain13/R.swift/issues/456

and this is apple document.(but this is old)

> Tasks that contain inputs that are the outputs of other tasks or outputs that are the inputs of other tasks are ordered tasks. These are tasks that are executed in the order determined by their inputs and outputs.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/XcodeBuildSystem/200-Build_Phases/bs_build_phases.html#//apple_ref/doc/uid/TP40002690-BBCCFEDJ

maybe this is related to these issues.
https://github.com/apollographql/apollo-ios/issues/423
https://github.com/apollographql/apollo-ios/issues/383

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -1010,10 +1010,11 @@
 			files = (
 			);
 			inputPaths = (
+				"$TEMP_DIR/apollo-lastrun",
 			);
 			name = "Generate Apollo Client API";
 			outputPaths = (
-				"$(SRCROOT)/Tests/GitHubAPI/API.swift",
+				$SRCROOT/Tests/GitHubAPI/API.swift,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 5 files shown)


paulmelnikow:
Another article on this issue: https://medium.com/zendesk-engineering/xcode-10s-build-system-and-code-generation-r-swift-f8d520e28875

Apollo.xcodeproj/project.pbxproj
```diff
@@ -1010,27 +1010,31 @@
 			files = (
 			);
 			inputPaths = (
+				"$TEMP_DIR/apollo-lastrun",
```
paulmelnikow: Out of curiosity, what does this do?

stNamco: sorry, I misunderstood about input file path. so this is not needed.

PR #434, opened 2019-01-14
Actually install the version of node we want via NVM/nodenv instead of just checking

bachand:
👋 

I am working on updating Airbnb to the latest version of Apollo. I was running into issues with this script.

The issue I faced was that sourcing `~/.nvm/nvm.sh` was causing Node to be reset to the version specified in my `.nvmrc`. The version in there is 6.x.x, which is necessary for other tooling. I couldn't find a way to set Node to 8.x.x outside of this script since sourcing `~/.nvm/nvm.sh` would continually reset us back to 6.x.x.

Since we have access to these version managers, let's just set up the version that we need so that the script can be run regardless of what Node versions you have installed. Making this change allowed me to remove the code that checks that the Node version is correct.

The one downside that I can see of this change is that if you had a version like 8.1.0 installed and active, the old script would have run fine since it was just checking the major version. As of this change, we will install and activate the exact version. I think there are pros/cons to each approach. The con of the new approach is that you may have to install a version of node that is very similar to one that you already have. The pro everyone is always running the same version of Node, which makes it easier to debug issues 🤷‍♂️ 

cc: @ljharb for NVM changes

### DIFF
--- a/scripts/check-and-run-apollo-cli.sh
+++ b/scripts/check-and-run-apollo-cli.sh
@@ -1,7 +1,7 @@
 # Only major and minor version should be specified here
 REQUIRED_APOLLO_CLI_VERSION=1.9
-# Specify fully qualified version here
-REQUIRED_NODE_VERSION=8.0.0
+# Specify fully qualified version here. Ideally this should be a LTS version.
+REQUIRED_NODE_VERSION=8.15.0
 
 # Using npx to execute 'apollo' looks for a local install in node_modules before checking $PATH (for a global install)
 APOLLO_CLI="npx --no-install apollo"
--- a/scripts/check-and-run-apollo-cli.sh
+++ b/scripts/check-and-run-apollo-cli.sh
@@ -1,7 +1,7 @@
 # Only major and minor version should be specified here
 REQUIRED_APOLLO_CLI_VERSION=1.9
-# Only major version should be specified here
-REQUIRED_NODE_VERSION=8
+# Specify fully qualified version here
+REQUIRED_NODE_VERSION=8.0.0
 
 # Using npx to execute 'apollo' looks for a local install in node_modules before checking $PATH (for a global install)
 APOLLO_CLI="npx --no-install apollo"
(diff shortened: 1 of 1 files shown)


scripts/check-and-run-apollo-cli.sh
```diff
@@ -1,7 +1,7 @@
 # Only major and minor version should be specified here
 REQUIRED_APOLLO_CLI_VERSION=1.9
-# Only major version should be specified here
-REQUIRED_NODE_VERSION=8
+# Specify fully qualified version here
+REQUIRED_NODE_VERSION=8.0.0
```
ljharb: I think sticking with the major is probably better; especially considering that 8.0.0 is EOL and only the latest 8 (8.15.0 at this writing) is considered LTS.

martijnwalraven:
@bachand Thanks for looking into this! I think using these version managers to install the required Node version makes a lot of sense. I'll let you address @ljharb's feedback first, but happy to merge this whenever you're ready.

scripts/check-and-run-apollo-cli.sh
bachand: Ah good context. Thanks @ljharb. Nodenv was requiring my to specify a fully qualified version. I’ll update this to be 8.15.0 since that is LTS.

scripts/check-and-run-apollo-cli.sh
```diff
@@ -17,18 +17,33 @@ if [[ -z "$CONFIGURATION" ]]; then
...
 if [[ -s "$HOME/.nvm/nvm.sh" ]]; then
   . "$HOME/.nvm/nvm.sh"
+  use_correct_node_via_nvm
 elif [[ -x "$(command -v brew)" && -s "$(brew --prefix nvm)/nvm.sh" ]]; then
   . "$(brew --prefix nvm)/nvm.sh"
+  use_correct_node_via_nvm
 fi
 
 # Set up the nodenv node version manager if present
 if [[ -x "$HOME/.nodenv/bin/nodenv" ]]; then
```
bachand: Full transparency: I just installed a clean version of Nodenv and this file doesn't exist, so I am wondering if this check is incorrect. That being said, I've tested the commands here for installing Node 8.15.0 via Nodenv and they work. Determining if in fact there is an issue in this check should be a separate investigation / PR.

bachand:
Thanks for the quick review @martijnwalraven. I have made the change suggested by @ljharb and signed the CLA. This is ready to be merged (from my perspective).

I look forward to working more closely together on Apollo moving forward :)

martijnwalraven:
Thanks, looks good to me! Also looking forward to working together.

bachand:
Thanks for merging @martijnwalraven. Do you have a sense of when you'll be creating a new version?

martijnwalraven:
@bachand I just released version 0.9.5 with these changes.

bachand:
Thanks 🙏

PR #437, opened 2019-01-15
Fixes #435 Invoking Query.Data(jsonObject:) fails for queries with params …

### LINKED ISSUE #435: Invoking Query.Data(jsonObject:) fails for queries with params
Following #51, I am trying to create a GraphQLSelectionSet via init(jsonObject: JSONObject)

The query I'm testing has a non-optional parameter 'Episode'

query Hero($episode: Episode, $withFriends: Boolean!) {
    hero(episode: $episode) {
        name
        friends @include(if: $withFriends) {
            name
        }
    }
}

and to JsonObject I'm passing is 

{
    "data": {
        "hero": {
            "__typename": "Droid",
            "name": "R2-D2",
            "friends": [{
                "__typename": "Human",
                "name": "Luke Skywalker"
            }, {
                "__typename": "Human",
                "name": "Han Solo"
            }, {
                "__typename": "Human",
                "name": "Leia Organa"
            }]
        }
    }
}

The call fails in GraphQLExecutor.swift https://github.com/apollographql/apollo-ios/blob/0928061a06147bf973a921d424a4d871c7c30bf0/Sources/Apollo/GraphQLExecutor.swift#L165

The error is "Variable withFriends was not provided". The variable info: GraphQLResolveInfo was never populated because the init method calls executor.execute without specifying the variables field, which is nil by default.

Should we skip validation of input parameters when calling this method? If so, which would be the recommended way?

Thanks,
Diego

diegorozen:
https://github.com/apollographql/apollo-ios/issues/435

Adds an optional 'variables' parameter to pass to the executor


diegorozen:
@martijnwalraven Added a new test group. It includes testing a conditional query with and without params as well as a simpler query built from JsonObject. We could repeat all the tests from ParseQueryResponseTests to run through Query.Data(jsonObject:), but it seemed like an overkill

martijnwalraven:
Thanks, this looks great!

PR #452, opened 2019-02-02
Added documentation for `read` and `update` operations

Nickersoft:
Gave a very brief overview of the cache read/update functionality that was just added just to prove its existence. Does not really contain a deep-dive into how the cache is internally managed and propagated like the React docs do, so this section I feel can definitely be expanded. This is really just to get something on the site.

### DIFF
--- a/docs/source/watching-queries.md
+++ b/docs/source/watching-queries.md
@@ -25,3 +25,43 @@ apollo.cacheKeyForObject = { $0["id"] }
 ```
 
 > In some cases, just using `cacheKeyForObject` is not enough for your application UI to update correctly. For example, if you want to add something to a list of objects without refetching the entire list, or if there are some objects that to which you can't assign an object identifier, Apollo cannot automatically update existing queries for you.
+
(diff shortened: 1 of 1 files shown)


martijnwalraven:
Thanks!

AnthonyMDev:
@Nickersoft Just read through #413 and the context around that. Thanks so much for making this fix. 
I'm looking for a little clarification, and the docs don't seem to answer my question here.

I'm not sure when to use the manual writing to the cache. If I want to do optimistic updates, then I would use this, but if I am implementing caching properly, it seems to me that writing directly to the cache to update after mutations isn't necessary.

From the docs:

>  This functionality is useful when performing mutations or receiving subscription data, as you should always update the local cache to ensure consistency with the operation that was just performed. 

If I'm properly implementing `cacheKeyForObject` on my `ApolloClient`, and my mutation includes fields to return all of the fields that would be mutated by the mutation, wouldn't my cache already be automatically updated when the mutation finishes?

PR #458, opened 2019-02-15
Update ApolloSQLite project to Swift 4.0

JosephDuffy:
I noticed that there was an update that enables support for Xcode 10.2 (with the Swift 5 compiler) but this did not work for Carthage because the project itself was still using Swift 3.0.

This PR updates the project to use Swift 4.0

### DIFF
--- a/ApolloSQLite.xcodeproj/project.pbxproj
+++ b/ApolloSQLite.xcodeproj/project.pbxproj
@@ -560,7 +560,7 @@
 				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos";
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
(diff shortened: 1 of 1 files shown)


JosephDuffy:
CI failed a single test on a single configuration, which feels a little random and not due to my changes. Can CI be re-run? Here's the job that failed: https://travis-ci.org/apollographql/apollo-ios/jobs/493816613

naknut:
I think this needs to be merged ASAP, from Xcode 10.2 and forward Swift 3 is no longer supported and Swift 4 is the lowest supported version. That means that without this merge, all Carthage builds will fail from today and forward (given you use the latests tools). Ive also opened my own [PR](https://github.com/apollographql/apollo-ios/pull/484) for this but I changed the Swift compiler to Swift 5 instead. From my side it doesn't really matter if this PR or mine gets merged since both will work with Xcode 10.2. I just want my Carthage builds to not fail.

martijnwalraven:
Closing this in favor of #475.

PR #475, opened 2019-03-21
Revise and update workspace build configuration

tonyarnold:
I have been having trouble getting various pieces of Apollo to build for macOS since I started using it, and from what I can see, it all comes down to how the build has been configured. 

This PR aims to do a few things:

1. Extract all build settings from the `.pbxproj` files so that they can be shared between the various projects and targets in the workspace.
2. Add a script to ensure no new build settings are added to the `pbxproj` files.
3. Ensure that (where appropriate) all projects build for iOS, macOS, tvOS and watchOS properly.
4. Bumps Swift support to version 4.2 (tested with both `swiftc` versions 4.2 and 5.0).

I know there's a fair bit in this PR, but it should simplify maintenance.

This fixes the following issues:

- #330
- #380 
- #386
- #418 
- Probably a host of others relating to Carthage, Swift and Xcode versions

PR #484, opened 2019-03-26
Changed swift versions to 5 for ApolloSQLite

naknut:
If you try to build the current master with Carthage using Xcode 10.2 you get the following error:

`error: SWIFT_VERSION '3.0' is unsupported, supported versions are: 4.0, 4.2, 5.0. (in target 'ApolloSQLite')`

This its because Swift 3 is unsupported from Xcode 10.2 and forward. Lowest supported version is 4.0. After this change we compile all the schemes using the Swift 5 compiler so Carthage won't fail any more.

### DIFF
--- a/ApolloSQLite.xcodeproj/project.pbxproj
+++ b/ApolloSQLite.xcodeproj/project.pbxproj
@@ -493,7 +493,6 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.apollographql.ApolloSQLiteIOSTests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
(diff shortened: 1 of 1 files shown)


jimisaacs:
@naknut hopefully those test failures are not caused by the build change :/
Our team would love this change to get in.
Do you have any Carthage work arounds in the mean time?

naknut:
@jimisaacs I think it’s related. It seems like Travis doesn’t have Swift 5 support yet. That’s why it fails. I guess you can point to my version your Cartfile in the meantime. You can find my updates and patches version here: https://github.com/naknut/apollo-ios

jimisaacs:
@naknut Great suggestion, I'll give that a shot a workaround.

jimisaacs:
@naknut I think I'd prefer if the version could stay 4.2 for the time being. I'll fork myself if I can get my build working without requiring 10.2

martijnwalraven:
Closing this in favor of #475.

Issue #294, opened 2018-06-25
operationName not set when doing fetch request


MarkMolina:
After generating code based on the downloaded schema I receive query's like:

```
public static let operationDefinition =
    "query QueryName($parameters: [String]!) {" +
    "  foo(bar: $lorem) {" +
    "    __typename" +
    "    x" +
    "    y" +
    "  }" +
    "}"
```

I can then easily tell the client to do a fetch request and pass an instance of this query. It works fine but for some reason the QueryName (which is the operationName right?) is not set in the request. Using Charles Proxy I found that the iOS SDK is also not setting any operationName on the request and therefore our request are marked as "unknown" in new relic. 

If I inspect our javascript client on our website it does actually add an operationName to the request. This is done automatically. Is there a way to also do this for the iOS SDK?

Using Apollo iOS version 0.5.6.

martijnwalraven:
Since [`operationName` is only required when multiple operations are included in a single query document](https://graphql.org/learn/serving-over-http/#post-request), and Apollo iOS never sends multiple operations, we don't currently set this. But it would be pretty easy to add to `HTTPNetworkTransport` if you have a need for it (like logging in New Relic).

MarkMolina:
Thanks Martijn. Do you know how I can easily retrieve the query name from the generated code? I see that I can easily change the send method to include an operationName but am not sure how to retrieve the matching name that goes with that specific request.

Maybe from the `GraphQLOperation` parameter?

martijnwalraven:
Ah, you're right, we don't currently generate an `operationName` on `GraphQLOperation` either. We could definitely do that, but it does require changes to codegen. It's a bit of a hack, but for now you might be able to derive it from `operationDefinition` (looking at `operationType` to cut off `query` or `mutation` and know where to find the name in the string).

MarkMolina:
Right, I'll implement it like suggested for our logging purposes for now and keep an eye on any updates. Maybe I'll dive into the code as well and see I can help out with this.

DarkDust:
According to [this document](https://graphql.org/learn/queries/): _"It is only required in multi-operation documents, but its use is encouraged because it is very helpful for debugging and server-side logging."_ 

We'd like to use them for subscriptions so support for them would be welcome. Does the note in [the draft section 5.2.3.1](https://facebook.github.io/graphql/draft/#sec-Single-root-field) mandate to send the operation name or am I misunderstanding this?

DarkDust:
[...]

PR #496, opened 2019-04-01
Send operationName for GraphQLOperations

js:
This PR adds `operationName` to the request payload. operationName is not really required per the graphql spec unless you send multiple operations in the same request. But it is _very_ useful for have when mapping up performance metrics and such so I'd like to have Apollo always send it. 

This PR indirectly relies on [changes](https://github.com/js/apollo-tooling/commit/637907e049c1b648939bd3cebc52f73cec214f85) to the apollo codegen, but I'm currently blocked there due to some silly npm tooling issues before I can verify that the `apollo codegen:generate` part actually works as intended. 

Current behaviour (not sending operationName) will be kept until the generator is updated.

The changes suggested in issue #294 only works if you name your operation definition (? what are these actually called) the same as the graphql operation you're calling, eg this is valid graphQL:
```
query nudgeUser($id: ID!) {
  pingUser(id: $id) { ... }
}
```
but would give an `operationName` of "nudgeUser" instead of the actual operation name "pingUser"

### DIFF
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -6,14 +6,15 @@ public enum GraphQLOperationType {
 
 public protocol GraphQLOperation: class {
   var operationType: GraphQLOperationType { get }
-  
+
   var operationDefinition: String { get }
   var operationIdentifier: String? { get }
-  
+  var operationName: String? { get }
+
   var queryDocument: String { get }
-  
+
   var variables: GraphQLMap? { get }
-  
+
   associatedtype Data: GraphQLSelectionSet
 }
 
@@ -23,6 +24,10 @@ public extension GraphQLOperation {
   }
 
   var operationIdentifier: String? {
+    return nil
+  }
+
+  var operationName: String? {
     return nil
   }
 
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -121,12 +121,17 @@ public class HTTPNetworkTransport: NetworkTransport {
   private let sendOperationIdentifiers: Bool
 
   private func requestBody<Operation: GraphQLOperation>(for operation: Operation) -> GraphQLMap {
+    var body: GraphQLMap = ["variables": operation.variables, "operationName": operation.operationName]
+
     if sendOperationIdentifiers {
       guard let operationIdentifier = operation.operationIdentifier else {
         preconditionFailure("To send operation identifiers, Apollo types must be generated with operationIdentifiers")
       }
-      return ["id": operationIdentifier, "variables": operation.variables]
+      body["id"] = operationIdentifier
+    } else {
+      body["query"] = operation.queryDocument
     }
-    return ["query": operation.queryDocument, "variables": operation.variables]
+
+    return body
   }
 }
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -249,13 +249,18 @@ public class WebSocketTransport: NetworkTransport, WebSocketDelegate {
   }
   
   private func requestBody<Operation: GraphQLOperation>(for operation: Operation) -> GraphQLMap {
+    var body: GraphQLMap = ["variables": operation.variables, "operationName": operation.operationName]
+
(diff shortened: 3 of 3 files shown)


martijnwalraven:
> but would give an operationName of "nudgeUser" instead of the actual operation name "pingUser"

Note that this isn't actually correct: the operation name in this case is `nudgeUser`, that is [what operation name means in the GraphQL spec](https://graphql.github.io/graphql-spec/draft/#sec-Language.Operations).

`pingUser` is known as a 'root field', and there can be multiple of these in a single operation. The other side of this is that different operations can include the same root field. So root fields can't be used to uniquely identify operations.

js:
aha, I missed that part, thank you for correcting me! :)

What threw me off-course is that our graphql server ("apollo-server-core": "1.3.6") rejects the requests unless the operationName matches this root field but that must be something else then…

js:
The reason I utterly confused myself when I did as suggested in #294 was because of the inflection the code generator does to operation names, eg: `mutation updateFCMToken(..){..}` will yield an `UpdateFcmTokenMutation` swift type, which the GraphQL backend will reject as the operation name `UpdateFcmToken`!=`updateFCMToken`.

Dispite my initial confusion/incompetence I think it makes sense to add it as a property which the generator can fill in, rather than relying on the typename which comes with a few gotchas like the above.

js:
@martijnwalraven Would love to land this, what do I need to do?

The CI failures for ios9.3 seems to be an unrelated build error:
`xctest (5240) encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted. (Underlying error: Test runner exited before starting test execution.))`

js:
@designatednerd now that it has been added to the codegen (strangely not from my four month old PR but whatever) could you revisit this so that the client actually sends the operation name in the request? Thanks and sorry for at’ing but you seem like the main driver these days

designatednerd:
@js Yeah there's a lot of back and forth on that repo - I'm sorry we missed your PR! You can close it if that's the only thing that was in there, or update it and tag me on it if it wasn't. 

I am going to be looking at the updated codegen stuff tomorrow (got sucked into a rabbit hole today). Will ping you!

designatednerd:
@js OK, switching to an updated version of the CLI which automatically generates operation names has now been merged into `master` - do you want to update this in the next day or so in order to get it shipped with that in `0.13.0`?

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -6,14 +6,15 @@ public enum GraphQLOperationType {
 
 public protocol GraphQLOperation: class {
   var operationType: GraphQLOperationType { get }
-  
+
   var operationDefinition: String { get }
   var operationIdentifier: String? { get }
-  
+  var operationName: String? { get }
```
skevy: Should this be nullable?

designatednerd: I think with the changes in codegen we *should* be able to make this non-null since it's now auto-generated.

designatednerd:
@js Would you have time to get this cleaned up today, or would you prefer I take this over so we can get it shipped with the 0.13.0 changes?

Sources/Apollo/GraphQLOperation.swift
js: I opted for an Optional In case the tooling wasn’t updated (either by the user or within the tooling itself), to soften the dependency

designatednerd: fair enough - we can certainly leave it optional for a bit but I think eventually we should make this required.

js:
I’m on vacation so not able to do much for the next week or two @designatednerd feel free to do what’s needed :)

designatednerd:
Nice! Will do and enjoy your vacation!

Sources/Apollo/GraphQLOperation.swift
designatednerd: heads up: Had to make this non-nullable because protocol resolution stuff made this always return `nil` even if there was an implemented `operationName`. I can give a longer explanation if desired but it's basically [this example of how protocols with default implementations resolve which implementation to use](https://speakerdeck.com/designatednerd/protocols-all-the-way-down-dotswift-paris-january-2018?slide=70) and why that's not always what you expect.

designatednerd:
Closing this one out in favor of #657 - thanks for your patience with all of us @js!

PR #505, opened 2019-04-07
Move response parsing to DispatchQueue

cltnschlosser:
This moves the response parsing onto a DispatchQueue. The queue was already there and looks like it was intended for this purpose. The queue stopped being used in this commit: https://github.com/apollographql/apollo-ios/commit/56f1674d065396ea978c2ec7b5d552ea9c03365a#diff-ceeef11df562c93c0080f63181a19188

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -152,21 +152,18 @@ public class ApolloClient {
         return
       }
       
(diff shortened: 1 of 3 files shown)


cltnschlosser:
@martijnwalraven You seem to be the guy that gets things merged. Looking for feedback on this change.

PR #513, opened 2019-04-17
use firstIndex(where:)

masashi-sutou:
- use `firstIndex(where:)` , because `index(where:)` will be deprecated in Swift 4.2 before being removed in Swift 5.

> In addition, the index(of:) and index(where:) methods will be renamed to firstIndex(of:) and firstIndex(where:), respectively. This change is beneficial for two reasons. First, it groups the first... and last... methods into two symmetric analogous groups, and second, it leaves the index... methods as solely responsible for index manipulation.

> The name changes to index(of:) and index(where:) are easily migratable, and will be deprecated in Swift 4.2 before being removed in Swift 5.

> ref: https://github.com/apple/swift-evolution/blob/master/proposals/0204-add-last-methods.md#detailed-design

Xcode10.2 Swift Conversion | After converting to swift5, Generating preview |
:-: | :-: | 
 |  |

### DIFF
--- a/Sources/Apollo/Collections.swift
+++ b/Sources/Apollo/Collections.swift
@@ -24,7 +24,7 @@ struct GroupedSequence<Key: Equatable, Value> {
   fileprivate var groupsForKeys: [[Value]] = []
   
   mutating func append(value: Value, forKey key: Key) -> (Int, Int) {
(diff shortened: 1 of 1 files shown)


masashi-sutou:
#522 was merged, so I close this issue.

PR #518, opened 2019-04-24
make `-[ApolloStore clearCache]` public

RolandasRazma:
followup to https://github.com/apollographql/apollo-ios/pull/517

make `-[ApolloStore clearCache]` public. 

At the moment there is no way to clear cache unless you have access to `ApolloClient`. Having `clearCache()` on `ApolloClient` is not exactly correct as multiple `ApolloClient` can share same cache.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -68,6 +68,7 @@ public class ApolloClient {
   /// Clears apollo cache
   ///
   /// - Returns: Promise
+  @available(*, deprecated, message: "No longer supported. User .store.clearCache() instead.")
   public func clearCache() -> Promise<Void> {
     return store.clearCache()
   }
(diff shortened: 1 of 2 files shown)


designatednerd:
So I'm definitely in agreement with making `clearCache` on the `Store` public for the reasons you've outlined. However, I don't think we should deprecate `clearCache` on the client - many users will only be using a single client with a single cache, and for that use case leaving it on the client makes sense. 

How about adding a note to the documentation that if you're using the same cache across multiple, you should call the method on the store instead?

RolandasRazma:
It's bit confusing if we leave it...

```swift
apolloClient.clearCache()
// do I need to call apolloClient.store.clearCache() as well?
```

designatednerd:
@RolandasRazma I think if you're only using one client and one cache (which is a fairly good chunk of our users), leaving it as a convenience would probably be better. 

Having it exposed separately is helpful for the people who need it, but I think for people who only need the one client and the one cache, it's easier to leave the existing behavior and expose the underlying `clear` as a benefit for the people who need it.

RolandasRazma:
IMO it brings unnecessary confusion, but it's your call. Removed deprecation.

designatednerd:
I'll update some of the docs once this is merged to try to make it clearer when to use which one - thank you!

Issue #514, opened 2019-04-19
Cancellation logic causing non-determinism when invoking result handler.


coreygrunewald:
Hello,

We are seeing some non-deterministic behavior in a unit test from cancelling a `.fetch()` call with a cache policy of `.returnCacheDataAndFetch`.

Most of the time, our result handler callback is never called. I'm assuming this is because we are hitting this line in the `FetchQueryOperation`: https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloClient.swift#L192.

Occasionally, we see our result handler callback is invoked, most likely hitting this line: https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloClient.swift#L206. Even though the task is in a cancelled state, the result handler is still invoked with a result.

I think the cancellation logic can be improved by:
1. Always invoking the result handler in the case of a task being cancelled. By not invoking the result handler, the caller could be stuck in a waiting state, etc.
2. When the result handler is invoked, if the task is cancelled, pass a specific Error instance, with no result, signaling to the user that the operation was cancelled.

Does this make sense? Am I missing any context around how cancellation should work? Any insight or feedback would be much appreciated, thank you!

michaelnisi:
You should be able to use [Operation’s](https://developer.apple.com/documentation/foundation/operation/1408085-completionblock) `completionBlock: (() -> Void)?` for cleanup, it should get called after the operation is finished, successfully or not.

Another thought. Considering that a cancelled operation doesn’t *really* have a result and cancellation isn’t *really* an error, how about introducing  a new cancellation handler `cancellationHandler: (() -> Void)?` for signalling to interested objects without changing the API? From what I understand from reading the code, always calling the result handler for cancelled operations would be a breaking change.

But it should be deterministic, of course, never calling the result handler when cancelled.

michaelnisi:
Forget the first part, `Cancellable` doesn’t expose `completionBlock`.

designatednerd:
@coreygrunewald Are you still seeing this issue on 0.11.1 or higher? We've added some methods which should be helpful for this, but I'm not positive it'd totally fix the whole issue here. Let me know what happens after you upgrade. Thanks!

designatednerd:
Since I haven't heard back on this for a couple weeks, I'm going to close this issue out. Please open a new issue if you're seeing this on newer versions of our SDK. Thank you!

PR #525, opened 2019-04-27
Test fetch cancelling

michaelnisi:
This adds a test showing how for cancelled fetches, the result handler is only called when `.fetchIgnoringCacheData` is the selected cache policy. Not exactly the same, but related to #514. Does it resolve it?

Although, this PR is suggesting a fix—the flag in line 92 toggles back to the previous state—improving documentation for cancelled fetches would help. But the question remains:

For cancelled fetches, should `resultHandler: ((GraphQLResult<Query.Data>?, Error?) -> Void)?` be called:

1) always with `(nil, nil)`
2) always with something else
3) never
4) sometimes (well documented)

🤔

**Do not merge**  but please discuss.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -8,7 +8,7 @@ public protocol Cancellable: class {
 }
 
 /// A cache policy that specifies whether results should be fetched from the server or loaded from the local cache.
-public enum CachePolicy {
+public enum CachePolicy: CaseIterable {
   /// Return data from the cache if available, else fetch results from the server.
   case returnCacheDataElseFetch
   ///  Always fetch results from the server.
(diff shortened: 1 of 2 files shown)


Sources/Apollo/ApolloClient.swift
```diff
@@ -87,7 +87,9 @@ public class ApolloClient {
   func _fetch<Query: GraphQLQuery>(query: Query, cachePolicy: CachePolicy, context: UnsafeMutableRawPointer? = nil, queue: DispatchQueue, resultHandler: OperationResultHandler<Query>?) -> Cancellable {
     // If we don't have to go through the cache, there is no need to create an operation 
     // and we can return a network task directly
-    if cachePolicy == .fetchIgnoringCacheData {
+    //
+    // But without an operation, we lose the ability to cancel properly.
+    if false, cachePolicy == .fetchIgnoringCacheData {
```
RolandasRazma: ? this is hardcoded false

michaelnisi: It’s the flag I’ve mentioned in my comment above. Removing the `false` reverts the code to the previous *indeterministic* behaviour, in which it doesn't pass the newly added test. With this PR I’m trying to find out how cancelled fetches should deal with the result handler. Once that’s defined, the tests can be updated, and eventually the implementation adjusted. Or we can decide to keep the current behaviour and just improve documentation.

designatednerd:
@michaelnisi Please check out `.fetchIgnoringCacheCompletely`, as well as [`EmptyCancellable`](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/Cancellable.swift#L16) - do either of these help your case here?

michaelnisi:
Thanks

PR #532, opened 2019-05-02
Update Spectrum link

michaelnisi:
Updates the chat channel link in the docs..


michaelnisi:
I just saw the dysfunct `.embed-container` on https://www.apollographql.com/docs/ios/ and included it here.

PR #539, opened 2019-05-09
Response parsing now done on the provided handlerQueue

Valpertui:
If the cache policy is `.fetchIgnoringCacheData`, the send method is executed on the caller queue/thread. ( see `_fetch` method) .   
It also means that the response parsing is done on the caller queue/thread which can be the main queue (and so, blocking it).   

This change executes the parsing on the provided handler queue every time.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -150,18 +150,21 @@ public class ApolloClient {
         return
       }
       
-      firstly {
-        try response.parseResult(cacheKeyForObject: self.cacheKeyForObject)
-      }.andThen { (result, records) in
-        notifyResultHandler(result: result, error: nil)
+      handlerQueue.async {
         
-        if let records = records {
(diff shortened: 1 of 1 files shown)


cltnschlosser:
Why would you want the response parsing to be done on the handler queue? In most cases that will be the main queue, which will block UI updates. See my now conflicted PR which moved response parsing to a background queue https://github.com/apollographql/apollo-ios/pull/505

Valpertui:
That’s your choice to use the main queue as handler queue, which is a terrible idea in my opinion, but in my case handler queue is always a background queue

Valpertui:
After looking at your PR, I agree that your version is better, since it use a dedicated queue for response parsing and not the user provided one (which will probably be the wrong one for beginners or inexperienced users since the library use the mainqueue as default queue for the ‘handlerQueue’, I suspect that it’s something that should be changed, but it would break so many apps using Apollo directly from a ViewController)

PR #552, opened 2019-05-28
use lock in InMemoryNormalizedCache to avoid race conditions

RolandasRazma:
we see some crashes trying to access `RecordSet` while doing `.clear()` on `InMemoryNormalizedCache`

```
* thread #35, queue = 'com.apollographql.DataLoader', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
  frame #0: 0x00000001c71bf530 libobjc.A.dylib`objc_msgSend + 16
  frame #1: 0x00000001f60ee364 libswiftCore.dylib`Swift.Dictionary._Variant.subscript.getter : (A) -> Swift.Optional<B> + 148
  frame #2: 0x0000000106044da8 Apollo`RecordSet.subscript.getter(key="User_12", self=Apollo.RecordSet @ 0x00000001701b20d8) at RecordSet.swift:24:19
  frame #3: 0x000000010602fc64 Apollo`closure #1 in InMemoryNormalizedCache.loadRecords($0="User_12", self=0x0000000282b1ef60) at InMemoryNormalizedCache.swift:9:42
  frame #4: 0x000000010602fce8 Apollo`thunk for @callee_guaranteed (@guaranteed String) -> (@owned Record?, @error @owned Error) at <compiler-generated>:0
  frame #5: 0x000000010602fdcc Apollo`partial apply for thunk for @callee_guaranteed (@guaranteed String) -> (@owned Record?, @error @owned Error) at <compiler-generated>:0
  frame #6: 0x00000001f6074ef4 libswiftCore.dylib`(extension in Swift):Swift.Collection.map<A>((A.Element) throws -> A1) throws -> Swift.Array<A1> + 748
  frame #7: 0x000000010602fb1c Apollo`InMemoryNormalizedCache.loadRecords(keys=1 value, self=0x0000000282b1ef60) at InMemoryNormalizedCache.swift:9:24
  frame #8: 0x0000000106030228 Apollo`protocol witness for NormalizedCache.loadRecords(forKeys:) in conformance InMemoryNormalizedCache at <compiler-generated>:0
  frame #9: 0x0000000106001c20 Apollo`partial apply at <compiler-generated>:0
  frame #10: 0x00000001060099c4 Apollo`closure #1 in DataLoader.dispatch(self=0x00000002831abd00) at DataLoader.swift:53:12
  frame #11: 0x0000000105ff0d50 Apollo`thunk for @escaping @callee_guaranteed () -> () at <compiler-generated>:0
  frame #12: 0x00000001084176f0 libdispatch.dylib`_dispatch_call_block_and_release + 24
  frame #13: 0x0000000108418c74 libdispatch.dylib`_dispatch_client_callout + 16
  frame #14: 0x0000000108420bf4 libdispatch.dylib`_dispatch_lane_serial_drain + 712
  frame #15: 0x00000001084218b4 libdispatch.dylib`_dispatch_lane_invoke + 456
  frame #16: 0x000000010842b77c libdispatch.dylib`_dispatch_workloop_worker_thread + 1148
  frame #17: 0x00000001c7bed114 libsystem_pthread.dylib`_pthread_wqthread + 304
  frame #18: 0x00000001c7befcd4 libsystem_pthread.dylib`start_wqthread + 4
```

sentry.io log
Screenshot 2019-07-29 at 13 18 32

This PR wraps access to `InMemoryNormalizedCache.records` in internal queue to avoid race conditions.

### DIFF
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -2,7 +2,7 @@ import Foundation
 
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
-  private let recordsLock = NSLock()
+  private let recordsLock = NSRecursiveLock()
 
   public init(records: RecordSet = RecordSet()) {
     self.records = records
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -2,6 +2,7 @@ import Foundation
 
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
+  private let recordsLock = NSLock()
 
   public init(records: RecordSet = RecordSet()) {
     self.records = records
@@ -10,7 +11,9 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   public func loadRecords(forKeys keys: [CacheKey],
                           callbackQueue: DispatchQueue?,
                           completion: @escaping (Result<[Record?], Error>) -> Void) {
+    self.recordsLock.lock()
     let records = keys.map { self.records[$0] }
+    self.recordsLock.unlock()
     DispatchQueue.apollo_returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(records))
@@ -19,7 +22,9 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   public func merge(records: RecordSet,
                     callbackQueue: DispatchQueue?,
                     completion: @escaping (Result<Set<CacheKey>, Error>) -> Void) {
+    self.recordsLock.lock()
     let cacheKeys = self.records.merge(records: records)
+    self.recordsLock.unlock()
     DispatchQueue.apollo_returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(cacheKeys))
@@ -27,7 +32,9 @@ public final class InMemoryNormalizedCache: NormalizedCache {
 
   public func clear(callbackQueue: DispatchQueue?,
                     completion: ((Result<Void, Error>) -> Void)?) {
+    self.recordsLock.lock()
     self.records.clear()
+    self.recordsLock.unlock()
     
     guard let completion = completion else {
       return
--- a/Tests/ApolloCacheDependentTests/FetchQueryTests.swift
+++ b/Tests/ApolloCacheDependentTests/FetchQueryTests.swift
@@ -383,4 +383,95 @@ class FetchQueryTests: XCTestCase {
       waitForExpectations(timeout: 5, handler: nil)
     }
   }
+  
+  func testThreadedCache() throws {
+    let cache = InMemoryNormalizedCache()
+    
+    let networkTransport1 = MockNetworkTransport(body: [
+      "data": [
+        "hero": [
+          "id": "1000",
+          "name": "Luke Skywalker",
+          "__typename": "Human",
+          "friends": [
+            ["__typename": "Human", "name": "Leia Organa", "id": "1003"],
+            ["__typename": "Human", "name": "Han Solo", "id": "1002"],
+          ]
+        ]
+      ]
+    ])
+    
+    let networkTransport2 = MockNetworkTransport(body: [
+      "data": [
+        "hero": [
+          "id": "1002",
+          "name": "Han Solo",
+          "__typename": "Human",
+          "friends": [
+            ["__typename": "Human", "name": "Leia Organa", "id": "1003"],
+            ["__typename": "Human", "name": "Luke Skywalker", "id": "1000"],
+          ]
+        ]
+      ]
+    ])
+    
+    let store = ApolloStore(cache: cache)
+    let store2 = ApolloStore(cache: cache)
+    
+    let client1 = ApolloClient(networkTransport: networkTransport1, store: store)
+    let client2 = ApolloClient(networkTransport: networkTransport2, store: store2)
+    
+    let group = DispatchGroup()
+    
+    let watcherQueue = DispatchQueue(label: "test watcher queue")
+    var watchers = [GraphQLQueryWatcher<HeroAndFriendsNamesWithIDsQuery>]()
+    
+    for _ in 0...1000 {
+      
+      group.enter()
+      DispatchQueue.global().async {
+        let watcher =
+          client1.watch(
+          query: HeroAndFriendsNamesWithIDsQuery(), cachePolicy: .returnCacheDataAndFetch) { result in
+            if result.value?.source == .some(.server) {
+              group.leave()
+            }
+        }
+        
+        watcherQueue.sync {
+          watchers.append(watcher)
+        }
+      }
+      
+      group.enter()
+      DispatchQueue.global().async {
+        let watcher =
+          client2.watch(
+          query: HeroAndFriendsNamesWithIDsQuery(), cachePolicy: .returnCacheDataAndFetch) { result in
+            if result.value?.source == .some(.server) {
+              group.leave()
+            }
+        }
+        
+        watcherQueue.sync {
+          watchers.append(watcher)
+        }
+      }
+      
+      group.enter()
+      DispatchQueue.global().async {
+        _ = client1.clearCache() { _ in
+          group.leave()
+        }
+      }
+      
+    }
+    
+    let expectation = self.expectation(description: "Fetching query")
+    group.notify(queue: .main) {
+      expectation.fulfill()
+    }
+    self.wait(for: [expectation], timeout: 10)
+    
+  }
 }


RolandasRazma:
I realised that not queue fixes race conditions but actual usage of Promises. Will open smaller PR for that

RolandasRazma:
actually only promises don't fix it ;/ Will open PR again...

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -1,21 +1,37 @@
+import Dispatch
```
designatednerd: I don't think you need to import this explicitly - it should be part of `Foundation`

RolandasRazma:
@designatednerd It should be possible, I can replicate crash very easily (not sure if its easy to test)
we have setup where `multiple clients use same store` and as soon as you start doing stuff on multiple threads it crashes. We could put everything on one just because we know about bug, but other people might be facing same problem without realising reasion

designatednerd:
@RolandasRazma I mean add an automated test so we can make sure a) This is doing what it needs to and b) This doesn't regress in the future 😄

RolandasRazma:
@designatednerd I got that :) Just was saying that its easy to replicate by hand but not sure how easy to replicate in test

designatednerd:
Maybe you could set up two clients that use the same store for the test and try to have them access/mutate things in a certain sequence?

RolandasRazma:
@designatednerd So I tried to write test to replicate bug (wanted to create separate PR) but because this is race condition test not always work, it slows down all other tests and it crashes everything when it works. Not sure if it should be merged together with all other tests. I will drop it here for now, let me know what you want to do... (Without changes in this PR I manage to replicate crash with this test every 3rd run or so)

```swift
  func testThreadedCache() throws {
    let cache = InMemoryNormalizedCache()
    
    let networkTransport = MockNetworkTransport(body: [
      "data": [
        "hero": [
          "name": "Luke Skywalker",
          "__typename": "Human"
        ]
      ]
    ])
    
    let loadingClient = ApolloClient(networkTransport: networkTransport, store: ApolloStore(cache: cache))
    let clearingClient = ApolloClient(networkTransport: networkTransport, store: ApolloStore(cache: cache))
    
    let group = DispatchGroup()
    
    for _ in 0...10000 {
      
      group.enter()
      DispatchQueue.global().async {
        loadingClient.fetch(query: HeroNameQuery(), cachePolicy: .returnCacheDataElseFetch) { (result, error) in
          group.leave()
        }
      }
      
      group.enter()
      DispatchQueue.global().async {
        _ = clearingClient.clearCache()
        group.leave()
      }
      
    }
    
    let expectation = self.expectation(description: "Fetching query")
    group.notify(queue: .main) {
      expectation.fulfill()
    }
    self.wait(for: [expectation], timeout: 5)
    
  }
```

designatednerd:
I'm really not sure why that would affect other tests, it certainly shouldn't. I'll try to take a look tomorrow.

RolandasRazma:
@designatednerd it affects in a sense that it has tight loop and that will delay other ones, also on crash it will take everything down and other tests won't have chance to run

designatednerd:
Ah I think I missed that the expectation was within the giant `for` loop when I was looking at this last night. 

It also doesn't seem like we're doing any asserting on the results - in theory we should be able to check after the client's been cleared that it wouldn't have anything in the cache.

RolandasRazma:
I don't think its about clearing, problem is that ivar can be accessed simultaneously.  It usually breaks in `records.merge` as it takes longer than clear

designatednerd:
OK, coming back to this, I'm a bit confused: Is the test passing when it's run with these changes in place, or is it still failing? 

Or are the failures happening on another branch that does **not** have these changes in place?

RolandasRazma:
Test in comments will crash on dev (without this PR) but it won't do it constantly (as this is race condition). I manage to crash every 3rd run or so
As far as I can see it will not crash if this PR is applied

designatednerd:
Yeah then I'd say add it to this PR - that'll at least let us know if we've borked anything in the future.

RolandasRazma:
@designatednerd do you know any better way to test for race conditions? This doesn't feel right...

RolandasRazma:
@designatednerd maybe just add check if ivar accesses from multiple threads (as it will lead to crash sooner or later)?

designatednerd:
Race conditions are really hard to test for - the main thing I'd do is probably check the results of queries from both clients performed in order. it could be something where you just run everything in sequence - and theoretically you should get results back sequentially. This is the sequence I'd probably be checking: 

Set `Luke Skywalker` with loading client
Fetch data with clearing client -> Should be `Luke Skywalker` because the set should finish before the fetch starts
Clear with clearing client
Fetch data with loading client -> should be cleared because the clear should finish before the fetch starts

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -1,21 +1,37 @@
...
+
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
-
+  private let queue: DispatchQueue
+  
   public init(records: RecordSet = RecordSet()) {
     self.records = records
+    self.queue = DispatchQueue(label: "com.apollographql.InMemoryNormalizedCache", attributes: .concurrent)
```
AnthonyMDev: Shouldn't we be using a serial queue to ensure that we don't have two blocks accessing it at the same time? Wouldn't concurrent still allow for race conditions?

RolandasRazma: I was thinking same, but serial queue would slow down things and I can't replicate crash with concurrent one... Maybe fact that it is accessed on same thread is enough...
We have this PR in production for past couple of months (way before I created PR) and see 0 crashes.

AnthonyMDev: Okay, you're right a serial queue is less performant, but this concurrent queue doesn't seem to necessarily address the issue. Of course it's up to @designatednerd in the end, and I'm glad that this workaround has been able to work for your team in production. But I wouldn't feel comfortable about pulling in a workaround for an issue that we can't consistently reproduce and that doesn't seem to make perfect sense.

It seems that this may _happen_ to avoid the race conditions present in your code base currently, but it doesn't completely avoid the possibility of them occurring. 

I'd feel much more comfortable going with `NSRecursiveLock` as a solution here than dispatching to another queue.

Additionally, though I understand it may not be possible, it would be great if we could work a bit more on trying to create a test case that consistently crashes.

I'd be willing to spend some time looking into this and try to reproduce a crashing test, but not sure I'll get much further than you did @RolandasRazma .

RolandasRazma: completely agree it needs better testing. But the only thing I can think off is add check if ivar is accessed from multiple threads

AnthonyMDev:
Just sent this as a message to @designatednerd, but probably should just put it here for discussion.

It looks like we are using both a DispatchQueue with .barrier AND a ReadWrite cache lock. That seems like overkill. Adding an additional queue on the cache itself seems like a crazy amount of unneeded locking.

If there are still crashes that are being reported currently, then it indicates to me that we aren't properly locking on the correct places. But if we are going to add new locks, I'd like to kill some of the old ones and just figure out the correct places to be locking, rather than continuing to add more and more locks on top of each other.

Does this seem to be right, or am I missing something about the locks and queues that are currently in place?

AnthonyMDev:
Alright, so I've written a modified test. This one does not crash, but it does fail the equal assertion a random number of times, and it fails consistently. It's failing because the cache is being cleared between the `write` and the `loadRecords()` calls.

When I implement the PR from @RolandasRazma this test passes! It's not showing the `EXC_BAD_ACCESS` crash, but I think I'm getting to the same underlying behavior.

```swift
func testMultipleStoresOneCacheConcurrentAccess() throws {
    let initialRecords: RecordSet = [
      "2001": ["name": "R2-D2", "__typename": "Droid", "primaryFunction": "Protocol"]
    ]

    let cache = InMemoryNormalizedCache(records: initialRecords)

    let store1 = ApolloStore(cache: cache)
    let store2 = ApolloStore(cache: cache)

    let expectation = self.expectation(description: "Read updated query")

    var data = HeroDetails(unsafeResultMap: ["name": "R2-D2",
                                             "__typename": "Droid",
                                             "primaryFunction": "Protocol"])

    for i in 1...1000 {
      DispatchQueue.global().async {
        _ = store1.withinReadWriteTransaction { transaction1 in
          data.name = "R4-D\(i)"
          try! transaction1.write(object: data, withKey: "2001")

          let heroName = try? transaction1.loadRecords(forKeys: ["2001"]).await().first??["name"] as? String

          XCTAssertEqual(heroName, "R4-D\(i)")
          if i == 1000 {
            expectation.fulfill()
          }

        }
      }

      DispatchQueue.global().async {
        store2.clearCache()
      }
    }

    self.wait(for: [expectation], timeout: 5)
  }
```

However the more I think about this, the more I think that both the `ApolloStore` and the cache itself shouldn't have their own locks. It seems like it's a lot of overkill and causes performance issues. Also, every future implementation of a new cache type will need to handle its own locking.
I understand using the same cache with multiple clients, but I think that in that case, you should always use one shared `ApolloStore`. Since the store is already handling thread safety here, using the same cache in different stores should be considered a developer error and not supported. 

If I do both the clearing of the cache and the read/write on the same store, this test passes without the new code from this PR.

AnthonyMDev:
I've just confirmed that in the test case @RolandasRazma provided, if you use one single store with both clients, the crash also goes away.

RolandasRazma:
I think I can reproduce it even with one store and cache, need to look into it

AnthonyMDev:
I saw a crash in production today and we are only using one store and cache. So it still does exist. I'm unable to create a test to reproduce it though so far. Continuing to try.

Sources/Apollo/InMemoryNormalizedCache.swift
jimisaacs: Hello, I've been looking around to see all the work being done on this library recently and saw this PR and this thread, so hope you don't mind me chiming in. The reason why concurrent works is because your using it with barriers. If you only even access a concurrent queue with barriers, you've essentially made it function exactly like a serial queue. A common approach to synchronizing with a queue, is using a concurrent queue, then use async barrier writes, and sync non-barrier reads. Keeps the happy-path fast, and the write path safe.

jimisaacs: IIRC, I think async reads would work with the pattern described too, but would definitely need testing because I'm not 100% sure of any side-effects inside the task blocks here... if any. Otherwise, might as well take the first suggestion, if it's all "barrier'ed up", might as well make it serial and remove the barriers.

RolandasRazma:
@AnthonyMDev We also have one store and cache in production (but but multiple clients). Trying to reproduce this as well in test but nut that simple. 
On app I can reproduce quite easy, looking now what is the difference between app and tests. One of difference is that we use `.watch` extensively...

AnthonyMDev:
We could definitely try using watch instead of fetch to see if we can make this crash. I'm still unable to get a crashing unit test with on store. But I do believe you that it's possible, since I've seen it in prod on my own app!
Lets try using watch and see what we can get.

Sources/Apollo/InMemoryNormalizedCache.swift
AnthonyMDev: Yeah, if you are always using barriers, then serial is the same! Thanks @jimissacs!

AnthonyMDev:
I switched to using `watch` and was able to make it crash one in every 3-5 runs. Turning on the thread sanitizer confirmed a suspicion I had on Friday!

Theoretically, the `ApolloStore` should be locking the read and write of it's cache, so we shouldn't have to ALSO have a separate lock on the cache itself. **But** the `DataLoader` object is getting direct access to the `loadRecords` function of the cache and calling it in it's own async queue. Which means that it's accessing the cache itself asynchronously outside of the store's locks! This is almost certainly what's causing the crash.

As I've said before, I don't think that we should make the cache and the apollo store both responsible for their own locking. Either find a way to move all locking to the underlying cache, or just make the apollo store handle all access to the cache, and make it thread safe there! I think we need to make the `DataLoader` object access the cache data THROUGH the apollo store, instead of having access to the underlying cache itself!

AnthonyMDev:
Here is what the test I'm using now looks like.
```
func testThreadedCache() throws {
    let cache = InMemoryNormalizedCache()

    let networkTransport1 = MockNetworkTransport(body: [
      "data": [
        "hero": [
          "id": "1000",
          "name": "Luke Skywalker",
          "__typename": "Human",
          "friends": [
            ["__typename": "Human", "name": "Leia Organa", "id": "1003"],
            ["__typename": "Human", "name": "Han Solo", "id": "1002"],
          ]
        ]
      ]
      ])

    let networkTransport2 = MockNetworkTransport(body: [
      "data": [
        "hero": [
          "id": "1002",
          "name": "Han Solo",
          "__typename": "Human",
          "friends": [
            ["__typename": "Human", "name": "Leia Organa", "id": "1003"],
            ["__typename": "Human", "name": "Luke Skywalker", "id": "1000"],
          ]
        ]
      ]
      ])

    let store2 = ApolloStore(cache: cache)
    
    let client1 = ApolloClient(networkTransport: networkTransport1, store: store)
    let client2 = ApolloClient(networkTransport: networkTransport2, store: store2)

    let group = DispatchGroup()

    let watcherQueue = DispatchQueue(label: "test watcher queue")
    var watchers = [GraphQLQueryWatcher<HeroAndFriendsNamesWithIDsQuery>]()

    for _ in 0...1000 {

      group.enter()
      DispatchQueue.global().async {
        let watcher =
          client1.watch(
          query: HeroAndFriendsNamesWithIDsQuery(), cachePolicy: .returnCacheDataAndFetch) { result in
            if result.value?.source == .some(.server) {
              group.leave()
            }
        }

        watcherQueue.sync {
          watchers.append(watcher)
        }
      }

      group.enter()
      DispatchQueue.global().async {
        let watcher =
        client2.watch(
        query: HeroAndFriendsNamesWithIDsQuery(), cachePolicy: .returnCacheDataAndFetch) { result in
          if result.value?.source == .some(.server) {
            group.leave()
          }
        }

        watcherQueue.sync {
          watchers.append(watcher)
        }
      }

      group.enter()
      DispatchQueue.global().async {
        _ = client1.clearCache().finally {
          group.leave()
        }
      }

    }

    let expectation = self.expectation(description: "Fetching query")
    group.notify(queue: .main) {
      expectation.fulfill()
    }

  }
```

RolandasRazma:
good job @AnthonyMDev

designatednerd:
@AnthonyMDev @RolandasRazma What do you all think it will take to get this straightened out so maybe we could land it?

RolandasRazma:
I guess decision has to be made who is responsible for locking properly. "Store" or its users. 
There is already locking mechanism in Apollo itself, so adding another one on store does look like redundant. On the other hand, if to look from separation point, there might be several different "stores" and some of them might not even require locking

designatednerd:
Reading back through the comments, I definitely agree we should put all the lock/unlock logic in one place to simplify. 

Looking at the general stack: 
- It definitely doesn't make sense to put it in `ApolloClient` since multiple clients can access the same `ApolloStore`. 
- It could make sense to put it in `ApolloStore` since this would cover all different types of things conforming to `NormalizedCache`. However, it seems like it's plausible to have multiple `ApolloStore` objects which access the same `NormalizedCache` object. We could warn folks away from doing this in the docs. 
- Making instances of `NormalizedCache` handle locking could be a real pain in the butt, since that means each individual cache type would need to manage this. However, if we were to (for example) move the `InMemoryCache` to being backed by an `NSCache`, then we wouldn't actually *need* to manage this since `NSCache` is thread-safe, and we wouldn't need to deal with it at all.

I'm torn between the "do it once and test the crap out of it" of putting it in `ApolloStore` vs the "But what if you don't even have to do it!" of fobbing it off on `NormalizedCache` instances. 

I'm mulling over an `isThreadSafe` property on `NormalizedCache` and protocol extensions to handle locking/unlocking at the `NormalizedCache` level, but that seems like it could be a **gargantuan** rabbit hole. 

Would love to hear other thoughts on options here.

AnthonyMDev:
Sorry I've been MIA on this for a while. I think the safest and most performant way to handle this is to handle locking as close to the source as possible. If certain types of caches don't even need the locking then we shouldn't force them to use it. 
Having the lock at any higher level than the cache itself lends itself to more issues like we are currently seeing where multiple stores are accessing the same cache.
Different caches may be able to implement locking differently to improve performance based on their implementation details as well. 

I strongly support locking at the cache level.

designatednerd:
@AnthonyMDev Makes sense - what do you think about having the locking work in a protocol extension?

AnthonyMDev:
I'm not a fan of that just because it limits the configuration of how the locking is implemented too much in my opinion. Writing locking code isn't fun, but it's not the most difficult thing in the world to do. When you are writing a caching layer, that's something that's expected. Locking for thread safety is too important and too particular to have a shared implementation for I think.

designatednerd:
I think Anthony makes some great points here - locking at the cache level rather than at any higher level is probably the least-risky way to go. I think he's also probably right that I'm trying too hard to throw Protocol Extensions at the problem. :)

Is it worth continuing this within this particular PR? Do any of you have some time to work on this?

AnthonyMDev:
Haha, to be fair, I LOVE extracting as much logic as possible into protocol extensions. It's definitely the right mindset to be in by default. I think this is just a rare exception to that. :)

AnthonyMDev:
As for your questions:
I think the direction this PR took isn't what we are going to go with, so I don't think we need to continue with it. Thanks so, so, so MUCH @RolandasRazma for bringing this up, researching it, coming up with the test case, and providing so much information! If you or @designatednerd have time to implement the locking in the cache and kill the locking in the store, go for it. 
I'm not sure when I'll be able to find time to take this on, but if no one else can, I will try to do it next week. 

If either of you work on this, let me know so we don't work over each other. I'd like to provide my test cases to you also so that you can use them to help verify that the new locking works. I riffed off of @RolandasRazma 's tests and even though they still don't fail 100% of the time, they fail more consistently now. :)

designatednerd:
Unfortunately this isn't my highest priority at the moment but I'm happy to provide support for whoever has time to work on it!

AnthonyMDev:
Roger that. It's not my highest priority right now either haha ;) But I'll try to make it happen when I get a chance.

RolandasRazma:
should we keep this open until it can be addressed properly so other people would know about problem?

designatednerd:
We can either keep it open or open a new issue to summarize what we've figured out and what needs to happen - I'll leave that up to you @RolandasRazma.

AnthonyMDev:
I’m thinking I would keep the issue open, but you’re the repo maintainer, so up to you!

RolandasRazma:
Was there any progress on this? I reverted to stock implantation (latest from GitHub) and we getting like 80 crashes a day. Stack is not very useful but it goes something like...

__hidden
__hidden
RecordSet.subscript.getter
<redacted>
_hidden
_hidden
InMemoryNormalizedCache.loadRecords

RolandasRazma:
Looks like simple `NSLock` around `self.records` access is enough to stop crashing with latest changes in `ApolloStore` and `InMemoryNormalizedCache`

@designatednerd what do you think about such fix? Can I get into deadlock somehow?

designatednerd:
I mean you always *can* get into deadlock, but I suspect if it doesn't deadlock in the tests which are doing a lot of different accesses of the code, it won't in practice. Probably worth adding a test which explicitly checks accessing from two different threads. 

Would love to see this update get over the line!

RolandasRazma:
I was testing with test @AnthonyMDev ended up with. Looks fine. Tried to use mutex as well but didn't noticed any speed difference and NSLock easier to use. We pushed it to production (live dangerously and stuff) but its unlikely we hear anything from users as people don't report deadlocks :) Will open PR so we can continue discussion.

designatednerd:
Awesome, looking forward to it! Heads up that I'm on vacation for the next week and a half so responses may be slow, but I'll try to get a look as I can.

RolandasRazma:
@designatednerd pushed changes, also included @AnthonyMDev test as it crashes 100% of a time now without lock (after locks were removed from ApolloStore)

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
+  private let recordsLock = NSLock()
```
AnthonyMDev: Is there a reason why we aren't using `NSRecursiveLock` here? If the implementation of accessing the records ever changes to need to access the cache recursively or anything, then this is going to deadlock. 

If there is a good reason I'm missing to use a regular `NSLock`, then great, but I prefer to default to recursive if possible for safety.

RolandasRazma: no good reason, switched

lukelabonte:
> Looks reasonable to me, and a _far_ simpler change than earlier. I'm interested in eventually abstracting some of this logic into a property wrapper, but for now let's get this going!
> 
> FYI this won't ship as part of a release until I'm back on the ground to support it, but I'll merge it now if you want to use it off `master`.

@designatednerd do you have an ETA on when you'll be back? I'm wondering when we should expect a new release that would include this fix.

designatednerd:
This shipped a year ago as part of 0.22.0. There have been substantial other changes around how we're handling this, I'd appreciate you filing a new issue with updated details if you're having problems. Thank you!

Issue #570, opened 2019-06-07
GET request could be a fetch option


dmandarino:
We are not able to use GET request as a fetch option like we do in others platforms. So, when we want to fetch information from an API that only allows this method, it fails. Also, we are not able to help API with cache, as most of them don't implement a specific header to allow cache with POST.

dmandarino:
This was done in issue #572

designatednerd:
Yep! Holding off on closing this until it gets released 🤓

designatednerd:
Shipped with 0.11.0!

PR #572, opened 2019-06-07
Create FetchOptions to allow GET Requests like in Web

dmandarino:
#570 

To give an option of choose between GET and POST as HTTP Method. 

So, requests can be cached by API and user can choose the method suits better for each case. Also, some API only GET.

By exposing `FetchOptions` in fetch method, users will be able to choose the appropriate method every time he use it. 

Like:  `fetch(query: GraphQLQuery(), fetchOptions: .GET)`

### DIFF
--- a/Sources/Apollo/FetchOptions.swift
+++ b/Sources/Apollo/FetchOptions.swift
@@ -1,0 +1,6 @@
+import Foundation
+
+public enum FetchOptions: String {
+    case GET
+    case POST
+}
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -95,7 +95,7 @@ public class HTTPNetworkTransport: NetworkTransport {
   ///
   /// - Parameters:
   ///   - operation: The operation to send.
-  ///   - fetchHTTPMethod: The HTTP Method to be used.
+  ///   - fetchHTTPMethod: The HTTP Method to be used in operation.
   ///   - completionHandler: A closure to call when a request completes.
   ///   - response: The response received from the server, or `nil` if an error occurred.
   ///   - error: An error that indicates why a request failed, or `nil` if the request was succesful.
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -69,22 +69,22 @@ public class HTTPNetworkTransport: NetworkTransport {
   ///
   /// - Parameters:
   ///   - operation: The operation to send.
-  ///   - fetchOptions: The HTTP Method to be used.
+  ///   - fetchHTTPMethod: The HTTP Method to be used.
   ///   - completionHandler: A closure to call when a request completes.
   ///   - response: The response received from the server, or `nil` if an error occurred.
   ///   - error: An error that indicates why a request failed, or `nil` if the request was succesful.
   /// - Returns: An object that can be used to cancel an in progress request.
-  public func send<Operation>(operation: Operation, fetchOptions: FetchOptions, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
+  public func send<Operation>(operation: Operation, fetchHTTPMethod: FetchHTTPMethod, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
     let body = requestBody(for: operation)
     var request = URLRequest(url: url)
     
-    if fetchOptions.rawValue == "GET", let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
+    if fetchHTTPMethod.rawValue == "GET", let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
         request = URLRequest(url: urlForGet)
-    } else if fetchOptions.rawValue == "POST" {
+    } else if fetchHTTPMethod.rawValue == "POST" {
         request.httpBody = try! serializationFormat.serialize(value: body)
     }
     
-    request.httpMethod = fetchOptions.rawValue
+    request.httpMethod = fetchHTTPMethod.rawValue
     request.setValue("application/json", forHTTPHeaderField: "Content-Type")
     
     let task = session.dataTask(with: request) { (data: Data?, response: URLResponse?, error: Error?) in
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -69,18 +69,23 @@ public class HTTPNetworkTransport: NetworkTransport {
   ///
   /// - Parameters:
   ///   - operation: The operation to send.
+  ///   - fetchOptions: The HTTP Method to be used.
   ///   - completionHandler: A closure to call when a request completes.
   ///   - response: The response received from the server, or `nil` if an error occurred.
   ///   - error: An error that indicates why a request failed, or `nil` if the request was succesful.
   /// - Returns: An object that can be used to cancel an in progress request.
-  public func send<Operation>(operation: Operation, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
-    var request = URLRequest(url: url)
-    request.httpMethod = "POST"
-    
-    request.setValue("application/json", forHTTPHeaderField: "Content-Type")
-
+  public func send<Operation>(operation: Operation, fetchOptions: FetchOptions, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
     let body = requestBody(for: operation)
-    request.httpBody = try! serializationFormat.serialize(value: body)
+    var request = URLRequest(url: url)
+    
+    if fetchOptions.rawValue == "GET", let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
+        request = URLRequest(url: urlForGet)
+    } else if fetchOptions.rawValue == "POST" {
+        request.httpBody = try! serializationFormat.serialize(value: body)
+    }
+    
+    request.httpMethod = fetchOptions.rawValue
+    request.setValue("application/json", forHTTPHeaderField: "Content-Type")
     
     let task = session.dataTask(with: request) { (data: Data?, response: URLResponse?, error: Error?) in
       if error != nil {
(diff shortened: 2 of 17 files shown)


Sources/Apollo/FetchOptions.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public enum FetchOptions: String {
```
designatednerd: I think it might be a little easier to understand this if it were something like `FetchHTTPMethod` - you're not offering any other options here, just letting the user set the HTTP method to use for a fetch.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,18 +69,23 @@ public class HTTPNetworkTransport: NetworkTransport {
...
   ///   - response: The response received from the server, or `nil` if an error occurred.
   ///   - error: An error that indicates why a request failed, or `nil` if the request was succesful.
   /// - Returns: An object that can be used to cancel an in progress request.
-  public func send<Operation>(operation: Operation, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
+  public func send<Operation>(operation: Operation, fetchOptions: FetchOptions, completionHandler: @escaping (_ response: GraphQLResponse<Operation>?, _ error: Error?) -> Void) -> Cancellable {
+    let body = requestBody(for: operation)
     var request = URLRequest(url: url)
-    request.httpMethod = "POST"
     
+    if fetchOptions.rawValue == "GET", let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
```
designatednerd: This might be easier to do as a `switch` on `fetchOptions` - that way you don't have to care about what the `rawValue` is.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,18 +69,23 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+        request = URLRequest(url: urlForGet)
+    } else if fetchOptions.rawValue == "POST" {
+        request.httpBody = try! serializationFormat.serialize(value: body)
```
designatednerd: I would reeeeeally avoid `try!` here, as it'll crash an app if it fails. 

What about a `do/try` and call the completion closure with any caught error?

Tests/ApolloCacheDependentTests/WatchQueryTests.swift
```diff
@@ -39,7 +39,7 @@ class WatchQueryTests: XCTestCase {
 
       var expectation = self.expectation(description: "Fetching query")
 
-      let watcher = client.watch(query: query) { (result, error) in
+      let watcher = client.watch(query: query, fetchOptions: .POST) { (result, error) in
```
designatednerd: I actually think you might get better test results here and in the two bits above if you leave out the `fetchOptions` parameter. That way you can validate that the default option is still `.POST`.

Sources/Apollo/FetchOptions.swift
dmandarino: Totally agree. I'm changing it.

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: Done :)

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: Agreed. Actually, I did not change this line, just put it inside my `if` (now a switch as you suggested). But I agree that it could be changed.

Tests/ApolloCacheDependentTests/WatchQueryTests.swift
dmandarino: You are right, the `watch` was not accepting default value as POST. Just made the changes for it.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -47,6 +47,32 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
   }
 }
 
+public struct GraphQLHTTPRequestError: Error, LocalizedError {
+  public enum ErrorKind {
```
designatednerd: You should be able to just make `GraphQLHTTPRequestError` an enum instead of needing a struct wrapper around the embedded enum. I can see we've got this set up in another place similarly but I'm not sure why - @martijnwalraven any thoughts?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -47,6 +47,32 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
   }
 }
 
+public struct GraphQLHTTPRequestError: Error, LocalizedError {
```
designatednerd: I don't think this needs to conform to `LocalizedError` since it's not providing any localized description of the error.

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
```diff
@@ -299,7 +299,7 @@ class StarWarsServerTests: XCTestCase {
 
       let expectation = self.expectation(description: "Performing mutation")
 
-      client.perform(mutation: mutation) { (result, error) in
+      client.perform(mutation: mutation, fetchHTTPMethod: .POST) { (result, error) in
```
designatednerd: I think all these ones which aren't in `FetchQueryTests` should probably just use the default rather than handing in the `.POST` parameter - glad to hear the earlier request to change this caught some issues!

dmandarino:
@designatednerd Thank you for your review! I just did the changes you asked for.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,18 +95,32 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+      if let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
+        request = URLRequest(url: urlForGet)
+      } else {
+        completionHandler(nil, GraphQLHTTPRequestError(kind: .serializedQueryParamsMessageError))
+      }
+    default:
+      do {
+        request.httpBody = try serializationFormat.serialize(value: body)
+      } catch {
+        completionHandler(nil, GraphQLHTTPRequestError(kind: .serializedBodyMessageError))
```
designatednerd: You should probably return here and above to prevent the rest of this from executing

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Looks like it actually was changed when I look at the diff - there's still a `try` but that's wayyy better than a `try!`, since it catches and returns the error instead of crashing the app.

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: Hi! 

This method `send<Operation>` should return a `Cancellable`. In this case, it should return a task after the request. If it fails, it will execute the request and returns another error. 

As the task must resume ( `task.resume()` ), what do you think of instead of `do/try`, we use a `try?` to let the request executes and returns the correct error? Another option would be change the return type to be optional as `Cancellable?`. But I don't like that much this approach because it would change the whole architecture of `apollo-client` and also, the request would never be executed in this case. And I'm not sure that it would be a good idea for clients.

What do you think? Do you have any suggestion?

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
dmandarino: Good point! Changing it!

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: I'll poke around - I think there's probably something else that conforms to `Cancelable` that we can return here

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: I did it in that way to confirm with the project's pattern. I didn't want to change it, although I know it's not necessary to put inside a struct wrapper.

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: Same reason: I did it because every customised error in project, confirm with both, even though `LocalizedError` already extends from `Error`, and I'm using `errorDescription` declared on this protocol. Of course i can remove it, if you think it would be better in this way. Just didn't want to go against the pattern while opening this PR.

Sources/Apollo/HTTPNetworkTransport.swift
dmandarino: Okay! Thank you, once again!

dmandarino: Right now the only non private object/struct is an extension of `URLSessionTask` declared inside this class. The others are the classes `GraphQLQueryWatcher` and `WebSocketTask` which are not supposed to be used in here. 

But I was thinking about it... A request can be executed without a request body or a query parameter. It's like a request to an endpoint without parameters. So, if this try fails, it won't have a request body, and so, when the task resume, it will just return an error and the reason for it.

But also, the only way for this to fail, right now, would be in the case that `Operation` fails to be constructed. Which means a bad format in query that would fail while compiling or maybe during the processing. 

Maybe, that's why it was using `try!` before. I do not like to force a try, but changing it for `try?` wouldn't cause a crash. And because we are confident that GraphQLOperation only exists if it succeeded parsing those values in query, the worst case scenario would be a request to an endpoint without parameter.

dmandarino: But once again, tell me what is your opinion, please! I want to help with it, but if you want to keep this change and find another way to solve it, please, be my guest to fix it.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Fair enough, i'll talk this one through with @martijnwalraven in meatspace but for this PR, it's fine to leave.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: (samesies - this can stay for this PR)

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: I think for this PR we're gonna leave it - I'll make a fix that addresses this in other areas as well in a separate PR.

dmandarino: Ok, thank you for the review. :) !!

PR #551, opened 2019-05-28
Add various performance improvements

martijnwalraven:
This PR adds various performance improvements, mostly focusing on response parsing. Among other things, it changes the response path representation to a linked list, and it adds a fast path for parsing results that don't need to be stored in the cache, which can be activated with a new `.fetchIgnoringCacheCompletely` cache policy. For large query results, this seems to show an 80-85% performance improvement.

PR #592, opened 2019-06-27
Swift 5: No, seriously

designatednerd:
In this PR: 
- Updated podspec to indicate correct swift version
- Removed a couple no-longer-necessary workarounds
- Add note about whining from CocoaPods for Websocket imports of the core lib


Configuration/Shared/Project-Version.xcconfig
```diff
@@ -1 +1 @@
-CURRENT_PROJECT_VERSION = 0.10.1
+CURRENT_PROJECT_VERSION = 0.10.2
```
designatednerd: @martijnwalraven I'll want to make this 0.11.0 if we get #551 merged in soon

PR #599, opened 2019-07-01
GET as a transport option instead of a parameter + ErrorCancellable 

designatednerd:
I had a chat with @martijnwalraven about how using `GET` vs `POST` was handled in #572, and he had a couple good points I missed in the review: First, passing the fetch method as a parameter doesn't make sense if it's going to eventually be handed off to a web socket, since it'll ignore that parameter anyway. Second, if someone wants to use `GET` for a fetch request, they almost certainly want to use it for all fetch requests. 

His suggestion (which I agreed with and implement here) was to use a parameter on the `HTTPNetworkTransport` to set whether fetches should use `GET` or `POST` - I left the default as `POST` to match existing behavior, but I think this will make things a lot simpler for the end user. 

@dmandarino I'd love to hear your thoughts on this as it's a pretty significant change to what you proposed, but I *think* with the same end result. 

Also in this PR: 
- Added `ErrorCancelable` return type which allows you to return a `Cancellable` after encountering an error that should early-exit. 
- Updated `.xcodeproj` files to ensure 2-space formatting for the Apollo projects (note: Didn't fix any existing 4-space formatting, will do that in a separate PR when I have more time)
- Fixed a couple of warnings in the web socket tests
- Started pulling a few things out into their own files to make them easier to find


Sources/Apollo/Cancellable.swift
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// MARK: - URL Session Conformance
+
+extension URLSessionTask: Cancellable {}
+
+// MARK: - Early-Exit Helper
+
+/// A class to return when an error that should cause us to bail out of something still needs to return `Cancellable`.
+public final class ErrorCancellable: Cancellable {
```
martijnwalraven: I'm wondering if this should really be tied to errors, or whether something more generic like an `EmptyCancellable` or  `NullCancellable` (from [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern)), would make sense.

designatednerd: Ooh good call - `EmptyCancellable` is way more accurate

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -1,125 +1,59 @@
...
-    case .GET:
+ 
+    if self.useGETForQueries && operation.operationType == .query {
       if let urlForGet = mountUrlWithQueryParamsIfNeeded(body: body) {
         request = URLRequest(url: urlForGet)
+        request.httpMethod = GraphQLHTTPMethod.GET.rawValue
       } else {
-        completionHandler(nil, GraphQLHTTPRequestError(kind: .serializedQueryParamsMessageError))
+        completionHandler(nil, GraphQLHTTPRequestError.serializedQueryParamsMessageError)
+        return EmptyCancellable()
```
dmandarino: Nice touch!

SirensOfTitan:
@designatednerd: Perhaps I'm thinking too big picture here, but it might be worthwhile to look how custom extensions will evolve as `apollo-ios` does:

1. Will `apollo-ios` adopt the `ApolloLink` strategy like other clients?  
2. Or perhaps something more idiomatic swift?

... we've been using NetworkTransports to compose different layers of behavior together into one larger NetworkTransport.  For example, we have one NetworkTransport for Authentication-retries (i.e. when the user has to request a token from the server to continue requests), and (as of yesterday) one for PersistedQuery management.

If the GET query option is baked into `HTTPNetworkTransport` (which does feel a lot cleaner as some transports like WebsocketTransport don't need it) as-is, we'd need to re-create network transports for the composition pattern I mentioned above to work (as we'd first want to optimistically send the GET request then send a POST for the full query if we don't have the query cached).

designatednerd:
@SirensOfTitan Definitely good to keep the big picture in mind but I think at this point, particularly given that a) We're at a `0.x` release and will be for a while, so "This may evolve" is kind of baked in, and b) People have been requesting this feature for literally years, I would prefer to err on the side of shipping this and making changes when we make decisions about the `ApolloLink` strategy later.

dmandarino:
I agree with @designatednerd. People are asking for it since 2017 and there were issues opened last year too. The implementation may change in the future while the project goes on, and the need changes.

SirensOfTitan:
@designatednerd @dmandarino  Sorry, I'm not arguing against GET, I'm saying that the general, fuzzy idea of how the API will evolve could act as a useful heuristic as to whether we should keep GET as a parameter in `send` or as a parameter on `HTTPNetworkTransport`.

designatednerd:
Totally understand your point @SirensOfTitan, but at this point I want to get this out the door for day to day usage. 

I'm definitely planning on revisiting some of this architecture when I get the bugs to a manageable point, but with 173 open issues (as of right this second) that may be a minute 😃

designatednerd:
Screen Shot 2019-07-02 at 9 04 59 PM
^ taking this as an approval and merging!

PR #602, opened 2019-07-01
Header configuration + Fix GET with null parameter

designatednerd:
I know this has been a request [for some time](https://github.com/apollographql/apollo-ios/issues/37), but this PR contains the ability to change parameters on the URL request before it gets sent out, the ability to have the raw `Data?`/`URLResponse`/`Error?` response from a `URLSessionTask` so you can grab headers and/or log things as you want, and the ability to retry asynchronously (for example, after re-authenticating). 

This work also uncovered some weirdness with sending GET requests with null parameters for optional variables, so I broke out the GET url constructor, simplified it, and tested it.

### DIFF
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -1,0 +1,63 @@
+//
+//  GraphQLGETTransformer.swift
+//  Apollo
+//
+//  Created by Ellen Shapiro on 7/1/19.
+//  Copyright © 2019 Apollo GraphQL. All rights reserved.
+//
+
+import Foundation
+
+struct GraphQLGETTransformer {
+  
+  let body: GraphQLMap
+  let url: URL
+  
+  private let variablesKey = "variables"
+  private let queryKey = "query"
+  
+  /// A helper for transforming a GraphQLMap that can be sent with a `POST` request into a URL with query parameters for a `GET` request.
+  ///
+  /// - Parameters:
+  ///   - body: The GraphQLMap to transform from the body of a `POST` request
+  ///   - url: The base url to append the query to.
+  init(body: GraphQLMap,
+       url: URL) {
+    self.body = body
+    self.url = url
+  }
+  
+  /// Creates the get URL.
+  ///
+  /// - Returns: [optional] The created get URL or nil if the provided information couldn't be used to access the appropriate parameters.
+  func createGetURL() -> URL? {
+    guard var components = URLComponents(string: self.url.absoluteString) else {
+      return nil
+    }
+    
+    guard let query = self.body.jsonObject[self.queryKey] as? String else {
+      return nil
+    }
+    
+    var queryItems = components.queryItems ?? [URLQueryItem]()
+
+    queryItems.append(URLQueryItem(name: self.queryKey, value: query))
+    components.queryItems = queryItems
+    
+    guard let variables = self.body.jsonObject[self.variablesKey] as? [String: AnyHashable] else {
+      return components.url
+    }
+    
+    guard
+      let serializedData = try? JSONSerialization.data(withJSONObject: variables),
+      let jsonString = String(bytes: serializedData, encoding: .utf8) else {
+        return components.url
+    }
+    
+    queryItems.append(URLQueryItem(name: self.variablesKey, value: jsonString))
+    components.queryItems = queryItems
+    
+    return components.url
+  }
+}
+
--- a/Sources/Apollo/GraphQLHTTPRequestError.swift
+++ b/Sources/Apollo/GraphQLHTTPRequestError.swift
@@ -7,7 +7,7 @@ public enum GraphQLHTTPRequestError: Error, LocalizedError {
   public var errorDescription: String? {
     switch self {
     case .cancelledByDeveloper:
-      return "The request was cancelled by the developer using the HTTPNetworkTransportDelegate."
+      return "The request was cancelled by the developer using the HTTPNetworkTransportPreflightDelegate."
     case .serializedBodyMessageError:
       return "JSONSerialization error: Error while serializing request's body"
     case .serializedQueryParamsMessageError:
--- a/Sources/Apollo/GraphQLHTTPRequestError.swift
+++ b/Sources/Apollo/GraphQLHTTPRequestError.swift
@@ -7,7 +7,7 @@ public enum GraphQLHTTPRequestError: Error, LocalizedError {
   public var errorDescription: String? {
     switch self {
     case .cancelledByDeveloper:
-      return "The request was cancelled by the developer using the HTTPNetworkTransportPreflightDelegate."
+      return "The request was cancelled by the HTTPNetworkTransportPreflightDelegate."
     case .serializedBodyMessageError:
       return "JSONSerialization error: Error while serializing request's body"
     case .serializedQueryParamsMessageError:
--- a/Sources/Apollo/GraphQLHTTPRequestError.swift
+++ b/Sources/Apollo/GraphQLHTTPRequestError.swift
@@ -1,10 +1,13 @@
 /// An error which has occurred during the serialization of a request.
 public enum GraphQLHTTPRequestError: Error, LocalizedError {
+  case cancelledByDeveloper
   case serializedBodyMessageError
   case serializedQueryParamsMessageError
   
   public var errorDescription: String? {
     switch self {
+    case .cancelledByDeveloper:
+      return "The request was cancelled by the developer using the HTTPNetworkTransportDelegate."
     case .serializedBodyMessageError:
       return "JSONSerialization error: Error while serializing request's body"
     case .serializedQueryParamsMessageError:
--- a/Sources/Apollo/GraphQLHTTPRequestError.swift
+++ b/Sources/Apollo/GraphQLHTTPRequestError.swift
@@ -1,12 +1,12 @@
 /// An error which has occurred during the serialization of a request.
 public enum GraphQLHTTPRequestError: Error, LocalizedError {
-  case cancelledByDeveloper
+  case cancelledByDelegate
   case serializedBodyMessageError
   case serializedQueryParamsMessageError
   
   public var errorDescription: String? {
     switch self {
-    case .cancelledByDeveloper:
+    case .cancelledByDelegate:
       return "The request was cancelled by the HTTPNetworkTransportPreflightDelegate."
     case .serializedBodyMessageError:
       return "JSONSerialization error: Error while serializing request's body"
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -106,23 +106,23 @@ public class HTTPNetworkTransport: NetworkTransport {
     
   private func mountUrlWithQueryParamsIfNeeded(body: GraphQLMap) -> URL? {
     guard let query = body.jsonObject["query"], var queryParam = queryString(withItems:  [URLQueryItem(name: "query", value: "\(query)")]) else {
-        return self.url
+      return self.url
     }
     if areThereVariables(in: body) {
-        guard let serializedVariables = try? serializationFormat.serialize(value: body.jsonObject["variables"]) else {
-            return URL(string: "\(self.url.absoluteString)?\(queryParam)")
-        }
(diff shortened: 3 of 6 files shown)


Sources/Apollo/GraphQLGETTransformer.swift
```diff
@@ -0,0 +1,63 @@
...
+//  GraphQLGETTransformer.swift
+//  Apollo
+//
+//  Created by Ellen Shapiro on 7/1/19.
+//  Copyright © 2019 Apollo GraphQL. All rights reserved.
+//
+
+import Foundation
+
+struct GraphQLGETTransformer {
```
martijnwalraven: Breaking this out and testing it is a great improvement. Just wondering why you decided to make this a struct instead of a top-level function. Were you thinking of extending this with additional methods or multi-stage building in the future?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -1,11 +1,32 @@
 import Foundation
 
+public protocol HTTPNetworkTransportDelegate: class {
```
martijnwalraven: I've blocked this way too long and I want to leave it to you to decide what to do, but [in the past](https://github.com/apollographql/apollo-ios/pull/297#issuecomment-400535981) I've seen it as a hard requirement that our solution should also support retries on error and token refresh, which probably requires at least some callbacks to be asynchronous. See [here](https://github.com/Alamofire/Alamofire/blob/master/Documentation/AdvancedUsage.md#adapting-and-retrying-requests) for the way Alamofire supports this for example. Maybe it would be enough to make `shouldSend` asynchronous?

martijnwalraven: Here are some other previous ideas for delegate methods: https://github.com/apollographql/apollo-ios/pull/257#issuecomment-381837750

martijnwalraven: There was also this idea about propagating custom attributes from client API methods in a context argument: https://github.com/apollographql/apollo-ios/pull/257#issuecomment-381400610

martijnwalraven: Not saying you should do anything with this, I know I've probably made it too complicated in the past :)

Sources/Apollo/GraphQLGETTransformer.swift
designatednerd: I prefer to make things named structs/classes/enums instead of free-floating functions. This is partly to make it easier to break things into additional methods in the future, but it's mostly for clarity of where the hell something is coming from.

Sources/Apollo/HTTPNetworkTransport.swift
RolandasRazma: retries on token error and refresh is supper useful, we have that in our code. Helps with wrong dates on client and token format changes on server

designatednerd: I based a bunch of this on [your suggestions here](https://github.com/apollographql/apollo-ios/pull/257#issuecomment-381837750). 

What am I missing about `Context` that wouldn't be covered by just altering the `Request` object? 

I'm gonna take a look at making response handling a separate delegate, since not everyone who needs request handling needs response handling, and optional delegate methods require all sorts of objc nonsense. 

I'll admit that I'm kiiiind of reluctant to do custom retry logic due to the additional complexity, but I'll see if I can make that work.

designatednerd:
@martijnwalraven Added a couple more delegates to allow retrying and/or handling the raw request + response

Sources/Apollo/GraphQLHTTPRequestError.swift
```diff
@@ -1,10 +1,13 @@
 /// An error which has occurred during the serialization of a request.
 public enum GraphQLHTTPRequestError: Error, LocalizedError {
+  case cancelledByDeveloper
   case serializedBodyMessageError
   case serializedQueryParamsMessageError
   
   public var errorDescription: String? {
     switch self {
+    case .cancelledByDeveloper:
+      return "The request was cancelled by the developer using the HTTPNetworkTransportPreflightDelegate."
```
koenpunt: Not sure if this should mention "the developer", because then you could attribute almost all implementation to "the developer".

Maybe something like `cancelledByPreflight` with "The request was cancelled by the the HTTPNetworkTransportPreflightDelegate" is more suitable?

designatednerd: Fair point

designatednerd:
This is now out of Fake Draft Because You Can't Turn An Open PR Into A Draft PR status and is ready for review 😃

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Response handling and retry delegates have now been added.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -1,11 +1,78 @@
...
+  /// NOTE: Don't just call the `retryHandler` with `true` all the time, or you can potentially wind up in an infinite loop of errors
+  ///
+  /// - Parameters:
+  ///   - networkTransport: The network transport which received the error
+  ///   - error: The received error
+  ///   - request: The URLRequest which generated the error
+  ///   - response: [Optional] Any response received when the error was generated
+  ///   - retryHandler: A closure indicating whether the operation should be retried. Asyncrhonous to allow for re-authentication or other async operations to complete.
+  func networkTransport(_ networkTransport: HTTPNetworkTransport,
+                        receievedError error: Error,
```
martijnwalraven: Small typo, should be `receivedError`.

Sources/Apollo/GraphQLHTTPRequestError.swift
```diff
@@ -1,10 +1,13 @@
 /// An error which has occurred during the serialization of a request.
 public enum GraphQLHTTPRequestError: Error, LocalizedError {
+  case cancelledByDeveloper
```
koenpunt: I think the enum case should be changed too, for the same reason the `errorDescription` did.

koenpunt:
Great to see some activity on this repo again, thanks!

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -1,11 +1,78 @@
...
+
+// MARK: -
+
 /// A network transport that uses HTTP POST requests to send GraphQL operations to a server, and that uses `URLSession` as the networking implementation.
 public class HTTPNetworkTransport: NetworkTransport {
   let url: URL
   let session: URLSession
   let serializationFormat = JSONSerializationFormat.self
   let useGETForQueries: Bool
+  let preflightDelegate: HTTPNetworkTransportPreflightDelegate?
```
martijnwalraven: I'm a little worried about a proliferation of delegate properties here. I can see the benefit of keeping these separate, but it may also be worth simplifying this to a single `delegate` property that can conform to any of these delegate protocols (similar to how `URLSession`s delegate can conform to `URLSessionDelegate`, `URLSessionTaskDelegate`, etc.).

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -1,11 +1,78 @@
...
+  /// A callback to allow hooking in URL session responses for things like logging and examining headers.
+  /// NOTE: This will call back on whatever thread the URL session calls back on, which is never the main thread. Call `DispatchQueue.main.async` before touching your UI!
+  ///
+  /// - Parameters:
+  ///   - networkTransport: The network transport that completed a task
+  ///   - request: The request which was completed by the task
+  ///   - data: [optional] Any data received. Passed through from `URLSession`.
+  ///   - response: [optional] Any response received. Passed through from `URLSession`.
+  ///   - error: [optional] Any error received. Passed through from `URLSession`.
+  func networkTransport(_ networkTransport: HTTPNetworkTransport,
```
martijnwalraven: For consistency with platform methods maybe something like `networkTransport:didCompleteTaskForRequest:withData:response:error`? So `didComplete` instead of `completed`, and I'm not sure whether the `raw` part is necessary? Alternatively, we could leave out the `task` altogether, and make it something like `networkTransport:didCompleteRequest:withData:response:error`.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: The problem is that `URLSessionDelegate` et al are Objective-C protocols, so they can have optional methods. If you only want to implement method `foo()`, then that's all you have to implement.

Our Swift protocols can't unless we mark the protocol `@objc`, which has a ton of other knock-on effects we don't want. This means if you're implementing a single protocol, you have to implement ALL of its methods even if you don't need or even want them. 

I agree that having to have things conforming to multiple protocols is annoying, but I think the separation of concerns (and the ability to avoid needing to implement EVERYTHING just to get one thing) is worth it.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Agree on `didComplete` vs `completed`, will update. 

The `Raw` is to indicate that this is unprocessed data (and that the consumer shouldn't necessarily be trying to process it themselves). I started with `didCompleteRequest` but I went to the bit with `Raw` because I didn't want it to be confused by users with the actual completion closure.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: 🤦‍♀️If i spell `received` right on the first try ONCE IN MY LIFE I will be happy.

Sources/Apollo/HTTPNetworkTransport.swift
martijnwalraven: Oh, I missed this before, but to answer your question about `Context`: this would allow you to pass in additional data from `ApolloClient` methods like `fetch` (so per request, at the point of use), as opposed to doing it from the delegate. Probably not needed right now, but for context :)

Sources/Apollo/HTTPNetworkTransport.swift
martijnwalraven: Ah, I'm ok with having different delegate protocols, but what I meant was that we could still have only one `delegate` property. I believe that is similar to what `URLSession`'s delegate does, so your delegate has to conform to `URLSessionDelegate` and optionally to `URLSessionTaskDelegate` etc.

designatednerd: `URLSessionDelegate` is the base protocol which inherits from `NSObjectProtocol` - `URLSessionTaskDelegate` et al inherit from `URLSessionDelegate. Since these are all Obj-C, they can have optional methods, which pure Swift protocols can't. 

We could set up some kind of inheritance with our pure Swift protocols to allow passing the highest level as a single delegate, but the user would still need to implement all protocols in the inheritance tree, which again may not be what they need or want.

designatednerd: AHA. Talking through with Martijn I realized there's a way around this: [An empty base protocol](https://gist.github.com/designatednerd/a3d5033a43ad9ac1fb3125b4589a7cfc). Implementing!

naknut:
Hi @designatednerd thanks so much for fixing this, I've been looking forward to the header reconfiguration for a long time. Is there any documentation on how to actually do reconfiguration of headers? I tried looking through the code in this PR but I feel a little lost.

designatednerd:
Nothing formal yet, but you can can see how we're altering the request if you look [at the tests' implementation of `willSend`](https://github.com/apollographql/apollo-ios/pull/602/files#diff-e5cf1b0debf2df0422fbac165b993ba5R202). Does that help?

naknut:
Yes I was looking through that but couldn’t really figure it out. But I’ll have an other look!

designatednerd:
Basically since `request.allHTTPHeaderFields` can't just have stuff appended to it, you copy the existing header fields, insert whatever you need, and re-assign the copy that now includes your additional fields to the request. 

Since `request` is an `inout` variable, you don't need to return anything, it just modifies the variable being used directly.  Does that help?

alouanemed:
Awesome work there @designatednerd, I wonder when we can see this shipped to the next update? Thanks!

designatednerd:
Hoping for early next week!

PR #608, opened 2019-07-02
Feature: Auto Persisted Queries with CDN Support

Codebear98:
**Usage**
--
```
let networkTransport = HTTPNetworkTransport(
      url: url,
      configuration: sessionConfiguration,
      useGETForQueries: false,
      enableAutoPersistedQueries: true,
      useHttpGetMethodForPersistedQueries: true)

let client = ApolloClient(networkTransport: networkTransport)
```
**Features**
- Support APQs
- Support use `HTTPGetMethod` for all APQs Requests by setting `useGETForQueries=true`
- Support use `HTTPGetMethod` only for supported APQs request by setting `useGETMethodForPersistedQueries=true` , according to https://www.apollographql.com/docs/apollo-server/features/apq/#cdn

`useGETMethodForPersistedQueries=true` is aligning with apollo-android [#1376](https://github.com/apollographql/apollo-android/pull/1376)

1. it sends GetMethod for the 1st Request.
2. If server dose not have cache or not support APQs with error message `PersistedQueryNotFound` or  `PersistedQueryNotSupported`
3. Then it fallbacks to retry with `POST` Request (retry with `GET` when `useGETForQueries=true`) and includes `QueryDocument` for registering APQs.

`useGETMethodForPersistedQueries` is introduced due to an error `exceeds limit on url query string` may occur on some sever when query string exceeds 2k limit with large amount of data such as QueryDoucment.

**Limitation**
--
1. To enable APQs,`sha256hash` of `QueryDocument` is required for each `QueryOperation`. 
You need to Modify Apollo runscript in `YOUR PROJECT` to also generate `operationIdentifier` (`sha256hash`) to your API.swift (please see below)

2. Must use Apollo 1.9.2, since there is a bug in `Apollo-tooling` while generating `operationIdentifier`
The Query and hash does not match when the `Query` contains `fragment`.  It's due to an extra `\n` added in `QueryDocument` for hashing. But the `\n` is missing in Swift CodeGen.
https://github.com/apollographql/apollo-tooling/issues/1362

There is a hacking workaround in `HTTPNetworkTransport.swift` line212, It adds back the `\n` to the `QueryDocument`. 
  ```
  if sendQueryDocument {
      // TODO: This work-around fix "operationId is invalid for swift codegen" (https://github.com/apollographql/apollo-tooling/issues/1362), please remove this work-around after it's fixed.
      let modifiedQuery = operation.queryDocument.replacingOccurrences(of: "fragment", with: "\nfragment")
      payload["query"] = modifiedQuery
    }
```

**Apollo runscript**
add `--operationIdsPath=operationIdsPath` for swift codegen 
```

# Type a script or drag a script file from your workspace to insert its path.
APOLLO_FRAMEWORK_PATH=$(eval find $FRAMEWORK_SEARCH_PATHS -name "Apollo.framework" -maxdepth 1 -print | head -n 1)

if [ -z "$APOLLO_FRAMEWORK_PATH" ]; then
echo "error: Couldn't find Apollo.framework in FRAMEWORK_SEARCH_PATHS; make sure to add the framework to your project."
exit 1
fi

# Download schema.json first 
# apollo schema:download --endpoint=http... --header="apikey: ..."
cd "${SRCROOT}/${TARGET_NAME}/DataSource/GraphQL"
$APOLLO_FRAMEWORK_PATH/check-and-run-apollo-cli.sh codegen:generate --passthroughCustomScalars --queries="$(find . -name '*.graphql')" --operationIdsPath=operationIdsPath --schema=schema.json Generated/GraphQLAPI.swift

```

### CODE REVIEW

**Comment by designatednerd** [comment]
@Codebear98 we've got some changes in progress with this with #599 and #602 - you may want to hold off until those get merged!

**Comment by Codebear98** [comment]
> @Codebear98 we've got some changes in progress with this with #599 and #602 - you may want to hold off until those get merged!

cool, make sense for #599, probably can align with apollo-android.

**Comment by Codebear98** [comment]
PR updated following the design of #599 and #602  @designatednerd please take a look, thanks!

File: Sources/Apollo/GraphQLGETTransformer.swift [hunk 1]
```diff
@@ -35,26 +32,23 @@ struct GraphQLGETTransformer {
...
+      _ = try body.compactMap({ arg in
+        if let value = arg.value as? GraphQLMap {
+          let data = try JSONSerialization.data(withJSONObject: value.jsonValue)
+          if let string = String(data: data, encoding: .utf8) {
+            queryItems.append(URLQueryItem(name: arg.key, value: string))
+          }
+        } else if let string = arg.value as? String {
+          queryItems.append(URLQueryItem(name: arg.key, value: string))
+        }
+      })
```
designatednerd: So one thing I would be very cautious about with this change is the fact that it makes the query strings non-deterministic, as shown by the fact that you had to add additional strings to the tests. 

Can you explain why this is necessary?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,31 +69,38 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate
...
   let url: URL
-  let session: URLSession
+  var session: URLSession
   let serializationFormat = JSONSerializationFormat.self
-  let useGETForQueries: Bool
   let delegate: HTTPNetworkTransportDelegate?
 
+  private let _useGETForQueries: Bool
+  private let _enableAutoPersistedQueries: Bool
+  private let _useHttpGetMethodForPersistedQueries: Bool
```
designatednerd: I'm trying to get us away from underscores for private properties - can you please remove those?

Tests/Apollo/Network/HttpNetworkTransportTests.swift
```diff
@@ -0,0 +1,342 @@
...
+  func testRequestBody() {
+    let network = HTTPNetworkTransport(url: URL(string: endpoint)!)
+    let mockSession = MockURLSession()
+    network.session = mockSession
+    let query = HeroNameQuery()
+    let _ = network.send(operation: query) { _,_ in }
+    let request = mockSession.lastRequest
+    
+    XCTAssertNotNil(request)
+    XCTAssert(request!.url?.host == network.url.host)
```
designatednerd: Heads up that failed force-unwraps in tests crash the whole test suite rather than failing the test. Generally I prefer to do this instead: 

```swift
guard let foo = bar else {
   XCTFail("No value for `bar`!")
   return 
}
```

That way the test with a nil value fails with a clear error message and other tests can still run.

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
```diff
@@ -3,9 +3,64 @@ import XCTest
...
+
+protocol TestConfig {
+  func network() -> HTTPNetworkTransport
+}
+
+class DefaultConfig: TestConfig {
+  func network() -> HTTPNetworkTransport {
+    return HTTPNetworkTransport(url: URL(string: "http://localhost:8080/graphql")!,
+                                enableAutoPersistedQueries: false,
+                                useHttpGetMethodForPersistedQueries: false)
```
designatednerd: since these both default to false, you can omit the params here. That also helps ensure that nobody changes the default values out from under us.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,31 +69,38 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate
...
+  
   /// Creates a network transport with the specified server URL and session configuration.
   ///
   /// - Parameters:
   ///   - url: The URL of a GraphQL server to connect to.
   ///   - configuration: A session configuration used to configure the session. Defaults to `URLSessionConfiguration.default`.
-  ///   - sendOperationIdentifiers: Whether to send operation identifiers rather than full operation text, for use with servers that support query persistence. Defaults to false.
   ///   - useGETForQueries: If query operation should be sent using GET instead of POST. Defaults to false.
+  ///   - enableAutoPersistedQueries: Whether to send persistedQuery extension. QueryDocument will be absent at 1st request, retry with QueryDocument if server respond PersistedQueryNotFound or PersistedQueryNotSupport. Defaults to false.
+  ///   - useHttpGetMethodForPersistedQueries: Whether to send PersistedQuery supported request with HTTPGETMethod, retry with HTTPPOSTMethod if PersistedQuery not support/not found in server.
```
designatednerd: Why does this need to be a separate parameter instead of respecting the user's preference in `useGETForQueries`?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -105,106 +112,166 @@ public class HTTPNetworkTransport: NetworkTransport {
...
   }
-
-  private let sendOperationIdentifiers: Bool
   
+  // retryByDefault: still retry if delegate is absent, otherwise repect the value from delegate
   private func handleErrorOrRetry<Operation>(operation: Operation,
                                              error: Error,
                                              for request: URLRequest,
                                              response: URLResponse?,
+                                             retryByDefault: Bool = false,
```
designatednerd: I think it might be better to keep this tied to the set value of APQ on the instance rather than passing it as a parameter - that'd make it considerably clearer where this is coming from.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -105,106 +112,166 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+        response: response,
+        retryHandler: { [weak self] shouldRetry in
+          guard let self = self, shouldRetry else {
+            completionHandler(nil, error)
+            return
+          }
+          _ = self.send(operation: operation, retryFor: error, completionHandler: completionHandler)
+      })
+    } else if retryByDefault {
+      _ = self.send(operation: operation, retryFor: error, completionHandler: completionHandler)
```
designatednerd: I think it would be clearer to check if APQ is enabled && the error is an APQ error first, then either re-send the query if it is, or continue going through and checking for the delegate if it's not. It'd probably mean moving the `retry` logic from that second `send` method down here - which think is probably clearer in the end.

Tests/Apollo/Network/HttpNetworkTransportTests.swift
```diff
@@ -0,0 +1,342 @@
+import XCTest
+@testable import Apollo
+import ApolloTestSupport
+import StarWarsAPI
+
+class HttpNetworkTransportTests: XCTestCase {
```
designatednerd: I think this is more "Automatic Persisted Queries Tests"  than something generic about the network transport - do you mind renaming?

Tests/Apollo/Network/HttpNetworkTransportTests.swift
```diff
@@ -0,0 +1,342 @@
...
+    
+    validateUrlParams(with: request!,
+                      query: query,
+                      queryDocument: false,
+                      persistedQuery: true,
+                      variable: true)
+  }
+}
+
+// Mark: Helpers
```
designatednerd: The `// MARK:` directive doesn't really do anything in Swift unless you use all uppercase and a semicolon.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -222,11 +289,16 @@ public class HTTPNetworkTransport: NetworkTransport {
...
   }
   
-  private func createRequest<Operation: GraphQLOperation>(for operation: Operation) throws -> URLRequest {
-    let body = requestBody(for: operation)
-    var request = URLRequest(url: self.url)
+  private func createRequest<Operation: GraphQLOperation>(for operation: Operation,
+                                                          httpMethod: GraphQLHTTPMethod,
+                                                          sendQueryDocument: Bool,
+                                                          autoPersistQueries: Bool
+        ) throws -> URLRequest {
```
designatednerd: can you pull these onto the same line as the last parameter to match the rest of the code please?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -259,15 +333,54 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+      let modifiedQuery = operation.queryDocument.replacingOccurrences(of: "fragment", with: "\nfragment")
+      payload["query"] = modifiedQuery
+    }
+    
+    return payload
+  }
+}
+
+extension URLSession {
+  func synchronousDataTask(with request: URLRequest) -> (Data?, URLResponse?, Error?) {
```
designatednerd: I'm not sure if I'm missing something here, but this does not appear to be used. Can we get rid of it?

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -257,6 +257,8 @@ extension HTTPTransportTests: HTTPNetworkTransportRetryDelegate {
       retryHandler(true)
     case .invalidResponse:
       retryHandler(false)
+    case .persistedQueryNotFound,  .persistedQueryNotSupported:
```
designatednerd: mind bumping the 2nd case onto a new line? Makes it easier to tell how many cases are being handled with a single statement.

designatednerd:
One thing that @sirensoftitan pointed out when I closed his PR is that this ties APQ to the `HTTPNetworkTransport` - it would probably be good to figure out a way to make this also work for the `WebSocketTransport` as well.

Codebear98:
> ould probably be good to figure out a way to make this also work for the `WebSocketTransport` as well.

Probably refactor to decouple the request payload part out from NetworkTransport?
I'm not familiar with ApolloWebSocketClient, does it handle the same way?
Can also update the WebSocketTransport to currently progress.

Sources/Apollo/GraphQLGETTransformer.swift
Codebear98: Oh, the original idea is assuming `GraphQLGETTransformer` should not have logic about the usage, so it has less dependency between APQs or other request, variables or query, but just convert the GraphQLMap to queryString with the support of multiple deep levels dictionary. As APQs requires 2 dimensions dict body.

To solve the `non-deterministic` issue which coming from the natural of dictionary. What do you think if sorting the allKeys first before constructing the queryItems array?

designatednerd: Yeah I tried that. It unfortunately didn't help much.

I wound up creating some stuff in #628 that uses sorted keys where possible and then does tests to make sure stuff actually gets shoved in there when sorted keys are not possible. Would that work for you?

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: Well, `useGETForQueries` is applying to all query requests, including those with query document.
Query doucement can be very large depends on usage. 

And query string will be super long too.
For those Graphql server which restricts the length of url for security reason would return error of `exceed url length limit`. 
In fact, many of http server restricts url length to 2k by default.

So we need another `useHttpGetMethodForPersistedQueries` for those who wanna set limit to their server for security or other reason. To only apply Get for those APQs succeed Requests (which without query document).

Sources/Apollo/GraphQLGETTransformer.swift
Codebear98: Cool! will try on that method.

Codebear98: updated with sorting and added a few test cases to cover 2lv GraphQLMap

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: updated

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
Codebear98: updated

Tests/Apollo/Network/HttpNetworkTransportTests.swift
Codebear98: updated

Tests/Apollo/Network/HttpNetworkTransportTests.swift
Codebear98: updated

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: removed

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: updated and other method too.

Tests/ApolloTests/HTTPTransportTests.swift
Codebear98: updated

Tests/Apollo/Network/HttpNetworkTransportTests.swift
Codebear98: updated and other too.

Sources/Apollo/GraphQLGETTransformer.swift
Codebear98: what do u mean by hand `body` straight to the `dataSortedIfPossible`

like this?
```
        let data = try JSONSerialization.dataSortedIfPossible(withJSONObject: body.jsonValue)
        if let string = String(data: data, encoding: .utf8) {
        
        }
```
you will got a string of the follow and can't pass as query string
`
"{\"extensions\":{\"persistedQuery\":{\"sha256Hash\":\"f6e76545cd03aa21368d9969cb39447f6e836a16717823281803778e7805d671\",\"version\":1}},\"query\":\"query HeroName($episode: Episode) {\\n  hero(episode: $episode) {\\n    __typename\\n    name\\n  }\\n}\",\"variables\":{\"episode\":null}}"
`

There will be query parameters of 
`query=xxx`, `variables=xxx`, `extensions=xxx`
and u can't  serialise it if `xxx` is already a string

designatednerd: AHA. That's what I was missing. Thank you!

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -69,30 +69,38 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate
 
 /// A network transport that uses HTTP POST requests to send GraphQL operations to a server, and that uses `URLSession` as the networking implementation.
 public class HTTPNetworkTransport: NetworkTransport {
+  
   let url: URL
-  let session: URLSession
+  var session: URLSession
```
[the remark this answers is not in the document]
Codebear98: Oh, It's used to assign MockSession to intercept the Request.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -104,110 +112,160 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+      
       do {
-        guard let body = try self?.serializationFormat.deserialize(data: data) as? JSONObject else {
+        guard let body = try self.serializationFormat.deserialize(data: data) as? JSONObject else {
           throw GraphQLHTTPResponseError(body: data, response: httpResponse, kind: .invalidResponse)
         }
-        let response = GraphQLResponse(operation: operation, body: body)
-        completionHandler(response, nil)
+        if self.enableAutoPersistedQueries,
+          let error = body["errors"] as? [JSONObject],
```
RolandasRazma: this is getting bit messy here... How about creating `GraphQLResponse` (else path) and adding method to it to get errors. That way you don't need to know here much about `JSONObject` and `GraphQLResponse` already knows about it.

```swift
let response = GraphQLResponse(operation: operation, body: body)

if response.errors.contain(...) {

} else {
   completionHandler(response, nil)
}

File: Tests/ApolloTests/GETTransformerTests.swift [hunk 15]
```diff
@@ -86,6 +88,133 @@ class GETTransformerTests: XCTestCase {
...
+        XCTFail("query should not nil")
+        return
+      }
+      XCTAssertTrue(query == operation.queryDocument)
+      
+      guard let variables = url?.queryItems?["variables"] else {
+        XCTFail("variables should not nil")
+        return
+      }
+      XCTAssert(variables == "{\"episode\":\"EMPIRE\"}")
```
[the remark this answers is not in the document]
Codebear98: updated thanks

Tests/Apollo/Network/AutomaticPersistedQueriesTests.swift
```diff
@@ -0,0 +1,369 @@
...
+extension URLQueryItem {
+  var dictionaryRepresentation: [String: String]? {
+    if let value = value {
+      return [name: value]
+    }
+    return nil
+  }
+}
+
+private func +<Key, Value> (lhs: [Key: Value], rhs: [Key: Value]) -> [Key: Value] {
```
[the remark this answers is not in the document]
Codebear98: no particular reason, just removed +<Key, Value> and included the logic in reduce.

Tests/Apollo/Network/AutomaticPersistedQueriesTests.swift
```diff
@@ -0,0 +1,369 @@
...
+    }
+    
+    XCTAssert(version == 1)
+    XCTAssert(sha256Hash == query.operationIdentifier)
+  } else {
+    XCTAssertNil(ext)
+  }
+}
+
+private final class MockURLSession: URLSession {
```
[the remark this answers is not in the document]
Codebear98: updated

Sources/Apollo/GraphQLHTTPResponseError.swift
```diff
@@ -5,13 +5,19 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
...
     var description: String {
       switch self {
       case .errorResponse:
         return "Received error response"
       case .invalidResponse:
         return "Received invalid response"
+      case .persistedQueryNotFound:
+        return "Persisted query not found"
+      case .persistedQueryNotSupported:
+        return "Persisted query not support"
```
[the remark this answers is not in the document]
Codebear98: updated

Apollo.xcodeproj/xcshareddata/xcschemes/GitHubAPI.xcscheme
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1020"
+   version = "1.3">
```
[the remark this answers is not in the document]
Codebear98: unshared

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
Codebear98: oh sorry, I missed that part, updated! thanks!

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: updated

Sources/Apollo/HTTPNetworkTransport.swift
Codebear98: good idea, will update accordingly!

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -258,15 +319,31 @@ public class HTTPNetworkTransport: NetworkTransport {
...
-  private func requestBody<Operation: GraphQLOperation>(for operation: Operation) -> GraphQLMap {
-    if sendOperationIdentifiers {
+  private func requestBody<Operation: GraphQLOperation>(for operation: Operation,
+                                                        sendQueryDocument: Bool,
+                                                        autoPersistQueries: Bool) -> GraphQLMap {
+    
+    var payload: GraphQLMap = [:]
+    
+    if autoPersistQueries {
       guard let operationIdentifier = operation.operationIdentifier else {
```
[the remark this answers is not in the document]
RolandasRazma: `--operationIdsPath` is path to JSON file that is tooling dependant and/or can be modified as you see fit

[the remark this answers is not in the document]
Codebear98: The question here is Do we need to support autoPersistQueries just for some request but not for all?
As the autoPersistQueries seems network traffic optimisation like CDN caching which should be applying to all or none because it depends server side ability to support or not.

I just worry if we don't `preconditionFailure` for the pre-gen code that's not fulfilling the APQs requirement (but dev opt to turn on APQs). Developers will not be able to get alert about the run script missing `operationIdsPath` until they look into the traffic logging and find out APQs isn't in place.

In addition, I think the JSON file will not affect the code gen? I mean even you modify the JSON file, it won't affect operation.operationIdentifier in the generated Code?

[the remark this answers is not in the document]
designatednerd: IIRC the operation identifier is based on the hash of the body, so it shouldn't change based on the location of the codegen. 

Personally I'm good with leaving this as a precondition failure - @codebear98 is correct, if someone wants APQ's and hasn't set up generating operation IDs, we want this to fail loud, hard, and early.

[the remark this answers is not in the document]
RolandasRazma: fair enough

Sources/Apollo/HTTPNetworkTransport.swift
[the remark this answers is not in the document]
designatednerd: ok - let's go with this for now. I'm planning to get some stuff allowing a URL session to be passed in with an initializer soon.

Tests/ApolloCacheDependentTests/StarWarsServerTests.swift
```diff
@@ -3,9 +3,61 @@ import XCTest
...
+class APQsConfig: TestConfig {
+  func network() -> HTTPNetworkTransport {
+    return HTTPNetworkTransport(url: URL(string: "http://localhost:8080/graphql")!,
+                                enableAutoPersistedQueries: true)
+  }
+}
+
+class APQsWithGetMethodConfig: TestConfig, HTTPNetworkTransportRetryDelegate{
+  func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedError error: Error, for request: URLRequest, response: URLResponse?, retryHandler: @escaping (Bool) -> Void) {
+    retryHandler(true)
```
designatednerd: Do you want to check that the received error is of the correct type? Otherwise this could lead to an infinite loop.

Tests/ApolloTestSupport/MockURLSession.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  MockURLSession.swift
+//  ApolloTestSupport
+//
+//  Copyright © 2019 Apollo GraphQL. All rights reserved.
+//
+
+import Foundation
+
+public final class MockURLSession: URLSession {
```
designatednerd: Does all this need to be `public`?

Tests/ApolloTests/GETTransformerTests.swift
```diff
@@ -25,7 +25,9 @@ class GETTransformerTests: XCTestCase {
     
     let url = transformer.createGetURL()
     
-    XCTAssertEqual(url?.absoluteString, "http://localhost:8080/graphql?query=query%20HeroName($episode:%20Episode)%20%7B%0A%20%20hero(episode:%20$episode)%20%7B%0A%20%20%20%20__typename%0A%20%20%20%20name%0A%20%20%7D%0A%7D&variables=%7B%22episode%22:%22EMPIRE%22%7D")
+    let queryString = url?.absoluteString == "http://localhost:8080/graphql?query=query%20HeroName($episode:%20Episode)%20%7B%0A%20%20hero(episode:%20$episode)%20%7B%0A%20%20%20%20__typename%0A%20%20%20%20name%0A%20%20%7D%0A%7D&variables=%7B%22episode%22:%22EMPIRE%22%7D"
+    
+    XCTAssertTrue(queryString)
```
designatednerd: Why was this changed?

Tests/ApolloTests/GETTransformerTests.swift
```diff
@@ -96,8 +225,69 @@ class GETTransformerTests: XCTestCase {
...
+    
+    if #available(iOS 11, macOS 13, watchOS 4, tvOS 11, *) {
+    let queryString = url?.absoluteString == "http://localhost:8080/graphql?extensions=%7B%22persistedQuery%22:%7B%22sha256Hash%22:%22f6e76545cd03aa21368d9969cb39447f6e836a16717823281803778e7805d671%22,%22version%22:1%7D%7D&query=query%20HeroName($episode:%20Episode)%20%7B%0A%20%20hero(episode:%20$episode)%20%7B%0A%20%20%20%20__typename%0A%20%20%20%20name%0A%20%20%7D%0A%7D&variables=%7B%22episode%22:null%7D"
+      XCTAssertTrue(queryString)
+    } else {
+      guard let variables = url?.queryItems?["variables"] else {
+        XCTFail("variables should not nil")
+        return
+      }
+      XCTAssertEqual(variables, "{\"episode\":null}")
```
designatednerd: you can also check `variables.keys.contains("episode")` and then check that the value for that is null rather than checking the raw string.

Tests/ApolloTests/GETTransformerTests.swift
```diff
@@ -109,6 +299,6 @@ class GETTransformerTests: XCTestCase {
     let transformer = GraphQLGETTransformer(body: body, url: self.url)
     
     let url = transformer.createGetURL()
-    XCTAssertNil(url)
+    XCTAssertEqual(url?.absoluteString, "http://localhost:8080/graphql?variables=%7B%22episode%22:%22EMPIRE%22%7D")
```
designatednerd: I think I wound up deleting this test in master, you're welcome to as well.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -235,14 +290,16 @@ public class HTTPNetworkTransport: NetworkTransport {
...
         request.httpMethod = GraphQLHTTPMethod.POST.rawValue
+        request.httpBody = try serializationFormat.serialize(value: body)
       } catch {
         throw GraphQLHTTPRequestError.serializedBodyMessageError
       }
     }
-    
+
     request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+    request.setValue(operation.operationIdentifier, forHTTPHeaderField: "X-APOLLO-OPERATION-ID")
```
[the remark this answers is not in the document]
Codebear98: yes, and I would also add `operationName` in APQs request body, as it's available now thanks for 2.16 tooling upgrade!

designatednerd:
@Codebear98 Do you have interest in continuing this, or do you mind if I take this PR over?

Codebear98:
@designatednerd sorry, I don’t wanna delay the progress, please take it over if u would, thanks!

designatednerd:
This feature can now be followed at #767

PR #610, opened 2019-07-03
Update documentation + Scripts for building

designatednerd:
This PR combines elements of #387 and #420 with some new stuff that I've validated will work for CocoaPods and Carthage, as well as updates some out of date documentation.


designatednerd:
(note that this may be a few days to merge due to 🇺🇸🎆🍻🎆🍻🎆🇺🇸 weekend)

PR #618, opened 2019-07-10
Update documentation to show how to use new delegates and cache policy

designatednerd:
In this PR: 
- Added instructions and example code on how to use the new sub-protocols of `HTTPTransportClientDelegate` that came out with v0.11.0
- Got rid of recommendation to use the app delegate as a singleton and just make an actual singleton object instead
- Updated `fetching-queries.md` to include information about using `GET` for queries and add the most recent cache policy addition.
- Fixed a 🤦‍♀️with inline documentation

### DIFF
--- a/docs/source/initialization.md
+++ b/docs/source/initialization.md
@@ -2,13 +2,21 @@
 title: Creating a client
 ---
 
-In most cases, you'll want to create a single shared instance of `ApolloClient` and point it at your GraphQL server. The easiest way to do this is to define a global variable in `AppDelegate.swift`:
+## Basic Client Creation
+
+In most cases, you'll want to create a single shared instance of `ApolloClient` and point it at your GraphQL server. The easiest way to do this is to create a singleton:
 
 ```swift
-let apollo = ApolloClient(url: URL(string: "http://localhost:8080/graphql")!)
+class Apollo {
+  static let shared = Apollo() 
+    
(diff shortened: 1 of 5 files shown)


docs/source/initialization.md
```diff
@@ -2,26 +2,173 @@
...
 
-In most cases, you'll want to create a single shared instance of `ApolloClient` and point it at your GraphQL server. The easiest way to do this is to define a global variable in `AppDelegate.swift`:
+## Basic Client Creation
+
+In most cases, you'll want to create a single shared instance of `ApolloClient` and point it at your GraphQL server. The easiest way to do this is to create a singleton:
 
 ```swift
-let apollo = ApolloClient(url: URL(string: "http://localhost:8080/graphql")!)
+class Apollo {
+  static let shared = Apollo()
```
martijnwalraven: Is a singleton class really needed? What is the benefit over a global variable?

martijnwalraven: Ah, too fast, I see the example below with the delegate implementations.

designatednerd: Yeah that plus I'm very much team "If you're gonna use a singleton, use a singleton. Don't use the App Delegate as a pseudo-singleton as that's not what it's for."

designatednerd:
Also removed old link to CocoaDocs (which is dead) and fixed some stray whitespace in the iOS build script.

designatednerd:
I guess this is an approval?

Screen Shot 2019-07-11 at 5 40 07 PM

😃 Merging.

Issue #623, opened 2019-07-11
GET is not working in 0.11.0


dmandarino:
GET method released in `0.11.0` is not passing variables correctly, so it always fails when you have a variable in your query.

designatednerd:
@dmandarino Can you please give an example of how it's failing - is it a more complex variable that's not working or is it just a single variable?

dmandarino:
@designatednerd Yes, of course! Sorry about it. 
I've just opened a PR proposing a solution. #624.

When we use GET with queryVariable the class `GraphQLGETTransformer` makes a cast of ```self.body.jsonObject[self.variablesKey] as? [String: AnyHashable]```. But if always fails and return the url as if there weren't any variable.

So the request fails and returns a BadRequest, as the server is expecting for variable such as ID.

```["Received error response (400 bad request): {\"errors\":[{\"message\":\"Variable \\\"$id\\\" of required type \\\"ID!\\\" was not provided```

designatednerd:
Yeah that makes more sense. If you can add tests around that, it'd be super helpful.

dmandarino:
Hi @designatednerd, do you have any idea will be the next release with this fix? 
I can't use 0.11.0 because of this error and I really need this GET feature :/

designatednerd:
Will be in 0.12.0 - you can point Carthage or Cocoapods at the specific commit where this is fixed if you're really stuck in the meantime!

PR #624, opened 2019-07-11
[Fix] GET Request are failing due to variablesKey cast that always fails.

dmandarino:
#623

GET request are always failing because of a cast for in [String: AnyHashable] `GraphQLGETTransformer`. So I'm changing it to `[AnyHashable : Any]` because I've checked that it's the type we need to use to use custom headers too.


dmandarino:
@designatednerd Sure, I'm taking a look on it.

dmandarino:
@designatednerd it's still stucked. Is it usual to take that long? The other time it was so fast.

designatednerd:
Yeah during the day in the US the wait time for builds spikes pretty significantly, unfortunately. 😭

designatednerd:
@dmandarino looks like there's a couple failures in there where occasionally the order of parameters is not the same. I'll take a longer look tomorrow but I suspect the dictionary you're encoding is being generated in a different order. 

I'm annoyed that [`.sortedKeys` for JSONSerialization](https://developer.apple.com/documentation/foundation/jsonserialization/writingoptions/2888322-sortedkeys) is iOS 11+, I might have to write something for that to support it on older devices.

Sources/Apollo/GraphQLGETTransformer.swift
```diff
@@ -44,7 +44,7 @@ struct GraphQLGETTransformer {
     queryItems.append(URLQueryItem(name: self.queryKey, value: query))
     components.queryItems = queryItems
     
-    guard let variables = self.body.jsonObject[self.variablesKey] as? [String: AnyHashable] else {
+    guard let variables = self.body.jsonObject[self.variablesKey] as? [AnyHashable : Any] else {
```
designatednerd: So messing around in a playground then looking at [the docs for `JSONSerialization`](https://developer.apple.com/documentation/foundation/jsonserialization) I think this needs to be `[String: Any]` - JSON Serialization barfs if the keys aren't strings.

designatednerd:
Ugh, so I looked into this and getting `JSONSerialization` to sort consistently without using `.sortedKeys` is *stupidly* difficult. 

I'm almost certainly going to move codegen stuff to `Codable` in the medium term, so I didn't want to go too far down the rabbit hole of trying to get this to work with `JSONSerialization`, but it made the test about 900% more annoying to write. 

I've opened #628 with my solution to this - do you mind if we move forward with that one since I've already dealt with the test nonsense? Your work here was absolutely critical to getting that to work, and I really appreciate it - I'll make sure this PR gets linked in the changelog!

dmandarino:
@designatednerd don't worry, just do it! I just wanted to help this project. We all need this feature, specially for persisted query. And I really need this GET, because right now I'm using a fork of my own so the company I'm working on can use Apollo Client for iOS.

designatednerd:
I don't mind at all - let me know what you think, and I'll close this one out.

PR #626, opened 2019-07-11
File upload concept - Taken from #116

kimdv:
This code is taken from #116

### DIFF
--- a/Sources/Apollo/GraphQLFile.swift
+++ b/Sources/Apollo/GraphQLFile.swift
@@ -6,38 +6,38 @@ public struct GraphQLFile {
   let mimeType: String
   let inputStream: InputStream
   let contentLength: UInt64
-
+  
   public init(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", data: Data) {
     self.init(fieldName: fieldName, originalName: originalName, mimeType: mimeType, inputStream: InputStream(data: data), contentLength: UInt64(data.count))
   }
-
+  
   public init?(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", fileURL: URL) {
     guard let inputStream = InputStream(url: fileURL) else {
       return nil
     }
-
+    
     guard let contentLength = GraphQLFile.getFileSize(fileURL: fileURL) else {
       return nil
     }
-
+    
     self.init(fieldName: fieldName, originalName: originalName, mimeType: mimeType, inputStream: inputStream, contentLength: contentLength)
   }
-
+  
   public init(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", inputStream: InputStream, contentLength: UInt64) {
     self.fieldName = fieldName
     self.originalName = originalName
     self.mimeType = mimeType
-
+    
     self.inputStream = inputStream
     self.contentLength = contentLength
   }
-
+  
   private static func getFileSize(fileURL: URL) -> UInt64? {
     guard let fileSizeAttribute = try? FileManager.default.attributesOfItem(atPath: fileURL.path)[.size],
       let fileSize = fileSizeAttribute as? NSNumber else {
         return nil
     }
-
+    
     return fileSize.uint64Value
   }
 }
--- a/Sources/Apollo/GraphQLFile.swift
+++ b/Sources/Apollo/GraphQLFile.swift
@@ -12,9 +12,6 @@ public struct GraphQLFile {
   }
 
   public init?(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", fileURL: URL) {
-    // TODO: Better error handling
-
-
     guard let inputStream = InputStream(url: fileURL) else {
       return nil
     }
@@ -36,14 +33,11 @@ public struct GraphQLFile {
   }
 
   private static func getFileSize(fileURL: URL) -> UInt64? {
-    do {
-      guard let fileSize = try FileManager.default.attributesOfItem(atPath: fileURL.path)[.size] as? NSNumber else {
+    guard let fileSizeAttribute = try? FileManager.default.attributesOfItem(atPath: fileURL.path)[.size],
+      let fileSize = fileSizeAttribute as? NSNumber else {
         return nil
-      }
-
-      return fileSize.uint64Value
-    } catch {
-      return nil
     }
+
+    return fileSize.uint64Value
   }
 }
--- a/Sources/Apollo/GraphQLFile.swift
+++ b/Sources/Apollo/GraphQLFile.swift
@@ -1,0 +1,49 @@
+import Foundation
+
+public struct GraphQLFile {
+  let fieldName: String
+  let originalName: String
+  let mimeType: String
+  let inputStream: InputStream
+  let contentLength: UInt64
+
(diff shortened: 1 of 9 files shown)


kimdv:
@designatednerd I have started the PR. 
Tests are passing. 

Now I need to figure out how to test the upload part. 
Can you guide me here?

Sources/Apollo/GraphQLFile.swift
```diff
@@ -0,0 +1,49 @@
...
+  public init(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", inputStream: InputStream, contentLength: UInt64) {
+    self.fieldName = fieldName
+    self.originalName = originalName
+    self.mimeType = mimeType
+
+    self.inputStream = inputStream
+    self.contentLength = contentLength
+  }
+
+  private static func getFileSize(fileURL: URL) -> UInt64? {
```
designatednerd: I think since you're not doing anything with the error, it might be clearer to do something where you make use of `try?`, such as: 

I think this might be clearer as: 

```swift
guard 
    let fileSizeAttribute = try? FileManager.default.attributesOfItem(atPath: fileURL.path)[.size],
    let fileSize = fileSizeAttribute as? NSNumber else {
      return nil 
}

return fileSize.uint64Value
```

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -270,4 +281,26 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+	        formData.appendPart(string: data.debugDescription, name: name)
+	      }
+	    }
+
+	    for f in files {
+	      formData.appendPart(inputStream: f.inputStream, contentLength: f.contentLength, name: f.fieldName, contentType: f.mimeType, filename: f.originalName)
+	    }
+
+	    return formData
+	  }
```
designatednerd: indentation on this is alllll over the place, can you please update it?

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -270,4 +281,26 @@ public class HTTPNetworkTransport: NetworkTransport {
...
+	        let data = try! serializationFormat.serialize(value: data)
+	        formData.appendPart(data: data, name: name)
+	      } else if let data = data as? String {
+	        formData.appendPart(string: data, name: name)
+	      } else {
+	        formData.appendPart(string: data.debugDescription, name: name)
+	      }
+	    }
+
+	    for f in files {
```
designatednerd: a) This would probably be more readable as a `files.forEach` loop
b) Please don't just use single letters on things like this - if you want to use a `$0` in the `forEach` closure that's fine, but if you're gonna name something, please give it a name that is a full word. 😄

Sources/Apollo/MultipartFormData.swift
```diff
@@ -0,0 +1,111 @@
+public class MultipartFormData {
+
+	  public static let CRLF = "\r\n"
```
designatednerd: Indentation of this seems to have had some problems, could you please reindent this whole file?

Sources/Apollo/NetworkTransport.swift
```diff
@@ -9,5 +9,22 @@ public protocol NetworkTransport {
...
+  ///
+  /// - Parameters:
+  ///   - operation: The operation to send.
+  ///   - files: A list of files to send as a multipart request.
+  ///   - progressHandler: A closure to call periodically as the request is sent.
+  ///   - completionHandler: A closure to call when a request completes.
+  ///   - response: The response received from the server, or `nil` if an error occurred.
+  ///   - error: An error that indicates why a request failed, or `nil` if the request was succesful.
+  /// - Returns: An object that can be used to cancel an in progress request.
+  func upload<Operation: GraphQLOperation>(operation: Operation, files: [GraphQLFile]?, progressHandler: ((Progress) -> Void)?, completionHandler: @escaping (GraphQLResponse<Operation>?, Error?) -> Void) -> Cancellable
```
designatednerd: This is probably the biggest thing: Does this actually need to be on `NetworkTransport`, or should it be specific to `HTTPNetworkTransport`? I'd be quite surprised if anyone's sending files over websockets.

Sources/Apollo/GraphQLFile.swift
```diff
@@ -0,0 +1,49 @@
...
+  let mimeType: String
+  let inputStream: InputStream
+  let contentLength: UInt64
+
+  public init(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", data: Data) {
+    self.init(fieldName: fieldName, originalName: originalName, mimeType: mimeType, inputStream: InputStream(data: data), contentLength: UInt64(data.count))
+  }
+
+  public init?(fieldName: String, originalName: String, mimeType: String = "application/octet-stream", fileURL: URL) {
+    // TODO: Better error handling
```
designatednerd: This `TODO` should be resolved.

Sources/Apollo/NetworkTransport.swift
kimdv: Hmm. I don't think that. 
Maybe it would be better to only add it in `HTTPNetworkTransport` and if people wan't it in web sockets for some reason, we can always add it. 🥳

kimdv:
@designatednerd I have resolved the comments in the code! 
I reached out on Spectrum with some questions 🚀

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Indentation on this is still off - We're using 2-space indentation throughout the project, and that's set at the project level. If you select the whole file and do `ctrl-I` then it should indent it properly.

Sources/Apollo/MultipartFormData.swift
designatednerd: Same issue on the indentation here

Sources/Apollo/MultipartFormData.swift
```diff
@@ -0,0 +1,111 @@
+public class MultipartFormData {
```
kimdv: @designatednerd will it make sense to make this internal? 
As it is used for internal stuff?

designatednerd: You'll need to `@testable import Apollo` in the tests in order to access it, but you're right, this is an implementation detail. Great catch!

kimdv: Will remove `public`. 🚀

kimdv:
@designatednerd rebased and made it ready for review 🚀

kimdv:
Thanks a lot! 🚀

designatednerd:
Time's up, let's ship this!

Drusy:
Hello @designatednerd, could you provide a concret example how to use file upload ?
I can't find any official documentation on this topic.
How the `.graphql` file should look ?
Where `GraphQLFile` instances should be passed ?

Thanks in advance

designatednerd:
@Drusy Mind opening the lack of documentation as a new issue? I'll try to get that going later this week. The short answer is that you should use the `upload` function in the `HTTPNetworkTransport`.

Drusy:
Linked to #681 (File upload v0.13 - Missing documentation )

PR #628, opened 2019-07-12
Fix type error + Use Sorted Keys Where Possible

### LINKED ISSUE #623: GET is not working in 0.11.0

designatednerd:
Based on #624 and some of the failures we saw when @dmandarino tried to add a test, this fixes #623. 

In this PR: 
- Switch `[String: AnyHashable]` to `[String: Any]` to allow for proper encoding of types like booleans
- Add a convenience method to use `.sortedKeys` when generating JSON on platforms where that's possible
- Adapt @dmandarino's tests from #624 to use a stable-sorted json structure where possible, and add test that can run on both older and newer operating systems. 

_/me shakes fist at sky for `.sortedKeys` only being on iOS 11 and above_


designatednerd:
@dmandarino wanna give me a ✅if you feel this is good or feedback if you have it?

dmandarino:
@designatednerd Looks good to me. I've tested here too using the project I'm working on and everything looks fine and passed :D

PR #635, opened 2019-07-15
Separate Carthage optional dependencies

manicmaniac:
## Problem

When I install `apollo-ios` from Carthage, it always builds unnecessary optional dependencies like `SQLite.swift` and `Starscream`.
Building unused dependencies is not only a waste of time, but also it might lead a conflict with the version of user's dependency.
In fact, the current version designation is too strict for those who want to use `SQLite.swift` or `Starscream` itself.

https://github.com/apollographql/apollo-ios/blob/3a91bee6be5c4c42e35a8e372d9645f06b439b93/Cartfile

```
github "stephencelis/SQLite.swift" == 0.12.2
github "daltoniam/Starscream" == 3.1.0
```

I think the most of users of `apollo-ios` don't always use both of websocket and SQLite features.
Thus it shouldn't depend on them by default and only users have the right to decide whether to use the feature or not or which version to use.

Currently `apollo-ios`'s Podfile resolves this problem by using subspec feature.

## Cause

Since Carthage resolves dependency graph by parsing Cartfile, including optional dependencies into Cartfile causes this kind of problem.
Fortunately, Carthage has the neat solution: `Cartfile.private`.

## Solution

Renamed Cartfile to Cartfile.private to make dependencies optional.


designatednerd:
@manicmaniac What happens when you try to build on an empty Carthage project here - does it only build the `Apollo-iOS` framework and not the other two? I think if it tries to build the other two, it'll bomb out.

manicmaniac:
@designatednerd In fact, frameworks can be built individually even without building their dependencies (they are only needed at the time they are linked into an executable).

I have created the example repository including an empty iOS project and Cartfile.
In [this commit](https://github.com/manicmaniac/ApolloDependencyExample/commit/467f48cc83bd51a36693ded00045872542311525) I specified `apollographql/apollo-ios` on master branch in Cartfile.
Then in [the next commit](https://github.com/manicmaniac/ApolloDependencyExample/commit/987108b788babc506be0c06ea2f5cffd13d421f1), I specified `manicmaniac/apollo-ios`, which indicates this PR.
Both of them are integrated to Travis CI so that we can see what happens.

And the result is that both of them succeeded to build dependencies.
You can see build log [here for the first one](https://travis-ci.org/manicmaniac/ApolloDependencyExample/builds/558907485), and [here for the second one](https://travis-ci.org/manicmaniac/ApolloDependencyExample/builds/558908585).

Unfortunately I wrote the project filename incorrectly so the entire build was broken... but it doesn't affect the result.

designatednerd:
Welp, thanks for helping me figure out I needed to do #639. 🙃

Once that gets merged I'll try installing via Carthage again with this to see what I can figure out.

designatednerd:
@manicmaniac Can you please pull in the changes from `master` so I can try to get this to build locally? Thanks!

PR #644, opened 2019-07-17
`Result` ALL THE THINGS!!!

designatednerd:
This PR changes pretty much all of our completion closures to take advantage of Swift 5's built-in `Result` type. This is going to be a **SUPER-BREAKING CHANGE™**, though I believe it will help significantly in disambiguating whether a particular request has succeeded or failed, which is why I've done it in the first place. 

If you have reasons you think I should abandon this path, I'd love to hear them - please let me know in the comments!

In this PR: 
- Changed basically anything which had the following signature:
   ```swift
   completionHandler: ((Object?, Error?) -> Void)
  ```
  to use this signature instead: 
  ```swift
  completionHandler: ((Result<Object, Error>) -> Void)
  ```
  This takes advantage of Swift's built in `Result` type to make it WAY easier to tell when a request has failed prior to parsing. This does cause a couple of significant changes: 
  - Previously, there were cases where we would return neither a result nor an error. These cases will now always return a result.
  - This leads to some extra-fun and inscrutable compiler errors when you're going through and updating everything: 

    Screen Shot 2019-07-16 at 6 25 28 PM

    What this error is attemping to tell you is that you need to change the two-parameter completion closure into a single-parameter one. It's telling you that in a very, very opaque fashion, though. I plan to document this in the release notes as well. 
- Updated all tests - I'll flag the ones which are significant changes in comments. 
- Removed deprecated method `OperationResultHandler` typealias since I'm already breaking the crap out of everything

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -18,8 +18,7 @@ public enum CachePolicy {
 /// A handler for operation results.
 ///
 /// - Parameters:
-///   - result: The result of the performed operation, or `nil` if an error occurred.
-///   - error: An error that indicates why the mutation failed, or `nil` if the mutation was succesful.
+///   - result: The result of a performed operation. Will have a `GraphQLResult` with the data of the requested type on `success`, and an `Error` on `failure`.
 public typealias GraphQLResultHandler<Data> = (Result< GraphQLResult<Data>, Error>) -> Void
 
 /// The `ApolloClient` class provides the core API for Apollo. This API provides methods to fetch and watch queries, and to perform mutations.
@@ -226,7 +225,7 @@ private final class FetchQueryOperation<Query: GraphQLQuery>: AsynchronousOperat
           self.resultHandler(result)
           self.state = .finished
           return
-        }  
+        }
       }
       
       self.fetchFromNetwork()
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -20,7 +20,7 @@ public enum CachePolicy {
 /// - Parameters:
 ///   - result: The result of the performed operation, or `nil` if an error occurred.
 ///   - error: An error that indicates why the mutation failed, or `nil` if the mutation was succesful.
-public typealias GraphQLResultHandler<Data> = (_ result: GraphQLResult<Data>?, _ error: Error?) -> Void
+public typealias GraphQLResultHandler<Data> = (Result< GraphQLResult<Data>, Error>) -> Void
 
 /// The `ApolloClient` class provides the core API for Apollo. This API provides methods to fetch and watch queries, and to perform mutations.
 public class ApolloClient {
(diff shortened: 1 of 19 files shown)


Tests/ApolloCacheDependentTests/FetchQueryTests.swift
```diff
@@ -241,10 +264,24 @@ class FetchQueryTests: XCTestCase {
...
+              case .missingValue:
+                // Correct error!
+                break
+              default:
+                XCTFail("Unexpected JSON error: \(error)")
+              }
+            } else {
+              XCTFail("Unexpected error: \(error)")
+            }
+          }
```
designatednerd: This test represents a breaking change: Before if you cleared the cache and requested the same info again, you would get `nil` for both the `GraphQLResult` and the `Error`. Now you will receive an error since there is nothing to decode for that result.

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -89,19 +72,35 @@ public class WebSocketTransport: NetworkTransport, WebSocketDelegate {
...
+          let responseHandler = subscribers[id] {
+          if let payload = payload {
+            responseHandler(.success(payload))
+          } else if let error = error {
+            responseHandler(.failure(error))
+          } else {
+            let websocketError = WebSocketError(payload: payload,
+                                                error: error,
+                                                kind: .neitherErrorNorPayloadReceived)
+            responseHandler(.failure(websocketError))
```
designatednerd: This is a new error type to help handle cases where we previously could have sent `nil` for both the payload and the error.

Tests/ApolloCacheDependentTests/FetchQueryTests.swift
```diff
@@ -277,11 +314,26 @@ class FetchQueryTests: XCTestCase {
...
+              case .missingValue:
+                // Correct error!
+                break
+              default:
+                XCTFail("Unexpected JSON error: \(error)")
+              }
+          } else {
+            XCTFail("Unexpected error: \(error)")
+          }
+        }
```
designatednerd: This test also represents a breaking change - I *think* that this has an outer `GraphQLResultError` since there's some data in the cache, but not all of it.

designatednerd:
Twitter feedback on this change has been positive so far - gonna let @martijnwalraven tell me what I'm doing wrong tomorrow and then leave it open for further feedback for a day or two.

Tests/ApolloCacheDependentTests/FetchQueryTests.swift
martijnwalraven: That definitely seems correct to me, I think the previous behavior was a mistake.

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
```diff
@@ -68,18 +78,43 @@ class CachePersistenceTests: XCTestCase {
...
+        do {
+          try client.clearCache().await()
+        } catch {
+          XCTFail("Error Clearing cache: \(error)")
+          emptyCacheExpectation.fulfill()
+          return
+        }
 
-        client.fetch(query: query, cachePolicy: .returnCacheDataDontFetch) { (result, error) in
+        client.fetch(query: query, cachePolicy: .returnCacheDataDontFetch) { innerResult in
```
martijnwalraven: Why is this called `innerResult` here?

Sources/Apollo/ApolloClient.swift
```diff
@@ -18,12 +18,8 @@ public enum CachePolicy {
 /// A handler for operation results.
 ///
 /// - Parameters:
-///   - result: The result of the performed operation, or `nil` if an error occurred.
-///   - error: An error that indicates why the mutation failed, or `nil` if the mutation was succesful.
-public typealias GraphQLResultHandler<Data> = (_ result: GraphQLResult<Data>?, _ error: Error?) -> Void
-
-@available(*, deprecated, renamed: "GraphQLResultHandler")
-public typealias OperationResultHandler<Operation: GraphQLOperation> = GraphQLResultHandler<Operation.Data>
+///   - result: The result of a performed operation. Will have a `GraphQLResult` with the data of the requested type on `success`, and an `Error` on `failure`.
```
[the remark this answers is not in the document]
designatednerd: Yeah good call

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -26,11 +26,17 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
   }
   
   func fetch(cachePolicy: CachePolicy) {
-    fetching = client?.fetch(query: query, cachePolicy: cachePolicy, context: &context) { [weak self] (result, error) in
+    fetching = client?.fetch(query: query, cachePolicy: cachePolicy, context: &context) { [weak self] outerResult in
```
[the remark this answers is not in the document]
designatednerd: yeah that works - i'll try to change that throughout when there's not a 2nd result being wrapped

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
designatednerd: Because this is a nested callback - we already have `outerResult` above.

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: Yeah I plan to look at that in more depth when I do some much more detailed looks at the WebSocket stuff-  this is at least to get this out the door in a way that still gives us some kind of semi-meaningful error about what went wrong.

mormahr:
Hey, not an expert wrt. swift, but maybe one could add the old signature with a `@deprecated` or something like that?

designatednerd:
@mormahr Definitely doable, but it's pretty significant maintenance overhead to maintain all of the old public method signatures and basically need to do this with every single one: 

```swift
@available(*, deprecated, message: "Use the method with the `Result` completion handler instead")
func oldMethod(completion: @escaping (_ parameter: Parameter?, _ error: Error?)) {
  self.newMethod { result in 
    switch result {
    case .success(let parameter):
      completion(parameter, nil)
    case .failure(let error):
      completion(nil, error)
    }
  }
}
```

Unless there's a real good reason not to, I'd kinda prefer to rip the band-aid off here.

designatednerd:
OK folks - leaving this open for further comment until 10pm Netherlands Time (1pm US-Pacific) on Monday. Speak now or forever be mad at me for breaking your shit!

designatednerd:
May [deity of choice] have mercy on us all. Merging!

JakeTorres:
what's wrong with that, get a schema.json, but  swift API

PR #653, opened 2019-07-20
Add "operationName" to GraphQLOperation protocol

skevy:
Now that https://github.com/apollographql/apollo-tooling/pull/1386 is released, update `GraphQLOperation` to have an `operationName` var, so we can use it :)


designatednerd:
Hi! We've actually already got an open PR (#496) for this that I was hoping to let @js finish up since he's had it open since April - Johan, how do you feel about finishing yours vs. proceeding with this one?

skevy:
Sure thing.

PR #655, opened 2019-07-21
Make `GraphQLResultError` public

p4checo:
## Motivation

Errors produced by Apollo SDK should be public so that clients can perform type safe error handling using pattern matching.

### DIFF
--- a/Sources/Apollo/GraphQLExecutor.swift
+++ b/Sources/Apollo/GraphQLExecutor.swift
@@ -26,7 +26,7 @@ struct GraphQLResolveInfo {
 
 public struct GraphQLResultError: Error, LocalizedError {
   let path: ResponsePath
-  let underlying: Error
+  public let underlying: Error
   
   public var errorDescription: String? {
     return "Error at path \"\(path))\": \(underlying)"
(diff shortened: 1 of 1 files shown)


Sources/Apollo/GraphQLExecutor.swift
```diff
@@ -24,7 +24,7 @@ struct GraphQLResolveInfo {
   }
 }
 
-struct GraphQLResultError: Error, LocalizedError {
+public struct GraphQLResultError: Error, LocalizedError {
   let path: ResponsePath
   let underlying: Error
```
designatednerd: Makes sense -worth making `path` and `underlying` public as well?

p4checo: Well spotted! It's definitely worth it 👍

p4checo: Now that I see, that will require to also make `ResponsePath` public. Is that ok?

designatednerd: Ehhh let's keep that private but I think making the `underlying` error public probably wouldn't hurt.

p4checo: I agree 👌

PR #666, opened 2019-07-24
Update cache clearing documentation

designatednerd:
Also added a bit of documentation for a couple undocumented methods. 

Gonna hold off on updating the generated docs until release to avoid confusion.

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -33,6 +33,9 @@ public final class ApolloStore {
 
   private var subscribers: [ApolloStoreSubscriber] = []
 
+  /// Designated initializer
+  ///
+  /// - Parameter cache: An instance of `normalizedCache` to use to cache results.
   public init(cache: NormalizedCache) {
     self.cache = cache
     queue = DispatchQueue(label: "com.apollographql.ApolloStore", attributes: .concurrent)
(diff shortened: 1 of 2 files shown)


Sources/Apollo/ApolloStore.swift
```diff
@@ -44,6 +47,9 @@ public final class ApolloStore {
     }
   }
 
+  /// Clears ths instance of the cache. Note that a cache can be shared across multiple `ApolloClient` objects, so clearing that underlying cache will clear it for all clients.
```
AnthonyMDev: Typo :)
```suggestion
  /// Clears the instance of the cache. Note that a cache can be shared across multiple `ApolloClient` objects, so clearing that underlying cache will clear it for all clients.
```

designatednerd: English may be my first language but typo is clearly my second 🤦‍♀️

designatednerd:
@AnthonyMDev We're always happy to accept contributions, especially since we're now using the inline docs to generate API documentation on our website 😁

AnthonyMDev:
> @AnthonyMDev We're always happy to accept contributions, especially since we're now using the inline docs to generate API documentation on our website 😁

Absolutely! As soon as my company finished approving my open source contribution request I'll be on it! I think that should be done tomorrow!

Issue #567, opened 2019-06-05
BUG: Writing to cache not throwing execution errors


AnthonyMDev:
When there is an error executing a write to the cache, the error is being suppressed, even though the function is meant to throw.

In `ApolloStore.swift` on line 246, it should use `await()` instead of `wait()`.

designatednerd:
@AnthonyMDev Wanna put a PR together for this one? Would love to see this with a test that confirms an error is actually being thrown.

AnthonyMDev:
I would love to, but my companies policy requires me to submit a form for approval on open source libraries we contribute to. I'm sending in that form today, so hopefully I'll be able to start contributing shortly. I'll get a test and a fix in place as soon as that goes through.

designatednerd:
Godspeed with the bureaucracy

AnthonyMDev:
I've been approved to contribute so I'll get a unit test and PR out to you today!

PR #673, opened 2019-07-26
Fix error not being thrown on failed cache write

AnthonyMDev:
This is the fix for #567.
Without the code change, the unit test created fails consistently.


AnthonyMDev:
Agreement has been signed.

Issue #665, opened 2019-07-24
Error: Introspection result missing directive locations


noahemmet:
Hi! I'm upgrading to [version 0.13.0](https://github.com/apollographql/apollo-ios/releases/tag/0.13.0) and I'm running into some issues with the build script. When I run it, I get the following error:

```shell
++ npx --no-install apollo codegen:generate Views/GraphQL/ViewQueries.generated.swift --target=swift '--includes=./**/*.graphql' --localSchemaFile=Views/GraphQL/Schema.json --passthroughCustomScalars

Loading Apollo Project [started]

CLIError: Error in "Loading schema for Unnamed Project": Error: Introspection result missing directive locations: { onField: true, onOperation: false, args: [[Object]], name: "include", description: "Directs the executor to include this field or fragment only when the `if` argument is true.", onFragment: true }

    at Object.error (Project/Views/node_modules/@oclif/errors/lib/index.js:22:17)

    at Generate.error (Project/Views/node_modules/@oclif/command/lib/command.js:57:23)

    at OclifLoadingHandler.showError (Project/Views/node_modules/apollo/lib/OclifLoadingHandler.js:28:22)

    at OclifLoadingHandler.handle (Project/Views/node_modules/apollo/lib/OclifLoadingHandler.js:13:18) {

  oclif: { exit: 2 },

  code: undefined

}

Loading Apollo Project [failed]
```

My build script is set up like so:

```shell

cd "${SRCROOT}"
$APOLLO_FRAMEWORK_PATH/scripts/check-and-run-apollo-cli.sh codegen:generate "Views/GraphQL/ViewQueries.generated.swift" --target=swift --includes=./**/*.graphql --localSchemaFile="Views/GraphQL/Schema.json" --passthroughCustomScalars
```

A few notes:
- I'm using Xcode 13 and Swift Package Manager, which was working with version 0.12.0.
- I'm not using any `@directives` in my `.graphql` files.
- The backend is running [SwiftGraphQL](https://github.com/GraphQLSwift/GraphQL), so it's possible it's generating a `Schema.json` file that's incompatible with the latest Apollo code gen?
- I've tried various combinations of quotes/no-quotes, removing `--passthroughCustomScalars`, and some `--includes` path globs.

Any ideas? Thanks!

noahemmet:
[...]

PR #674, opened 2019-07-26
Add ApolloSQLite and ApolloWebSockets to Package.swift

noahemmet:
This exposes the ApolloSQLite and ApolloWebSockets sub-frameworks to clients.

This comes after I tried to set up disk-caching and realized `SQLiteNormalizedCache` wasn't available for use in SPM, although it was for Carthage and Cocoapods. I also included ApolloWebSockets since it seemed to be in the same situation.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,14 +9,27 @@ let package = Package(
         .library(
             name: "Apollo",
             targets: ["Apollo"]),
+		.library(
+			name: "ApolloSQLite",
+			targets: ["ApolloSQLite"]),
+		.library(
+			name: "ApolloWebSocket",
+			targets: ["ApolloWebSocket"]),
     ],
     dependencies: [
-
+		.package(url: "https://github.com/stephencelis/SQLite.swift.git", from: .init(0, 12, 2)),
+		.package(url: "https://github.com/daltoniam/Starscream", from: .init(3, 1, 0)),
     ],
     targets: [
         .target(
             name: "Apollo",
             dependencies: []),
+		.target(
+			name: "ApolloSQLite",
+			dependencies: ["Apollo", "SQLite"]),
+		.target(
+			name: "ApolloWebSocket",
+			dependencies: ["Starscream"]),
         .testTarget(
             name: "ApolloTestSupport",
             dependencies: ["Apollo"]),
--- a/Package.swift
+++ b/Package.swift
@@ -17,8 +17,8 @@ let package = Package(
 			targets: ["ApolloWebSocket"]),
     ],
     dependencies: [
-		.package(url: "https://github.com/stephencelis/SQLite.swift.git", from: .init(0, 12, 2)),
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -9,14 +9,27 @@ let package = Package(
...
+			name: "ApolloSQLite",
+			targets: ["ApolloSQLite"]),
+		.library(
+			name: "ApolloWebSocket",
+			targets: ["ApolloWebSocket"]),
     ],
     dependencies: [
-
+		.package(url: "https://github.com/stephencelis/SQLite.swift.git", from: .init(0, 12, 2)),
+		.package(url: "https://github.com/daltoniam/Starscream", from: .init(3, 1, 0)),
```
designatednerd: IIRC you can do these as `"3.1.0"` string literals - I kinda prefer that from a readability standpoint but I can also fix that myself if you'd prefer

noahemmet: Ah no problem; would you prefer `.upToNextMajor` or `.exact`? (Or did you mean _just_ the literal? I couldn't get that to work in Xcode 11b4…)

designatednerd: `exact` would be lovely - it's possible they yanked the ability to just use a string literal.

noahemmet:
Would you mind (if you haven't already) checking that this builds and compiles generated code? I should have mentioned this in the description, but I uh forgot 😬 :

I was unable to verify that this works, cause I was [having problems](https://github.com/apollographql/apollo-ios/issues/665) using the code gen with my weird Swift backend. I got to a certain point and started running into some name clashing errors. Hold off on merging one minute, lemme pull up the branch.

noahemmet:
Here it is, it's a runtime error in [`ApolloStore.ReadTransaction`](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloStore.swift#L189):

`
failed to demangle witness for associated type 'FinalResult' in conformance 'Apollo.Zip2Accumulator<Apollo.GraphQLSelectionSetMapper<Views.MainTabBarQuery.Data>, Apollo.GraphQLDependencyTracker>: GraphQLResultAccumulator' from mangled name '11FinalResult6Apollo24GraphQLResultAccumulatorPQz_AaDQy_t'
`

I assumed it was just an issue with my setup and didn't pursue it further, and then forgot I had put up the PR. But, it could also be an SPM issue. Sorry to push untested work, it was real late! 😦

designatednerd:
OK - I'll take a look tomorrow! I'm gonna mark this as changes requested so I don't accidentally merge it

designatednerd:
Gonna take another swing through this for `0.15.0` - would love to make sure this is building properly with SPM before I move to 🇺🇸in a couple weeks.

designatednerd:
OK! I was able to confirm this is working in Xcode 11b5 with the attached dummy project - run it and you'll see both web sockets and sqlite working nicely!

[SPMTest2.zip](https://github.com/apollographql/apollo-ios/files/3485837/SPMTest2.zip)

PR #688, opened 2019-08-02
Support Catalyst Builds

### LINKED ISSUE #687: Support Catalyst builds
We're using Apollo for our iOS project and want to bring it over to macOS using Catalyst. With the current build configuration, Xcode does not recognise Apollo to be Catalyst compatible.

simonboots:
This fixes #687 and allows Apollo to be used in Catalyst apps.

### DIFF
--- a/Configuration/Shared/Workspace-Universal-Framework.xcconfig
+++ b/Configuration/Shared/Workspace-Universal-Framework.xcconfig
@@ -3,9 +3,6 @@
 
 SUPPORTED_PLATFORMS = iphoneos iphonesimulator appletvsimulator appletvos watchsimulator watchos macosx
 
-// Set default SDKROOT
-SDKROOT = iphoneos
-
 // iOS-specific default settings
 SDKROOT[sdk=iphone*] = iphoneos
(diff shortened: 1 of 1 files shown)


Configuration/Shared/Workspace-Universal-Framework.xcconfig
```diff
@@ -3,6 +3,9 @@
 
 SUPPORTED_PLATFORMS = iphoneos iphonesimulator appletvsimulator appletvos watchsimulator watchos macosx
 
+// Set default SDKROOT
+SDKROOT = iphoneos
```
designatednerd: I'm slightly concerned this may eff things up if someone's using this in a `macosx` app - can you point me to the docs that talk about the default SDKRoot and how it affects catalyst stuff?

simonboots: Unfortunately, there's currently not much documentation on how the SDKROOT relates to Catalyst being available or not. However, I think I have found a better way. I have reverted setting the `SDKROOT` and added the `SUPPORTS_MACCATALYST = YES` configuration. That seems a much better solution.

PR #693, opened 2019-08-05
Create a protocol for public methods of Apollo client

gsabran:
This is so that developers can refer to the protocol instead of the concrete type, and use dependency injection for unit testing. It doesn't break any public APIs.

### DIFF
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -1,0 +1,82 @@
+// Created by guillaume_sabran on 8/5/19.
+// Copyright © 2019 Airbnb Inc. All rights reserved.
+
+import Dispatch
+
+// MARK: - Protocol
+
+/// Describes the Apollo client interface
+public protocol ApolloClientProtocol: class {
+
+  var store: ApolloStore { get }
+
+  var cacheKeyForObject: CacheKeyForObject? { get set }
+
+  /// Clears the underlying cache.
+  /// Be aware: In more complex setups, the same underlying cache can be used across multiple instances, so if you call this on one instance, it'll clear that cache across all instances which share that cache.
+  ///
+  /// - Returns: Promise which fulfills when clear is complete.
+  func clearCache() -> Promise<Void>
+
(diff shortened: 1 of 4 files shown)


Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -0,0 +1,82 @@
...
+  /// - Returns: An object that can be used to cancel an in progress subscription.
+  @discardableResult func subscribe<Subscription: GraphQLSubscription>(subscription: Subscription, queue: DispatchQueue, resultHandler: @escaping GraphQLResultHandler<Subscription.Data>) -> Cancellable
+}
+
+// MARK: - Extension
+
+public extension ApolloClientProtocol {
+
+  @discardableResult func fetch<Query: GraphQLQuery>(query: Query, cachePolicy: CachePolicy = .returnCacheDataElseFetch, context: UnsafeMutableRawPointer? = nil, queue: DispatchQueue = DispatchQueue.main, resultHandler: GraphQLResultHandler<Query.Data>? = nil) -> Cancellable {
+    return fetch(query: query, cachePolicy: cachePolicy, context: context, queue: queue, resultHandler: resultHandler)
```
designatednerd: One thing to watch out for with these default implementations is that they will override the local implementation anywhere that the client is passed as `ApolloClientProtocol`. [This is an extremely oversimplified example of that behavior](https://speakerdeck.com/designatednerd/protocols-all-the-way-down-dotswift-paris-january-2018?slide=70).

I would recommend removing the default implementations here and just setting the default params on the actual implementations, since it seems like the default implementations are just there to re-add the default parameters.

designatednerd:
Hey @gsabran - If I don't hear back from you by tomorrow, I'm going to take over this PR so I can get it shipped. Excellent idea!

designatednerd:
Closing in favor of #715 since there hasn't been any movement on this and I'd still like to get it out with `0.15.0`.

gsabran:
Hi thanks a lot for taking over, and sorry for not following up at all.. I left for vacations :)

designatednerd:
Haha, yep, I figured. 'Tis August after all. Thanks for the idea!

PR #699, opened 2019-08-08
Swap out session configuration for URLSession

designatednerd:
The code changes here are pretty minimal: Swapping out a `URLSessionConfiguration` for the actual `URLSession`. However, by being able to pass in a `URLSession` that you control the `delegate` of, you can more easily support a number of things: 

- Certificate pinning
- Self-signed certificates
- Authentication challenge responses
- Metrics inspection

Note that this is going to be a breaking change if you were previously passing in a session configuration to the `HTTPNetworkTransport` - you'll need to hand that configuration to a `URLSession` yourself first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,8 +49,8 @@ cache:
     - ../starwars-server
 
 before_install:
-  - nvm install 8
-  - nvm alias default 8
+  - nvm install 10
+  - nvm alias default 10
   - (./scripts/install-or-update-starwars-server.sh)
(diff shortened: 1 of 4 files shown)


designatednerd:
Blergh, something is up with Travis - it's refusing to install the `starwars-server` locally, so all the tests dependent on local are 💣ing out. I'll have to get a look at this late tonight or tomorrow.

designatednerd:
Woot, clearing the cache did the trick. Merging!

Issue #382, opened 2018-10-09
Change Promise public class to internal clarification.


cadizjavier:
Apollo iOS has a custom implementation of Promises that could potentially conflict with existing one in the host project of this dependency and is kinda confusing specifying every time which implementation of `Promise` to use.
I have done a quick look at the source code and it seems that the `Promise` class could be implemented as internal. 
- Is this issue happening to someone else? What is your workaround in your host project?
- Is it desirable to implement this as internal and refactor de `firstly` and `whenAll` public functions into something else? I can definitely give it a try as a Pull Request but need some clarifications about the status of this "issue".

js:
To solve the naming clash you need to specify the module name of the type you want to use, eg `MyApp.Promise` or `ThePromiseLib.Promise`, or `Apollo.Promise` if you want to use Apollo's.

The Promise is exposed in some of the other public API, such as the `DataLoader` so another approach would be needed there if this where to be hidden as you say.

cadizjavier:
Yes, this is exactly what are we doing in our project, specifying the module name which is kinda inconveniente and confusing. We asked ourselves, why we have now an implementation of Promise exposed to us now ?
I see now that you mention that is used for example in `DataLoader` so yes, agree, a refactor to hide this as an implementation detail would be needed.

cadizjavier:
I see now that a `Promise` is returned if we use for example `apolloClient.clearCache()` so `Promise` can't be internal in this case.
Would it make sense to return for the public API only callbacks and keep `Promise` as an internal implementation only for the use of the library ?

paulmelnikow:
What about simply using PromiseKit? It's well tested and as a community we benefit from standardization.

designatednerd:
Woof, yeah that almost certainly should never have been made public- it was built as an internal dependency and not for external use. I may look at doing this as a breaking change, but it's probably going to be a bit lower on my priority list.

designatednerd:
Shipped with `0.16.0`!

PR #709, opened 2019-08-13
Remove Public Promise

designatednerd:
_NOTE: Leaving this as a draft because I'd like @martijnwalraven to take a look at this when he's back from vacation, and I'd love to hear any other feedback in the meantime._

This PR addresses #382. The promises Martijn created were originally intended to only be for framework-internal consumption, not for use by outside apps or SDKs. This is why we're using a custom `Promise` type rather than existing `Promise` libraries.

However, our `Promise` type conflicts with those existing libraries and can cause all sorts of weirdness and confusion when it's exposed publicly, and it also somewhat hamstrings future implementation choices (including removing Promises under the hood entirely). 

For this PR, we're only removing the "public" aspect of promises and continuing to use them under the hood. This should give us more flexibility in the future, and make it easier to wrap these methods in their preferred promise library. 

Note that this is going to be a BREAKING CHANGE!

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -50,14 +50,17 @@ public final class ApolloStore {
   /// Clears the instance of the cache. Note that a cache can be shared across multiple `ApolloClient` objects, so clearing that underlying cache will clear it for all clients.
   ///
   /// - Returns: A promise which fulfills when the Cache is cleared.
-  public func clearCache() -> Promise<Void> {
-    return Promise<Void> { fulfill, reject in
-      queue.async(flags: .barrier) {
-        self.cacheLock.withWriteLock {
-          self.cache.clear()
+  public func clearCache(callbackQueue: DispatchQueue = .main, completion: (() -> Void)? = nil) {
+    queue.async(flags: .barrier) {
+      self.cacheLock.withWriteLock {
+          self.cache.clearPromise()
         }.andThen {
-          fulfill(())
-        }
+          guard let completion = completion else {
+            return
+          }
+          callbackQueue.async {
+            completion()
+          }
       }
     }
   }
@@ -66,7 +69,7 @@ public final class ApolloStore {
     return Promise<Void> { fulfill, reject in
       queue.async(flags: .barrier) {
         self.cacheLock.withWriteLock {
-          self.cache.merge(records: records)
+          self.cache.mergePromise(records: records)
         }.andThen { changedKeys in
           self.didChangeKeys(changedKeys, context: context)
           fulfill(())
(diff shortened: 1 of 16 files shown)


Sources/Apollo/ApolloStore.swift
```diff
@@ -258,3 +301,48 @@ public final class ApolloStore {
     }
   }
 }
+
+internal extension NormalizedCache {
```
martijnwalraven: It's probably not worth the trouble, but I wonder if we could use something like [Node's `promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) to avoid writing these wrapper methods manually.

Sources/Apollo/ApolloStore.swift
```diff
@@ -50,14 +50,17 @@ public final class ApolloStore {
...
-        self.cacheLock.withWriteLock {
-          self.cache.clear()
+  public func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, Error>) -> Void)? = nil) {
+    queue.async(flags: .barrier) {
+      self.cacheLock.withWriteLock {
+          self.cache.clearPromise()
         }.andThen {
-          fulfill(())
-        }
+          guard let completion = completion else {
```
martijnwalraven: Maybe `apollo_performAsyncIfNeeded` could be extended with a guard for an optional callback?

Sources/Apollo/ApolloStore.swift
```diff
@@ -99,14 +102,32 @@ public final class ApolloStore {
...
+  /// Performs an operation within a read transaction
+  ///
+  /// - Parameters:
+  ///   - body: The body of the operation to perform.
+  ///   - callbackQueue: [optional] The callback queue to use to perform the completion block on. Will perform on the current queue if not provided. Defaults to nil.
+  ///   - completion: [optional] The completion block to perform when the read transaction completes. Defaults to nil.
+  public func withinReadTransaction<T>(_ body: @escaping (ReadTransaction) throws -> T,
+                                       callbackQueue: DispatchQueue? = nil,
+                                       completion: ((Result<T, Error>) -> Void)? = nil) {
+    _ = self.withinReadTransactionPromise {
```
martijnwalraven: This may not be worth it, but maybe a generic wrapper method like [Node's `callbackify`](https://nodejs.org/api/util.html#util_util_callbackify_original) could be used to abstract this conversion.

Sources/Apollo/ApolloStore.swift
designatednerd: Since I'd eventually like to remove our promises system I'm not entirely convinced making them easier to create is worth it. 😆

Sources/Apollo/DispatchQueue+Optional.swift
```diff
@@ -0,0 +1,24 @@
...
+//  DispatchQueue+Optional.swift
+//  Apollo
+//
+//  Created by Ellen Shapiro on 8/13/19.
+//  Copyright © 2019 Apollo GraphQL. All rights reserved.
+//
+
+import Foundation
+
+public extension DispatchQueue {
```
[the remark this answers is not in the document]
designatednerd: Yes, in order for it to be usable in the `ApolloSQLite` framework.

designatednerd:
may [deity or lack thereof of choice] have mercy on us all

Issue #645, opened 2019-07-17
Create Roadmap document


designatednerd:
I realize this is a `TODO` to make more `TODO`s, but I should really write up the longer term plans I have for this repo at some point soon once I've got existing issues a little more under control.

designatednerd:
To-Done!

PR #717, opened 2019-08-16
Add ROADMAP.md

designatednerd:
Addresses #645. Note that the feedback I mention in here we plan to collect is going to be posted Monday as a Pinned Issue on this repo.


designatednerd:
#723 is now live!

PR #751, opened 2019-09-03
Include `PLATFORM_NAME` in bundle identifier

nchase:
As of ~5 days ago, App Store submissions will fail with the ITMS-90806 publishing error code if there are multiple bundles with the same identifier. This is a problem for apps that e.g. include an iOS version alongside a watchOS version.

This change just appends the `PLATFORM_NAME` to the bundle identifier to get around the issue.

Here's a thread from Apple's message boards about the issue: https://forums.developer.apple.com/thread/122048


designatednerd:
@nchase There's a thread on [CocoaPods' Github](https://github.com/CocoaPods/CocoaPods/issues/9135) about this - Are you using CP to integrate or another method? If it's only CocoaPods having this issue, I'm inclined to recommend this be handled with one of the [`post_install` workarounds](https://github.com/CocoaPods/CocoaPods/issues/9135#issuecomment-526326326) that fixes this problem for every framework rather than just for ours. 

If you're not using CocoaPods, what are you using?

Issue #742, opened 2019-08-29
User-friendly localizedDescription from query fetch Error


jordanhbuiltbyhq:
This is a refile of #477. Apologies for not getting back on time. We continue to see the `error.localizedDescription` is not something you want to show to the user. Testing it on 0.15.0 we get:
`"Received error response (500 internal server error): {\"errors\":[{\"message\":\"Invalid client auth token.\",\"extensions\":{\"code\":\"INTERNAL_SERVER_ERROR\"}}]}\n"`

designatednerd:
@jordanhbuiltbyhq Thanks for re-filing! I'll take a look and see what I can figure out. I suspect that's from `GraphQLError` not using the `message` as the localized description

Issue #477, opened 2019-03-22
User-friendly localizedDescription from query fetch Error


jordanhbuiltbyhq:
I'm fetching a query in Apollo and the result handler provides an optional result and error object.

`apollo.fetch(query: SampleQuery()) { (result, error) in`

If the error is non-nil I present this error's `localizedDescription` to the user (and if it's not then I check if the result's errors is not empty and use the first error's `localizedDescription`). At times the user can be shown a very unfriendly message from that outer error, such as 

> Received error response (500 internal server error): {"errors":[{"message":"Context creation failed: Invalid app info headers.","extensions":{"code":"INTERNAL_SERVER_ERROR"}}]}

It seems this is the case when the error is a `GraphQLHTTPResponseError`. Could we get a friendly `localizedDescription` out of these errors?

designatednerd:
@jordanhbuiltbyhq - I think I saw some changes around this recently - can you let us know how this is working for you on version 0.11.1 or higher?

designatednerd:
Since I haven't heard back here in a couple of weeks, I'm going to close this issue out. If you're still having problems with newer versions of Apollo, please open a new issue. Thank you!

PR #756, opened 2019-09-09
Better Handling of `HTTPResponseError` with json + Tests

designatednerd:
This PR addresses #742 (which is a re-file of #477), and makes sure that when an `HTTPResponseError` is received, errors are properly parsed. 

I also added some tests on this as a net for the future.

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		54DDB0921EA045870009DD99 /* InMemoryNormalizedCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54DDB0911EA045870009DD99 /* InMemoryNormalizedCache.swift */; };
 		5AC6CA4322AAF7B200B7C94D /* GraphQLHTTPMethod.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5AC6CA4222AAF7B200B7C94D /* GraphQLHTTPMethod.swift */; };
 		9B708AAD2305884500604A11 /* ApolloClientProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B708AAC2305884500604A11 /* ApolloClientProtocol.swift */; };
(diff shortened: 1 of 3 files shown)


designatednerd:
@jordanhbuiltbyhq Mind taking a look at this to make sure this is the behavior you were looking for?

jordanhbuiltbyhq:
This is lovely, thanks! Any expectation for when this would go live?

designatednerd:
I'll take a look and see if anything else needs to be patched but otherwise within the next day or two

jordanhbuiltbyhq:
oh I wanted to check, is there a better way to detect an unauthenticated error, to take the user back to the login screen? Currently I have this not-so-pretty chunk of code:

```
        if let graphQLError = error as? GraphQLHTTPResponseError {
            if let bodyData = graphQLError.body {
                if let jsonObject = try? JSONSerialization.jsonObject(with: bodyData, options: []) as? [String: Any] {
                    if let errors = jsonObject["errors"] as? [[String: Any]] {
                        if let extensions = errors.first?["extensions"] as? [String: Any] {
                            if let code = extensions["code"] as? String {
                                if code == "UNAUTHENTICATED" {
                                    //TODO: Go to login screen
                                }
                            }
                        }
                    }
                }
            }
        }
```

designatednerd:
Woo, pyramid of doom! Takes me back to the early days of Swift 😃 

So a couple suggestions: First, use `guard let` in order to stack your unwraps instead of pyramiding them (you'll see what I mean in the example below). Once this feature lands, you should be able to do something like: 

```swift
guard 
    let responseError = error as? GraphQLHTTPResponseError,
    let graphQLError = responseError.graphQLErrors?.first, 
    let extensions = graphQLError.extensions,
    let code = extensions["code"] else {
       // No error code to deal with 
      return
}

switch (code) {
case "UNAUTHENTICATED": 
    // Kick to login screen
default: 
   // Whatever else you need to do
}
```

designatednerd:
OK gonna merge this, it'll go out in the next patch release - I'll ping your other issue once it ships.

PR #758, opened 2019-09-11
Update installation docs to make where you need to put `schema.json` clearer

designatednerd:


designatednerd:
@aamir-nazir Please feel free to leave more comments requesting clarifications even though I already merged this!

PR #766, opened 2019-09-16
Include Node Binaries Rather Than Fighting With NPM

designatednerd:
The need to install things via `npm` in order to run code generation is unnecessarily plunging iOS developers into Node dependency hell, a place nobody wants to go.

This PR changes things up by including pre-built binaries of both our `apollo` CLI and of all its dependencies (including Node itself). This allows a few things: 

- Devs don't have to care about whether the version of the CLI they're using is compatible with the SDK, since they're tied together. 
- Devs don't have to worry about some kind of accidental breaking change being committed to `tooling` breaking their build. 
- Lays the foundation for moving the actual generation of code out of Node and into Swift (for now, we're keeping the parsing in Node). 

This of course comes with a few trade-offs: 

- These binaries are not small, so it will make downloading the SDK take more time. However, since they are not included in the framework, it does not affect the size of the framework itself.
- Swift Package Manager and Carthage will need to depend on the checked-out version of the SDK rather than something bundled within the framework. This is more of an issue for Carthage since some Carthage users don't run `carthage checkout` on a regular basis, but it also vastly simplifies the build script, and in turn the number of places where things can go wrong. 

I'm opening this as a draft PR - I'd love some feedback, particularly from Carthage users, on how migrating from the current setup to migrating to this setup would affect your project. You should be able to point to the branch this PR is opened off of in order to see the updated documentation instructions and get the binaries.


designatednerd:
OK - gonna leave this up for another day for comment - if we don't hear back from folks by Wednesday the 18th at 2pm US-Central time, I'm gonna merge it.

koenpunt:
As much as I like to see the installation to be simplified for non-js users, I think it's a really bad idea to add this much javascript files to a swift project.

As I mentioned in a different thread; it should be possible to bundle all these files in a single binary file, and that binary should then be downloaded separately of the library, but this could of be handled by a script in the library.

koenpunt:
But if you really want to do this, you at least remove add typescript declaration files (`.d.ts`) and sourcemap files (`.map`) since those are not necessary to run the code.

designatednerd:
@koenpunt This is a fair point. I'll see what i can do about getting just a raw binary rather than the tarball that has some binary and some stuff zipped

designatednerd:
Ahhhh, frak. I remembered the main reason why: GitHub doesn't accept files larger than 100mb. 

I agree having all the files in here is Less Than Ideal™, but they're not included in the project itself, they aren't included in the framework, and in most cases developers will never see these files directly. A big piece of the goal with using a local version of both node and the CLI is to remove the need to do anything besides `pod install` to be able to access the CLI, and get started. 

Having to download the binary separately kind of defeats the point there, but would be required if the file size limit remains at 100mb (which I'm almost positive it will for a while).

koenpunt:
> but they're not included in the project itself

What do you mean by this? Because what I see is that this PR adds about 15k javascript files to the repo.

> Having to download the binary separately kind of defeats the point there, but would be required if the file size limit remains at 100mb (which I'm almost positive it will for a while).

You could use GitHub LFS, but that would require users to have that configured. Another option would be to add a wrapper script that downloads the version from the github releases page, from the top of my head;
```sh
#!/bin/sh

APOLLO_VERSION=0.15.3
CLI_PATH="$PODS_ROOT/bin/apollo-cli-$APOLLO_VERSION"

if [ ! -f "$CLI_PATH" ]; then
  curl -o "$CLI_PATH" https://github.com/apollographql/apollo-ios/releases/download/$APOLLO_VERSION/apollo-cli.tgz
  chmod +x "$CLI_PATH"
fi

$CLI_PATH $@
```

designatednerd:
> what I see is that this PR adds about 15k javascript files to the repo.

They're added to the **repo**, not to the `xcworkspace` or any of the `xcodeproj` files. They are **never** included in any Apollo framework which is built using any dependency manager. That's what I mean as "not part of the project". 

In my experience, even with LFS, 100mb files are rejected when you try to push them up to GH. 

Again, I'd like to avoid users having a potential hole in terms of needing to download things separately from installing via a dependency manager. I can easily see the following scenario: 

- Developer is connected to the internet, [dependency manager of choice] installation completes successfully. Developer assumes they are done.
- Developer winds up without internet (gets on a flight, goes to work in a park, etc) or GitHub goes down
- This script is run, and the developer cannot build their project at all because they don't have the proper version of the CLI downloaded, and attempting to download bombs out. 
- Developer is sad/angry. 

Especially the "gets on a flight" bit I've personally experienced trying to work on stuff, and it's maddening.

koenpunt:
Oh I just checked, archiving the CLI results in only 35M (gzipped tar). But I still am not a fan of including it in this repo. Does cocoapods maybe support some after install hook to download additional resources?

designatednerd:
@koenpunt CocoaPods might, but Carthage and SPM don't have that kind of functionality, so it wouldn't really solve this problem.

koenpunt:
Hm, alright, well then building a single binary, archiving it, and adding that to the repo might be the least intrusive option.

designatednerd:
@koenpunt How would you propose we ensure old versions of the CLI get removed so as to not blow up the user's disk space once they've used a few different versions?

koenpunt:
> @koenpunt How would you propose we ensure old versions of the CLI get removed so as to not blow up the user's disk space once they've used a few different versions?

If we're going to bundle the archived CLI we just always extract it to the same file. And we can check the current version of the CLI by running `apollo -v`, and compare that to the version we need. Or more performant, compare it to the shasum of the binary.

```
#!/bin/sh

set -e

APOLLO_VERSION=0.15.3
CLI_PATH="$PODS_ROOT/Apollo/bin/apollo-cli-$APOLLO_VERSION"
CLI_SHASUM="9d71d6be53e2f06de9c70f2fe1d5f1e124e12ae8  $CLI_PATH"

extract_cli() {
  tar xzf apollo.tgz
}

if [ -f "$CLI_PATH" ]; then
  if ! /usr/bin/shasum -c <<< "$CLI_SHASUM"  ; then
    # incorrect binary, extract from archive
    extract_cli
  fi
else
  extract_cli
fi

$CLI_PATH $@
```

designatednerd:
`shasum` of the binary is probably better to prevent potential security holes by someone using a binary that impersonates the proper version. 

Few questions on this: 
- Why is `set -e` necessary in this script? It doesn't appear to be setting any environment variables, which AFAICT is what that does.
- Would it make more sense to have it unzip to `bin/apollo` instead of `bin/apollo-cli-${APOLLO_VERSION}`?
- Why does `$CLI_PATH` need to be part of the `CLI_SHASUM` variable? 

I do recall one of the tooling guys having some concerns around the packaging functionality, I'll see if I can get them to elaborate a bit here.

koenpunt:
> * Why is `set -e` necessary in this script? It doesn't appear to be setting any environment variables, which AFAICT is what that does.

This [makes the script exit if any of the subcommands fail](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script), but now I'm uncertain if that would mean it exits when the shasum doesn't match



> * Would it make more sense to have it unzip to `bin/apollo` instead of `bin/apollo-cli-${APOLLO_VERSION}`?

Yeah my bad, already updated that before you commented.

> * Why does `$CLI_PATH` need to be part of the `CLI_SHASUM` variable?

That's how `shasum` works, it verifies checksums based on path. I now use stdin, but if you would be verifying multiple files you'd probably have a checksum file, with multiple entries;

```
9d71d6be53e2f06de9c70f2fe1d5f1e124e12ae8  file1
70f2fe1d5f1e124e12ae89d71d6be53e2f06de9c  file2
2f06de9c70f2fe1d5f1e124e12ae89d71d6be53e  file3
```

designatednerd:
OK, I think I managed to hack something together. I'm starting by taking the tarball I was already unzipping and leaving it zipped instead of checking the unzipped version in. That takes us from almost a million lines of added code to ~120. 

Then when codegen is called, we check to make sure we have the current version of the tarball by checking the zip's SHA256 hash. If we do, we check to see if it's already installed, and install it if it's not. When it's installed, a `.shasum` file is added to the `apollo` directory so we know what version it was unzipped from. 

If the `.shasum` file and the SHA256 hash of the zip file don't match, we delete the current version of codegen and unzip the updated version. This will also allow for downgrading if needed, since it's only checking "is this the same?" rather than "is this newer?". 

If it does match, it proceeds with code generation. 

@koenpunt How does this look to you?

abernix:
👋 

Ok, so I **may be** one of the tooling guys having some concerns around the packaging functionality, as this long message will probably highlight. 😉 I hope you'll understand that it's in the best interest of overall maintainability and stability across not just this repo, but similar models which will surface elsewhere.

@koenpunt Your desire of avoiding the large number of files explicitly included in the repository is certainly worth addressing.  I completely agree.  I think you've proposed some alternatives above which are worthy inspiration, but I'd like to note some concern with some parts and try to also note a couple other details that I think are worth considering for the solution (or at least worth noting out loud).

> **Disclosure:** I am far from an iOS or Android engineer.  I have Xcode on my computer because it compiles other things. 😺  That said, I hope you'll bear with me!

These are some specific things which seem important to me and for the longer-term of not just this project, but other Apollo projects and interoperability:

0. **Don't pack the entirety of the module into this repo**

   First, I think these build on top of @koenpunt's very valid desire of not including the entire extracted `apollo` Node.js module with all of its dependencies within this repository.

1. **Debuggability**

   While I admire the seemingly sleek output of Zeit's `pkg`, I am concerned about using it as a critical piece in this picture.  Amongst other things, it has some caveats which are a bit concerning to me, all of which are advertised in its `README`: [some files not getting packaged](https://github.com/zeit/pkg#debug), [base images which use patched versions of `node`](https://github.com/zeit/pkg#build) (staying on top of this seems important.  how soon are versions released when Node.js releases security patches), [hides the source code](https://github.com/zeit/pkg#scripts) (sometimes a concern for auditing in enterprises), a [snapshot filesystem](https://github.com/zeit/pkg#snapshot-filesystem) (could further complicate hot-patching or debugging) and [potentially problematic native addon support](https://github.com/zeit/pkg#native-addons).

   That's not to say `pkg` isn't a cool tool — I actually think it's pretty neat that it does all this, but that doesn't mean that it's the right tool for this job.

2. **Simplicity of packaging**

   We already use [Heroku's `oclif`](https://github.com/oclif/oclif) for building the CLI.  It comes with the ability to package the tarballs and release packages _based on those tarballs_ which can be installed at the system level (with auto-updating) for macOS, Windows and Debian.  Installing at the system level doesn't accomplish the lock-step versioning step that I'll mention in a minute, but it's an option for those who want it.  My point for _this_ bullet-point is that it packages things up across those various distribution techniques in a simple way: It tars up the Node.js module and its dependencies, couples it with a `node` binary downloaded directly from the Node.js foundation and write a script to invoke it automatically behind the scenes.  Again,  everything that it does is based on the same tarball, so utilizing that same tarball here would help on the scale of keeping things relatively aligned across various deployments.  My hope would be that we could use a similar approach for Android, but again, see my disclaimer above. :smile:

3. **Longevity and stability**

   Not to say that `pkg` is going away, but I can't be certain if it's going to be actively maintained.  It has [a _**lot** of open issues_](https://github.com/zeit/pkg/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) of varying severity and hasn't had a commit since May of this year.  If we can leverage the option I've outlined above, it is at least backed by Heroku who uses identical tooling within their CLI and again automatically uses the latest version of Node.js on its own.

4. **Repo bloat / offline support**

   This is a bit of a continuation of point "zero" above, but whether it's an extracted copy (i.e. "vendored"), a full tarballed copy of it, a ZIP file, or any other series of bytes representing the same: I think it might be good to keep it locked outside.  If anything, this avoids the need to constantly update this repository each time the `apollo` CLI releases updates.  Updates might be important, but it would be great if it were up to the user, and not this repository, as to when they should be made (unless this repo has integration tests which test the `apollo` CLI?)  (More on this in a step four.)

   Offline support seems critical, but I wonder it's _okay_ if the (two) steps prior to going offline were: 1) Clone this repository and 2) Run `./scripts/download_and_extract_apollo_if_not_extracted_already.sh` to bootstrap it.  (There's a chance that download and extract should be two scripts, but that's an implementation detail.)

   The (badly-named) script I suggested could download the "latest" stable tarball from a known location, save it within the _clone_ (of this repository) in a directory and extract it.  Both the tarball and the extracted version would be `.gitignore`d, so they wouldn't get checked in.

   The `apollo` command itself already has update notifications built into it, so when an update notification comes in, the step to update would be to 1) remove the tarball and the extracted version, 2) run the script again to download a new version.  Some SHA algorithm could be introduced here, but I'd probably want to think about it just a bit more.  If it were important, I might suggest that an artifact was written out into the extracted directory by the extractor so it knows what version is currently installed.  (i.e. `tar xf tarball.tgz && shasum -c tarball.tgz > apollo/.tarball_hash`) and comparing that before doing re-extraction.

   Again, there's one more bit that I think ties in here that might be important, but that's the next bullet point!

4. **Lock-step versioning**

   I think whatever version of the `apollo` CLI that's installed should remain that version until updated.  It should be a version that the user clearly understands and can upgrade or downgrade as necessary.  The `apollo` CLI gets bug-fixes on a somewhat regular basis (and it even breaks sometimes! — though we of course try to avoid that).  I think that ideally, it would be possible for the project to just define in some location that it wants `apollo@<version>`.  If such a definition was set, _that_ would be the version that's downloaded by the download/ extract scripts.

I know this is a lot of words, but I hope it makes sense.  I don't actually think most of what I just wrote would be complicated to script out, but hopefully you'll help me understand what parts of this are good or bad ideas for this repo.

> And finally, cc @designatednerd

designatednerd:
@abernix as to Keeping things tied to a specific version of the CLI is why I'd like to include the tarball and not give the users a mechanism to update the CLI without updating the iOS lib. 

That keeps the CLI tied to a version we know works with the current version of the iOS lib.

abernix:
Also worth noting that I hadn't noticed @designatednerd's https://github.com/apollographql/apollo-ios/pull/766#issuecomment-533268937 while I was typing that up!

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
+# Bail if errors occur
+set -e
```
abernix: This is often a best practice which takes this a step further:

```suggestion
# Exit on all errors, undeclared variables and pipefailures.
set -euo pipefail
```

This isn't POSIX, but it is compatible with `bash` and `zsh`, and `fish`.

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Get the path to the script directory
+SCRIPT_DIR="$(eval dirname $0)"
```
abernix: Why is `eval` is necessary here?  The double quoted string would already interpolate variables.

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Get the SHASUM of the tarball
+ZIP_FILE="${SCRIPT_DIR}/apollo.tar.gz"
+SHASUM="$(eval /usr/bin/shasum -a 256 ${ZIP_FILE})"
```
abernix: Same `eval` question from above.

abernix: It's worth noting that the output of in the full path to the tarball (i.e. `123abc /Users/you/path/you/have/zippy-dee-doo-dah`), which means the path is baked into the `.shasum` and might be less portable (it would re-extract if you just moved the parent folder to another folder).  If it was worth avoiding that, this could be:

```suggestion
SHASUM="$(/usr/bin/shasum -a 256 ${ZIP_FILE})"
```

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+SHASUM_FILE="${SCRIPT_DIR}/apollo/.shasum"
+APOLLO_DIR="${SCRIPT_DIR}"/apollo
+
+# Helper functions
+remove_existing_apollo() {
+  rm -r "${APOLLO_DIR}"
+}
+
+extract_cli() {
+  tar xvzf "${SCRIPT_DIR}"/apollo.tar.gz -C "${SCRIPT_DIR}"
```
abernix: Nice for debugging, but I'll note that `tar` extraction _with_ the `-v`(erbose) flag much slower due to STDOUT buffering.  Though if really quantifying if the extraction of _x_ files happened, this might be a good idea to keep!

(Another less noisy but possibly still performant and helpful option might be to print the resulting folder size in KBytes with `du -ks folder/`. 🤷‍♂)

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+  touch "${SHASUM_FILE}"
```
abernix: `touch` doesn't seem necessary since `tee` will write it if it doesn't exist.  (This is one of the big things that makes `tee` desirable.

```suggestion
```

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+  tee "${SHASUM_FILE}" <<<"${SHASUM}"
```
abernix: I'm pretty sure that `<<<` is not POSIX and that it might be preferable to:

```suggestion
  echo "${SHASUM}" | tee "${SHASUM_FILE}"
```

This would accomplish the same thing and also work in `/bin/sh`, `dash`, `zsh`, etc.

Also, if printing the actual SHA in the debug output is also not desired, this could just be:

```suggestion
  echo "${SHASUM}" > "${SHASUM_FILE}"
```

This would create it and write the SHA silently. 🗣 🚫

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+}
+
+check_shasum() {
+  # Make sure the SHASUM matches the release for this version
+  EXPECTED_SHASUM="13febaa462e56679099d81502d530e16c3ddf1c6c2db06abe3822c0ef79fb9d2  ${ZIP_FILE}"
```
abernix: Ah, I see you're aware of the filename in the output now. 😄

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+    echo "Correct version of the CLI tarball is included, checking if it's already been extracted..."
+  else
+    echo "Error: The SHASUM of this zip file does not match the official released version from Apollo! This may present security issues. Terminating code generation."
+    exit 1
+  fi
+
+  # Check if the SHASUM file has already been written for this version
+  if [ -f "${SHASUM_FILE}" ]; then
+    # The file exists, let's see if it's the same SHASUM
+    FILE_CONTENTS="$(eval cat ${SHASUM_FILE})"
```
abernix: I love the composition of `eval cat` in terms of words 🐈, but I don't think `eval` is necessary.

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
...
+}
+
+# Make sure we're using an up-to-date and valid version of the Apollo CLI
+check_shasum
+
+# Use the bundled executable of the Apollo CLI to generate code
+APOLLO_CLI="${SCRIPT_DIR}/apollo/bin/run"
+
+# Print version
+echo "Apollo CLI Information: $(eval ${APOLLO_CLI} --version)"
```
abernix: Same `eval`-y.

Apollo.xcodeproj/project.pbxproj
```diff
@@ -1141,7 +1143,8 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
```
abernix: Should this _not_ be `/bin/sh`? 🤔

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,64 @@
+# Bail if errors occur
```
abernix: It's usually a good idea to have a `#!` line at the top of these to allow the interpreter to run the right script if it's invoked directly (e.g. `./run-bundled-codegen.sh`)

While Catalina will switch the default shell to `zsh`, it might actually be more defensive by just running this in `bash`, always?

```suggestion
#!/usr/bin/env bash
# Bail if errors occur
```

scripts/run-bundled-codegen.sh
abernix: Sorry for three separate comments on the same line!  I think including the `/usr/bin` might be less ideal here, but it's hard to say.  This wouldn't respect a users `PATH` though.

I'd probably say leave it for now, but this might break on my own computer.  Specifically, `shasum` is a **PERL SCRIPT** 😱, but the version of Perl that Apple ships is not the the same as the verison of Perl that my Homebrew has installed at the mercy of its own dependency needs.

That's to say, when I run `which shasum`, I and most things actually use `/usr/local/bin/shasum` which uses its matching version of Perl.

I guess I'm just saying the `/usr/bin` could be removed.

designatednerd: The problem is that this is being executed from a different directory - so it needs the full rather than the relative path (and will be different based on where Carthage, Cocoapods, and Swift Package manager check things out). I agree that's not ideal, but adding a relative path that works for all those different dependency managers isn't really feasible.

scripts/run-bundled-codegen.sh
designatednerd: I feel like I was running into problems without it, but that may have been on an older version, lemme test without

scripts/run-bundled-codegen.sh
designatednerd: I think something was fucking up without the V before but that might have been something else - hold plz...

scripts/run-bundled-codegen.sh
designatednerd: aha! good to know, I think I was trying to use something else that wouldn't write if it didn't exist before switching to `tee`

scripts/run-bundled-codegen.sh
designatednerd: but if you're using everyone's `/usr/bin/shasum` then you should be able to have a consistent value there, right?

Apollo.xcodeproj/project.pbxproj
designatednerd: that's the default that apple has been using pre-Catalina. I'll have to see what they're doing for scripts on Catalina, but for most intents and purposes this should work fine.

scripts/run-bundled-codegen.sh
designatednerd: removed.

scripts/run-bundled-codegen.sh
designatednerd: also removed

scripts/run-bundled-codegen.sh
designatednerd: yep didn't need it. removed

scripts/run-bundled-codegen.sh
designatednerd: removed.

scripts/run-bundled-codegen.sh
designatednerd: removed.

scripts/run-bundled-codegen.sh
designatednerd: added shebang

scripts/run-bundled-codegen.sh
designatednerd: yep!

scripts/run-bundled-codegen.sh
abernix: I was just generally highlighting that using explicit paths is sometimes less desirable since it doesn't respect a user's `PATH`.  For example, one might also consider using the full path to `tar`, `dirname`, `echo`, (etc.), but those aren't specified and I didn't see a specific reason for `shasum` being an exception.

My scenario is likely an extreme edge case, and possibly wouldn't break.  Anyhow, I'm certainly not objecting, just pointing out the subtle nuanced difference in file resolution patterns.

Apollo.xcodeproj/project.pbxproj
abernix: Yeah, leaving this is fine, and I wouldn't be surprised if it remains the same in Catalina since I think their intention is just to change the user's default shell.  Either way, the presence of the shebang in the scripts will ensure that the OS runs the desired interpreter.  In other words, this can be surely resolved. 😉

scripts/run-bundled-codegen.sh
koenpunt: Btw; these scripts should only be run from a Xcode build phase, so I think we can be pretty sure which shell it will run in.

scripts/run-bundled-codegen.sh
koenpunt: I deliberately set it to `/usr/bin/shasum`, because that's the one shipped with the OS. `/usr/local/bin/shasum` often is a custom installation, or something from Homebrew, which might yield different output.

koenpunt:
@abernix thanks for weighing in on this matter.

Some brief responses on your points;

> 0. **Don't pack the entirety of the module into this repo**

Well yes that, but mainly because it's unrelated source code. It think it would even make the repo identify itself as being javascript.

> 1. **Debuggability**

I wasn't necessarily saying we have to use `pkg` (there are other tools with similar functionality), but since I recently discovered that the `firebase` cli (by Google) is also a single binary node application, I figured this might be a good solution.

> 2. **Simplicity of packaging**

Yeah I saw that, and I looked at their documentation first, hoping to find a single binary compilation command or something. But there isn't. 

> 3. **Longevity and stability**

Same as 1.

> 4. **Repo bloat / offline support**

> I think it might be good to keep it locked outside

I agree on this 100%. Maybe not for the same reasons, but adding binaries to the repo will increase its absolute size. And every time the CLI is updated, the absolute repo size will increase by the full archive size (currently 21mb), since there are not going to be any diffs between those archives. So depending on how often you intend to update the CLI, this could get out of hand pretty quickly.

And about offline support; I'd say that we can inform users about the additional step(s) for offline support in the setup instructions. For CocoaPods users this would be less of a concern, since there we could set up an after-install hook.

> 6. **Lock-step versioning**

Ruby (or actually rubygems) has a cool feature where you can specify which version of a command you want to run;

If I run `pod --version`, it would execute with and output my default version (currently `1.8.0.beta.2`), but when I run `pod _1.7.5_ --version`, it will give me `1.7.5`. 

Unfortunately the node/npm ecosystem doesn't provide such a mechanism.

But yes, I think it would be good for the library to define the required version of the cli, and let a script download and install that version.

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+check_shasum() {
+  # Make sure the SHASUM matches the release for this version
+  EXPECTED_SHASUM="13febaa462e56679099d81502d530e16c3ddf1c6c2db06abe3822c0ef79fb9d2  ${ZIP_FILE}"
+
+  if [ "${SHASUM}" == "${EXPECTED_SHASUM}" ]; then
+    echo "Correct version of the CLI tarball is included, checking if it's already been extracted..."
+  else
+    echo "Error: The SHASUM of this zip file does not match the official released version from Apollo! This may present security issues. Terminating code generation."
```
koenpunt: You might want to echo this to stderr instead of stdout. You can do so by redirecting the output by appending ` >&2` to the end of the line.

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,65 @@
...
+remove_existing_apollo() {
+  rm -r "${APOLLO_DIR}"
+}
+
+extract_cli() {
+  tar xzf "${SCRIPT_DIR}"/apollo.tar.gz -C "${SCRIPT_DIR}"
+  echo "${SHASUM}" | tee "${SHASUM_FILE}"
+}
+
+check_shasum() {
```
koenpunt: This function does a bit more than checking the shasum, so could be named more aptly

koenpunt:
> I think it might be good to keep it locked outside

One thing that came to mind; what if we make it a git-submodule, would that be supported by cocoapods, carthage and swift pm?

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "Correct version of the CLI tarball is included, checking if it's already been extracted..."
+  else
+    echo "Error: The SHASUM of this zip file does not match the official released version from Apollo! This may present security issues. Terminating code generation."
+    exit 1
+  fi
+
+  # Check if the SHASUM file has already been written for this version
+  if [ -f "${SHASUM_FILE}" ]; then
+    # The file exists, let's see if it's the same SHASUM
+    FILE_CONTENTS="$(cat ${SHASUM_FILE})"
```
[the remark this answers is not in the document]
designatednerd: I'll try this against my `Carthage Test Two` sample project and find out...

scripts/run-bundled-codegen.sh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+# Make sure we're using an up-to-date and valid version of the Apollo CLI
+check_shasum
+
+# Use the bundled executable of the Apollo CLI to generate code
+APOLLO_CLI="${SCRIPT_DIR}/apollo/bin/run"
+
+# Print version
+echo "Apollo CLI Information: $(${APOLLO_CLI} --version)"
```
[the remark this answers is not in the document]
designatednerd: I think this is probably over-optimizing. It adds a trivial amount of time compared to how long codegen itself takes, and it'll make hunting down problems considerably clearer.

scripts/check-and-run-apollo-cli.sh
```diff
@@ -1,3 +1,5 @@
+echo "warning: The check-and-run-apollo-cli.sh script is deprecated and will be removed. Please see https://www.apollographql.com/docs/ios/installation/#adding-a-code-generation-build-step for updated build instructions which use a bundled version of the Apollo CLI."
```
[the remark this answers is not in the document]
designatednerd: That's why I prefaced this with `warning`, so it would show up in the Xcode sidebar and not just in the build log like so: 

Screen Shot 2019-09-20 at 4 30 13 PM

scripts/run-bundled-codegen.sh
designatednerd: Yeah I think explicitly using the system one here is probably the best way to handle it.

scripts/run-bundled-codegen.sh
designatednerd: Doesn't seem to have any outward effect when building with Xcode directly, but couldn't hurt for command line builds.

scripts/run-bundled-codegen.sh
designatednerd: the `eval` is gone, at least

scripts/run-bundled-codegen.sh
designatednerd: renamed a positively Obj-Cian `validate_codegen_and_extract_if_needed`

scripts/run-bundled-codegen.sh
designatednerd: Added

scripts/run-bundled-codegen.sh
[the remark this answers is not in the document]
designatednerd: It did not! Updated the script to quote ALL THE THINGS!

designatednerd:
> what if we make it a git-submodule, would that be supported by cocoapods, carthage and swift pm?

I think *maybe* Carthage might support that, but I don't believe SPM does and I'm almost positive CP doesn't. Also the thought of having to deal with submodules ON TOP of three dependency managers is giving me a migraine just thinking about it.

I started thinking about having a repo as a dependency that just grabs the current binary and its contents, and then setting that up as a dependency here. My concern is that just pushes the repo bloat into a different repo that people are still going to have to check out, causing the exact same problem but with the additional overhead of having it be a separate repo.

scripts/check-and-run-apollo-cli.sh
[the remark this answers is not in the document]
koenpunt: Ah, cool, didn't know about that :+1:

koenpunt:
Well I'm just trying to suggest things to prevent us littering the main repo. 

But the best solution IMO still is to upload the binary as a release asset to the github release, and download it from there.

designatednerd:
"Best" has a lot of possible meanings here, but the two biggest ones that are competing in this case are: 

- Is it the strategy that clogs up the repo the least? 
- Is it the strategy where developers can use the codegen the most easily? 

The absolute easiest thing would be to just include the unzipped binaries directly in the repo, and then there are no extra steps in using it after downloading. However, you correctly pointed out that clogs up the repo from both a binary and file perspective, which is why we went to the `.zip` file with a script which extracts it. It's an extra step, but it's a local-only one which is not prone to many errors. People are still basically ready to build immediately after installing the dependency. 

Having getting the CLI be a separate download step to get it means when people do `pod install` or `carthage build` or install via SPM, they are *not* ready to go immediately - they have to download the zip file and extract it. If they can't download it for whatever reason (no internet, GitHub is down, whatever artifact storage we're using is down), then they can't build.

If it only affected building Apollo stuff that'd be one thing, but because the version of the iOS SDK is tightly coupled to a particular version of the CLI, it will almost certainly kill the build for **your whole project**. Even if you have an earlier version of the CLI locally, the generated code and what the version of the SDK you have is expecting are going to be different, and you will almost certainly get errors at compile time. If you don't have an older version locally, then your code doesn't generate and you can't do anything. This is extra-true for CI builds, since there may be considerations about downloading to a CI machine that aren't necessarily the case for a local. 

That scenario is definitely not as bad as the Node Hell™ we've been dealing with, but in terms of "Which of these is the lesser of two evils, particularly if you are dealing with developers new to the iOS ecosystem?", I'm willing to sacrifice some repo bloat to reduce the number of potential points of failure.

koenpunt:
I do understand what you want to achieve, but I think "Offline Support" is solving an annoying problem, but it's probably not a very common problem. 

In terms of progressive enhancement we could start with downloading an archive from the script (as opposed to installing modules using npm, which also requires the network), which would already be a very big improvement to the workflow of the majority of users of the library.

And again, adding a binary to the repo with change its size indefinitely. Update the CLI 5 times, and users are pulling in 100mb of old CLI versions when they clone the repo, which is a lot of data for something they don't need. And you won't be able to remove this from the history, ever, without rewriting the git history, which is something you probably don't want. Old, but related post; https://robinwinslow.uk/2013/06/11/dont-ever-commit-binary-files-to-git/

designatednerd:
It's not *just* offline, though. Offline is the easiest-to-explain reason it'd happen, but "some way in which the developer cannot download the zip even though they were able to successfully download the rest of the dependency" is the problem I'm solving for here. 

A few things I've seen problems with in the past that could cause issues even if the user has a perfectly good internet connection: 

- [Place hosting artifact download] goes down because of a bad deploy, and rolling back fails for several hours to days. 
- [Place hosting artifact download] gets DDOS'd and is inaccessible for several hours to days.
- [Country developer lives in] decides to block off a huge block of IP addresses at the national level, which includes [place hosting artifact download]. 
- [Place developer works] has an overly-agressive content filter which blocks [place hosting artifact download]. 

I'll mull this over with the team on Monday. The real way to have this not be in the repo is to use NPM, but we've seen that NPM is complex and unreliable enough, particularly for iOS developers, that we need a different solution.

koenpunt:
I think all four points come down to the same if we host the binaries on github as release assets, because then the same can happen for the initial clone.
And I'm not saying these things don't happen, but I think we're trying to solve a problem that shouldn't really be a concern of the library.

Also I don't think you responded to my size concern now or before (or did I maybe miss a comment somewhere in this thread?), but I'm interested to hear how you think that's going to work out over time.

designatednerd:
Somewhere in this nest of comments I basically said I'm fine with the repo bloat as a trade-off, but we're still discussing internally (I was out sick yesterday). 

And while I get the thought about hosting the binaries as release assets on GH, that makes it really hard to do testing of new versions before we actually make a release, which is less than ideal.

designatednerd:
OK, here's the plan:  

- This PR will be merged as-is - it's a considerable improvement in developer experience that I really don't want to hold up any longer. 
- I will open an issue to look into future options in terms of potentially having the zip download from an artifact repository or a GH release resource in a way that can be used from any package manager. 

TL;DR - We'll bloat the repo for now, but not for eternity.

designatednerd:
#791 has been created as a pin to come back to this.

koenpunt:
> TL;DR - We'll bloat the repo for now, but not for eternity.

This isn't true, because the binaries we add, can never be removed, so will change the size of cloning the repo indefinitely. Or you must be fine with rewriting the history of the repo, but I can't imagine that's something you want? 

But let's see what #791 yields.

designatednerd:
I meant the additions to the bloat will not continue eternally. I recognize what we're doing now will add *some* bloat.

PR #767, opened 2019-09-16
Support Auto-Persisted Queries

designatednerd:
This is an update of work started in #608. However, we're running into an issue where stripping whitespace out of queries is causing APQ validation to fail, because the SHA256 hash no longer matches. I need to figure out if there's a way around this or if I need to just suck it up and use multiline string literals (cc @lilyball). 

Leaving this in draft until I can figure out, but @codebear98 would love feedback on some of the restructuring I did in the merge.


designatednerd:
This is now waiting on [some changes to tooling](https://github.com/apollographql/apollo-tooling/pull/1537)

designatednerd:
Okay! Will need to merge this again once #821 gets merged, but this shouuuuuuld be about ready to go

fiznool:
This is really great. ❤️ 

So far I guess we can look at the test files for examples, but are there any plans to add a section in the docs regarding how to set this up?

designatednerd:
Added #951 to my todo list @fiznool

PR #770, opened 2019-09-18
Introducing delegate if there is GraphQL errors

kimdv:
The idea of this delegate is based on 
- https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-error
- https://blog.apollographql.com/full-stack-error-handling-with-graphql-apollo-5c12da407210

It adds ability for better error handling on the clients when there is Graphql errors  
Etc when a token is expired. 

The diffrent from this delegate and `HTTPNetworkTransportRetryDelegate` is that some errors might come from the GraphQL API and not from middleware etc. 

 - [x] Add tests

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -65,6 +65,21 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate 
                         retryHandler: @escaping (_ shouldRetry: Bool) -> Void)
 }
 
+/// Methods which will be called after some kind of response has been received and it contains GraphQLErrors
+public protocol NetworkGraphQLErrorDelegate: HTTPNetworkTransportDelegate {
+
+
+  /// Called when response contains one or more GraphQL errors.
+  /// NOTE: Don't just call the `retryHandler` with `true` all the time, or you can potentially wind up in an infinite loop of errors
+  ///
+  /// - Parameters:
+  ///   - networkTransport: The network transport which received the error
+  ///   - errors: The received GraphQL errors
+  ///   - retryHandler: A closure indicating whether the operation should be retried. Asyncrhonous to allow for re-authentication or other async operations to complete.
+  func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedGraphQLErrors errors: [GraphQLError], retryHandler: @escaping (_ shouldRetry: Bool) -> Void)
+}
+
+
 // MARK: -
 
 /// A network transport that uses HTTP POST requests to send GraphQL operations to a server, and that uses `URLSession` as the networking implementation.
@@ -193,6 +208,25 @@ public class HTTPNetworkTransport {
     task.resume()
     
     return task
+  }
+
+  private func handleGraphQLErrorsOrComplete<Operation>(operation: Operation,
+                                                        response: GraphQLResponse<Operation>,
+                                                        completionHandler: @escaping (_ result: Result<GraphQLResponse<Operation>, Error>) -> Void) throws {
+    if let delegate = self.delegate as? NetworkGraphQLErrorDelegate,
+      let graphQLErrors = try response.parseResultFast().errors,
+      !graphQLErrors.isEmpty {
+      delegate.networkTransport(self, receivedGraphQLErrors: graphQLErrors, retryHandler: { [weak self] shouldRetry in
+        guard shouldRetry else {
+          completionHandler(.success(response))
+          return
+        }
+
+        _ = self?.send(operation: operation, completionHandler: completionHandler)
+      })
+    } else {
+      completionHandler(.success(response))
+    }
   }
   
   private func handleGraphQLErrorsIfNeeded<Operation>(operation: Operation,
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -146,11 +146,11 @@ public class HTTPNetworkTransport {
       
       if let receivedError = error {
         self.handleErrorOrRetry(operation: operation,
-                                 files: files,
-                                 error: receivedError,
-                                 for: request,
-                                 response: response,
-                                 completionHandler: completionHandler)
+                                files: files,
+                                error: receivedError,
+                                for: request,
+                                response: response,
+                                completionHandler: completionHandler)
         return
       }
       
@@ -163,11 +163,11 @@ public class HTTPNetworkTransport {
                                                          response: httpResponse,
                                                          kind: .errorResponse)
         self.handleErrorOrRetry(operation: operation,
-                                 files: files,
-                                 error: unsuccessfulError,
-                                 for: request,
-                                 response: response,
-                                 completionHandler: completionHandler)
+                                files: files,
+                                error: unsuccessfulError,
+                                for: request,
+                                response: response,
+                                completionHandler: completionHandler)
         return
       }
       
@@ -176,11 +176,11 @@ public class HTTPNetworkTransport {
                                              response: httpResponse,
                                              kind: .invalidResponse)
         self.handleErrorOrRetry(operation: operation,
-                                 files: files,
-                                 error: error,
-                                 for: request,
-                                 response: response,
-                                 completionHandler: completionHandler)
+                                files: files,
+                                error: error,
+                                for: request,
+                                response: response,
+                                completionHandler: completionHandler)
         return
       }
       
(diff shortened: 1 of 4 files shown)


Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -167,6 +183,25 @@ public class HTTPNetworkTransport {
     
     return task
   }
+
+  private func handleGraphQLErrorsOrComplete<Operation>(operation: Operation,
+                                                        response: GraphQLResponse<Operation>,
+                                                        completionHandler: @escaping (_ result: Result<GraphQLResponse<Operation>, Error>) -> Void) throws {
+    if let delegate = self.delegate as? NetworkGraphQLErrorDelegate,
```
designatednerd: Personally I'd `guard let` this for clearer control flow.  

You may also want to look at some of the changes proposed in #767 - there's some stuff in there that's similar to this just without the delegate, including something to only parse the errors (rather than parsing the entire data structure as you're doing below).

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -167,6 +183,25 @@ public class HTTPNetworkTransport {
     
     return task
   }
+
+  private func handleGraphQLErrorsOrComplete<Operation>(operation: Operation,
```
designatednerd: I think this needs to include a `files` parameter in case an operation with files needs to get retried

kimdv: Just saw the `handleErrorOrRetry` also misses that? 
Want me to add here or in a new PR?

designatednerd: Add here - I think the reason it doesn't handle it there is that it's more specifically looking at retries for APQ stuff rather than in general, and APQ can only be done with Queries, not mutations.

Sources/Apollo/HTTPNetworkTransport.swift
kimdv: Added `guard let`

Related to the #767 are you thinking of the `func parseErrorsFast()`?

designatednerd: yep!

kimdv:
Will create tests when we can mock. Depending on #767

designatednerd:
@kimdv #767 is finally merged if you wanna use the stuff from that still

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -131,10 +146,11 @@ public class HTTPNetworkTransport {
...
         self.handleErrorOrRetry(operation: operation,
-                                error: receivedError,
-                                for: request,
-                                response: response,
-                                completionHandler: completionHandler)
+                                 files: files,
+                                 error: receivedError,
+                                 for: request,
+                                 response: response,
+                                 completionHandler: completionHandler)
```
designatednerd: I think the indentation is off here and below

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -183,17 +203,47 @@ public class HTTPNetworkTransport {
...
     return task
   }
+
+  private func handleGraphQLErrorsOrComplete<Operation>(operation: Operation,
+                                                        files: [GraphQLFile]?,
+                                                        response: GraphQLResponse<Operation>,
+                                                        completionHandler: @escaping (_ result: Result<GraphQLResponse<Operation>, Error>) -> Void) throws {
+    guard let delegate = self.delegate as? HTTPNetworkTransportGraphQLErrorDelegate,
+      let graphQLErrors = response.parseErrorsOnlyFast(),
+      !graphQLErrors.isEmpty else {
```
designatednerd: this is more of a note to self than to you: I need to add my readability `isNotEmpty` boolean extension to this code 🙃

kimdv:
@designatednerd I rebased and added two tests. One with error and one with none.

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -193,6 +196,50 @@ class HTTPTransportTests: XCTestCase {
                                                      delegate: self)
     XCTAssertNotEqual(self.networkTransport, nonIdenticalTransport)
   }
+
+  func testErrorDelegateWithErrors() throws {
+    self.graphQlErrors = []
+    let query = HeroNameQuery()
+    let body = ["errors": [["message": "Test graphql error"]]]
```
kimdv: Because `GraphQLError` is not encodable or NSObject we can not use it. 
Maybe we should consider to make `GraphQLError` encodable?

Tests/ApolloTestSupport/MockURLSession.swift
```diff
@@ -9,19 +9,25 @@ import Foundation
 
 public final class MockURLSession: URLSession {
   public private (set) var lastRequest: URLRequest?
-  
+
+  public var data: Data?
+  public var response: URLResponse?
+  public var error: Error?
```
kimdv: This way we can handle what the completion should return

kimdv:
@designatednerd I rebased and added two tests. One with error and one with none. 
I'm not sure that the tests are good enough. If not, please send some hints! 🚀

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -266,3 +313,11 @@ extension HTTPTransportTests: HTTPNetworkTransportRetryDelegate {
     }
   }
 }
+
+// MARK: - HTTPNetworkTransportGraphQLErrorDelegate
+
+extension HTTPTransportTests: HTTPNetworkTransportGraphQLErrorDelegate {
+  func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedGraphQLErrors errors: [GraphQLError], retryHandler: @escaping (Bool) -> Void) {
+    self.graphQlErrors = errors
```
designatednerd: doesn't the retry handler need to be called here?

kimdv: Doh 🤦‍♂️
Will be fixed

designatednerd:
@kimdv can you merge in and/or rebase off `master` so the test suite can run on Circle please?

kimdv:
@designatednerd done! 
Let's see what explodes 🔥🙈

designatednerd:
🎉 Nothing! 🎉 Mergin'

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Done in #872

RolandasRazma:
how do you know if failure after retrying? 
if request fails and you call `retryHandler(true)` and it fails again I don't see how to know that you should just give up

designatednerd:
That's why we explicitly recommend against just calling `retryHandler(true)` without any context in the documentation for this method and the retry delegate for network errors. 

If you're not looking at the actual errors that have come in, you could easily wind up in an infinite retry loop.

RolandasRazma:
I am. I'm thinking about case where server say "token invalid", I update token, and server say "still invalid". Correct thing to do is to give up, but I don't know that I already tried to update token

designatednerd:
It sounds a bit more like a bug in your server if a just-updated token is being rejected as invalid, but you could theoretically add some tracking of what the last (n) errors received were, and if some aspect of them (status code, message, etc) is the same, then stop retrying.

RolandasRazma:
maybe it's me, but I like to code defensively 🤷‍♂

designatednerd:
That's fair, but I think the "how do I handle situation x" stuff is more of an application-level decision, so we're letting developers handle that as defensively as they'd like to.

kimdv:
Hi! 
We had the same problem but solved it in the client. 
``` swift
final class GraphQLClient {
    private var retryHandlers = [(Bool) -> Void]()
    private let isolationQueue = DispatchQueue(label: "GraphQLClient.isolationQueue",
                                               qos: .userInitiated,
                                               attributes: [],
                                               autoreleaseFrequency: .workItem,
                                               target: nil)
}
```

``` swift
extension GraphQLClient: HTTPNetworkTransportGraphQLErrorDelegate {
    func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedGraphQLErrors errors: [GraphQLError], retryHandler: @escaping (Bool) -> Void) {
        let errors = errors.map { OurGraphQLError(error: $0) }
        
        guard error.contains(.authTokenInvalid) else {
            retryHandler(false)
            return
        }
        
        let isTokenExchangeInProgress = scheduleForExecution(retryHandler: retryHandler)
        if isTokenExchangeInProgress { return }
        
        exchangeRefreshToken { [weak self] result in
            guard let strongSelf = self else { return }
            
            switch result {
            case .success:
                strongSelf.executeRetryHandlers(retry: true)
            
            case .failure(let error):              
                strongSelf.executeRetryHandlers(retry: false)
            }
        }
    }

    private func scheduleForExecution(retryHandler: @escaping (Bool) -> Void ) -> Bool {
        var isTokenExchangeInProgress = true
        isolationQueue.sync {
            isTokenExchangeInProgress = !retryHandlers.isEmpty
            retryHandlers.append(retryHandler)
        }
        return isTokenExchangeInProgress
    }

    private func executeRetryHandlers(retry: Bool) {
        isolationQueue.async { [weak self] in
            guard let strongSelf = self else { return }
            for handler in strongSelf.retryHandlers {
                handler(retry)
            }
            strongSelf.retryHandlers = []
        }
    }
}
```

In this way we can cancel if refresh token exchange fails.

PR #771, opened 2019-09-18
Introducing RequestCreator protocol

kimdv:
This will make it possible to modify how requests are created. 

Useful in file uploads as the specs can be different from framework to framework

### DIFF
--- a/Sources/Apollo/GraphQLFile.swift
+++ b/Sources/Apollo/GraphQLFile.swift
@@ -2,11 +2,11 @@ import Foundation
 
 /// A file which can be uploaded to a GraphQL server
 public struct GraphQLFile {
-  let fieldName: String
-  let originalName: String
-  let mimeType: String
-  let inputStream: InputStream
-  let contentLength: UInt64
+  public let fieldName: String
+  public let originalName: String
+  public let mimeType: String
+  public let inputStream: InputStream
+  public let contentLength: UInt64
   
   /// A convenience constant for declaring your mimetype is octet-stream.
   public static let octetStreamMimeType = "application/octet-stream"
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -74,6 +74,7 @@ public class HTTPNetworkTransport {
   let serializationFormat = JSONSerializationFormat.self
   let useGETForQueries: Bool
   let delegate: HTTPNetworkTransportDelegate?
+  private let requestCreator: RequestCreator
   private let sendOperationIdentifiers: Bool
   
   /// Creates a network transport with the specified server URL and session configuration.
@@ -88,12 +89,14 @@ public class HTTPNetworkTransport {
               session: URLSession = .shared,
               sendOperationIdentifiers: Bool = false,
               useGETForQueries: Bool = false,
-              delegate: HTTPNetworkTransportDelegate? = nil) {
+              delegate: HTTPNetworkTransportDelegate? = nil,
+              requestCreator: RequestCreator? = nil) {
     self.url = url
     self.session = session
     self.sendOperationIdentifiers = sendOperationIdentifiers
     self.useGETForQueries = useGETForQueries
     self.delegate = delegate
+    self.requestCreator = requestCreator ?? ApolloRequestCreator()
   }
 
   private func send<Operation>(operation: Operation, files: [GraphQLFile]?, completionHandler: @escaping (_ results: Result<GraphQLResponse<Operation>, Error>) -> Void) -> Cancellable {
@@ -213,7 +216,7 @@ public class HTTPNetworkTransport {
   }
   
   private func createRequest<Operation: GraphQLOperation>(for operation: Operation, files: [GraphQLFile]?) throws -> URLRequest {
-    let body = RequestCreator.requestBody(for: operation, sendOperationIdentifiers: self.sendOperationIdentifiers)
+    let body = requestCreator.requestBody(for: operation, sendOperationIdentifiers: self.sendOperationIdentifiers)
     var request = URLRequest(url: self.url)
     
     // We default to json, but this can be changed below if needed.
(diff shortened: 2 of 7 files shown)


Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -88,12 +89,14 @@ public class HTTPNetworkTransport {
...
               useGETForQueries: Bool = false,
-              delegate: HTTPNetworkTransportDelegate? = nil) {
+              delegate: HTTPNetworkTransportDelegate? = nil,
+              requestCreator: RequestCreator? = nil) {
     self.url = url
     self.session = session
     self.sendOperationIdentifiers = sendOperationIdentifiers
     self.useGETForQueries = useGETForQueries
     self.delegate = delegate
+    self.requestCreator = requestCreator ?? ApolloRequestCreator()
```
designatednerd: I think it would make more sense to provide `ApolloRequestCreator` as  the default value here - then the parameter doesn't have to be nullable

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -88,12 +89,14 @@ public class HTTPNetworkTransport {
               session: URLSession = .shared,
               sendOperationIdentifiers: Bool = false,
               useGETForQueries: Bool = false,
-              delegate: HTTPNetworkTransportDelegate? = nil) {
+              delegate: HTTPNetworkTransportDelegate? = nil,
+              requestCreator: RequestCreator? = nil) {
```
designatednerd: can you please update the docs to include this parameter?

Sources/Apollo/MultipartFormData.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 /// A helper for building out multi-part form data for upload
-class MultipartFormData {
+public class MultipartFormData {
```
designatednerd: 1) Why does this need to be public? 
2) If it needs to be public, why aren't any other methods or properties public?

Sources/Apollo/RequestCreator.swift
```diff
@@ -50,48 +74,34 @@ public struct RequestCreator {
...
   }
 }
+
+// Helper struct to create requests independently of HTTP operations.
+struct ApolloRequestCreator: RequestCreator {
+
+  /// Default Apollo request creator
+  ///
+  /// - Returns: Default apollo request creator
+  static func `default`() -> RequestCreator {
```
designatednerd: does this need to exist? Wouldn't you just be able to create the default one using `ApolloRequestCreator()`?

Sources/Apollo/MultipartFormData.swift
kimdv: 1: Because the compiler is not happy otherwise :(
2: The actually should. If anyone make their own multipart creator they should be accessible

kimdv: Pushed public methods.

Sources/Apollo/MultipartFormData.swift
```diff
@@ -61,7 +61,7 @@ class MultipartFormData {
                     filename: filename)
   }
 
-  func appendPart(inputStream: InputStream,
+  public func appendPart(inputStream: InputStream,
```
designatednerd: If this needs to be public, can you please add documentation?

Sources/Apollo/MultipartFormData.swift
```diff
@@ -47,7 +47,7 @@ class MultipartFormData {
   ///   - name: The name of the part to pass along to the server
   ///   - contentType: [optional] The content type of this part. Defaults to nil.
   ///   - filename: [optional] The name of the file for this part. Defaults to nil.
-  func appendPart(data: Data,
+  public func appendPart(data: Data,
```
designatednerd: Can you please update the indentation here and below?

Sources/Apollo/RequestCreator.swift
```diff
@@ -50,48 +74,29 @@ public struct RequestCreator {
...
-    
-    for (index, file) in files.enumerated() {
-      formData.appendPart(inputStream: file.inputStream,
-                          contentLength: file.contentLength,
-                          name: "\(index)",
-                          contentType: file.mimeType,
-                          filename: file.originalName)
+
+    files.forEach {
+      formData.appendPart(inputStream: $0.inputStream, contentLength: $0.contentLength, name: $0.fieldName, contentType: $0.mimeType, filename: $0.originalName)
```
designatednerd: Why has all of this changed - making this public should not require a change to its functionality

Sources/Apollo/MultipartFormData.swift
kimdv: Pushed. Please check for typos. My English is not the best 😅

Sources/Apollo/RequestCreator.swift
kimdv: Sorry about that! Fixed 😅

kimdv: That is code for Absinthe file upload.

designatednerd:
@kimdv I think you missed a couple in the websocket tests

kimdv:
Should be fixed! 
Sorry 🤦‍♂️

Sources/Apollo/GraphQLFile.swift
```diff
@@ -2,11 +2,11 @@ import Foundation
...
-  let fieldName: String
-  let originalName: String
-  let mimeType: String
-  let inputStream: InputStream
-  let contentLength: UInt64
+  public let fieldName: String
+  public let originalName: String
+  public let mimeType: String
+  public let inputStream: InputStream
+  public let contentLength: UInt64
```
kimdv: Those fields need to be public if users have to make things like 
```
files.forEach {
            formData.appendPart(inputStream: $0.inputStream, contentLength: $0.contentLength, name: $0.fieldName, contentType: $0.mimeType, filename: $0.originalName)
        }
```

designatednerd:
:shipit: This will ship with 0.16.0 (hopefully early next week)

PR #784, opened 2019-09-23
Make buildable when using SPM in XCode 11

cheif:
I tried adding apollo to our project using SPM in XCode 11 and got some build issues.

The first issue was due to the swift version being specified to 4.0 in `Package.swift`, making XCode trying to build this with the old toolchain, leading to a build error at:
https://github.com/apollographql/apollo-ios/blob/beb6127feada2064dd25dc225d97549e4354462a/Sources/Apollo/GraphQLHTTPResponseError.swift#L39
Due to this depending on the nil-coalescing feature in swift 5.0, changing the `swift-tools-version` in `Package.swift` fixes this.

When this was fixed I got some build-errors due to `Foundation` not being imported in two files, so I added this as well.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:5.0
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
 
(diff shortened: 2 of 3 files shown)


designatednerd:
@cheif I have not seen this error in my usage of the betas - is this on a brand new project or on an existing project?

humblehacker:
@designatednerd I'm not the PR author, but I get the reported error on a brand new project.

cheif:
It’s on an existing project for me, I’ve been looking into migrating from CocoaPods to SPM with the arrival of a stable XCode 11.

designatednerd:
Bizarre! I wonder why this wasn't showing up in any of the betas. However, I don't see how these changes can hurt, so let's do this.

cheif:
Yeah, I agree that it's strange that it didn't show up in the betas, but maybe they were always building with swift 5.1 or something, since even if this could be built with swift 4.0, it wouldn't be possible to link against it from swift 5.1 due to the lack of module stability.

PR #804, opened 2019-10-02
Added correct way of default parameters

kimdv:
How I missed that I don't know.. 

Because the `HTTPNetworkTransport` didn't use `manualBoundary` it will use de protocol extension and not the requestCreator that is provided and implements the method.

I think this is the best way to implement "default" parameters

### DIFF
--- a/Sources/Apollo/MultipartFormData.swift
+++ b/Sources/Apollo/MultipartFormData.swift
@@ -34,7 +34,12 @@ public class MultipartFormData {
     self.init(boundary: "apollo-ios.boundary.\(UUID().uuidString)")
   }
 
-  func appendPart(string: String, name: String) throws {
+  /// Appends the passed-in string as a part of the body.
+  ///
+  /// - Parameters:
+  ///   - string: The string to append
+  ///   - name: The name of the part to pass along to the server
+  public func appendPart(string: String, name: String) throws {
     self.appendPart(data: try self.encode(string: string),
                     name: name,
                     contentType: nil)
--- a/Sources/Apollo/RequestCreator.swift
+++ b/Sources/Apollo/RequestCreator.swift
@@ -31,6 +31,15 @@ extension RequestCreator {
   ///
   /// - Parameters:
   ///   - operation: The operation to use
+  /// - Returns: The created `GraphQLMap`
+  public func requestBody<Operation: GraphQLOperation>(for operation: Operation) -> GraphQLMap {
+    return requestBody(for: operation, sendOperationIdentifiers: false)
+  }
+
+  /// Creates a `GraphQLMap` out of the passed-in operation
+  ///
+  /// - Parameters:
+  ///   - operation: The operation to use
   ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Defaults to false.
   /// - Returns: The created `GraphQLMap`
   public func requestBody<Operation: GraphQLOperation>(for operation: Operation, sendOperationIdentifiers: Bool = false) -> GraphQLMap {
@@ -59,6 +68,26 @@ extension RequestCreator {
   ///   - files: An array of files to use.
   ///   - sendOperationIdentifiers: True if operation identifiers should be sent, false if not.
   ///   - serializationFormat: The format to use to serialize data.
+  /// - Returns: The created form data
+  /// - Throws: Errors creating or loading the form  data
+  public func requestMultipartFormData<Operation: GraphQLOperation>(for operation: Operation,
+                                                                    files: [GraphQLFile],
+                                                                    sendOperationIdentifiers: Bool,
+                                                                    serializationFormat: JSONSerializationFormat.Type) throws -> MultipartFormData {
+    return try requestMultipartFormData(for: operation,
+                                        files: files,
+                                        sendOperationIdentifiers: sendOperationIdentifiers,
+                                        serializationFormat: serializationFormat,
+                                        manualBoundary: nil)
+  }
+
+  /// Creates multi-part form data to send with a request
+  ///
+  /// - Parameters:
+  ///   - operation: The operation to create the data for.
+  ///   - files: An array of files to use.
+  ///   - sendOperationIdentifiers: True if operation identifiers should be sent, false if not.
+  ///   - serializationFormat: The format to use to serialize data.
   ///   - manualBoundary: [optional] A manual boundary to pass in. A default boundary will be used otherwise.
   /// - Returns: The created form data
   /// - Throws: Errors creating or loading the form  data
@@ -66,7 +95,7 @@ extension RequestCreator {
                                                                     files: [GraphQLFile],
                                                                     sendOperationIdentifiers: Bool,
                                                                     serializationFormat: JSONSerializationFormat.Type,
-                                                                    manualBoundary: String? = nil) throws -> MultipartFormData {
+                                                                    manualBoundary: String?) throws -> MultipartFormData {
     let formData: MultipartFormData
 
     if let boundary = manualBoundary {
--- a/Sources/Apollo/RequestCreator.swift
+++ b/Sources/Apollo/RequestCreator.swift
@@ -42,7 +42,7 @@ extension RequestCreator {
   ///   - operation: The operation to use
   ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Defaults to false.
   /// - Returns: The created `GraphQLMap`
-  public func requestBody<Operation: GraphQLOperation>(for operation: Operation, sendOperationIdentifiers: Bool = false) -> GraphQLMap {
+  public func requestBody<Operation: GraphQLOperation>(for operation: Operation, sendOperationIdentifiers: Bool) -> GraphQLMap {
     var body: GraphQLMap = [
       "variables": operation.variables,
       "operationName": operation.operationName,
--- a/Sources/Apollo/RequestCreator.swift
+++ b/Sources/Apollo/RequestCreator.swift
@@ -31,15 +31,6 @@ extension RequestCreator {
   ///
   /// - Parameters:
   ///   - operation: The operation to use
-  /// - Returns: The created `GraphQLMap`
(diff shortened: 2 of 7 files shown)


kimdv:
Should I see if I can come up with some tests to test it works well?

Sources/Apollo/RequestCreator.swift
```diff
@@ -27,6 +27,15 @@ public protocol RequestCreator {
 }
 
 extension RequestCreator {
+  /// Creates a `GraphQLMap` out of the passed-in operation
+  ///
+  /// - Parameters:
+  ///   - operation: The operation to use
+  /// - Returns: The created `GraphQLMap`
+  public func requestBody<Operation: GraphQLOperation>(for operation: Operation) -> GraphQLMap {
+    return requestBody(for: operation, sendOperationIdentifiers: false)
```
designatednerd: Why is `sendOperationIdentifiers` hard-coded to `false` here?

kimdv: Because of this change I missed to correct: https://github.com/apollographql/apollo-ios/pull/804/commits/6e4cc7f333d554af00b8845244da46b92cee5c82

designatednerd: I would rather force callers to provide that `sendOperationIdentifiers` parameter if they're using the default implementation than create another method with such a similar signature.

Sources/Apollo/RequestCreator.swift
```diff
@@ -52,6 +61,26 @@ extension RequestCreator {
...
+  /// - Throws: Errors creating or loading the form  data
+  public func requestMultipartFormData<Operation: GraphQLOperation>(for operation: Operation,
+                                                                    files: [GraphQLFile],
+                                                                    sendOperationIdentifiers: Bool,
+                                                                    serializationFormat: JSONSerializationFormat.Type) throws -> MultipartFormData {
+    return try requestMultipartFormData(for: operation,
+                                        files: files,
+                                        sendOperationIdentifiers: sendOperationIdentifiers,
+                                        serializationFormat: serializationFormat,
+                                        manualBoundary: nil)
```
designatednerd: this also seems like something that could be done by having outside callers choose whether to pass nil or not

Sources/Apollo/RequestCreator.swift
kimdv: Will remove the two methods that "removes" the default parameter.

Sources/Apollo/RequestCreator.swift
kimdv: Will remove this method

kimdv:
@designatednerd added some tests. Don't know if it is enough? 
It is actually just to test if the correct implementation is used. 

Also changed name of test file to be more explicit as it also tests custom and requestBody now, 

Let me hear what you think 🚀

Tests/ApolloTests/GETTransformerTests.swift
```diff
@@ -90,7 +90,7 @@ class GETTransformerTests: XCTestCase {
   
   func testEncodingQueryWithNullDefaultParameter() {
     let operation = HeroNameQuery()
-    let body = requestCreator.requestBody(for: operation)
+    let body = requestCreator.requestBody(for: operation, sendOperationIdentifiers: false)
```
designatednerd: Probably worth adding one test to this file where `sendOperationIdentifers` is `true` just to make sure that's actually including the operation identifier

Tests/ApolloTests/RequestCreatorTests.swift
```diff
@@ -283,4 +284,135 @@ Bravo file content.
...
+    let alphaFileUrl = self.fileURLForFile(named: "a", extension: "txt")
+
+    let alphaFile = GraphQLFile(fieldName: "upload",
+                                originalName: "a.txt",
+                                mimeType: "text/plain",
+                                fileURL: alphaFileUrl)
+
+    let data = try customRequestCreator.requestMultipartFormData(
+      for: HeroNameQuery(),
+      files: [alphaFile!],
```
designatednerd: Probably worth `guard let` ing this on creation, since a force-unwrap failing in a test causes the whole test suite to crash rather than failing the single test.

Tests/ApolloTests/RequestCreatorTests.swift
```diff
@@ -283,4 +284,135 @@ Bravo file content.
...
+
+  func testRequestBodyWithCustomRequestCreator() {
+    let query = HeroNameQuery()
+    let req = customRequestCreator.requestBody(for: query, sendOperationIdentifiers: false)
+
+    XCTAssertEqual(query.queryDocument, req["test_query"] as? String)
+  }
+}
+
+fileprivate struct TestRequestCreator: RequestCreator {
```
designatednerd: Hm...I'm torn on whether to suggest pulling this out into a separate file - though if you do I'd probably call it something more like `TestCustomRequestCreator` to clarify that it's not the request creator that should be used for all tests.

Tests/ApolloTests/RequestCreatorTests.swift
```diff
@@ -283,4 +284,135 @@ Bravo file content.
...
+      guard let operationIdentifier = operation.operationIdentifier else {
+        preconditionFailure("To send operation identifiers, Apollo types must be generated with operationIdentifiers")
+      }
+
+      body["test_id"] = operationIdentifier
+    } else {
+      body["test_query"] = operation.queryDocument
+    }
+
+    return body
```
designatednerd: Are this implementation and the one below substantially different from the real implementation? Might be worth making them more obviously different to help make sure the "correct" implementation is being used.

Tests/ApolloTests/RequestCreatorTests.swift
```diff
@@ -283,4 +284,135 @@ Bravo file content.
...
+  }
+
+  public func requestMultipartFormData<Operation: GraphQLOperation>(for operation: Operation,
+                                                                        files: [GraphQLFile],
+                                                                        sendOperationIdentifiers: Bool,
+                                                                        serializationFormat: JSONSerializationFormat.Type,
+                                                                        manualBoundary: String?) throws -> MultipartFormData {
```
designatednerd: Alignment went a little goofy here

Tests/ApolloTests/RequestCreatorTests.swift
kimdv: This is just copy/paste from other code made by you 🙈
Should I remove all force unwrap in this class?

designatednerd: 🤦‍♀️Nah I'll fix it in a follow-up PR.

Tests/ApolloTests/RequestCreatorTests.swift
kimdv: The multipart is the one before the Apollo "standard" was added. It is actually the one that works with Absinthe GraphQL 😅

I have changed the names of the keys. 
I think it is okay as we only want to ensure the right method is called?
So all keys in the `requestBody` starts with `test_`. I think that is good enough to ensure that we not call the default implementation

Tests/ApolloTests/GETTransformerTests.swift
kimdv: Well, the default implementation here was actually false. 
Maybe that is for another PR? 
I can open an issue to you can track it?

kimdv: Just not to add something that is not related to this PR and RequestCreator? 😁

Tests/ApolloTests/RequestCreatorTests.swift
kimdv: Will pull out. I was a little should I do or not. 
Some people want to keep in file other want it out 🤷‍♂️

I like the second one.

Tests/ApolloTests/RequestCreatorTests.swift
kimdv: Sowwy 🙈

Tests/ApolloTests/GETTransformerTests.swift
designatednerd: Yeah I'll add it in a follow-up PR - you don't need to open an issue for it, I'll leave this comment open

Tests/ApolloTests/RequestCreatorTests.swift
designatednerd: Haha, I'm the moron who did it first, no worries

kimdv:
Ping me if you need help with clean up 😁

designatednerd:
@kimdv Tests seem to have bombed out on MacOS - can you double check that the files are added to the proper targets?

kimdv:
@designatednerd struct was fileprivate 🤦‍♂️

Sources/Apollo/MultipartFormData.swift
```diff
@@ -34,7 +34,12 @@ public class MultipartFormData {
     self.init(boundary: "apollo-ios.boundary.\(UUID().uuidString)")
   }
 
-  func appendPart(string: String, name: String) throws {
+  /// Appends the passed-in string as a part of the body.
+  ///
+  /// - Parameters:
+  ///   - string: The string to append
+  ///   - name: The name of the part to pass along to the server
+  public func appendPart(string: String, name: String) throws {
```
kimdv: @designatednerd this was only internal. The `TestCustomRequestCreator` used it and I know I use in my project too. 
https://github.com/apollographql/apollo-ios/pull/804/files#diff-b1b6d9ae6149908bf9768bbc4bc4b3b0R50

kimdv:
@designatednerd it seems that the multipart tests are a little unstable?
`// Operation parameters may be in weird order, so let's at least check that the files and single parameter got encoded properly.`

I ran it once on my machine. All good and pushed. 

Saw it failed and now it fails from time to time?

kimdv:
Wow. I saw the comment that the there is a weird order. But it took some time to find out why! 
But solved the flaky test. It should pass now 🧙‍♂️ (Gandalf emoji)

designatednerd:
Still approved! :shipit:

PR #812, opened 2019-10-04
use awk to split shasum from file

koenpunt:
This removes the need of converting it to a bash array.

### DIFF
--- a/scripts/run-bundled-codegen.sh
+++ b/scripts/run-bundled-codegen.sh
@@ -8,7 +8,7 @@ SCRIPT_DIR="$(dirname "$0")"
 
 # Get the SHASUM of the tarball
 ZIP_FILE="${SCRIPT_DIR}/apollo.tar.gz"
-SHASUM="$(/usr/bin/shasum -a 256 "${ZIP_FILE}")"
+SHASUM="$(/usr/bin/shasum -a 256 "${ZIP_FILE}" | /usr/bin/awk '{ print $1 }')"
 SHASUM_FILE="${SCRIPT_DIR}/apollo/.shasum"
 APOLLO_DIR="${SCRIPT_DIR}"/apollo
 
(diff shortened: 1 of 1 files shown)


scripts/run-bundled-codegen.sh
```diff
@@ -20,16 +20,14 @@ remove_existing_apollo() {
...
-  echo "${ARRAY[0]}" | tee "${SHASUM_FILE}"
+  echo "${SHASUM}" | tee "${SHASUM_FILE}"
 }
 
 validate_codegen_and_extract_if_needed() {
   # Make sure the SHASUM matches the release for this version
   EXPECTED_SHASUM="13febaa462e56679099d81502d530e16c3ddf1c6c2db06abe3822c0ef79fb9d2"
 
-  if [[ ${SHASUM} = ${EXPECTED_SHASUM}* ]]; then
+  if [[ ${SHASUM} = ${EXPECTED_SHASUM} ]]; then
```
designatednerd: The asterisk here (and below) is intentional - it's designed to match `0.16.0` produced versions of the `.shasum` file that also have the file name in there as well.

scripts/run-bundled-codegen.sh
```diff
@@ -8,7 +8,7 @@ SCRIPT_DIR="$(dirname "$0")"
 
 # Get the SHASUM of the tarball
 ZIP_FILE="${SCRIPT_DIR}/apollo.tar.gz"
-SHASUM="$(/usr/bin/shasum -a 256 "${ZIP_FILE}")"
+SHASUM="$(/usr/bin/shasum -a 256 "${ZIP_FILE}" | /usr/bin/awk '{ print $1 }')"
```
designatednerd: Not super-familiar with `awk` - why `$1` instead of `$0`?

koenpunt: `$0` would print the full input, `$1` prints the shasum, and `$2` would print the filename.

scripts/run-bundled-codegen.sh
koenpunt: Ah got ya.

PR #828, opened 2019-10-11
Consistancy on script variables in documentation

eridem:
The documentation make reference to `SCRIPT_PATH` in several parts.
It could be great to keep the consistency on the names for SPM script.


designatednerd:
Good catch, thank you!

PR #839, opened 2019-10-17
Update CI to use new Xcode and add test runs for iOS 13

designatednerd:
Gonna try to do some work to make this more efficient (possibly including switching CI providers), but first I'd at least like to get this up and running on current Xcode for testing.


designatednerd:
Flargh. I'm closing this out and just switching us to Circle, this is madness.

PR #854, opened 2019-10-22
Use weak references to Apollo client in closures

gsabran:
This PR changes a few closures that were keeping a strong reference to the Apollo client. This is to make sure that as soon as the user of Apollo client releases its reference, all Apollo operations will stop.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -64,7 +64,10 @@ public class ApolloClient {
   }
   
   fileprivate func send<Operation: GraphQLOperation>(operation: Operation, shouldPublishResultToStore: Bool, context: UnsafeMutableRawPointer?, resultHandler: @escaping GraphQLResultHandler<Operation.Data>) -> Cancellable {
-    return networkTransport.send(operation: operation) { result in
+    return networkTransport.send(operation: operation) { [weak self] result in
+      guard let self = self else {
+        return
+      }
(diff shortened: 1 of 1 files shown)


Sources/Apollo/ApolloClient.swift
```diff
@@ -206,7 +215,7 @@ private func wrapResultHandler<Data>(_ resultHandler: GraphQLResultHandler<Data>
 }
 
 private final class FetchQueryOperation<Query: GraphQLQuery>: AsynchronousOperation, Cancellable {
-  let client: ApolloClient
+  weak var client: ApolloClient?
```
designatednerd: Why does this need to be `weak` here?

gsabran: I think you're adding the operation to the client's queue
https://github.com/apollographql/apollo-ios/blob/509b2220c7a2e5dc5d0940b11361df0068ed4a0c/Sources/Apollo/ApolloClient.swift#L137-L138
so both the client and the operation will be retained here

designatednerd: oh woof, good point.

mitchellporter:
Anyone else who ends up here wondering why they're not getting any data back anymore... you're probably not holding onto a strong reference of the `ApolloClient` now that the retain cycle has been fixed.

designatednerd:
Ha, good catch! I'll make a note of it on the release notes for `0.18.0`

PR #857, opened 2019-10-23
Add default apollo engine headers

js:
If you're using Apollo Engine, there's an option to [identify clients and their versions](https://www.apollographql.com/docs/graph-manager/client-awareness/). 

To be perfectly frank I'm not quite sure how such an API should actually look in the iOS SDK, since the URLSession is passed in and there's really no existing API (?) for mutating the request outside of that. So not quite sure where this belongs, if at all, or how the API should look. So here's a getter with the headers and some default values that users can merge into their url session or whatever.

is this even worth doing? 🤷‍♂ 😄

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -78,7 +78,25 @@ public class HTTPNetworkTransport {
   let delegate: HTTPNetworkTransportDelegate?
   private let requestCreator: RequestCreator
   private let sendOperationIdentifiers: Bool
-  
+
+  /// The default client headers that Apollo Engine uses to identify its clients
+  /// Please see https://www.apollographql.com/docs/graph-manager/client-awareness/
+  ///
+  /// This should be set as headers on your URLRequests, if you're using Apollo Engine, eg:
+  /// sessionConfiguration.httpAdditionalHeaders = HTTPNetworkTransport.apolloEngineClientHeaders
+  static let apolloEngineClientHeaders: [String: String] = {
+    func bundleValue(forKey key: String) -> String? {
+      return Bundle.main.object(forInfoDictionaryKey: key) as? String
+    }
+    let identifier = bundleValue(forKey: String(kCFBundleIdentifierKey)) ?? ""
+    let version = bundleValue(forKey: "CFBundleShortVersionString") ??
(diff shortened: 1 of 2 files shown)


designatednerd:
Ha I was actually planning on doing something similar today. You can use the [pre-flight delegate's `willSend` method](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/HTTPNetworkTransport.swift#L24) on the `HTTPNetworkTransport` to alter headers yourself.

However I'd like to make these variables so that people can pass in their own values here - this could be helpful for people who want to specifically identify a beta client or A/B test variant or whatever else in a way that can't really be identified through the bundle identifier alone.

Would you mind if I took a stab at it on my end?

js:
Yeah saw the delegate, but since I need it on all GQL request anyway I just set it on the session config. 

Feel free to stab away at it if you got a plan, I certainly don’t :)

designatednerd:
🔪away shortly!

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -78,7 +78,25 @@ public class HTTPNetworkTransport {
...
+  static let apolloEngineClientHeaders: [String: String] = {
+    func bundleValue(forKey key: String) -> String? {
+      return Bundle.main.object(forInfoDictionaryKey: key) as? String
+    }
+    let identifier = bundleValue(forKey: String(kCFBundleIdentifierKey)) ?? ""
+    let version = bundleValue(forKey: "CFBundleShortVersionString") ??
+      bundleValue(forKey: String(kCFBundleVersionKey)) ?? ""
+    return [
+      "apollographql-client-name": identifier,
+      "apollographql-client-version": version,
```
cltnschlosser: Is this suppose to be `apollo-ios` and `0.18.0` or `com.example.app` and `<app-version>`?

designatednerd:
@js my 🔪at this went down a tiiiiiny bit of a 🐇🕳️, but it's up as #858.

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: It should probably be `com.example.app` and `<app-version>` to distinguish between different apps hitting the same graph - that's what I went with in #858

js:
Closing in favour of #858

PR #858, opened 2019-10-24
GraphManager/Engine Headers + minor WebSocket cleanup

designatednerd:
In this PR: 

- Added default and ability to set non-default name and version headers for Apollo GraphManager and Apollo Engine
- Pulled some web socket components out of the giant class file into their own classes, start adding inline documentation (to the extent that I understand what's happening 🙃). 
- Added a convenience extension for getting stuff out of a bundle's info dictionary.


Sources/Apollo/NetworkTransport.swift
```diff
@@ -8,8 +8,60 @@ public protocol NetworkTransport {
...
+    var version = String()
+    if let shortVersion = Bundle.main.shortVersion {
+      version.append(shortVersion)
+    }
+    
+    if let buildNumber = Bundle.main.buildNumber {
+      if version.isEmpty {
+        version.append(buildNumber)
+      } else {
+        version.append("-\(buildNumber)")
```
js: I actually decided against this on my end as I'm increasing the build number on each build (based of number of commits) and it makes the "Clients" view in Apollo Engine a bit noisy for the non-public environments/variants. YMMV of course and it's always configurable anyway.

Issue #874, opened 2019-10-29
Automate CocoaPods Releases


designatednerd:
I am tired of having my laptop tied up by `pod trunk push` 🙃. Should [use this guide](https://fuller.li/posts/automated-cocoapods-releases-with-ci/) to make CI do it instead.

PR #878, opened 2019-11-01
Add workflow for publishing to Cocoapods trunk automatically when a semver tag is added

designatednerd:
Won't really be able to fully test this until the next version lands, but this *should* address #874. Will leave that issue open until I actually publish a new version.


designatednerd:
[holding until we get our circle plan poked and I can run this job]

designatednerd:
OK well by accidentally not restricting what branches were being built I was able to validate that this both builds and pushes as the correct user to trunk, so I'm taking this off the milestone since it's been validated to work. 🎉

PR #880, opened 2019-11-03
Fixing data races in subscriptions

aivcec:
Current implementation of WebSocketTransport is not thread-safe so I took the steps to identify shared resources and possible race condition scenarios.  A lot of this was done using ideas from [the past pr](https://github.com/apollographql/apollo-ios/pull/307).

I wrote a couple of tests that help illustrate the problems. If you run them without ThreadSanitizer, sometimes they will pass with a success, but there is a chance that you will get an EXC_BAD_ACCESS error due to race conditions. In order to identify the issues check the ThreadSanitizer in the ApolloWebSocket/Test phase and run the tests that way.

Also, make sure to only run the tests in the StarWarsSubscriptionTests file, as the StarWarsWebSocketTests uses InMemoryNormalizedCache which causes TS problems on its own. I did not address those issues in this PR, this PR is only about making WebSocketTransport thread-safe.

Its important to note that transport methods can be called from two sources. Transport methods can be called from the outside/client (**A**) who can add subscribers, cancel them and close the connection. He can call the transport from any thread/queue he wants. The underlying socket (Starscream), however, only calls the transport from the specified callback queue (**B**) which defaults to Main. It is from that queue that all WebSocketDelegate methods are called from.

I will go through each set of shared resources and how I fixed the data races:

**1) sequenceNumber variable**

This variable is used to enumerate all subscribers. The data race arises if two subscribers are concurrently being added to the transport (both try to fetch and increment the counter).

Solution: I added a simple AtomicCounter class that NSLocks the current thread upon incrementing this value.

**2) Subscription arrays (subscribers and subscriptions)**

These arrays are being accessed/modified when:
  **A**) A new subscriber is being added, disconnected or the client is closing connection
  **B**) The underlying socket sends a new message, an error or unsubscribes a subscriber 

Solution: I think it is a bad idea to process starscream responses on the main queue so I added a separate serial processing queue to handle those tasks. This queue is set as a callbackQueue on the Websocket so all **B**) calls are done serially. To fix data races that stem from **A**) I made those perform on the same queue manually as well (`sendHelper`, `unsubscribe`, `closeConnection`).

**3) Message queue array**

This was fairly similar to the issue above and was solved by performing all write() calls on the callbackQueue as well. Some were made explicit, some are being done through the WebSocketDelegate/callbackQueue (**B**).

**4) Various state variables (reconnect, error)**

TSan was also reporting issues with concurrent access to these variables so I wrapped each of them with an NSLock (Atomic class). There are some other state variables (acked etc.) but they don't seem to be an issue atm as they are always called from the processing queue (from WebSocketDelegate methods).

As far as I can tell this fixes all the current concurrency issues in ApolloWebSocket, but I'm interested if you have other ideas for how to solve these issues. Some of these could be avoided by different architecture of the WebSocketTransport.

I also made some changes to MockWebSocket to support the addition of callback queue property.


Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
```diff
@@ -252,4 +254,107 @@ class StarWarsSubscriptionTests: XCTestCase {
...
+  
+  func testConcurrentSubscriptionCancellations() {
+    let firstSubscription = ReviewAddedSubscription(episode: .empire)
+    let secondSubscription = ReviewAddedSubscription(episode: .empire)
+    
+    let expectation = self.expectation(description: "Subscriptions cancelled")
+    expectation.expectedFulfillmentCount = 2
+    
+    let sub1 = client.subscribe(subscription: firstSubscription) { _ in }
+    let sub2 = client.subscribe(subscription: secondSubscription) { _ in }
```
designatednerd: Worth adding a post of a review, then failing either of these if something comes through since they're both being cancelled immediately?

Utilities/Atomic.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
```
designatednerd: This and the other file should be in `Sources/ApolloWebSocket` so they're included with Swift Package Manager.

Alternately it could go in `Sources/Apollo` if you make the class `public` - I have a feeling this may be a good replacement for some other locking code we have that's not quite working the way we want it to.

Utilities/AtomicCounter.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+class AtomicCounter {
+  private let lock = NSLock()
+  private var _value = 0
```
designatednerd: I see why you used the underscore in `Atomic<T>` for `_value` to disambiguate between the private property and the public one, but here I think the underscore is unnecessary since there's nothing to disambiguate and the property is entirely private.

Utilities/AtomicCounter.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+class AtomicCounter {
+  private let lock = NSLock()
```
designatednerd: Why does this need to have its own lock rather than having an `Atomic<Int>` for the value?

Sources/ApolloWebSocket/ApolloWebSocket.swift
```diff
@@ -14,6 +14,9 @@ public protocol ApolloWebSocketClient: WebSocketClient {
   
   /// The URLRequest used on connection. 
   var request: URLRequest { get set }
+  
+  /// Queue where the callbacks are executed
+  var callbackQueue: DispatchQueue { get set }
```
designatednerd: In other places like in the `HTTPNetworkTransport`, we've allowed the callback queue to be specified at the point where a method like `query` is called so that different information can be processed on different callback queues. 

I'm not sure how straightforward it would be to do that with `subscription` since I haven't had a chance to dive into the real nitty gritty of this code, but what do you think about that architecture vs setting the callback queue per websocket instance?

aivcec: I was mimicking the implementation of the underlying Starscream `WebSocket`. This is the queue from which a websocket calls its `WebSocketDelegate` methods. As for the queue where the subscription callbacks (`Result<...> -> Void`) are processed, they could be added to the subscriber list to be used when notifying subscribers. I feel like this is beyond the scope of this PR.

I added this to be able to easily set the Starscream callback queue. It didn't see like a big deal since the `ApolloWebSocketClient` implementation is only ever changed during testing. (AFAIK)

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
aivcec: I can do that if you think it looks better, but I think it wouldn't really matter.

I feel like these tests only make sense if TSAN is turned on anyway.

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
```diff
@@ -252,4 +254,107 @@ class StarWarsSubscriptionTests: XCTestCase {
...
+    
+    DispatchQueue.global().async {
+      sub1.cancel()
+      expectation.fulfill()
```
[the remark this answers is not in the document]
aivcec: You mean that you would like to test if the subscriber is actually removed from the `WebSocketTransport`? This could be done if the processing in `disconnect()` is done using `sync` instead of `async`. 

But I think we are just testing here if two subscribers can _attempt_ to cancel concurrently.
Maybe the description is flawed?

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
```diff
@@ -252,4 +254,107 @@ class StarWarsSubscriptionTests: XCTestCase {
...
+    
+    waitForExpectations(timeout: 10, handler: nil)
+  }
+  
+  func testConcurrentSubscriptionAndConnectionClose() {
+    let empireReviewSubscription = ReviewAddedSubscription(episode: .empire)
+    let expectation = self.expectation(description: "Connection closed")
+    
+    DispatchQueue.global().async {
+      let sub = self.client.subscribe(subscription: empireReviewSubscription) { _ in }
```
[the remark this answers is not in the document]
aivcec: Same as above. I can do that.

Utilities/Atomic.swift
aivcec: Oops, wasn't aware of the folder structure. I'll go with the first option for now.

Utilities/AtomicCounter.swift
aivcec: Sure

Utilities/AtomicCounter.swift
aivcec: That would only work if `Atomic<Int>` had a method that increments the value in one lock/unlock, as incrementing "manually" would imply that you get the original and set the incremented value - thus locking twice. I can add an extension that does just that. In that case I feel like I don't need this class at all.

My initial idea was to subclass `Atomic<Int>`, but that only worked with a recursive lock.

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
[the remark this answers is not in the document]
designatednerd: More that `sub1` was, in fact, cancelled. But I suppose that'd require subclassing the subscription to check whether `cancel` got called

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
designatednerd: I'm not particularly concerned about the way it looks, but it'd be nice to have a way to test these without TSAN turn on since right now we're not able to turn it on for the library as a whole (yet)

Utilities/AtomicCounter.swift
designatednerd: ahhhh yeah because you need one lock to access and another to modify. Hmmmm....

I like the idea of not having to have a separate class to modify the value. I wonder if we could do something like a `updateValue(_ value: @autoclosure () -> T)`, which would allow access+modify with the built-in lock. I'm not sure that's totally possible with an overridden setter though...

Sources/ApolloWebSocket/ApolloWebSocket.swift
designatednerd: Ahhh right I forgot about that implementation detail. You're right, this makes more sense!

Utilities/AtomicCounter.swift
aivcec: You can work with _value directly so it would be possible.
It would look like `updateValue(_ value: (T) -> T)` though, at least for this use case.

aivcec: `updateValue(_ value: (T) -> T) -> T` actually, which might be a bit convoluted :D

aivcec: I did it with a simple extension of the Atomic<Int> type, to keep it simple for now.

Sources/ApolloWebSocket/Utilities/Atomic.swift
```diff
@@ -0,0 +1,36 @@
...
+    set {
+      lock.lock()
+      defer { lock.unlock() }
+      
+      _value = newValue
+    }
+  }
+}
+
+extension Atomic where T == Int {
```
designatednerd: Love it!

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
designatednerd: ^still thinking on this - is this something where we should at least be able to say "These should never get called, even without TSAN on?'

aivcec: Yeah, I see what you mean. Kinda misunderstood you before.

I will perform a mutation after calling `cancel` to confirm.

aivcec: This is what I'm trying to do:  

    let firstSubscription = ReviewAddedSubscription(episode: .empire)
    let secondSubscription = ReviewAddedSubscription(episode: .empire)
    let expectation = self.expectation(description: "Subscriptions cancelled")
    expectation.expectedFulfillmentCount = 2
        
    let sub1 = client.subscribe(subscription: firstSubscription) { _ in
      XCTFail("Unexpected subscription response")
    }
    let sub2 = client.subscribe(subscription: secondSubscription) { _ in
      XCTFail("Unexpected subscription response")
    }
    
    let queue = DispatchQueue(label: "com.apollographql.testing", attributes: .concurrent)
    
    queue.async {
      sub1.cancel()
      expectation.fulfill()
    }
    queue.async {
      sub2.cancel()
      expectation.fulfill()
    }
    queue.async(flags: .barrier) {
      _ = self.client.perform(mutation: CreateReviewForEpisodeMutation(episode: .empire, review: ReviewInput(stars: 5, commentary: "The greatest movie ever!")))
    }
    
    waitForExpectations(timeout: 10, handler: nil)

But what happens is that final mutation sometimes triggers a subscription response in another test - `testMultipleSubscriptions`. Its callbacks get called too many times and the whole test fails. Testing this way can be achieved if the test sleeps after `waitForExpectations` for a while, but I feel thats just messy.

designatednerd: I think waiting for expectations is totally reasonable - you also can do an inverted expectation where the test fails if it actually gets called - so instead of having the explicit `XCTFail`s you can just have an `invertedExpectation.fulfill()`. You'd probably only need to wait that one 1-2 seconds, since the perform mutation should happen a lot faster than that.

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
```diff
@@ -252,4 +254,107 @@ class StarWarsSubscriptionTests: XCTestCase {
...
+      sub.cancel()
+    }
+    
+    _ = self.client.perform(mutation: CreateReviewForEpisodeMutation(episode: .empire, review: ReviewInput(stars: 5, commentary: "The greatest movie ever!")))
+    
+    DispatchQueue.global().async {
+      self.webSocketTransport.closeConnection()
+      expectation.fulfill()
```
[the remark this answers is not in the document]
aivcec: Do you think it's enough to check if `websocketTransport` is disconnected after waiting for expectation?

I don't like testing a mocked wrapper of a class and I don't think theres a way to confirm the transport is closed right away, as `closeConnection()` is being performed asynchronously.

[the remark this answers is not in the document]
designatednerd: Yeah, I suppose that's fair. We should punt on this for now, hopefully we'll be able to get TSAN running in full in the not-insanely-distant-future.

Tests/ApolloWebsocketTests/StarWarsSubscriptionTests.swift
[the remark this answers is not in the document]
designatednerd: (punting on this too)

designatednerd:
This will go out with 0.20.0, so keep an eye out for that release (hopefully later today)

PR #881, opened 2019-11-04
add advanced codegen tip using a temporary file

koenpunt:
> In addition to letting Xcode to decide whether or not to run the codegen based on input and output files, you can also decide to first write the file to a temporary location, and then compare this temporary file to the current one. And then only when the files differ you move the temporary file into place.

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -257,7 +257,9 @@ Make sure to uncheck the "Copy Files If Needed" checkbox, since it should alread
 
 Here are a few things you can do to make your code generation even better once you've gotten up and running with the basics above:
 
-### Set Up Input And Output Files In Your Build Phase
+### Prevent Unnecessary Recompilation 
+
+#### Set Up Input And Output Files In Your Build Phase
 
 Particularly if you're using Interface Builder or SwiftUI to talk to a module which also has a code generation build step, this is helpful to prevent the `API.swift` file from causing an auto-regeneration loop. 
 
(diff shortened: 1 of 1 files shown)


docs/source/installation.md
```diff
@@ -267,14 +269,29 @@ For example, if you're using something like this to run your code generation for
 "${SCRIPT_PATH}"/run-bundled-codegen.sh codegen:generate --target=swift --includes=./**/*.graphql --localSchemaFile="schema.json" API.swift
 ```
 
-Assuming youv'e set the script out to run from `$(SRCROOT)/YourTarget`, you can add `$(SRCROOT)/YourTarget/**/*.graphql` (the path you're running it from + the glob you're passing to the `includes` CLI parameter) to the list of `Input Files` for your Apollo Run Script Build phase. Then you can add `$(SRCROOT)/YourTarget/API.swift` (the path you're running it from + the output file) to the list of `Output Files`: 
+Assuming you've set the script out to run from `$(SRCROOT)/YourTarget`, you can add `$(SRCROOT)/YourTarget/**/*.graphql` (the path you're running it from + the glob you're passing to the `includes` CLI parameter) to the list of `Input Files` for your Apollo Run Script Build phase. Then you can add `$(SRCROOT)/YourTarget/API.swift` (the path you're running it from + the output file) to the list of `Output Files`:
```
designatednerd: 🤦‍♀️good catch

docs/source/installation.md
```diff
@@ -267,14 +269,29 @@ For example, if you're using something like this to run your code generation for
...
+
+```
+"${SCRIPT_PATH}"/run-bundled-codegen.sh codegen:generate --target=swift --includes=./**/*.graphql --localSchemaFile="schema.json" "${SRCROOT}/YourTarget/API.swift.new"
+if ! diff -q "${SRCROOT}/YourTarget/API.swift.new" "${SRCROOT}/YourTarget/API.swift"; then
+  mv "${SRCROOT}/YourTarget/API.swift.new" "${SRCROOT}/YourTarget/API.swift"
+else
+  rm "${SRCROOT}/YourTarget/API.swift.new"
+fi
+```
+
```
designatednerd: I'm a little torn on whether to make this an official suggestion, because it does almost all the work of the CLI in order to prevent the recompile loop. 

Setting stuff up file inputs/outputs with Xcode should prevent the codegen from even spinning up, which would both avoid the recompile loop and save a bunch of time on the parsing. 

What do you see as the advantage of this method over specifying input/outputs?

koenpunt: I've seen it happening that sometimes the graphql source files are touched somehow, but the contents didn't change, and then the API is still regenerated, and then Xcode will recompile the file.

It would of course be better if the CLI only writes the file if it actually changed, then this would for sure not be necessary.

designatednerd: Yeah not rewriting if nothing changed is definitely something I'd like to do in Swift Codegen but that's probably still a bit out. Unfortunately I don't think this is terribly easy to set up at the CLI level, but I'll reach out to the tooling team.

As to why it's rewriting, I wonder if it's something around auto-trimming of whitespace or some nonsense like that.

How about this: Throw in a note that you should only use this if for some reason the Xcode file input/output method isn't working, since that method doesn't even spin up the codegen if it's working, which is way more efficient. Then I think we can call it good.

koenpunt: Done!

Issue #348, opened 2018-08-13
Enum type causes auto generated API file to fail


Dmurph24:
**Issue:**

Apollo creates an invalid auto generated file from a graphql query when it contains a `enum` type. When the class with type `enum` is generated, it extends the `Apollo.JSONEncodable` and `Apollo.JSONDecodable` classes, which do not appear to actually exist.

![image](https://user-images.githubusercontent.com/1217029/44051866-0cffa052-9f09-11e8-89a6-ad2caa52a793.png)

```
'JSONEncodable' is not a member type of 'Apollo'
'JSONDecodable' is not a member type of 'Apollo'
```

**Partial Solution:**

To fix this temporarily I auto generated the invalid code for this query, copied the contents, and put them into a separate swift file. To fix the `'JSONEncodable' is not a member type of 'Apollo'` warning I simply changed `Apollo.JSONEncodable` to `JSONEncodable`. It seems as though if you just remove the `Apollo` namespace from the class extensions, this fixes the issue.

**GraphQL Query:**
The graphql query works fine in postman / insomnia. I can't give out the exact query, but it would be similar to this, given the knowledge of the schema below:
```graphql
query Icons {
    icons {
       icon
    }
}

```

**Schema:**

```json
        ...
        {
          "kind": "ENUM",
          "name": "Icon",
          "description": "An enum of icons used by the system",
          "fields": null,
          "inputFields": null,
          "interfaces": null,
          "enumValues": [
            {
              "name": "icon1",
              "description": null,
              "isDeprecated": false,
              "deprecationReason": null
            },
            {
              "name": "icon2",
              "description": null,
              "isDeprecated": false,
              "deprecationReason": null
            },
            ...
          ],
          "possibleTypes": null
        },
        ...
```

The partial solution is starting to be very tedious. I found it strange that no once else has ran into this issue.

Vic-L:
I had this error because I have a class with the same name in an attempt to make a singleton.

```
class Apollo {
    static let shared = Apollo()
    let client: ApolloClient

    init() {
        client = ApolloClient(url: URL(string: ProcessInfo.processInfo.environment["GRAPHQL_BASE_URL"]!)!)
    }
}
```

It was fine after I renamed it. It is probably interfering with the namespace.

Dmurph24:
I am doing the exact same thing. Thanks for the response. Ill change the class name later but I'm certain that is the issue that is happening.

cczufish:
[...]

PR #889, opened 2019-11-07
Update singleton naming suggestions in docs

designatednerd:
Some recent replies to #348 uncovered that our default naming suggestion for a singleton, which had been `Apollo`, could cause some conflicts with enums, which need to conform to `Apollo.JSONEncodable` and `Apollo.JSONDecodable`. 

I've updated the suggested singleton naming to `Network.shared.apollo`, since that at least will not cause conflicts with our own damned library 🤦‍♀.


designatednerd:
In b4 the "Never use a singleton!" crowd shows up 😆

PR #890, opened 2019-11-07
New Tutorial, Parts 0 and 1

designatednerd:
In this PR: 

- Added introduction
- Added getting started, which walks you from creating a new project to having a query that successfully hits the server. 

Notes: 

- ~~I'm going to be doing this with SwiftUI, mostly because for a more prototype-y UI like we're going to be building, it's a LOT less code.~~ UNDO! UNDO! Gonna take me way longer to get up to speed than I was hoping, so I'll be doing things the Olde Skool way. That doesn't really affect these two parts though.
- We're eventually going to deploy this on heroku to avoid some of the silliness with CodeSandbox nuking and restarting containers on a regular basis, but until then, I've got it all using CS. 
- The files for parts 2 and 3 are mostly there as placeholders - I moved some overflow from pt 1 in there already but those are not ready for a review of content. 
- Code is gonna go in a separate repo - I'm gonna do starter/final projects per part but I want to do that when I have the final version of the tutorial written, adjusting all that as you go is absolutely maddening.


designatednerd:
(Anyone who wants to be a guinea pig and try this out please feel free!)

PR #902, opened 2019-11-14
New Tutorial, Part 2

designatednerd:
Here's the part showing how to use queries! Notes from the [Part 1 PR](https://github.com/apollographql/apollo-ios/pull/890) apply, I'm going to leave this in draft until that one gets merged since it has a fairly large number of the same commits.


designatednerd:
@StephenBarlow Heads up that I'm gonna merge this part and then you can do a separate PR off master to split it up

PR #903, opened 2019-11-15
fix for swift 4.2

mishagray:
fixing a compilation issue with swift 4.2

### DIFF
--- a/Sources/Apollo/GraphQLHTTPResponseError.swift
+++ b/Sources/Apollo/GraphQLHTTPResponseError.swift
@@ -41,8 +41,9 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
   public var graphQLErrors: [GraphQLError]? {
     guard
       let data = self.body,
-      let json = try? self.serializationFormat.deserialize(data: data) as? JSONObject,
-      let errorArray = json?["errors"] as? [JSONObject] else {
+      let deserializedData = try? self.serializationFormat.deserialize(data: data),
+      let json = deserializedData as? JSONObject,
+      let errorArray = json["errors"] as? [JSONObject] else {
         return nil
     }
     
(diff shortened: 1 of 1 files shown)


Sources/Apollo/GraphQLHTTPResponseError.swift
```diff
@@ -42,7 +42,7 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
     guard
       let data = self.body,
       let json = try? self.serializationFormat.deserialize(data: data) as? JSONObject,
-      let errorArray = json["errors"] as? [JSONObject] else {
+      let errorArray = json?["errors"] as? [JSONObject] else {
```
designatednerd: This shouldn't be necessary - we've unwrapped `json` directly above.

What version of Swift are you using?

mishagray: I'm just adding the code via cocoapods, and building it with XCode 11.2.1

mishagray: Using swift '5' 
5.1.2
I think 'some' swift variants are casting the 'try?' differently now.   Pushing another 'attempt'.

designatednerd: Right - what version of Swift is your project using though, Swift 4 or Swift 5? If you're on 4, `try?` and `as?` on the same line can create a double-optional, whereas on 5, it should only create a single optional. This is related to the version of Swift you're using rather than the version of Xcode. 

Please note that this project supports Swift 5 and higher.

mishagray: oops...  I was setting it to swift 4.2...

mishagray: yeah.. this is my bad.   There was a bad post_install cocoa pod script that clobbered this to swift 4.2  
Not sure if you even want to support 4.2, so take the code or not...

mishagray:
I guess this PR isn't needed... it does fix swift 4.2 compatibility, but I'm not sure you want to add that as a thing you are going to support anyways.  I'm fixing my bad Podfile to not need this branch...

designatednerd:
I'd rather leave this issue as a sort of "canary in the coal mine" indicator that you're using Swift 4.2 instead of 5.0. I'm gonna go ahead and close this out. 

Thank you for the PR though!

PR #906, opened 2019-11-16
ApolloWebSocket should depend on Apollo since it imports Apollo.

JimRoepcke:
Trying to profile my iOS app project failed because SplitNetworkTransport.swift does `import Apollo` but ApolloWebSocket doesn't depend on Apollo in Package.swift like it needs to.

With this change I am able to profile my app. I'm using Xcode 11.1.


designatednerd:
Good catch!

PR #908, opened 2019-11-16
Export version number as a constant

manicmaniac:
## Motivation

I'm a developer of [ApolloDeveloperKit](https://github.com/manicmaniac/ApolloDeveloperKit), one of libraries that depends on `apollo-ios`.
Sometimes I want to get the version of `Apollo` that is linked with my library, usually in order to support wide versions of `Apollo`.

Fortunately Xcode supports `apple-generic` versioning system.
Xcode will automatically export the version number set by `CURRENT_PROJECT_VERSION` build setting variable if `apple-generic` is enabled.

## Overview

I changed one of Xcode build configuration files so as to enable `apple-generic` versioning system.
Then I wrote a test to ensure the existence of a constant `ApolloVersionNumber`.

## Details

There's few documentations about `apple-generic` versioning system so I describe what I know.

Once `apple-generic` versioning system is enabled, Xcode automatically generates `Apollo_vers.c` file in `$DERIVED_FILES_DIR` and it looks like the following:

```c
 extern const unsigned char ApolloVersionString[];
 extern const double ApolloVersionNumber;

 const unsigned char ApolloVersionString[] __attribute__ ((used)) = "@(#)PROGRAM:Apollo  PROJECT:Apollo-0.19.0" "\n";
 const double ApolloVersionNumber __attribute__ ((used)) = (double)0.19;
```

At the compile time, any files in `$DERIVED_FILES_DIR` will be compiled and linked with the built binary so users of the library can access exported `ApolloVersionNumber`.

Objective-C users can access `ApolloVersionString[]` as well but somehow Swift users can't.

Some famous libraries like [Alamofire](https://github.com/Alamofire/Alamofire/blob/262fc46a273c4e01483aee933a8230b4b253ef49/Alamofire.xcodeproj/project.pbxproj#L1751) has the same setting.

I don't know any disadvantages brought from `apple-generic` (maybe nothing).


designatednerd:
Ooh, neat! 

One critical question: Where does this grab the value for the library version from? We've got the [`Project-Version.xcconfig`](https://github.com/apollographql/apollo-ios/blob/master/Configuration/Shared/Project-Version.xcconfig) file setting this essentially globally, I'd prefer to avoid having a second place to need to set it if at all possible.

manicmaniac:
> Where does this grab the value for the library version from?

I confirmed that it is `CURRENT_PROJECT_VERSION`.

designatednerd:
OK cool, and that's what's set  in the `xcconfig` file.  Lovely. Let's do it!

PR #911, opened 2019-11-19
Content edit for first tutorial batch

StephenBarlow:


docs/gatsby-config.js
```diff
@@ -25,8 +25,10 @@ module.exports = {
             'api-reference'
           ],
           Tutorial: [
-            'tutorial/tutorial_0',
-            'tutorial/tutorial_1'
+            'tutorial/tutorial-introduction',
+            'tutorial/tutorial-create-project',
+            'tutorial/tutorial-obtain-schema',
+            'tutorial/tutorial-execute-query'
```
designatednerd: i assume these are semantic instead of numbered for SEO reasons?

docs/source/tutorial/tutorial-create-project.md
```diff
@@ -0,0 +1,51 @@
...
+
+    ![select the first and third targets](images/select_libs.png)
+
+6. Click **Finish**. Swift Package Manager (SPM) fetches your dependencies. When it completes, you can see them in the project navigator:
+
+    ![screenshot of installed dependencies](images/installed_dependencies.png)
+
+> **Note:** Because Swift Package Manager has not yet implemented [Target-Based Dependency Resolution](https://github.com/apple/swift-evolution/blob/master/proposals/0226-package-manager-target-based-dep-resolution.md), you'll see the `SQLite` dependency even though you didn't select it.
+
+Now you've got your project set up, and the Apollo dependencies added. It's time to start pulling together the things you need to build your API!
```
designatednerd: Can you talk about why you'd like to break this bit out as a separate step (beyond, I assume, length)? I really feel like we need to emphasize that you're not done setting up until you've got all the codegen stuff set up, because that's a *huge* difference between this library and others on iOS.

docs/source/tutorial/tutorial-introduction.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "Step 0: Introduction"
```
designatednerd: I think "Step" is redundant here, especially for the left sidebar. In the [fullstack tutorial](https://www.apollographql.com/docs/tutorial/introduction/) we just use `0.`, `1.`, etc. as the prefixes. Should we do that here as well?

docs/source/tutorial/tutorial-obtain-schema.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Step 2: Obtain your GraphQL schema"
+---
+
+## Explore your server's schema
+
+This tutorial uses the same GraphQL server that's provided as part of [the Apollo full-stack tutorial](https://www.apollographql.com/docs/tutorial/introduction/). You can visit [`https://n1kqy.sse.codesandbox.io/`](https://n1kqy.sse.codesandbox.io/) to start it up and open the GraphiQL tool to explore its schema:
```
designatednerd: This is...not correct - it's a modified version of the server that you wind up _creating_ in the full-stack tutorial, it is not provided during that tutorial at all. 

For issues like this, would you prefer I address this after merge and then have you take another quick run-through?

docs/gatsby-config.js
StephenBarlow: Yep, SEO plus more reader-friendly URLs

docs/source/tutorial/tutorial-introduction.md
StephenBarlow: Ah! Yes. Will address

docs/source/tutorial/tutorial-create-project.md
StephenBarlow: Length is the reason, yeah. Understood that this step alone now does not constitute being done with "setup". Let's leave them divided for now and revisit the total structure when we're nearly ready to go live.

docs/source/tutorial/tutorial-obtain-schema.md
StephenBarlow: Issues like this are great to resolve before merge! Resolving now. And yes I am bound to make more edits along the way that are not correct; please do not hesitate to let me know when I do :D

StephenBarlow:
@designatednerd Feedback (hopefully) addressed! Thanks for the note on separating content edits from file restructuring, that was my bad.

docs/source/tutorial/tutorial-create-project.md
designatednerd: Yeah "Now that you've got your project set up" is definitely not correct given that there are two more tutorial parts before you're done setting up 🙃

docs/source/tutorial/tutorial-obtain-schema.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: "Step 2: Obtain your GraphQL schema"
+title: "2. Obtain your GraphQL schema"
 ---
 
 ## Explore your server's schema
 
-This tutorial uses the same GraphQL server that's provided as part of [the Apollo full-stack tutorial](https://www.apollographql.com/docs/tutorial/introduction/). You can visit [`https://n1kqy.sse.codesandbox.io/`](https://n1kqy.sse.codesandbox.io/) to start it up and open the GraphiQL tool to explore its schema: 
+This tutorial uses the same GraphQL server that you build as part of [the Apollo full-stack tutorial](https://www.apollographql.com/docs/tutorial/introduction/). You can visit [`https://n1kqy.sse.codesandbox.io/`](https://n1kqy.sse.codesandbox.io/) to start it up and open the GraphiQL tool to explore its schema:
```
designatednerd: Nit: Not the same server, a modified version of the server.

docs/source/tutorial/tutorial-execute-query.md
```diff
@@ -0,0 +1,124 @@
...
+    hasMore
+  }  
+}
+```
+
+If you run this query by pressing the play button in GraphiQL, the query returns results as a JSON object on the right-hand side of the page: 
+
+![basic query JSON in GraphiQL](images/completed_basic_query.png)
+
+This query executes successfully, but it doesn't include any information about the `launches`! That's because we didn't include the necessary field in our query.
```
designatednerd: I think it's worth keeping the explanation of the distinction between `Launch` and `LaunchConnection` here - it explains why you hit the `launches` endpoint but didn't get an array of launches back as the top-level object

docs/source/tutorial/tutorial-execute-query.md
```diff
@@ -0,0 +1,124 @@
...
+
+2. Click **Next** and name the file `LaunchList.graphql`. Make sure it's saved at the same level as your `schema.json` file. As previously, don't add it to any target.
+
+3. Copy your final query from GraphiQL and paste it into `schema.json`. 
+
+You're now ready to generate code from the combination of your saved query and schema.
+
+## Running code generation
+
+1. Return to your project's **Apollo CLI** Run Script build phase. Comment out the line that you added to the bottom (that includes `apollo:schema`) and _uncomment_ the line you previously commented out (that includes `codegen:generate`).
```
designatednerd: I think it's worth leaving the explanation of why you don't need the schema command anymore (basically, because the schema's not changing here)

docs/source/tutorial/tutorial-execute-query.md
```diff
@@ -0,0 +1,124 @@
...
+        print("Failure! Error: \(error)")
+      }
+    }
+```
+
+Build and run your application. CodeSandbox might take a few seconds to spin up your GraphQL server if nobody's been using it recently, but once it's up, you should see a response that resembles the following: 
+
+![success log output](images/success_log_barf.png)
+
+You're now successfully fetching data from the network using generated code! Now it's time to move on to [displaying query results in your UI](tutorial_2).
```
designatednerd: since you took out the links at the bottom of the other tutorials, should we get rid of this one as well?

docs/source/tutorial/tutorial-introduction.md
```diff
@@ -0,0 +1,31 @@
...
+* Error handling
+* Query watching
+* Subscriptions
+* File uploads
+
+A pre-built GraphQL server is provided for your app to execute operations against.
+
+## Ready to start?
+
+[🚀 Let's go! 🚀](./tutorial-create-project)
```
designatednerd: why does this need `./` instead of the name of the file?

StephenBarlow: This is purely a personal preference thing; all links between articles in the same docset should be relative to ensure that the link checker can identify broken ones, and this looks more like a relative path to me

docs/source/tutorial/tutorial-create-project.md
```diff
@@ -0,0 +1,51 @@
...
+
+5. Name the project **RocketReserver**. Make sure the language is **Swift** and the User Interface is **Storyboard**:
+
+    ![options for creating a new project](images/options_for_project.png)
+
+6. Click **Next**. Select a location to save your new project and click **Finish**.
+
+Xcode creates and opens your new project, which has the following structure:
+
+<img alt="Initial project structure" src="images/initial_file_setup.png" width="300">
```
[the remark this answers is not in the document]
StephenBarlow: Correct!

docs/source/tutorial/tutorial-execute-query.md
StephenBarlow: Ah, hm, I'll add links back to the others instead :D

docs/source/tutorial/tutorial-obtain-schema.md
```diff
@@ -0,0 +1,62 @@
...
+
+8. Build your project to execute the script. In Finder, navigate to the folder that contains your `AppDelegate.swift` file. The folder should now include the downloaded `schema.json` file. Drag this file from Finder into Xcode:
+
+    ![where to drag the schema file](images/drag_schema_into_xcode.png)
+
+    When Xcode offers to add the schema file, make sure **all targets are unchecked** to reduce the size of your application bundle: 
+
+    ![all targets unchecked in dialog](images/dont_add_to_target.png)
+
+Your server's schema is now available locally. It defines what it's possible to request from your server. Next, you'll create your first **operation** against the schema, which fetches or updates information in your graph.
```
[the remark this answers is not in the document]
StephenBarlow: 👍 moving the second sentence up to the very top

docs/source/tutorial/tutorial-obtain-schema.md
```diff
@@ -0,0 +1,62 @@
...
+    
+4. Drag the newly created phase up between "Dependencies" and "Compile Sources":
+
+    ![where to drag the run script](images/drag_run_script.png)
+
+5. Double-click the name of the build phase to rename it to **Apollo CLI**:
+
+    ![UI for renaming](images/rename_run_script.png)
+
+6. Expand the Apollo CLI phase. Paste the **Swift Package Manager Run Script** from [Adding a code generation build step](/installation/#adding-a-code-generation-build-step) into the text area. This script uses your schema to generate the code that the Apollo iOS SDK uses to interact with your server.
```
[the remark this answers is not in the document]
StephenBarlow: Yeah, it's for the reasons you mention that it's probably safer for the time being to link to the parent header.

StephenBarlow:
@designatednerd Another feedback-addressing commit!

https://github.com/apollographql/apollo-ios/pull/911/commits/e909942a1cef4344fbe944da09d18a43f585f91a

docs/source/tutorial/tutorial-obtain-schema.md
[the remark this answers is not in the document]
designatednerd: Is that an "LOL different browsers hate your CSS" thing or what?

docs/source/tutorial/tutorial-introduction.md
designatednerd: OK, even though functionally `(tutorial-create-project)` is the same?

docs/source/tutorial/tutorial-obtain-schema.md
```diff
@@ -0,0 +1,64 @@
...
+title: "2. Obtain your GraphQL schema"
+---
+
+## Explore your server's schema
+
+This tutorial uses a modified version of the GraphQL server you build as part of [the Apollo full-stack tutorial](https://www.apollographql.com/docs/tutorial/introduction/). You can visit [`https://n1kqy.sse.codesandbox.io/`](https://n1kqy.sse.codesandbox.io/) to start it up and open the GraphiQL tool to explore its schema:
+
+![the GraphiQL query explorer](images/graphiql.png)
+
+The schema defines which GraphQL operations your server can execute. Click **Schema** on the right-hand side to view a list of the possible queries and mutations you can run:
```
designatednerd: I'm gonna update this on my end a bit in the follow-up PR - it's true but it's incomplete, since it defines not just the operations, but the types returned by those operations, and the types the fields on those returned types.

PR #918, opened 2019-11-23
Add List-related watch tests

gilbox:
Confirms that Apollo supports automatic updates to lists. Addresses reordering and deletion.

### DIFF
--- a/Tests/ApolloCacheDependentTests/WatchQueryTests.swift
+++ b/Tests/ApolloCacheDependentTests/WatchQueryTests.swift
@@ -191,6 +191,7 @@ class WatchQueryTests: XCTestCase {
 
       var verifyResult: GraphQLResultHandler<HeroAndFriendsNamesQuery.Data>
 
+      // verify initial state
       verifyResult = { result in
         switch result {
         case .success(let graphQLResult):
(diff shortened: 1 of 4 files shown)


Tests/ApolloCacheDependentTests/WatchQueryTests.swift
```diff
@@ -149,11 +149,101 @@ class WatchQueryTests: XCTestCase {
...
+      var expectation = self.expectation(description: "Fetching query")
+
+      _ = client.watch(query: query) { result in
+        verifyResult(result)
+        expectation.fulfill()
+      }
+
+      waitForExpectations(timeout: 5, handler: nil)
+
+      verifyResult = { result in
```
designatednerd: I think the ordering here is throwing me wayyyy off - so the first `verifyResult` is verifying what's already in the initial records, and this second one is verifying the result after the the request you set up with `networkTransport` actually hits the cache? 

I realize due to the construction of `withCache` this may not be possible to reorder, but some comments in here would probably make this a LOT easier to follow.

gilbox: Thanks for the feedback, I added a few comments.

PR #921, opened 2019-11-23
Expose internal members of MultipartFormData

levi:
Implementing a custom `NetworkTransport` that mimics similar behavior to the default `HTTPNetworkTransport` depends on these members being public. Targets linking to the library can now access the full functionality `MultipartFormData` objects.

### DIFF
--- a/Sources/Apollo/MultipartFormData.swift
+++ b/Sources/Apollo/MultipartFormData.swift
@@ -3,7 +3,7 @@ import Foundation
 /// A helper for building out multi-part form data for upload
 public class MultipartFormData {
   
-  enum FormDataError: Error, LocalizedError {
+  public enum FormDataError: Error, LocalizedError {
     case encodingStringToDataFailed(_ string: String)
     
     var errorDescription: String? {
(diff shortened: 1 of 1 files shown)


Sources/Apollo/MultipartFormData.swift
```diff
@@ -15,9 +15,9 @@ public class MultipartFormData {
   }
 
   /// A Carriage Return Line Feed character, which will be seen as a newline on both unix and Windows servers.
-  static let CRLF = "\r\n"
+  public static let CRLF = "\r\n"
```
designatednerd: Why does this need to be public?

Sources/Apollo/MultipartFormData.swift
```diff
@@ -3,10 +3,10 @@ import Foundation
 /// A helper for building out multi-part form data for upload
 public class MultipartFormData {
   
-  enum FormDataError: Error, LocalizedError {
+  public enum FormDataError: Error, LocalizedError {
```
designatednerd: If you're doing a custom implementation, would it make more sense to declare your own error type instead of making this one public?

designatednerd:
@levi Any thoughts on my comments here?

designatednerd:
@levi Do you have a second to take a look at this PR? Thus far you've been the only person requesting this functionality, and I'd like to better understand why this was necessary for you before deciding whether to try to add this on our end or to close this PR.

PR #924, opened 2019-11-26
Sb/tutorial 2 edits

StephenBarlow:


docs/source/tutorial/tutorial_2.mdx
```diff
@@ -239,35 +249,37 @@ override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
...
+Go back to `LaunchList.graphql`. Your query is already fetching most of the information you want to display, but it would be nice to display both the name of the mission and an image of the patch. 
 
-Looking at the schema in GraphiQL, you can see that `Launch` has a property of `mission`, which will allow you to get details of the mission. A mission has both a name and a `missionPatch` property, and the `missionPatch` can optionally take a parameter about what size something needs to be.
+Looking at the schema in GraphiQL, you can see that `Launch` has a property of `mission`, which allows you to get details of the mission. A mission has both a name and a `missionPatch` property, and the `missionPatch` can optionally take a parameter about what size something needs to be.
 
-Since loading a TableView with large images can be a real performance hog, ask for the name and a small mission patch. Update your query to look like the following:
+Since loading a `TableView` with large images can be a real performance hog, ask for the name and a small mission patch. Update your query to look like the following:
```
designatednerd: The `TableView` here either shouldn't be backticked or should be `UITableView` - it could also be `table view`

docs/source/tutorial/components/detail_ui_setup_panel.mdx
```diff
@@ -60,11 +60,6 @@ And update it to match the following:
 
 ![constant 0, priority 1000, multiplier 1:1](../images/aspect_ratio_multiplier.png)
 
-<<<<<<< HEAD
->**Note**: You may still see some sizing errors at this point, but those will go away in the next step.
```
designatednerd: We should keep this note, otherwise people may be like "WTF is going on here" when it has a million IB warnings.

docs/source/tutorial/tutorial_2.mdx
```diff
@@ -549,32 +549,38 @@ case .loading:
...
+- You could provide the identifier of an individual launch to a _different_ query to request all the details you want to display. 
 
-The first option *can* seem easier if there's not a substantial difference in size between what you're requesting for a list vs. a detail page. 
+The first option *can* seem easier if there isn't a substantial difference in size between what you're requesting for the list versus the detail page. 
 
-However, remember that part of the advantage of GraphQL over REST is that you only need to get the minimum amount of data needed to display a certain page. If you're not going to be displaying additional information, how much of your user's battery and bandwidth can you save, and how many unnecessary database calls on your backend can you prevent by not asking for something until you need it?
+However, remember that one of the advantages of GraphQL is that you can query for _exactly_ the data you need to display on a page. If you're not going to be displaying additional information, you can save bandwidth and execution time on both your frontend and backend by not asking for data until you need it.
```
designatednerd: Would definitely like to keep the battery thing in here - less bandwidth = less battery and that's critical for mobile

docs/source/tutorial/tutorial_2.mdx
```diff
@@ -588,17 +594,15 @@ You'll see a warning in GraphiQL:
 
 ![variable dollar sign id of type ID was used in position expecting type ID exclamation point](images/graphiql_nullable_conflict.png)
 
-This bit can be VERY confusing for Swift developers in particular, because GraphQL's assumptions about nullability are essentially the opposite of Swift's. 
+This bit can be confusing, because GraphQL's assumptions about nullability are different from Swift's:
```
designatednerd: I think i'd like to keep `for Swift Developers in particular` here, because it's not an inherently confusing thing

docs/source/tutorial/tutorial_2.mdx
```diff
@@ -588,17 +594,15 @@ You'll see a warning in GraphiQL:
 
 ![variable dollar sign id of type ID was used in position expecting type ID exclamation point](images/graphiql_nullable_conflict.png)
 
-This bit can be VERY confusing for Swift developers in particular, because GraphQL's assumptions about nullability are essentially the opposite of Swift's. 
+This bit can be confusing, because GraphQL's assumptions about nullability are different from Swift's: 
 
-In Swift, if you don't annotate a property's type with either a question mark or an exclamation point, that property is non-nullable. 
+* In Swift, if you annotate a property's type with either a question mark or an exclamation point, that property is nullable.
```
designatednerd: This is not correct. 

It's nullable if there's a question mark. It's implicitly unwrapped if it's got an exclamation point. In theory you can set either to null but calling it "nullable" will make Swift devs go wtf.

I'd like to go back to the original explanation on these two because I think for Swift developers this framing will make a lot more sense.

Maybe "That property can be set to null".

docs/source/tutorial/components/detail_ui_setup_panel.mdx
StephenBarlow: ok cool this was me messing up conflict resolution

PR #933, opened 2019-11-29
Update tests to use `XCTUnwrap` instead of guard let else fail

designatednerd:
Apple added a way to have your tests error out if something doesn't unwrap in Xcode 11 called [`XCTUnwrap`](https://developer.apple.com/documentation/xctest/3380195-xctunwrap) - thanks to @basthomas for reminding me this existed, and inspiring me to clean all this up.


BasThomas:
Yay!

Issue #729, opened 2019-08-21
Build error when including a type and set of types


MatrixSenpai:
Running into an issue with conflicting types.

I have a schema definition that follows this format:
```
type Match {
  id: ID
  name: String
  start: Date
  winner: Team
  teams: [Team]
}
type Team {
  id: ID
  name: String
}
```

When trying to define it in Apollo on the iOS client, I get the build error 'Cannot convert value of type 'Match.Winner.Type' to 'GraphQLFragment.Type'. My definitions are
```
# TeamFragment.graphql
fragment Team on Team {
  id
  name
}

# MatchFragment.graphql
fragment Match on Match {
  id
  name
  start
  winner {
    ... Team
  }
  teams {
    ... Team
  }
}
```

If I remove the definition for the `teams` field, `Team` gets created as a `GraphQL Fragment`. If I add it, `Team` is defined as a `GraphQLSelectionSet` and throws a type error.

Am I simply defining these wrong?

MatrixSenpai:
As an addendum to this, I created two new fragments, `WinningTeam` and `ParticipatingTeams` and assigned them to `winner` and `teams`, respectively. However, those fragments are exactly the same, and this is not necessarily expected behavior. If I try to assign either of those fragments to both fields, the build fails the same way

designatednerd:
I'll take a deeper look into this when I get back from moving, but I know we've had some issues with fragments that have the same name as the type name. 

In general I'd expect the fragment name to be something more like `TeamDetails` or `TeamIDAndName` rather than having the same exact name as the type itself.

MatrixSenpai:
Fair enough, but either way, it produces unexpected results

designatednerd:
Yeah agreed. I’ll see what I can do when I’m back.

designatednerd:
OK sorry I dropped the ball on this one. @MatrixSenpai I believe that you've basically found an undocumented edge case by naming the fragment the same thing as the type. I'm updating our documentation to tell people not to do this explicitly.

MatrixSenpai:
I've not had an issue with naming the fragments exactly the same as they are defined in the schema. The issue comes in when I try to use the same fragment in two different places as children of an object. For example, if I have the following schema
```
fragment Match on Match {
  // ...
  winner {
    ... Team
  }
  participants {
    ... Team
  }
}
```
where `winner` is of type `Team` and `participants` is of type `Array<Team>`, that's where I run into an issue

designatednerd:
[...]

PR #942, opened 2019-12-11
Document fragment naming issue

designatednerd:
This PR (sorta) addresses #729 - the user on that issue was confused as to why they couldn't name a fragment the same thing as the type of the fragment, because there was no documentation explicitly telling them not to do this or why. Now there is!


designatednerd:
After further discussion in #729 I misunderstood this issue. Closing.

Issue #917, opened 2019-11-22
Why are `clientName` and `clientVersion` on `NetworkTransport` `set` properties? 


bwhiteley:
Wouldn't `{ get }` be sufficient for these protocol properties? 
The `{ set }` complicates some implementations because I want them to be computed properties. I'm currently doing ` set { fatalError() }`, which works, but...

bwhiteley:
Also, couldn't you then provide default implementations?

designatednerd:
Hey @bwhiteley - we've already had [one extensive discussion around this in #877](https://github.com/apollographql/apollo-ios/issues/877). Take a look through that discussion, and me know if you have further questions after reading that. 

In terms of providing a default value, there is a default provided through the `defaultClientName` and `defaultClientVersion` computed properties defined in the `NetworkTransport` protocol, but you can't provide a default getter and setter for a computed property in a protocol. Basically this is a limitation of protocol-oriented programming when it comes to settable properties.

designatednerd:
@bwhiteley Does this answer your question?

bwhiteley:
I read the discussion in #877 but it still seems like the wrong decision. Apollo itself doesn't need the `clientName` and `clientVersion` to be mutable, and IMO that's what should drive the protocol requirements. If they are defined as `get`-only on the protocol, that doesn't prevent an implementation from changing them at runtime. 

Prior to this change it was easy to conform an existing REST client to `NetworkTransport` in an extension (which is what we do). The REST client in question is an API gateway client that handles things like authentication and token refresh. It comes from different framework from where we have Apollo code, so it is nice to add `NetworkTransport` conformance in an extension next to our Apollo code. Now we can't provide `NetworkTransport` conformance in an extension without doing something weird like having setters that do nothing or call `fatalError()` (since we have no need to modify the properties at runtime). 

Anyway, the workaround is easy enough so it's not the end of the world, but now you know our use case.

designatednerd:
It's definitely helpful to understand the use case - the use case that I've been looking at has been focused on the people who do need to change this on a regular basis and use our batteries-included classes, so it's helpful to hear where the hard edges come together who don't. 

Unsurprisingly given my decision to do it the way we're doing it, I disagree with this assessment (which is fairly foundational to your argument):

> Apollo itself doesn't need the `clientName` and `clientVersion` to be mutable

[...]

Issue #877, opened 2019-11-01
NetworkTransport's new `client*` property requirements require setters?


cysp:
The `clientName` and `clientVersion` requirements that have been added to the `NetworkTransport` protocol are marked as requiring both getters and setters, is that intentional? 
The upshot of requiring setters is that conforming implementations must expose those fields as externally mutable properties of at least as high visibility as the type itself, which they probably don't want to do.

I've modified that protocol locally to only require getters and everything compiles so if there's no reason to require the setter, I'm happy to raise a PR reducing the requirement to just the getter. 🙂

```diff
diff --git i/Sources/Apollo/NetworkTransport.swift w/Sources/Apollo/NetworkTransport.swift
index 44adee10..09f2806d 100644
--- i/Sources/Apollo/NetworkTransport.swift
+++ w/Sources/Apollo/NetworkTransport.swift
@@ -12,10 +12,10 @@ public protocol NetworkTransport: class {
   func send<Operation>(operation: Operation, completionHandler: @escaping (_ result: Result<GraphQLResponse<Operation>, Error>) -> Void) -> Cancellable
   
   /// The name of the client to send as the `"apollographql-client-name"` header.
-  var clientName: String { get set }
+  var clientName: String { get }
   
   /// The version of the client to send as the `"apollographql-client-version"` header
-  var clientVersion: String { get set }
+  var clientVersion: String { get }
 }
 
 public extension NetworkTransport {
```

designatednerd:
Hey Scott, this was an explicit architectural decision to support being able to change the property after initialization. The anticipated use case is something like this:

I'm running an A/B test, and I want to make sure I can identify which of the A/B buckets the current user is in when I call my GQL API, but I can't get the information of _which_ bucket this user should be in without first calling my GQL API. 

Without the ability to set after instantiation, I'd have to completely tear down and rebuild my `ApolloClient` almost immediately after setting it up, or I'd have to use separate client instances to get the A/B bucket information and then do the rest of my calls.

This is certainly not a default setup, but particularly in larger companies, this is a frequent enough situation that I think it's worth explicitly exposing this with a setter.  

Does that make sense? Or do you have alternate thoughts on downsides of exposing the setter?

cysp:
[...]

PR #954, opened 2019-12-16
Make `clientName` and `clientVersion` get-only protocol requirements

designatednerd:
**BREAKING CHANGE** 

This has been much-discussed in #877 and #917, and it's clear that for people implementing their own `NetworkTransport` method rather than using one of ours, having to provide a setter for this is really annoying. If implementers still want to make the property get/set, they can (and we do with `HTTPNetworkTransport`). 

Also added a convenience method to add the headers to any `URLRequest`, and added a note to the `send` method that these should probably be sent along if you want to use monitoring tools like Apollo Graph Manager.

### DIFF
--- a/Sources/Apollo/NetworkTransport.swift
+++ b/Sources/Apollo/NetworkTransport.swift
@@ -23,12 +23,12 @@ public protocol NetworkTransport: class {
 public extension NetworkTransport {
   
   /// The field name for the Apollo Client Name header
-  var headerFieldNameApolloClientName: String {
+  static var headerFieldNameApolloClientName: String {
     return "apollographql-client-name"
   }
   
   /// The field name for the Apollo Client Version header
-  var headerFieldNameApolloClientVersion: String {
+  static var headerFieldNameApolloClientVersion: String {
     return "apollographql-client-version"
   }
   
@@ -74,8 +74,8 @@ public extension NetworkTransport {
   /// Adds the Apollo client headers for this instance of `NetworkTransport` to the given request
   /// - Parameter request: A mutable URLRequest to add the headers to.
   func addApolloClientHeaders(to request: inout URLRequest) {
-    request.setValue(self.clientName, forHTTPHeaderField: self.headerFieldNameApolloClientName)
-    request.setValue(self.clientVersion, forHTTPHeaderField: self.headerFieldNameApolloClientVersion)
+    request.setValue(self.clientName, forHTTPHeaderField: Self.headerFieldNameApolloClientName)
+    request.setValue(self.clientVersion, forHTTPHeaderField: Self.headerFieldNameApolloClientVersion)
   }
 }
 
--- a/Sources/Apollo/NetworkTransport.swift
+++ b/Sources/Apollo/NetworkTransport.swift
@@ -22,13 +22,13 @@ public protocol NetworkTransport: class {
 
 public extension NetworkTransport {
   
-  /// The header field name for the Client Name
-  var headerFieldNameClientName: String {
+  /// The field name for the Apollo Client Name header
+  var headerFieldNameApolloClientName: String {
     return "apollographql-client-name"
   }
   
-  /// The header field name for the client version
-  var headerFieldNameClientVersion: String {
+  /// The field name for the Apollo Client Version header
+  var headerFieldNameApolloClientVersion: String {
     return "apollographql-client-version"
   }
   
@@ -63,11 +63,11 @@ public extension NetworkTransport {
     return version
   }
   
-  /// Adds the client headers for this instance of `NetworkTransport` to the given request
+  /// Adds the Apollo client headers for this instance of `NetworkTransport` to the given request
   /// - Parameter request: A mutable URLRequest to add the headers to.
-  func addClientHeaders(to request: inout URLRequest) {
-    request.setValue(self.clientName, forHTTPHeaderField: self.headerFieldNameClientName)
-    request.setValue(self.clientVersion, forHTTPHeaderField: self.headerFieldNameClientVersion)
+  func addApolloClientHeaders(to request: inout URLRequest) {
+    request.setValue(self.clientName, forHTTPHeaderField: self.headerFieldNameApolloClientName)
+    request.setValue(self.clientVersion, forHTTPHeaderField: self.headerFieldNameApolloClientVersion)
   }
 }
 
--- a/Sources/Apollo/NetworkTransport.swift
+++ b/Sources/Apollo/NetworkTransport.swift
@@ -12,10 +12,10 @@ public protocol NetworkTransport: class {
   func send<Operation>(operation: Operation, completionHandler: @escaping (_ result: Result<GraphQLResponse<Operation>, Error>) -> Void) -> Cancellable
   
   /// The name of the client to send as the `"apollographql-client-name"` header.
-  var clientName: String { get set }
+  var clientName: String { get }
   
   /// The version of the client to send as the `"apollographql-client-version"` header
-  var clientVersion: String { get set }
+  var clientVersion: String { get }
 }
 
 public extension NetworkTransport {
--- a/Sources/Apollo/NetworkTransport.swift
+++ b/Sources/Apollo/NetworkTransport.swift
@@ -5,7 +5,7 @@ public protocol NetworkTransport: class {
   
   /// Send a GraphQL operation to a server and return a response.
   ///
-  /// Note: The `clientName` and `clientVersion` should be sent with any URL request which needs headers so your client can be identified by tools like Apollo Graph Manager. The `addClientHeaders` method is provided below to do this for you automatically, and this is handled for you by batteries-included versions of `NetworkTransport`. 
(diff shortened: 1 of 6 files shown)


designatednerd:
@bwhiteley @cysp

gsabran:
Thanks @designatednerd for putting this PR out so quickly.

IMO those requirements should be totally removed from the protocol. I think that folks using custom implementations are doing so because they want to be in control, and are accepting the corresponding responsibilities. I see protocols as describing the requirements that Apollo's SDK needs to work properly, and not as guiding the external implementation.

It gives me the inaccurate impression that the Apollo needs to read those values to work properly, while the documentation of `send<Operation>` is actually what matters most in this case.

Sources/Apollo/NetworkTransport.swift
```diff
@@ -5,28 +5,30 @@ public protocol NetworkTransport: class {
   
   /// Send a GraphQL operation to a server and return a response.
   ///
+  /// Note: The `clientName` and `clientVersion` should be sent with any URL request which needs headers so your client can be identified by tools like Apollo Graph Manager. The `addClientHeaders` method is provided below to do this for you automatically, and this is handled for you by batteries-included versions of `NetworkTransport`.
```
gsabran:
```suggestion
  /// Note: The sender should properly set the headers `apollographql-client-name` and `apollographql-client-version` when sending the related request. This is so that Graph Manager can properly attribute traffic.`. 
```

designatednerd:
> It gives me the inaccurate impression that the Apollo needs to read those values to work properly

The iOS SDK doesn't, but Apollo Graph Manager does (as would basically anything else that's trying to look at what version/client are sending a particular request). 

It's certainly a best practice to include these in some way, shape, or form so that anything looking at where requests are coming from is able to quickly tell what's coming from what version of what client. I would like to steer people *real* hard towards following this best practice, and making them at least implement these properties, even if they want to send them with other headers.

Sources/Apollo/NetworkTransport.swift
designatednerd: I'm going to decline this suggestion - the header names are not what specifically need to be set by the client, since those are included in the default implementation. It's the actual values which need to be set. I'll see if I can make this clearer though.

designatednerd:
@gsabran Better?

gsabran:
👍Ok thanks!

designatednerd:
@bwhiteley @cysp any further thoughts? I'll leave this open until tomorrow morning for more comments.

cysp:
As an implementor of a custom `NetworkTransport`, my mind didn't make the connection from "the `NetworkTransport` protocol requires me to implement some client identification properties, I assumed that the properties were going to be read (and set) by somewhere else in the iOS SDK. The comment added to `send()` here indicating that it's actually _the network transport itself_ which is supposed to use them seems like it might be helpful in that regard. 🙂

We don't use the graph manager (though we're interested in its capabilities and will be thinking more about it "soon"), but perhaps a dashboard on there with a list of user-agents from which it has seen queries without client identification would be a good way to remind implementors to supply the client identification headers?

Now that these properties are `get`-only requirements, would it be worth giving them default implementations that forward through to the `defaultClient(Name|Version)` static properties?

<details>
<summary>networktransport-default-client-vars-light.diff</summary>

```diff
diff --git c/Sources/Apollo/NetworkTransport.swift i/Sources/Apollo/NetworkTransport.swift
index 63a90fe3..766944fc 100644
--- c/Sources/Apollo/NetworkTransport.swift
+++ i/Sources/Apollo/NetworkTransport.swift
@@ -41,6 +41,10 @@ public extension NetworkTransport {
     return "\(identifier)-apollo-ios"
   }
   
+  var clientName: String {
+    return Self.defaultClientName
+  }
+
   /// The default client version to use when setting up the `clientVersion` property.
   static var defaultClientVersion: String {
     var version = String()
@@ -62,6 +66,10 @@ public extension NetworkTransport {
     
     return version
   }
+
+  var clientVersion: String {
+    return Self.defaultClientVersion
+  }
   
   /// Adds the Apollo client headers for this instance of `NetworkTransport` to the given request
   /// - Parameter request: A mutable URLRequest to add the headers to.
```
</details>

designatednerd:
>Now that these properties are get-only requirements, would it be worth giving them default implementations that forward through to the defaultClient(Name|Version) static properties? 

I can't believe I didn't think of this (I think because I was still thinking about the "override to get/set" situation), this is an excellent idea.

Sources/Apollo/NetworkTransport.swift
```diff
@@ -5,28 +5,30 @@ public protocol NetworkTransport: class {
...
+  /// The version of the client to send as a header value.
+  var clientVersion: String { get }
 }
 
 public extension NetworkTransport {
   
-  /// The header field name for the Client Name
-  static var headerFieldNameClientName: String {
+  /// The field name for the Apollo Client Name header
+  var headerFieldNameApolloClientName: String {
```
cysp: I have two questions about these header name properties:
- what was the reason that these changed to be instance variables, rather than static variables? (since `addApolloClientHeaders()` is in a generic context, I think it has access to `Self` to access the static properties, it certainly does with newer versions of Swift)
- at the moment these fields are not protocol requirements, so despite the fact that `addApolloClientHeaders()` accesses them through `self`, because that's happening in a generic context the `NetworkTransport` default implementation is used and any specialisation in an implementation is ignored, is that intentional? if that is intentional (and it would make sense to me that it is, because why would implementors want to customise the header field _names_?) it would be less surprising (at least to me) when reading the code if those were just static variables somewhere, not accessed through the protocol. 🙃

designatednerd: So there's a couple thoughts happening here: 

- It seemed goofy to me that at the call site you had to specify the class name of the current class, although that's less bad with being able to call `Self` within the protocol (I thought the `Self` outside of the protocol thing hadn't landed yet?), and there was no real reason for them to be static rather than instance other than it seemed good for namespacing. However, now that by default they're mostly accessed within the protocol, I'll rethink this change.
- This is certainly in the realm of "my opinion" vs "the right way to do things", but I *hate* free-floating static variables. It's almost always primarily related to something, so I think it's way better to have static variables on whatever it is they're related to rather than floating in space.

Sources/ApolloWebSocket/SplitNetworkTransport.swift
```diff
@@ -8,34 +8,22 @@ public class SplitNetworkTransport {
...
-    }
-    set {
-      self.httpNetworkTransport.clientVersion = newValue
-      self.webSocketNetworkTransport.clientVersion = newValue
+    let httpVersion = self.httpNetworkTransport.clientVersion
+    let websocketVersion = self.webSocketNetworkTransport.clientVersion
+    if httpVersion == websocketVersion {
+      return httpVersion
+    } else {
+      return "SPLIT_HTTPVERSION_\(httpVersion)_WEBSOCKETNAME_\(websocketVersion)"
```
cysp: Hah, this is existing but this should probably be `WEBSOCKETVERSION`?

```suggestion
      return "SPLIT_HTTPVERSION_\(httpVersion)_WEBSOCKETVERSION_\(websocketVersion)"
```

Sources/Apollo/NetworkTransport.swift
cysp: Cool, namespacing is good.
From reading the code I was unsure whether it was intentional that implementations of `NetworkTransport` were _not_ capable of specialising those properties because I saw them being accessed statically on implementation types:
https://github.com/apollographql/apollo-ios/search?q=headerFieldNameClientName&unscoped_q=headerFieldNameClientName

🙃

bwhiteley:
> > Now that these properties are get-only requirements, would it be worth giving them default implementations that forward through to the defaultClient(Name|Version) static properties?
> 
> I can't believe I didn't think of this (I think because I was still thinking about the "override to get/set" situation), this is an excellent idea.

https://github.com/apollographql/apollo-ios/issues/917#issuecomment-557708746
;)

bwhiteley:
I too somehow missed that it's up to an implementation of `NetworkTransport` to actually include these headers in the request.

bwhiteley:
> It's certainly a best practice to include these in some way, shape, or form so that anything looking at where requests are coming from is able to quickly tell what's coming from what version of what client. I would like to steer people _real_ hard towards following this best practice, and making them at least implement these properties, even if they want to send them with other headers.

Are you saying it doesn't have to be these particular headers? Is this what `User-Agent` is typically used for?

bwhiteley:
Ultimately I agree with @gsabran now that I understand better what's going on here. This is basically some documentation and a convenience function. 

However, by changing the properties to get-only and providing default implementations, the properties no longer impose an undue burden on custom implementations of `NetworkTransport`. As a result, these are solid improvements.

Sources/ApolloWebSocket/SplitNetworkTransport.swift
designatednerd: DERRRP good catch

designatednerd:
It doesn't have to be these particular headers if you're not using Apollo Graph Manager. It does if you're using Apollo graph manager. Given that the library is owned and maintained by Apollo, we're certainly going to do everything we can to make it easy for people to use Graph Manager. 😇

`User-Agent` could theoretically be used for this, but separating client and version out in the same string is annoying, and having them be separate is *really* helpful for diagnostics. 

Having this information passed as its own custom header also allows any existing use of `User-Agent` to continue without interference.

PR #963, opened 2020-01-03
Fixing some weirdness with circle

designatednerd:
Suddenly some WebSocket tests that build fine on a local 11.3 weren't building at all on an 11.1 image. Not sure what's going on there but we should update to the 11.3 image anyway.


designatednerd:
*shakes fist in the general direction of git submodules*

PR #972, opened 2020-01-13
Edit auth on web socket for style and language

designatednerd:
This is an edit for language and style to match the rest of the docs on the stuff added in #966.


designatednerd:
@rodrigoelp Mind giving this a once-over to make sure I didn't butcher your work too badly here?

rodrigoelp:
@designatednerd I don't think you butchered anything. If anything I wrote it in a very colloquial way and this look good.

designatednerd:
No worries at all, I'm on the road so my responses are gonna be slow no matter what.

PR #986, opened 2020-01-31
Update Xcode version.

abdullahselek:
Thanks for creating this library. This PR aims to use latest version of Xcode and simulators on CircleCi.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ commands:
 jobs:
   macOS10_15:
     macos:
-      xcode: "11.3.0"
+      xcode: "11.3.1"
     environment:
       DESTINATION: platform=macOS,arch=x86_64
       CIRCLE_XCODE_SCHEME: Apollo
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,107 +35,107 @@ commands:
 jobs:
   macOS10_15:
     macos:
-      xcode: "11.3.0"
+      xcode: "11.3.1"
     environment:
       DESTINATION: platform=macOS,arch=x86_64
       CIRCLE_XCODE_SCHEME: Apollo
-      CIRCLE_XCODE_SDK: macosx10.15
+      CIRCLE_XCODE_SDK: macosx10.15.1
```
designatednerd: The SDKs shouldn't be changed here - only the version of Xcode has been bumped, not the SDK itself. This is why the builds are all failing. 

Here's a screenshot from my machine of the SDKs included with 11.3.1: 
Screen Shot 2020-01-31 at 9 47 20 AM

designatednerd:
Thanks for the PR! This stuff is always a bit of a pain. :P

PR #991, opened 2020-02-03
Fix SQLite Cache Key Bug

giantramen:
If your query has "." in it this would
create an undefined cache key because it would
just drop the query after the ".". For all "_ROOT" queries
we just want the first part anyways so just return that.

For example here is our query: 
QUERY_ROOT.searchPois([criteria:[area:[northeastCorner:latitude:31.882253225338154,longitude:-75.24958693102437],[southwestCorner:latitude:18.87160867771246,longitude:-84.49958923798633]],types:[\"SITE\"]])
And we normally want "QUERY_ROOT" as the key but with the old logic we were getting 
"QUERY_ROOT.searchPois([criteria:[area:[northeastCorner:latitude:31.882253225338154,longitude:-75.24958693102437],[southwestCorner:latitude:18.87160867771246,longitude:-84"

If there are more elegant solutions I'm all ears, we were running into issues with this so fixed it the easier way I could find :)

### DIFF
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -33,19 +33,16 @@ public final class SQLiteNormalizedCache {
   }
 
   private func recordCacheKey(forFieldCacheKey fieldCacheKey: CacheKey) -> CacheKey {
-    let components = fieldCacheKey.components(separatedBy: ".")
-    var updatedComponents = [String]()
-    for component in components {
-      if updatedComponents.last?.last?.isNumber ?? false && component.first?.isNumber ?? false {
-        updatedComponents[updatedComponents.count - 1].append(".\(component)")
-      } else {
-        updatedComponents.append(component)
-      }
+    var components = fieldCacheKey.components(separatedBy: ".")
+    if components.count > 1, let lastComponent = components.last,
+      lastComponent.first?.isNumber ?? false && components[components.count - 2].last?.isNumber ?? false {
+      components[components.count - 2].append(".\(lastComponent)")
+      components.removeLast()
     }
-    if updatedComponents.count > 1 {
-      updatedComponents.removeLast()
+    if components.count > 1 {
+      components.removeLast()
     }
-    return updatedComponents.joined(separator: ".")
+    return components.joined(separator: ".")
   }
 
   private func createTableIfNeeded() throws {
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -33,14 +33,19 @@ public final class SQLiteNormalizedCache {
   }
 
   private func recordCacheKey(forFieldCacheKey fieldCacheKey: CacheKey) -> CacheKey {
-    var components = fieldCacheKey.components(separatedBy: ".")
-    if fieldCacheKey.contains("_ROOT"), let rootKey = components.first {
-        return rootKey
+    let components = fieldCacheKey.components(separatedBy: ".")
+    var updatedComponents = [String]()
+    for component in components {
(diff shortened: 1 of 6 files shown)


designatednerd:
Looks like we've got some test failures - do you mind also adding an explicit test around this change?

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -33,11 +33,19 @@ public final class SQLiteNormalizedCache {
...
-      components.removeLast()
+    let components = fieldCacheKey.components(separatedBy: ".")
+    var updatedComponents = [String]()
+    for component in components {
+      if updatedComponents.last?.last?.isNumber ?? false && component.first?.isNumber ?? false {
+        updatedComponents[updatedComponents.count - 1].append(".\(component)")
+      } else {
+        updatedComponents.append(component)
+      }
+    }
```
designatednerd: Does this need to loop through every single component, or does it only need to look at the last couple of components? I'm concerned if we loop through every single component every single time, that's going to wind up chewing up a bunch of energy we may not need to.

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -319,7 +319,44 @@ class LoadQueryFromStoreTests: XCTestCase {
...
+        case .success(let graphQLResult):
+          XCTAssertNil(graphQLResult.errors)
+
+          guard let data = graphQLResult.data else {
+            XCTFail("No data returned with result")
+            return
+          }
+
+          XCTAssertEqual(data.starshipLength?.name, "Millennium Falcon")
+          XCTAssertEqual(data.starshipLength?.length, starshipLength)
```
designatednerd: should the coordinates also be the same?

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -319,7 +319,44 @@ class LoadQueryFromStoreTests: XCTestCase {
...
+
+      load(query: query) { result in
+        switch result {
+        case .success(let graphQLResult):
+          XCTAssertNil(graphQLResult.errors)
+
+          guard let data = graphQLResult.data else {
+            XCTFail("No data returned with result")
+            return
+          }
```
designatednerd: You can use `let data = try XCTUnwrap(graphQLResult.data)` here and not have to deal with writing your own failure mechanism

giantramen: Maybe I'm missing something but I don't think I can use that in the completion block.

designatednerd: I haven't had an issue with that in other tests

giantramen: It's not used in completion blocks anywhere else, it seems like this test file uses the completion block with XCTFail pattern for all the tests.

designatednerd:
Test failures seem to be tied to #993 - going to put this into a holding pattern for a minute until we can figure out what's going on there.

giantramen:
I am going to take another look at this as well, the changes that were made during review have caused the issue to reappear.

giantramen:
> I am going to take another look at this as well, the changes that were made during review have caused the issue to reappear.

Okay this test case actually matches the original issue so should be good now 👍

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
giantramen: I revisited this, need to loop through but only if we see it's a "_ROOT" so it shouldn't happen very often.

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
designatednerd: no worries i'll fight with this once this is merged

Issue #993, opened 2020-02-03
Flaky cache tests


designatednerd:
We're seeing some really flaky cache tests, particularly since #552 was merged - it may be related to the changes in that or it may not.

designatednerd:
@RolandasRazma @AnthonyMDev Would either of you have a second to take a look at this? Obviously flakiness is obnoxious to diagnose but we're getting intermittent and weird failures on a bunch of tests on CI.

RolandasRazma:
Yes, it probably related, don't like that tight loop `0...10000` at all. 
We definitely can reduce that loop to only few interactions as in new code base there is no locking at all

designatednerd:
It's not failing consistently on one particular test though - that's the bit that makes this a bit obnoxious to try to figure out

RolandasRazma:
If tests running in parallel maybe we slow down is enough to knock them off

designatednerd:
I don't think that's a great solution - we do want the tests to reflect potentially very high loads, and if we just slow them down in order to make them pass, we're absolutely going to miss something.

designatednerd:
[This is a good example of some of the failures we're seeing](https://circleci.com/gh/apollographql/apollo-ios/4074): There's no change except to our documentation, but it's randomly having a bunch of tests time out.

AnthonyMDev:
If that test is failing intermittently, it almost certainly means that there are bugs in the locking. That’s what it was designed to test. It has to loop a large number of times to trigger the edge case of the dead lock. If you reduce the loop to 10, the test becomes useless. If you reduce it to 500, it will maybe sometimes deadlock, just not quite as often. Because it’s a race condition we are testing for, the test will not ALWAYS fail if the bug exists. It’s near impossible to make that happen. But that test shouldn’t be failing randomly. It most likely indicates something is wrong.

designatednerd:
Does seem like we're getting some repeated crashes. Going to keep updating links as these repeat: 

`testThreadedCache`: 
[1](https://circleci.com/gh/apollographql/apollo-ios/4167) 
[2](https://circleci.com/gh/apollographql/apollo-ios/4166)

`testLoadingHeroAndFriendsNamesQueryWithIDs`:
[1](https://circleci.com/gh/apollographql/apollo-ios/4158)
[2](https://circleci.com/gh/apollographql/apollo-ios/4174)

designatednerd:
[...]

PR #999, opened 2020-02-05
Update syntax highlighting guide

kristiyandobrev:


designatednerd:
@kristiyandobrev - I appreciate the update but I think this duplicates the work in the `xcode-graphql` repo. Your update actually underlines why that's not a good idea: the instructions might change, and then we'd have to remember to change them in two places. 

Do you mind updating to just point at the README for that repo for the instructions?

kristiyandobrev:
@designatednerd sure !

designatednerd:
Failures are tied to #993, going to merge since this is wholly unrelated.

PR #1002, opened 2020-02-05
Remove delegate from `HTTPNetworkTransport` initializer

designatednerd:
Suggested by @RolandasRazma in #990, let's make this a fully breaking change to the delegate so people are more clearly ware of the change. 

Also I *think* I might have figured something out with one of the tests that keeps failing as part of #993, so I threw that fix in here.

### DIFF
--- a/Tests/ApolloCacheDependentTests/FetchQueryTests.swift
+++ b/Tests/ApolloCacheDependentTests/FetchQueryTests.swift
@@ -473,5 +473,8 @@ class FetchQueryTests: XCTestCase {
     }
     self.wait(for: [expectation], timeout: 10)
     
+    for watcher in watchers {
+      watcher.cancel()
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


Tests/ApolloCacheDependentTests/FetchQueryTests.swift
```diff
@@ -473,5 +473,8 @@ class FetchQueryTests: XCTestCase {
     }
     self.wait(for: [expectation], timeout: 10)
     
+    for watcher in watchers {
+      watcher.cancel()
+    }
```
designatednerd: I have a sneaking suspicion that the watchers are getting overcalled because they're not necessarily getting killed off by ARC in time to avoid calls from other tests.

designatednerd: [Nope - still failing occasionally](https://circleci.com/gh/apollographql/apollo-ios/4380)

RolandasRazma: not sure if its the case here, but we had to add to wrap to `autoreleasepool { ... }` to workaround ARC

designatednerd: what'd you need to add that to?

PR #1003, opened 2020-02-06
Use screenshot class and split query UI article into multiple sections

StephenBarlow:
Before jumping into #932 , just wanted to take advantage of our new `screenshot` class for applicable images and get our existing tutorial sections into roughly equal-sized articles.


StephenBarlow:
This PR edits actual content almost zero, just split existing content up.

PR #1004, opened 2020-02-06
Improvements for Websockets

fassko:
* allow to opt-out for reconnection
* allow to disable duplicate messages when reconnecting


Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -74,12 +75,16 @@ public class WebSocketTransport {
...
               clientVersion: String = WebSocketTransport.defaultClientVersion,
               sendOperationIdentifiers: Bool = false,
+              reconnect: Bool = true,
               reconnectionInterval: TimeInterval = 0.5,
+              allowSendingDuplicates: Bool = true,
               connectingPayload: GraphQLMap? = [:],
               requestCreator: RequestCreator = ApolloRequestCreator()) {
     self.connectingPayload = connectingPayload
     self.sendOperationIdentifiers = sendOperationIdentifiers
+    self.reconnect.value = reconnect
```
designatednerd: Since `reconnect` is initialized with `false`, it has to have been getting flipped somewhere if it's reconnecting. I don't see anywhere that's being changed here - is that on purpose?

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -291,13 +296,14 @@ extension WebSocketTransport: WebSocketDelegate {
...
     self.error.value = nil
-    initServer()
+    initServer(reconnect: reconnect.value)
     if reconnected {
       self.delegate?.webSocketTransportDidReconnect(self)
       // re-send the subscriptions whenever we are re-connected
       // for the first connect, any subscriptions are already in queue
       for (_,msg) in self.subscriptions {
-        write(msg)
+        let id = allowSendingDuplicates ? nil : queue.first { $0.value == msg }?.key
```
designatednerd: why send a `nil` id if duplicates are disallowed instead of just not sending at all?

fassko: If Websocket reconnects then there is no entry in the queue so we need to call `write(msg, id: id)` which actually adds to queue and when `acked` then sends to the GraphQL Websockets. There is function `writeQueue()` that does it.

If it hasn't `acked` then queue isn't empty and we can check for duplicates. We need to call `write` in this case as well because while looping trough server can send `acked` back.

Does it make sense?

Sources/ApolloWebSocket/WebSocketTransport.swift
fassko: I think I didn't get the question. `reconnect` has default value to `true` in initilizer - WebSocketTransport.swift:78 

or I'm doing something wrong here?

designatednerd: I was looking at WebSocketTransport.swift:28 where the `Atomic` var is initialized with a value of `false`. 

Because `reconnect.value` wasn't being flipped to `true` in the initializer previously, there's got to be someplace it's getting flipped elsewhere, which could make it ignore a `false` passed in with the initializer now that this has been explicitly added.

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: Ah I think I flipped the ternary in my mind - I've always been shit with ternaries. 

So in this case if we *are* allowing the sending of duplicates, then we send a `nil` ID because it doesn't matter. If we're *not* allowing the sending of duplicates then we check through everything for a matching message to find its ID? 

It still strikes me as a bit weird to check the messages instead of the IDs, but that makes moderately more sense.

fassko: Yes exactly what you wrote.

Ideally maybe we should link subscription IDs and IDs in the queue, but currently it  is not the case. That's why I figured out searching for message. As this could be opt-in approach I think for now it's fine.

What do you think? Are there any other folks who can help to review? Maybe I'm missing something.

I agree about ternary operator also not my cup of tea :P

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -291,13 +296,14 @@ extension WebSocketTransport: WebSocketDelegate {
 
   public func websocketDidConnect(socket: WebSocketClient) {
     self.error.value = nil
-    initServer()
+    initServer(reconnect: reconnect.value)
```
designatednerd: OK, I figured out where `reconnect.value` was being set before: In `initServer` at line ~193, where `reconnect` was defaulted to `true`. I think maybe if we can take out the loop of setting the reconnect value, that could help.

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: might be clearer if you got rid of the ternary and just explicitly sent `nil`

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: [figured it out](https://github.com/apollographql/apollo-ios/pull/1004#discussion_r377881133) - further stuff can go through that comment

Sources/ApolloWebSocket/WebSocketTransport.swift
fassko: Changed to simpler approach

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -297,7 +303,14 @@ extension WebSocketTransport: WebSocketDelegate {
...
       // for the first connect, any subscriptions are already in queue
       for (_,msg) in self.subscriptions {
-        write(msg)
+        if allowSendingDuplicates {
+          write(msg)
+        } else {
+          // search duplicate message from the queue
+          // do not send multiple times
+          let id = queue.first { $0.value == msg }?.key
+          write(msg, id: id)
```
designatednerd: I'm confused - this looks like it'll just `write` the same `message` with the same `id` rather than bailing if a duplicate is found, but the comment says "do not send multiple times". How is this preventing duplicates from being sent?

fassko: I agree. I removed misleading comment.

The actual place where it checks for previous message is from line 225.

Basically how it should work if we not allowing to send duplicates is:
1. Try to find previous message (currently just by content because we don't match subscription IDs in the queue)
	1. a If found then use the ID from the queue
	1. b If not found then send nil for `ID` to `write`
3. In `write` function
	1. If `ID` is not `nil` we update previous message
	2. If `ID` is null then add new message to the queue, this did happen before
	3. If queue is empty then add first entry

Ideally we need to somehow link together `subscriptions` dictionary and internal `queue` of messages. But I would like to do it gradually. What do you think?

designatednerd: OK so right now the `write` queue is doing all the work. Got it. Thanks for changing the comment to make that clearer.

PR #1009, opened 2020-02-11
Edits to Mutations section of iOS tutorial

StephenBarlow:


docs/source/tutorial/tutorial-mutations.md
```diff
@@ -37,7 +37,7 @@ func networkTransport(_ networkTransport: HTTPNetworkTransport,
 }
 ```
 
-The `willSend` request is the last thing which can manipulate the request before it goes out to the network. Since the request is passed as an `inout` variable, you can manipulate its contents directly. 
+The `willSend` request is the last thing that can manipulate the request before it goes out to the network. Because the request is passed as an `inout` variable, you can manipulate its contents directly.
```
designatednerd: some day i'm gonna make you send me a doc on how to use which vs that because I always pick the wrong one 😆

Issue #1011, opened 2020-02-12
SQLite tests having intermittent failures in the threading test


designatednerd:
Background: While investigating #993 I determined that the intermittent failures of `testThreadedCache` appear to only be happening in the SQLite tests when the `DispatchGroup` gets `leave` called more times than it should. I added counters to the three closures and was seeing wayyyy more than the expected 1001 `leave` calls, which would cause a crash since `enter` and `leave` calls to `DispatchGroup` need to be balanced. 

In the discussion for #1010, there was a general agreement that this test uncovered existing incorrect behavior rather than having the incorrect behavior introduced in the corresponding PR, and this test was updated to not run on SQLite tests in order to keep further forward movement unblocked. 

I have *no* idea why a) This is happening b) It only seems to be happening on the SQLite tests, even though the client in the `testThreadedCache` is hard-coded to be an in-memory cache. The failures are intermittent, but they do still happen even when the test is run in isolation, so it doesn't seem like it's leakage from another test. 

Any help we can get on this would be really great.

cysp:
I enabled TSan and ran the test suite and it pointed to data races on properties of `GraphQLQueryWatcher`:

<details>
<summary>Enabling TSan:</summary>

```diff
diff --git i/Apollo.xcodeproj/xcshareddata/xcschemes/Apollo.xcscheme w/Apollo.xcodeproj/xcshareddata/xcschemes/Apollo.xcscheme
index 1e839927..a375da33 100644
--- i/Apollo.xcodeproj/xcshareddata/xcschemes/Apollo.xcscheme
+++ w/Apollo.xcodeproj/xcshareddata/xcschemes/Apollo.xcscheme
@@ -27,6 +27,7 @@
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       shouldUseLaunchSchemeArgsEnv = "NO"
+      enableThreadSanitizer = "YES"
       codeCoverageEnabled = "YES"
       onlyGenerateCoverageForSpecifiedTargets = "YES">
       <MacroExpansion>
@@ -81,6 +82,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
[...]

PR #1012, opened 2020-02-12
Cancel existing operation in a watcher before starting a new one

designatednerd:
Thanks to @RolandasRazma for the suggestion in #1011 - this removes the hack needed to make things pass added in #1010. 

There's still some pretty significant thread safety issues going on here as pointed out by @cysp, but I think there's a BIG rabbit hole there, and I'd at least like to remove the dirty hack ASAP.


designatednerd:
@cysp It doesn't appear that we have tests specifically for either `AsynchronousOperation` or for `FetchQueryOperation`, which seems to be where the main issue is. We also don't have TSAN turned on for tests in general because there's a bunch of tangled stuff causing races we haven't (so far) been able to detangle. Again, any help we can get on that would be much appreciated.

PR #1016, opened 2020-02-13
Fix memory leaks in Promise whenAll

AnthonyMDev:
When any of the promises reject, we were calling reject, but never leaving the `DispatchGroup`. This means the `notify` block was never being scheduled. This _is_ what we wanted, but the unintended consequence of that is that the `notify` block was then retained indefinitely, and the `group.notify` block is retaining the `promises` or the `resultOrPromises`. This way, the notify block gets called and releases it's reference to the `promises`, but if any rejects occured, it does not call `fulfill`.

@designatednerd I was seeing memory leaks in my application and was able to track it down to this. I wrote a unit test for `Promise`, but unit testing memory leaks of value semantic entities is impossible, so I can't provide a unit test for `ResultOrPromise`. I tested the fix with my app, and the memory leaks are gone now.

### DIFF
--- a/Tests/ApolloTests/PromiseTests.swift
+++ b/Tests/ApolloTests/PromiseTests.swift
@@ -249,4 +249,34 @@ class PromiseTests: XCTestCase {
     
     waitForExpectations(timeout: 1)
   }
+
+  func testWhenAllRejectsWhenAnyOfThePromisesRejectsAsync_doesNotCreateMemoryLeak() throws {
+    var executeReject: ((Error) -> Void)?
+
+    var promises: [Promise<String>] = [Promise(fulfilled: "foo"),
+                                       Promise<String> { _, reject in executeReject = reject },
+                                       Promise(fulfilled: "bar")]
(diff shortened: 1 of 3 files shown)


Tests/ApolloTests/PromiseTests.swift
```diff
@@ -249,4 +249,34 @@ class PromiseTests: XCTestCase {
...
+    }
+
+    promises = []
+    executeReject?(TestError())
+    executeReject = nil
+
+    waitForExpectations(timeout: 1)
+
+    XCTAssertNil(weakPromise0)
+    XCTAssertNil(weakPromise1)
```
designatednerd: So before this fix, this particular `assert` would fail, correct?

AnthonyMDev: Actually, all three of the asserts were failing. The notify block references the entire array of promises, so all of them are retained.

designatednerd: oof, ouch. well this is definitely an improvement.

PR #1027, opened 2020-02-19
Edits to Swift scripting article

StephenBarlow:


docs/source/swift-scripting.md
```diff
@@ -1,30 +1,29 @@
 ---
-title: Swift Scripting
+title: Swift scripting
+sidebar_title: Swift scripting (beta)
 ---
 
-⚠️ **PLEASE NOTE: THIS FUNCTIONALITY IS IN BETA** ⚠️
+> **This functionality is in beta.**
```
designatednerd: I would prefer to keep the warning symbols here - the `>` renders this warning so lightly that it's easy to miss.

docs/source/swift-scripting.md
```diff
@@ -1,30 +1,29 @@
 ---
-title: Swift Scripting
+title: Swift scripting
+sidebar_title: Swift scripting (beta)
```
designatednerd: ooh neat, I didn't know you could do that.

docs/source/swift-scripting.md
```diff
@@ -124,87 +123,84 @@ One of the convenience wrappers available to you in the target is `ApolloSchemaD
...
-5. Build and run using the Xcode project. Note that if you're on Catalina you may get a warning asking if your executable can access files in a particular folder like this:
+5. Build and run using the Xcode project. Note that if you're on Catalina you might get a warning asking if your executable can access files in a particular folder like this:
 
    ![permission prompt](screenshot/would_like_to_access.png)
    
-   Click the "OK" button. Your CLI output will look something like this: 
+   Click **OK**. Your CLI output will look something like this: 
    
-   ![log barf for successful run](screenshot/schema_download_success.png)
+   ![log output for successful run](screenshot/schema_download_success.png)
```
designatednerd: aw you're no fun 🤢

docs/source/swift-scripting.md
```diff
@@ -124,87 +123,84 @@ One of the convenience wrappers available to you in the target is `ApolloSchemaD
...
 
-Note the warning: This isn't relevant for schema downloading, but it *is* relevant for generating code: In order to generate code, you need both the schema and some kind of operation. Now that you've got the schema, it's time to 
+## Using codegen to create a `.graphql` file with an operation
 
-## Codegen theory + creating a `.graphql` file with an operation
+Code generation requires both of the following to run:
 
-Code generation takes a combination of the **schema**, which defines what it's *possible* for you to request from or send to your server, and your **operations**, which define what you are *actually* requesting from the server. 
+* Your **schema**, which defines what it's *possible* for you to request from or send to your server
+* A collection of **operations**, which define what you are *actually* requesting from the server
```
designatednerd: `collection of operations` makes it sound like there should be more than one when we need *at least one*. Any thoughts on how we could phrase that more clearly? Maybe `non-empty collection`?

docs/source/swift-scripting.md
```diff
@@ -124,87 +123,84 @@ One of the convenience wrappers available to you in the target is `ApolloSchemaD
...
 - A **mutation**, which changes data on the server and then receives updated data back
 - A **subscription**, which allows you to listen for changes to a particular object or type of object
 
-The code generation takes your operations and compares them to the schema to validate that what you are asking for is, in fact, possible. If it's not possible, the whole process errors out. If it is possible, it generates Swift code that gives you end-to-end type safety for each operation. 
-
-Thus, a simple equation can be used to describe generating code: 
-
-`schema + operations = code`
-
-If you're missing either of the first two parts, the code can't be generated. If there's operations but no schema, the operations can't be validated, so we can't know if code should be generated. If there's a schema but no operations, then there's nothing to validate or generate code for.
```
designatednerd: I would like to re-add this paragraph and the `schema + operations = code` bit, since it explains *why* this doesn't work if you don't have both, which a lot of people gloss over in my experience.

docs/source/swift-scripting.md
```diff
@@ -247,12 +243,12 @@ Now, you're able to generate code from a debuggable Swift Package Manager execut
     swift run
     ```
     
-    >**NOTE**: If your package ever seems to have problems with caching, run `swift package clean` before `swift run` for a totally clean build. It is not recommended to do this by default since it substantially increases build time.
+    >**Note**: If your package ever seems to have problems with caching, run `swift package clean` before `swift run` for a totally clean build. It is not recommended to do this by default, because it substantially increases build time.
     
-4. Build your target. Since `swift run` is being called from within your target, all of the pieces of the environment, including `$SRCROOT`, will automatically be passed to the environment of the executable, and you don't have to worry about passing anything manually. 
+4. Build your target. Because `swift run` is called from within your target, all of the pieces of the environment, including `$SRCROOT`, are automatically passed to the environment of the executable, and you don't have to worry about passing anything manually.
```
designatednerd: I think I can remove everything after `build your target` here - the SRCROOT thing is no longer necessary.

docs/source/swift-scripting.md
```diff
@@ -247,12 +243,12 @@ Now, you're able to generate code from a debuggable Swift Package Manager execut
...
 
-Now, every time you build your project, this script will get called. Since Swift knows not to recompile everything unless something's changed, it should not have a significant impact on your build time. 
+Now, every time you build your project, this script gets called. Because Swift knows not to recompile everything unless something's changed, it should not have a significant impact on your build time. 
 
 ## Swift-specific troubleshooting
 
-- If at any point you start seeing a bunch of errors around `SecTaskLoadEntitlements` resulting in an immediate exit of the script rather than showing the permission prompt, validate that all the folders you're looking for exist and are at the *exact* path you think they are. You may have a typo in one of your paths. 
\ No newline at end of file
+If you encounter errors around `SecTaskLoadEntitlements` that result in an immediate exit of the script instead of showing the permission prompt, validate that all the folders you're looking for exist at the correct path.
```
designatednerd: i'd like to throw something about typos in here since a typo is what me fight with this for an hour 🙃

docs/source/swift-scripting.md
StephenBarlow: It's pretty new! :D

docs/source/swift-scripting.md
```diff
@@ -1,30 +1,29 @@
 ---
-title: Swift Scripting
+title: Swift scripting
+sidebar_title: Swift scripting (beta)
 ---
 
-⚠️ **PLEASE NOTE: THIS FUNCTIONALITY IS IN BETA** ⚠️
+> ⚠️ **This functionality is in beta.** ⚠️
```
designatednerd: I still would prefer to get rid of the `>` here if possible

PR #1029, opened 2020-02-20
Some Disassembly Required

designatednerd:
This PR switches us from using an ~unholy~ unwieldy combination of Carthage, embedded Xcode projects, and Git Submodules to build the library locally over to using Swift Package Manager. 

This has led to several changes:  
- Everything has been collapsed into a single Xcode project instead of three different projects for the main lib, the SQLite lib, and the WebSockets Lib. ~Still needs to be a workspace at this point, but maaaaybe I can get rid of that too in a future PR? We'll see.~ Nope, got rid of it here!
- Moved several `Test` frameworks to the `Sources` folder since `testTarget` is only for test targets, not test-only frameworks, according to Swift Package manager.
- Got rid of unused targets (performance tests that haven't been run in at least a year, a test host iOS application that nothing was using).
- Changed the way cache-dependent tests are run so that they're subclassed for SQLite tests rather than doing runtime magic to figure out which cache to use. 
- Used `#file` to get access to the filesystem for tests in several places where we were previously having to use environment variables, simplifying code and making it much easier to run the tests from the `Package.swift`. 
- Added shared schemes to allow running tests if you open the `Package.swift` instead of the `xcworkspace`. 

Still trying to figure out: 
- Swift Package Manager is not seeing the `Apollo.h` header, which means the `VersionNumberTests` won't build when run from the SPM package, which is less than ideal. There's probably a better way to do this, still working on what that is (suggestions welcome!)


designatednerd:
For those who were wondering why I was interested in dropping Carthage support: 

Verifying this was working with SPM and CocoaPods: 5 minutes each
Verifying this was working with Carthage: Four hours and having to disassemble and reassemble my Carthage testbed project

😆😭

designatednerd:
Still looking into the version number thing, but since that didn't work when installed via SPM before, I'm gonna go ahead and merge.

PR #1046, opened 2020-02-27
Codegen Parsing 1: The beginning

designatednerd:
In this PR: 
- Added a switch to be able to turn on Swift-based codegen. Basically, nobody but me should use this for now. 😇 
- Got the AST output of the Star Wars Schema + queries
- Got initial models of the parsed objects
- Got initial set of tests up and running
- Added some workarounds for the fact that my computer has lost its damned mind and is randomly deciding to not open files during testing


sneakyness:
FWIW, I switched over to the Swift codegen mid last week and am enjoying it so far w/o any issues. Writing build scripts in Swift is like drinking clean water for the first time, and I'm someone who enjoys writing shell scripts!

If there's anything specific you need done with that feel free to point the way!

designatednerd:
Aw yay! Glad to hear that. And keep an eye out for things labeled `help-wanted` - those are definitely things I need help with

Issue #516, opened 2019-04-23
GraphQLQueryWatcher retained by ApolloStore


AnthonyMDev:
I might be missing something here, but it seems to me like a `GraphQLQueryWatcher` that is not cancelled explicitly is retained by the `ApolloStore` and will continue to receive updates, even if the object that created the watcher (by calling `watch()`) releases it.

Do I need to be ensuring that I manually `cancel()` my `GraphQLQueryWatcher` every time I'm done with it? It would be nice if, when I have no retain on the watcher, it automatically cancels.

If I'm missing something and it already works this way, then please let me know and close this!

Thanks

niraj1991:
Hi @AnthonyMDev,

I have same issue do you find any solution?

Thanks

designatednerd:
Yeah not having to manually cancel would certainly be ideal, but I don't think that's how things are set up at the moment. I'll have to see how plausible that is, though I don't have an ETA for you at the moment.

AnthonyMDev:
It seems to me like the solution would just be having the watcher call `cancel()` on itself in it's own `deinit` block.

designatednerd:
If it's being retained by the Apollo store, wouldn't `deinit` never get called?

AnthonyMDev:
Ahhhh I looked at this last quite a while ago. I forgot about that part. Yes, we would need to figure that out.

casey-chow:
Sounds like it's time for some good old DisposeBags.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

designatednerd:
@casey-chow That...might be difficult given that we don't (and don't intend to) use Rx on this project 😃

koenpunt:
Would be a breaking change, but making the store not retain the watcher, and require the developer to retain a reference to the watcher for as long as it's needing the watcher could work, right?

wnagrodzki:
I also expected that query watcher would unsubscribe itself once I release all references to it. Maybe documentation needs to stress this even more, with a `- warning: `.

One can create a dedicated object to unsubscribe query watcher from Apollo store upon it's deallocation. That simplified the code in my project a bit.

```swift
final class ApolloWatcherHandle<Query>: QueryWatcher where Query: Apollo.GraphQLQuery {
    
    private let watcher: GraphQLQueryWatcher<Query>
    
    init(watcher: GraphQLQueryWatcher<Query>) {
        self.watcher = watcher
    }
    
    deinit {
[...]

PR #1048, opened 2020-02-28
Add note about the store retaining the query watcher

designatednerd:
Trying to add a bit more information about what's happening until I can get around to fixing #516.

### DIFF
--- a/Sources/Apollo/GraphQLQueryWatcher.swift
+++ b/Sources/Apollo/GraphQLQueryWatcher.swift
@@ -1,6 +1,8 @@
 import Dispatch
 
 /// A `GraphQLQueryWatcher` is responsible for watching the store, and calling the result handler with a new result whenever any of the data the previous result depends on changes.
+///
+/// NOTE: The store retains the watcher and does not automatically unsubscribe you. You must call `cancel()` on your query watcher when you no longer need results.
 public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, ApolloStoreSubscriber {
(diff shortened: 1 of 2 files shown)


designatednerd:
@AnthonyMDev @koenpunt @wnagrodzki would appreciate feedback on the language here

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -1,6 +1,8 @@
 import Dispatch
 
 /// A `GraphQLQueryWatcher` is responsible for watching the store, and calling the result handler with a new result whenever any of the data the previous result depends on changes.
+///
+/// NOTE: The store retains the watcher and does not automatically unsubscribe you. You must call `cancel()` on your query watcher when you no longer need results.
```
AnthonyMDev:
```suggestion
/// NOTE: The store retains the watcher while subscribed. You must call `cancel()` on your query watcher when you no longer need results. Failure to call `cancel()` before releasing your reference to the returned watcher will result in a memory leak.
```

wnagrodzki:
Thank you. I think that's perfect :)

What do you think about updating [this tutorial page](https://www.apollographql.com/docs/ios/caching/#watching-queries) with as well? Maybe replacing _kill of_ with mention about `cancel()` method and memory leak will draw a bit more attention to this issue for newcomers.

> Remember to kill off watchers when their parent object is deallocated! This is not (presently) done automatically.

designatednerd:
@wnagrodzki Yep great point.

PR #1061, opened 2020-03-04
Make GraphQLResponse a generic on the response type instead of on the operation type

gsabran:
It looks like there's little reasons to have `GraphQLResponse` be tied to the operation type, instead of more simply to the response type (`.Data`)

I was interested in this change as I use a query wrapper class around the `GraphQLOperation` generated by the codegen. This wrapper allows me to easily set context for my `NetworkTransport` that receives the operation through Apollo. While it shares the same response type as the wrapped operation, it is of a different type and this was not playing well with the `GraphQLResponse`

This slightly changes the public API of `NetworkTransport` as well, but all the changes are straightforward.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -78,10 +78,10 @@ public class ApolloClient {
     }
   }
 
-  private func handleOperationResult<Operation>(shouldPublishResultToStore: Bool,
+  private func handleOperationResult<Data: GraphQLSelectionSet>(shouldPublishResultToStore: Bool,
                                                 context: UnsafeMutableRawPointer?,
-                                                _ result: Result<GraphQLResponse<Operation>, Error>,
-                                                resultHandler: @escaping GraphQLResultHandler<Operation.Data>) {
+                                                _ result: Result<GraphQLResponse<Data>, Error>,
+                                                resultHandler: @escaping GraphQLResultHandler<Data>) {
     switch result {
     case .failure(let error):
       resultHandler(.failure(error))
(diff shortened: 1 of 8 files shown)


designatednerd:
> While it shares the same response type as the wrapped operation, it is of a different type and this was not playing well with the `GraphQLResponse`

Can you clarify a bit what you mean by this, perhaps by giving an example? 

This looks like a reasonably straightforward changeset, but it is definitely breaking so I'd like to mull it over a little. In the meantime, do you mind fixing the alignment of parameters in the method declarations that broke because you changed the name of the generic type?

gsabran:
Sure. At a high level the wrapper strategy looks like:

```swift
// networking service
class NetworkingService {
  init() {
    self.apolloClient = ApolloClient(transport: NetworkTransport())
  }

  func fetch(query: GraphQLQuery, requestConfiguration: RequestConfiguration, completion: ...) {
    // wrappedQuery is GraphQLQuery with the same Data type, but of different type than query
    let wrappedQuery = WrappedQuery(query, requestConfiguration)
    apolloClient.fetch(wrappedQuery, completion)
  }
}

class NetworkTransport {
  // called from Apollo
  func send(query: GraphQLQuery, completion: ...) {
    let requestConfiguration = (query as? WrappedQuery)?.requestConfiguration
    let request = buildRequest(query.operationName, query.params, requestConfiguration)
    apiSession.send(request, completion)
  }
}
```

designatednerd:
Ahhh ok so this is enabling a custom network transport, that's the piece I was missing in my mental model.

designatednerd:
Updated code looks good, I do want to think this over a bit since it'd be a breaking change.

LooTKinGz:
File update

Issue #1039, opened 2020-02-25
SwiftUI Preview Fails


peck:
I think I've followed the installation instructions on the docs page, https://www.apollographql.com/docs/ios/installation/, and my builds works and runs on the similar just fine, but my swiftui preview breaks with this error in the diagnostics. I'm new to swiftui (though fairly familiar with Apollo/graphql from other projects) and searching didn't find similar issue.

apollo-ios -> 0.23.0
Xcode -> 11.3.1

```
use of unresolved identifier \'Process\'

----------------------------------------

SchemeBuildError: Failed to build the scheme "practice"

use of unresolved identifier 'Process'

Compile /Users/peck/Library/Developer/Xcode/DerivedData/practice-ffshylstkbdpuwdygwokyrmihofw/SourcePackages/checkouts/apollo-ios/Sources/ApolloCodegenLib/Basher.swift:
/Users/peck/Library/Developer/Xcode/DerivedData/practice-ffshylstkbdpuwdygwokyrmihofw/SourcePackages/checkouts/apollo-ios/Sources/ApolloCodegenLib/Basher.swift:27:16: error: use of unresolved identifier 'Process'
    let task = Process()
               ^~~~~~~
```

On apollo-ios 0.22.0 I don't see the issue and preview seems to generally work as expected

designatednerd:
Hi! What package manager are you using? Can you validate that the `ApolloCodegenLib` is not included in the libraries that are being added to your iOS application? It's designed to only be used with MacOS, though it seems like I should probably add some defensive `if` statements around that.

designatednerd:
@peck The defensive `@available` statements have shipped with `0.23.1` which should prevent compilation of any of the classes that aren't available on iOS. I'm gonna close this out, please reopen if you're still seeing this issue when using 0.23.1. Thank you!

peck:
Apologies for not seeing your first reply, didn’t get a notification from github for that one.

I’ll give the new version a try and see if it fixes it.

To answer the question, I’m using swift package manager and did not have codegen lib included.

Thanks!

peck:
Following up, I tried with the latest release and get the same error

iancanderson:
👋 I got the same error using `0.23.0`, and still in `0.23.1`. Downgrading to `0.22.0` fixed the issue for me as well - SwiftUI preview is working again

designatednerd:
[...]

PR #1066, opened 2020-03-05
SwiftUI previews fix

TizianoCoroneo:
Hello @designatednerd  👋 
Thank you for your amazing work on this library! I had the [same issue reported here](https://github.com/apollographql/apollo-ios/issues/1039) and decided to try to help out.

This PR fixes an issue that started with the inclusion of `ApolloCodegenLib` in `0.23.0`, where all SwiftUI previews for files in targets that also import Apollo stopped working with the following error: 
```
use of unresolved identifier 'Process'

Compile /Users/peck/Library/Developer/Xcode/DerivedData/practice-ffshylstkbdpuwdygwokyrmihofw/SourcePackages/checkouts/apollo-ios/Sources/ApolloCodegenLib/Basher.swift:
/Users/peck/Library/Developer/Xcode/DerivedData/practice-ffshylstkbdpuwdygwokyrmihofw/SourcePackages/checkouts/apollo-ios/Sources/ApolloCodegenLib/Basher.swift:27:16: error: use of unresolved identifier 'Process'
    let task = Process()
```
In my case, I have Apollo as a dependency using SPM. I don't know if this issue is reproducible with cocoapods or other dependency managers.

This happens because `ApolloCodegenLib` is meant to be built only on macOS, and this check is only enforced with a runtime `@available` annotation instead of a compile-time `#if os(macOS)`.
I looked for a way to specify a specific platform per-target directly in the SPM manifest, but according to [this PR in the SPM repo](https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md) there is no way for now (and he recommends to strip away the platform specific code using `#if`): 
> Platform-specific targets or products: Consider that a package supports multiple platforms but has a product that should be only built for Linux. There is no way to express this intent and the build system may end up trying to build such targets on all platforms. A simple workaround is to use #if to conditionalize the source code of the target. Another use case comes up when you want to keep deployment target of a certain target lower than other targets in a package. A workaround is factoring out the target into its own package. A proper solution to this problem will be explored in a separate proposal, which should provide target-level settings.

This PR applies the suggestion specified above. I'm sorry it is not below 20 lines and I'm not really following the contributor guide, but this seems an "innocuous" bug fix. 😄 

I don't know how to test this behavior with SwiftUI previews to be honest.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCLI.swift
+++ b/Sources/ApolloCodegenLib/ApolloCLI.swift
@@ -1,7 +1,9 @@
 import Foundation
 
+// Only available on macOS
+#if os(macOS)
+
 /// Wrapper for calling the bundled node-based Apollo CLI.
-@available(OSX, message: "Only available on macOS")
 public struct ApolloCLI {
   
   /// Creates an instance of `ApolloCLI`, downloading and extracting if needed
@@ -47,3 +49,5 @@ public struct ApolloCLI {
     return try Basher.run(command: command, from: folder)
   }
 }
+
+#endif
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -1,7 +1,9 @@
 import Foundation
 
+// Only available on macOS
+#if os(macOS)
+
 /// A class to facilitate running code generation
-@available(OSX, message: "Only available on macOS")
 public class ApolloCodegen {
   
   /// Errors which can happen with code generation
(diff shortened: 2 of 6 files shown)


designatednerd:
Hey thank you for doing this! I think you’re being slightly too aggressive adding it to every file - it should really only be the files I marked with `@available` which should need this. I can’t believe I forgot that was runtime only. 🤦‍♀️

It does seem to be a combo SPM/SwiftUI bug that this lib gets compiled for SwiftUI even when it’s not a dependency of the target being compiled though. Will file a radar on that one.

TizianoCoroneo:
I removed the extra `#if(macOS` from the source files that didn't contain `@available` before.
Tested with my SwiftUI previews and it works 👍

designatednerd:
Awesome, thanks so much for this!

PR #1081, opened 2020-03-13
Uploading files. Change variables map

vandyshev:
Fix files uploading with different fieldName's

Changed map in Request payload based on [https://github.com/jaydenseric/graphql-multipart-request-spec#multipart-form-field-structure](https://github.com/jaydenseric/graphql-multipart-request-spec#multipart-form-field-structure)
from:
`{"0":["variables.avatar.0"],"1":["variables.background.1"]}`
to:
`{"0":["variables.avatar"],"1":["variables.background"]}`


Sources/Apollo/RequestCreator.swift
```diff
@@ -124,17 +124,23 @@ extension RequestCreator {
...
-    var map = [String: [String]]()
-    if files.count == 1 {
-      let firstFile = files.first!
-      map["0"] = ["variables.\(firstFile.fieldName)"]
-    } else {
-      for (index, file) in files.enumerated() {
-        map["\(index)"] = ["variables.\(file.fieldName).\(index)"]
+    var map = [String: [String: String]]()
+    files.enumerated().forEach { (index, file) in
+      var field = map[file.fieldName, default: [:]]
```
designatednerd: One thing that keeps throwing me off when I read this is `field.count`, partly because `field` is singular so I don't expect it to have a `count`. Would recommend renaming it something like `fieldsForFieldName` for clarity

Sources/Apollo/RequestCreator.swift
```diff
@@ -124,17 +124,23 @@ extension RequestCreator {
...
+        field = field.mapValues { return $0 + ".0" }
+        field["\(index)"] = "variables.\(file.fieldName).\(field.count)"
+      } else {
+        field["\(index)"] = "variables.\(file.fieldName).\(field.count)"
       }
+      map[file.fieldName] = field
     }
 
-    let mapData = try serializationFormat.serialize(value: map)
+    let mapValue = Dictionary(uniqueKeysWithValues: map.flatMap { $0.1 }.map { ($0.0, [$0.1]) })
```
designatednerd: This is impressively short, but it's awfully hard to tell what's going on here. Can you please give some of these variables names so it's easier to tell what's going on here?

designatednerd:
@vandyshev Have you had a chance to look at the questions I asked here?

vandyshev:
Change variables mapping for fix bug with two files uploading. Adding tests for FileUpload based on [android version](https://github.com/apollographql/apollo-android/tree/master/apollo-integration/src/main/graphql/com/apollographql/apollo/integration/upload) and [spec](https://github.com/jaydenseric/graphql-multipart-request-spec#multipart-form-field-structure)

designatednerd:
Hey @vandyshev - I appreciate the effort you've put in here, but this changeset has gotten enormous, and it's harder to tell what you've changed to fix the issue you ran into vs. what you've changed to work with the second framework you've added. 

Is there a way to back this up to where it's just the changes to the existing code rather than this whole changeset?

vandyshev:
@designatednerd If you look closely, only Sources/Apollo/RequestCreator.swift has changed.
The rest of the code is tests. The previous RequestCreator.swift didn’t work correctly to load two or nested files.

designatednerd:
I mostly see that, but it'd be helpful to see what code in the tests is being changed because it's wrong vs what code in the tests is being changed because it's looking at a different lib. 

Honestly if we're going to do another test lib for the uploading stuff, I'd rather set up an upload server and do it that way rather than just duplicating our existing library and messing around with it.

vandyshev:
> I mostly see that, but it'd be helpful to see what code in the tests is being changed because it's wrong vs what code in the tests is being changed because it's looking at a different lib.

Changed tests in Tests/ApolloTests/RequestCreatorTests.swift. 
Before there were no tests for upload two files. I did not add them to the StarWars API, because it has a different purpose.

What to do next it's up to you. But uploading files is sucks

designatednerd:
I agree the upload situation isn't ideal. IMHO, the upload spec is somewhat bolted on to GraphQL as a concept since GraphQL itself doesn't really have any notion of files. We've started [advising people to use other upload methods for anything beyond proof of concept](https://www.apollographql.com/docs/ios/mutations/#an-important-caveat-about-file-uploads), because we've found that to be a much, much better long term solution. 

That doesn't mean I don't want this to be working better - it does mean that I want to be really careful about what we do wind up merging into our support. Merging huge pull requests when smaller ones will still fix the issue is against that policy. 

I understand that you don't feel like this is a huge change, but I respectfully disagree: It's adding yet another framework when we may or may not have to, and we definitely don't have to in order to get the current solution working better. The framework as added may also not be the correct solution: A framework that more specifically targets uploads (and doesn't have the rest of the Star Wars stuff in it) is probably going to be easier to deal with in the long term. 

If you'd like to continue with this PR, please trim it back down to just the code and test fixes, and remove the new framework. If you'd prefer not to, I understand.

designatednerd:
Thank you very much for making that change. It looks like the test changes for `testMultipleFilesWithApolloRequestCreator` to validate we're using the multiple files stuff correctly are what needs to go back in, and then we should be good to go.

designatednerd:
@vandyshev Do you mind if I take over this PR to get it across the finish line?

designatednerd:
Hi! I've opened #1279 to address this. One thing to note is that if you're uploading [an array of files with the same name](https://github.com/jaydenseric/graphql-multipart-request-spec#file-list), you do still need the `.0` etc for the files in that variable. I'm going to close this one out but would love your review on that one if you have time. Thank you for poking me on this!

PR #1091, opened 2020-03-19
Documentation + Package.resolved updates

designatednerd:
In this PR: 
- Added major caveat about uploading to reflect general Apollo guidance about using GraphQL-based uploads to documentation
- Regenerated docs in preparation for 0.24.0
- Committed some changes to the `Package.resolved` I'd like to run through CI before shipping all this


docs/source/api/ApolloCodegenLib/classes/EnumGenerator.md
```diff
@@ -0,0 +1,58 @@
...
+
+```swift
+public class EnumGenerator
+```
+
+## Properties
+### `enumTemplate`
+
+```swift
+open var enumTemplate = """
```
designatednerd: Looks like our doc generator will put the full value of a directly assigned variable into the docs. Good to know 😛

PR #1092, opened 2020-03-20
Fix bash paths with spaces

sfoulston:
The `Basher.run` command is failing when it contains paths with spaces. This change encloses string interpolated paths in single quotes to allow paths with spaces to work with the bash script.

Changed `CodegenTestHelper.cliFolderURL` to use a path with a space in it (changed `"scripts"` to `"scripts directory"`) to test this change.


sfoulston:
Re-opening PR, accidentally deleted branch.

PR #1095, opened 2020-03-22
Workaround CLI argument parsing issue

sfoulston:
Due to an underlying issue with the Oclif framework that the Apollo CLI uses, the `schema:download` command (at least) fails when `--header` arguments are not placed at the end of the command, such as when specifying a non-default output path. The ApolloCodegenLib places the output path after the header argument, so the schema download always fails.

This PR just moves the header argument to the end of the bash command by reordering the array of arguments in `ApolloSchemaOptions#arguments`.

See also:
https://github.com/apollographql/apollo-tooling/issues/844#issuecomment-547143805


sfoulston:
PR was created against wrong branch. Please see https://github.com/apollographql/apollo-ios/pull/1096

Issue #1102, opened 2020-03-29
Swift Scripting: error: unknown package 'Apollo' in dependencies of target 'Codegen'


alexandrethsilva:
Hi @designatednerd!
First of all, thanks for the great effort on this script. It's really great to have someone looking into how to improve the existing process and making it better.

I'm not sure what may be up as I'm relatively new to Swift and Xcode development in general, but there's a problem when doing [**the setup described in the docs**](https://www.apollographql.com/docs/ios/swift-scripting/) that is really cryptic as I can't find much about it somewhere else.

As for the setup, I'm currently on **Xcode Version 11.4 (11E146)** and therefore **`swift-tools` 5.2**.

When getting to the **step 4** in the process (**Update the dependencies section to grab the Apollo iOS library**), I get the following message on the sidebar:

`Showing All Messages: dependency 'Apollo' is not used by any target`

Okay, makes sense, since I didn't add the target dependency yet. So I go ahead and do it. But then, immediately after adding it, I get the following:

`unknown package 'Apollo' in dependencies of target 'Codegen'`

My Package.swift currently looks like this:

```swift
// swift-tools-version:5.2
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "Codegen",
    dependencies: [
        // Dependencies declare other packages that this package depends on.
        // .package(url: /* package url */, from: "1.0.0"),
        .package(url: "https://github.com/apollographql/apollo-ios.git",
        from: "0.22.0")
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(name: "Codegen",
            dependencies: [
            .product(name: "ApolloCodegenLib", package: "Apollo")
            ]),
        .testTarget(
            name: "CodegenTests",
            dependencies: ["Codegen"]),
    ]
)
```

As a sidenote, if I comment the dependency out, Xcode downloads the package from github and adds the deps to the sidebar, like here:

![image](https://user-images.githubusercontent.com/117481/77854784-9ab51280-71ec-11ea-9326-6f7e1c25e3db.png)

But when I add it:

![image](https://user-images.githubusercontent.com/117481/77854791-a1438a00-71ec-11ea-8a4e-116ad1df92bd.png)

[...]

PR #1105, opened 2020-03-30
Fix package dependency syntax

alexandrethsilva:
@designatednerd I went ahead and made the PR related to #1102 in case you'd like to have it merged as I will probably be offline for a few hours.

One thing I'd like to ask though, is whether it would make sense to have the officially suggested syntax pointing to an exact version rather then a range and updating the docs as new versions are released, as in:

```swift
.package(name: "Apollo", url: "https://github.com/apollographql/apollo-ios.git", .exact("0.24.0"))
```
 I ask for a few reasons:
- You point out right below that the version should be an exact match to the one in the main project, but the unaware user may end up mixing things? That would also make it more in line with [**the iOS SDK setup docs**](https://github.com/apollographql/apollo-ios/blame/master/docs/shared/spm-installation-panel.mdx#L28) which favour the version to be up to the next minor only?
- It makes it easier to debug opened issues, since you'd know the exact version installed instead of having to guess or depend on the reporter double-checking it?
- It avoids propagating new bugs to people that setup their projects earlier and end up pulling a later version in the range (even if minor) which contains a bug?

Let me know if any of this makes sense and whether I should update the PR to make it align with these points.

Cheers!

### DIFF
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -28,8 +28,7 @@ To begin, let's set up a Swift Package Manager executable:
 4. Update the `dependencies` section to grab the Apollo iOS library:
 
     ```swift
-    .package(url: "https://github.com/apollographql/apollo-ios.git", 
-             from: "0.22.0")
+    .package(name: "Apollo", url: "https://github.com/apollographql/apollo-ios.git", from: "0.23.0")
     ```
   **NOTE**: The version should be identical to the version you're using in your main project. 
 


designatednerd:
I think this issue points to me needing to update something in `Package.swift`- you absolutely shouldn't need to do this manually. 

Good call on the `from` though - I should probably update that to get up to the next minor to match our recommendation for the main repo

alexandrethsilva:
@designatednerd should I drop this PR for the time being or would you like me to maybe take a different approach?

designatednerd:
Let's leave this up for now, I'm working on a fix but until it's actually ready this will at least help anyone who's stuck

designatednerd:
@alexandrethsilva I'd like to close this out in favor of #1106 - I really appreciate you taking the time to make this but I want to have the extra context I've added in that PR. is that cool?

designatednerd:
Went ahead and merged #1106 so I'm gonna close this out - thanks again for helping figure this out!

alexandrethsilva:
Hi @designatednerd, sure! Glad that I could help out a bit. 🤓

Issue #1104, opened 2020-03-30
Can't set enableSOCKSProxy true


KonstantinEfimenko:
I can't track queries by Charles Proxy. Looks like it already was solved in Starscream https://github.com/daltoniam/Starscream/pull/423/commits/b074c6c8330527b95f1ad9169b1d9ea9eea935e1, but I can't reach this property due to internal access level of websocket in WebSocketTransport

designatednerd:
I think we're trying to keep from directly exposing the web socket, but I can probably add a property to update this.

designatednerd:
Hey @fassko Can you think of any major wormholes I'm opening doing this? 😇

fassko:
@designatednerd Hmm, honestly hard to say. I'm not so proficient about proxies. :)

About Starscream and Websockets our long term plan should be using Apple's provided Websocket solution anyways.

designatednerd:
Isn't Starscream already doing that under the hood? 

Also we support back to iOS 9 so it's gonna be a WHILE before we can drop the non-apple stuff 😭

fassko:
Maybe it does, but I have not used it myself. Currently I won't have time to look into that cause I'm dealing with all the mess going on on the planet right now.

designatednerd:
TOTALLY fair! My productivity has been 💩the last few weeks myself.

designatednerd:
Shipped with `0.25.0`. Up for SPM and Carthage, publishing to CocoaPods trunk now.

Robuske:
Hey @KonstantinEfimenko , how did you setup Charles to watch for subscriptions? I tried several different configurations, but none seem to work 😞

designatednerd:
Heads up that this has been removed in more recent versions because Starscream itself removed support for it - I have a sneaking suspicion it may have been a bit flaky.

Robuske:
Oh... I am using an outdated version at the moment because we had no time to upgrade =/
Is there any alternative to subscription debugging @designatednerd?

designatednerd:
Using the debugger to pause on new items or to print new data received is probably going to be your best bet.

DanielKrofchick:
It's a bit unclear to me from the discussion above. Can someone confirm if Apollo Subscriptions can or cannot be monitored using Charles, and if so how?

designatednerd:
It was possible before we upgraded to Starscream 4.0. Starscream 4.0 removed support for SOCKS proxying, and so we consequently had to remove support for it as well, unfortunately.

PR #1108, opened 2020-03-31
Add ability to get to Starscream's underlying SOCKS proxy property.

designatednerd:
Addresses #1104. Definitely glad I went this way - it's not just that you have to set it on the web socket, you have to set it on the underlying stream. Ouch. 

Anyway that's all abstracted away now so yay.

### DIFF
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -61,6 +61,28 @@ public class WebSocketTransport {
       self.addApolloClientHeaders(to: &self.websocket.request)
     }
   }
+  
+  /// Determines whether a SOCKS proxy is enabled on the underlying request.
+  /// Mostly useful for debugging with tools like Charles Proxy.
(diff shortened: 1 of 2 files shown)


designatednerd:
@KonstantinEfimenko Wanna take a look?

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -61,6 +61,28 @@ public class WebSocketTransport {
...
+    get {
+      guard let socket = self.websocket as? SOCKSProxyable else {
+        // If it's not proxyable, then the proxy can't be enabled
+        return false
+      }
+      
+      return socket.enableSOCKSProxy
+    }
+    set {
+      guard var socket = self.websocket as? SOCKSProxyable else {
```
designatednerd: this has to be `guard var` as `guard let` wouldn't let me change `enableSOCKSProxy` since this protocol is not class-bound.

KonstantinEfimenko:
Looks good. Thank you!

Issue #1100, opened 2020-03-27
Automatic persisted queries does not work with mutations


DaniilSokolyuk:
Automatic persisted queries does not work with mutations, [only for query](https://github.com/apollographql/apollo-ios/blob/ae85cb0993c17d9e41214efaa968444bc55a3814/Sources/Apollo/HTTPNetworkTransport.swift#L345).

There are no restrictions in the [documention ](https://www.apollographql.com/docs/apollo-server/performance/apq/), only 
> For full-length queries and for all mutations, Apollo Client will continue to use POST requests.

designatednerd:
It looks like when it was initially implemented it was only implemented for queries and didn't send the APQ info for mutations. Looks like there's not a *technical* limit to this, just that we're not handling that at the moment.

My bandwidth is somewhat limited at the moment, but if you're interested in doing a PR to address this I'd be happy to take a look!

designatednerd:
Shipped with `0.25.0`. Up for SPM and Carthage, publishing to CocoaPods trunk now.

PR #1110, opened 2020-03-31
Fix APQs Automatic persisted queries does not work with mutations

IvanStebletsov:
Issue: APQs Automatic persisted queries does not work with mutations https://github.com/apollographql/apollo-ios/issues/1100

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -343,13 +343,15 @@ public class HTTPNetworkTransport {
     let sendQueryDocument: Bool
     let autoPersistQueries: Bool
     switch operation.operationType {
-    case .query:
+    case .query, .mutation:
       if isPersistedQueryRetry {
-        useGetMethod = self.useGETForPersistedQueryRetry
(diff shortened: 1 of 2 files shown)


Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -343,13 +343,15 @@ public class HTTPNetworkTransport {
...
       if isPersistedQueryRetry {
-        useGetMethod = self.useGETForPersistedQueryRetry
+        useGetMethod = operation.operationType == .mutation ? false : self.useGETForPersistedQueryRetry
         sendQueryDocument = true
         autoPersistQueries = true
       } else {
-        useGetMethod = self.useGETForQueries || (self.enableAutoPersistedQueries && self.useGETForPersistedQueryRetry)
+        useGetMethod = operation.operationType == .mutation ? false :
+                                                   self.useGETForQueries ||
+                                                   (self.enableAutoPersistedQueries && self.useGETForPersistedQueryRetry)
```
designatednerd: I'm not a giant fan of these huge ternaries, but I'll deal with that post-merge.

Issue #1116, opened 2020-04-01
Expose URL response in HTTPNetworkTransportGraphQLErrorDelegate


erudel:
Hi,

I am trying to implement a delegate to intercept responses that contain a GraphQL error and it seems that `HTTPNetworkTransportGraphQLErrorDelegate` would be the perfect place to do that.
In my particular use case, it would be also helpful to have access to the headers from the (HTTP) response to perform some additional logging. However, I noticed that the response is not exposed in the delegate method and instead other higher level objects are passed.

Do you think it would be reasonable to also expose the response? Or is there another way to get this information?

If you think this makes sense, I would be happy to start working on a PR to address this issue.

designatednerd:
I think what you're looking for is the [`HTTPNetworkTransportTaskCompletedDelegate`](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/HTTPNetworkTransport.swift#L30) - that gives you the raw data back from the request.

erudel:
Thank you for your reply @designatednerd. I looked into that delegate, and as you mentioned, you get the raw data at that point so we don't know yet whether the response contains a GraphQL error or not.
Is there a way to construct the result and inspect errors from the task completed delegate?

designatednerd:
The design of these delegates assumes that if you have a GraphQL error, the information you'd need to deal with a GraphQL error is actually included in the error itself. This is a best practice if you're going to be propagating errors using the GraphQL error system. 

I'm looking longer-term at a different system that would allow easier hooking into responses at any point in the stack, but that's going to be...a while.

erudel:
I agree with you on the idea that the information provided by these delegates should be enough to deal with error handling on the client side.

The reason I was looking to expose the response is that it would allow to perform error tracing (by looking at the HTTP headers) and help pinpoint the potential breakage on the backend side.

Maybe a better way to solve this issue would be to include the trace id in the error, rather than exposing the raw response and looking it up there. I'm curious to hear your thoughts on this approach.

designatednerd:
Yep, that sort of thing is exactly why the error can contain arbitrary JSON.

erudel:
Sounds good, thank you very much for taking the time to answer this and steer me in the right direction. Feel free to close the issue since the original question is not relevant anymore

designatednerd:
You're welcome!

xma000:
[...]

PR #1120, opened 2020-04-02
add request to GraphQLHTTPResponseError for easier debugging

RolandasRazma:
Was trying to trace some problems with auth and it was very useful to see what we actually send to server (headers in particular)

### DIFF
--- a/Sources/Apollo/GraphQLHTTPResponseError.swift
+++ b/Sources/Apollo/GraphQLHTTPResponseError.swift
@@ -22,6 +22,8 @@ public struct GraphQLHTTPResponseError: Error, LocalizedError {
     }
   }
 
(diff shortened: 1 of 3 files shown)


designatednerd:
Hey @RolandasRazma - I'm inclined to decline this one for similar reasons to the discussion in #1116 - you should be able to get the headers from another delegate and if you need more information, your server should really be putting it into the GraphQL error on its end.

RolandasRazma:
thought you will say so :)

designatednerd:
Do you have any other thoughts on this or should I just close it out?

RolandasRazma:
I think you can close it as its limited use, I personally can't use it from delegate but that's more of architectural limitation than Apollo problem

PR #1127, opened 2020-04-07
Add instructions to import KeychainSwift

ck3g:
The implementation of `DetailViewController.swift` requires the import of `KeychainSwift`. 

Adds instructions to include `KeychainSwift` library at the top of `DetailViewController.swift`.


ck3g:
That's not a problem figure out that one needs to include `KeychainSwift`. The compiler will yell about that anyway. 

It doesn't make the tutorial unusable or hard-to-follow. But I think it would be nice to have. 😄

designatednerd:
yeah just had to rerun a flaky test before merging 😇

Issue #1121, opened 2020-04-03
returning different error in -[HTTPNetworkTransportRetryDelegate networkTransport:receivedError:for:response:retryHandler]


RolandasRazma:
The problem I'm facing:
doing apolloClient.fetch()
request fails for reason Y (getting 401 to delegate)
trying to update token
token update fails for reason X
calling retryHandler(false)
request fails and callback in apolloClient.fetch() see error Y, where real error is X

What I would like to do is to modify (add new one and deprecate existing one) -[HTTPNetworkTransportRetryDelegate networkTransport:receivedError:for:response:retryHandler]
where you could return different error if you don't want to retry request

does that make sense @designatednerd or maybe there is better way to do that?

How it works now:

```swift
// HTTPNetworkTransportPreflightDelegate

func networkTransport(_ networkTransport: HTTPNetworkTransport, willSend request: inout URLRequest) {
  oauth.authenticate(&request)
}

// HTTPNetworkTransportRetryDelegate

func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedError error: Swift.Error, for request: URLRequest, response: URLResponse?, retryHandler: @escaping (Bool) -> Void) {

  guard
    (response as? HTTPURLResponse)?.statusCode == 401 // Unauthorized - meaning token is probably expired
  else {
    retryHandler(false)
    return
  }
  
  oauth.updateToken { updateTokenError in
    retryHandler(updateTokenError == nil)
  }
}

// App

apolloClient.fetch(query: ...) { (result: Result<GraphQLResult<Query.Data>, Swift.Error>) in
  // Error is 401, and it is correct from apollo perspective, but its not real error if oauth.updateToken fails for whatever reason - lets say internet connectivity problem
}
```

What I would like to do is:

```swift
// HTTPNetworkTransportRetryDelegate

func networkTransport(_ networkTransport: HTTPNetworkTransport, receivedError error: Swift.Error, for request: URLRequest, response: URLResponse?, continueHandler: @escaping (ContinueAction) -> Void) {

  guard
    (response as? HTTPURLResponse)?.statusCode == 401 // Unauthorized - meaning token is probably expired
  else {
    continueHandler(.fail(error))
    return
  }
  
  oauth.updateToken { updateTokenError in
    if let updateTokenError = updateTokenError {
      continueHandler(.fail(updateTokenError))	
    } else {
      continueHandler(.retry)
    }
  }
  
}

// App

[...]

PR #1128, opened 2020-04-07
improve HTTPNetworkTransportRetryDelegate to enable returning custom error

RolandasRazma:
solves use case described in https://github.com/apollographql/apollo-ios/issues/1121 where delegate might want to return different error depending on logic

### DIFF
--- a/Sources/Apollo/HTTPNetworkTransport.swift
+++ b/Sources/Apollo/HTTPNetworkTransport.swift
@@ -59,11 +59,52 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate 
   ///   - request: The URLRequest which generated the error
   ///   - response: [Optional] Any response received when the error was generated
   ///   - retryHandler: A closure indicating whether the operation should be retried. Asyncrhonous to allow for re-authentication or other async operations to complete.
+  @available(*, deprecated, message: "Use networkTransport(_:receivedError:for:,response:continueHandler:) instead")
   func networkTransport(_ networkTransport: HTTPNetworkTransport,
                         receivedError error: Error,
                         for request: URLRequest,
                         response: URLResponse?,
                         retryHandler: @escaping (_ shouldRetry: Bool) -> Void)
+  
+  /// Called when an error has been received after a request has been sent to the server to see if an operation should be retried or not.
+  /// NOTE: Don't just call the `retryHandler` with `true` all the time, or you can potentially wind up in an infinite loop of errors
+  ///
+  /// - Parameters:
+  ///   - networkTransport: The network transport which received the error
+  ///   - error: The received error
+  ///   - request: The URLRequest which generated the error
+  ///   - response: [Optional] Any response received when the error was generated
+  ///   - retryHandler: A closure indicating whether the operation should be retried. Asyncrhonous to allow for re-authentication or other async operations to complete.
+  func networkTransport(_ networkTransport: HTTPNetworkTransport,
+                        receivedError error: Error,
+                        for request: URLRequest,
+                        response: URLResponse?,
+                        continueHandler: @escaping (_ action: HTTPNetworkTransport.ContinueAction) -> Void)
+}
+
+public extension HTTPNetworkTransportRetryDelegate {
+  
+  func networkTransport(_ networkTransport: HTTPNetworkTransport,
+                        receivedError error: Error,
+                        for request: URLRequest,
+                        response: URLResponse?,
+                        retryHandler: @escaping (_ shouldRetry: Bool) -> Void) {
+    retryHandler(false)
+  }
+  
+  func networkTransport(_ networkTransport: HTTPNetworkTransport,
+                        receivedError error: Error,
+                        for request: URLRequest,
+                        response: URLResponse?,
+                        continueHandler: @escaping (_ action: HTTPNetworkTransport.ContinueAction) -> Void) {
+    self.networkTransport(networkTransport, receivedError: error, for: request, response: response) { (_ shouldRetry: Bool) in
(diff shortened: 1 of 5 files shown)


Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -59,11 +59,52 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate
   ///   - request: The URLRequest which generated the error
   ///   - response: [Optional] Any response received when the error was generated
   ///   - retryHandler: A closure indicating whether the operation should be retried. Asyncrhonous to allow for re-authentication or other async operations to complete.
+  @available(*, deprecated, message: "Use networkTransport(_:receivedError:for:,response:continueHandler:) instead")
```
designatednerd, line 62: So this is going to be a bit of a mess because even if we deprecate this method, people are still going to have the deprecation warning on the default implementation below. 

I'm wondering if this might be a "Let's take advantage of the fact that we're not at 1.0" moment to just straight up replace this method.

Sources/Apollo/HTTPNetworkTransport.swift
```diff
@@ -59,11 +59,52 @@ public protocol HTTPNetworkTransportRetryDelegate: HTTPNetworkTransportDelegate
...
   func networkTransport(_ networkTransport: HTTPNetworkTransport,
                         receivedError error: Error,
                         for request: URLRequest,
                         response: URLResponse?,
                         retryHandler: @escaping (_ shouldRetry: Bool) -> Void)
+  
+  /// Called when an error has been received after a request has been sent to the server to see if an operation should be retried or not.
+  /// NOTE: Don't just call the `retryHandler` with `true` all the time, or you can potentially wind up in an infinite loop of errors
```
designatednerd, line 70: This should be updated with the  appropriate `continueAction` that you shouldn't just call

Sources/Apollo/HTTPNetworkTransport.swift
RolandasRazma: hmm, I don't see deprecation warning unless I implement method in my delegate. Or I'm missing something?

Sources/Apollo/HTTPNetworkTransport.swift
RolandasRazma: done

RolandasRazma:
> Can you please add a test validating that your feature request is working the way you think it should - that is, on a retry, you get the correct error?

added test. Not rly in style you doing, so let me know if you want me to change it

Sources/Apollo/HTTPNetworkTransport.swift
designatednerd: Hm...in theory this should still cause a deprecation warning where it's in the same file but maybe that's been disabled to prevent exactly this scenario...

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -190,6 +190,65 @@ class HTTPTransportTests: XCTestCase {
...
+    }
+    
+    mockRetryDelegate.mockError = MockError.customError
+    
+    let expectationCustomError = self.expectation(description: "Send operation completed")
+    
+    let _ = transport.send(operation: HeroNameQuery()) { result in
+      switch result {
+      case .success:
+        XCTAssertTrue(false)
```
designatednerd, line 239: This should probably just be `XCTFail` with an explanation of why it's failing as the message

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -190,6 +190,65 @@ class HTTPTransportTests: XCTestCase {
...
+        expectationErrorResponse.fulfill()
+      case .failure(let error):
+        XCTAssertTrue(error is GraphQLHTTPResponseError)
+        expectationErrorResponse.fulfill()
+        break
```
designatednerd, line 222: the break here and below are unnecessary

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -190,6 +190,65 @@ class HTTPTransportTests: XCTestCase {
...
+    
+    let transport = HTTPNetworkTransport(url: URL(string: "http://localhost:8080/graphql_non_existant")!)
+    transport.delegate = mockRetryDelegate
+    
+    let expectationErrorResponse = self.expectation(description: "Send operation completed")
+    
+    let _ = transport.send(operation: HeroNameQuery()) { result in
+      switch result {
+      case .success:
+        XCTAssertTrue(false)
```
designatednerd, line 217: Samesies on `XCTFail`

Tests/ApolloTests/HTTPTransportTests.swift
```diff
@@ -190,6 +190,65 @@ class HTTPTransportTests: XCTestCase {
     self.wait(for: [expectation], timeout: 10)
   }
   
+  func testRetryDelegateReturnsCorrectError() throws {
```
designatednerd, line 193: I think this should probably be split into two tests: One that tests the scenario where you want the returned error and one that tests the scenario where you want the retry error.

RolandasRazma: There is "retry" test already (StarWarsServerTests:28, HTTPTransportTests:412). The only test that was needed is returning correct error.

designatednerd: Ah, I phrased that poorly: One that tests the scenario where you want the original error received from the first call to the server returned, and one that tests the scenario where you want the error that was returned during the login instead. Does that make more sense?

PR #1135, opened 2020-04-09
Add instructions to import KeychainSwift

ck3g:
Adds instructions to import `KeychainSwift` library to the `Network.swift`.

Some thoughts: Not sure if that's the best place for that instruction. It might seem that moves focus from the `networkTransport` method.


designatednerd:
@ck3g Thanks a lot for opening these PRs for stuff that's not quite working right!

PR #1136, opened 2020-04-09
Update instructions to add cancelTrip method

ck3g:
The `cancelTrip` didn't exist before, so instead of update, we need to create it.


ck3g:
The failing test doesn't seem to be related.

designatednerd:
Yeah...there's a couple that are tempermental. Rerunning that test then I'll merge.

PR #1141, opened 2020-04-10
Merges two parts of cancelTrip method together

ck3g:
The instructions for `cancelTrip` method consist of two parts. The first part has a `// TODO` placeholder, which a user needs to replace in the next step.

It may make sense to merge those two parts.

Also fixes https://github.com/apollographql/apollo-ios/pull/1137


designatednerd:
This one I'm going to decline - This was split up on purpose to keep the amount of code in each block smaller. Thank you for the PR though!

PR #1143, opened 2020-04-10
Make BookTrip mutation example consistent

ck3g:
Changes `$tripIDs` to `$id` in [BookTrip mutation](https://www.apollographql.com/docs/ios/tutorial/tutorial-mutations/#add-the-booktrip-mutation) to keep code examples and screenshots consistent


docs/source/tutorial/tutorial-mutations.md
```diff
@@ -130,7 +130,7 @@ This is helpful because the Swift code generation will now generate a method tha
 In the `Query Variables` section of GraphiQL, update variables to use `tripID` as the key, and remove the array brackets from around the identifier:
```
designatednerd, line 130: looks like this needs to update `tripID` as well

ck3g: Right! Good catch. I've pushed new changes

PR #1144, opened 2020-04-10
ApolloSQLite -> Fix deserialize method for custom scalars

ppowers10:
Fix:

[See Spectrum conversation regarding this issue](https://spectrum.chat/apollo/apollo-ios/custom-scalar-field-failing-when-deserialize-fieldjsonvalue-jsonvalue~7ce9fa3a-a93e-4d8b-9c78-0d300196fa92)

Changes:

- Removed `invalidRecordValue` as it is no longer used anywhere
- Updated `deserialize` method to return `fieldJSONValue` when a lack of a `$reference` value is found.  This allows custom scalar types to return properly rather than throwing the error.
- Updated tests to remove any `invalidRecordValue` thrown errors from `SQLiteNormalizedCacheError ` as they are not used anymore.  Instead, this test fails in the same manner that `InMemoryNormalizedCache` due to a `JSONDecodingError`.

- [x] Tests Passing


Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -298,26 +298,15 @@ class LoadQueryFromStoreTests: XCTestCase, CacheTesting {
...
-            XCTFail("Incorrect error type for underlying: \(graphQLError.underlying)")
+          guard let decodingError = graphQLError.underlying as? JSONDecodingError else {
+            XCTFail("Invalid error type.")
+            return
+          }
+          switch decodingError {
+          case .couldNotConvert(value: _, to: _):
+            break
+          case .missingValue, .nullValue, .wrongType:
+            XCTFail("Invalid error type.")
```
designatednerd, line 309: I think you can just do this and have this be a lot simpler: 

```swift
switch graphQLError.underlying {
case JSONDecodingError.couldNotConvert(value: _, to: _)?:
  break
default: 
   XCTFail("Invalid error type")
}

ppowers10: @designatednerd - The above change was made, but it looks like CircleCI is complaining about `ci/circleci: Apollo iOS 12.4` build.  I ran the tests locally and didn't see any issues.  I'm unable to trigger the build process again to see if it might pass and it was a glitch.  My previous code passed all the tests on all the CircleCI checks.

designatednerd: Yeah there's one test that gets flaky a decent amount - I'll kick the build

designatednerd:
(approved, obviously)

gsabran:
👍

gsabran:
👍

PR #1148, opened 2020-04-11
weak self in FetchQueryOperation

RolandasRazma:


RolandasRazma:
also added weak to not retain client

Sources/Apollo/ApolloClient.swift
```diff
@@ -281,9 +284,13 @@ private final class FetchQueryOperation<Query: GraphQLQuery>: AsynchronousOperat
...
 
   func fetchFromNetwork() {
+    networkTask?.cancel()
     networkTask = client?.send(operation: query,
                                shouldPublishResultToStore: true,
-                               context: context) { result in
+                               context: context) { [weak self] result in
+      guard let self = self else {
+        return
+      }
```
designatednerd, line 293: I don't think this and the above `weak self` guards are necessary because the operation will be held on to by the operation queue, so it can't get deallocated before it finishes or cancels (and below we're cancelling the network task if the operation is cancelled).

RolandasRazma: but don't you create circular retain here? FetchQueryOperation -> client -> FetchQueryOperation

designatednerd: 🤔Yeah I suppose that would create a retain cycle if there's a cancellation. Fair point!

designatednerd:
OK fair point on the retain cycle but I stand by my point about these operations being single-serve.

RolandasRazma:
ok, fair point. Whenever I see something reassigned without cleanup red light comes on :)

designatednerd:
Totally understand, that's normally a good instinct.

PR #1153, opened 2020-04-14
Add support for multiple headers

LachlanMcCulloch:


.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
```
designatednerd, line 7: I don't believe this needs to be committed

Tests/ApolloCodegenTests/ApolloSchemaTests.swift
```diff
@@ -54,7 +57,8 @@ class ApolloSchemaTests: XCTestCase {
         "--endpoint=http://localhost:8080/graphql",
         "--key=\(apiKey)",
         "'\(expectedOutputURL.path)'",
-        "--header='\(header)'"
+        "--header='\(headers.first!)'",
+        "--header='\(headers.last!)'"
```
designatednerd, line 61: I'd prefer these not have force-unwraps - maybe declare it as two vars and then shove those into the header array. However, this is something I can fix post-merge if you prefer.

Tests/ApolloTests/VersionNumberTests.swift
```diff
@@ -4,6 +4,6 @@ import XCTest
 class VersionNumberTests: XCTestCase {
   func testVersionNumberExists() {
     // It would be the first 2 digits of version number like `0.19`.
-    XCTAssertGreaterThanOrEqual(ApolloVersionNumber, 0)
+//    XCTAssertGreaterThanOrEqual(ApolloVersionNumber, 0)
```
designatednerd, line 7: This should not be committed - I know it doesn't run right if you're not using the schemes, but I'd still prefer to run it with the schemes.

.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
LachlanMcCulloch: Have removed - I'm unfamiliar with SPM 😅

Tests/ApolloCodegenTests/ApolloSchemaTests.swift
LachlanMcCulloch: Too easy - have created two separate variables labelled `firstHeader` and `lastHeader` for clarity and to remove the force unwraps

Tests/ApolloTests/VersionNumberTests.swift
LachlanMcCulloch: Yep, this was my bad, I meant to uncomment and should be back to normal now

Issue #1155, opened 2020-04-14
Cache write on watcher dependent key can cancel server fetch


flobories:
I recently updated my project to `0.25` from `0.20` and observed some unexpected behavior in request cancelations.

**Intended outcome**
Network requests on watched queries are not interrupted by cache updates. 

**Actual outcome**
If a cache update comes in while a server fetch is in flight, the server fetch is canceled, and a `.failure` is sent to the completion handler.

**How to reproduce the issue**
- Watch a query with a cache policy that includes a server fetch
- While the server fetch is in flight, modify a dependent key in the cache in a read/write transaction
- The fetch is canceled

I believe this behavior was introduced by #1012 in `0.22`
It appears that if a dependent key of the watcher is written to, [a cache fetch is triggered](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/GraphQLQueryWatcher.swift#L69), effectively [canceling](https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/GraphQLQueryWatcher.swift#L40) the in-flight server fetch, and issuing a `.failure` to the completion handler.

Here is a concrete example of how this is playing out in my project, which features a list of chat conversations:
- If some conversations are unread, the user can navigate to a carousel of unread conversations. This list of conversation was previously fetched and is available in the cache.
- When the users navigates to the carousel of unread conversations:
    - `UnreadConversationQuery` is watched with `.returnCacheDataAndFetch`
    - The cached list is returned and the first conversation displayed. The server fetch is issued.
    - The first conversation marks itself as read
         - `isRead` is set to true optimistically in the cache for this conversation
    -  Since `UnreadConversationQuery` depends on this same `isRead` key, the server fetch is canceled and replaced by a cache fetch.
    - The watch result handler is called with a `.failure`

RolandasRazma:
considering https://github.com/apollographql/apollo-ios/pull/1156 was merged is this still open issue?

designatednerd:
@RolandasRazma I leave these open until the version ships, but thanks for asking!

designatednerd:
This has been released with `0.28.0`!

PR #1156, opened 2020-04-14
Do not cancel server fetch of watcher on dependent key update

flobories:
This is an attempt at solving #1155
By only canceling fetches when they are replaced by a server fetch, we are making sure that cache updates do not cancel any inflight fetches.


Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -70,3 +74,14 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
     }
   }
 }
+
+private extension CachePolicy {
+  var alwaysIncludesServerFetch: Bool {
+    switch self {
+    case .fetchIgnoringCacheCompletely, .fetchIgnoringCacheData, .returnCacheDataAndFetch:
+      return true
+    case .returnCacheDataDontFetch, .returnCacheDataElseFetch:
```
designatednerd, line 83: If you don't mind, can you put each case on its own line to match existing code? If you're tied up I can fix it post-merge.

flobories: Will do! Thank you for the reactivity!

designatednerd: Thank you for giving me something to procrastinate with 😆

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -78,9 +78,15 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
+    case .fetchIgnoringCacheCompletely:
       return true
-    case .returnCacheDataDontFetch, .returnCacheDataElseFetch:
+    case .fetchIgnoringCacheData:
+      return true
+    case .returnCacheDataAndFetch:
+      return true
+    case .returnCacheDataDontFetch:
+      return false
+    case .returnCacheDataElseFetch:
```
designatednerd, line 89: Sorry, should have been clearer here: I meant more: 

```swift
case .fetchIgnoringCacheCompletely,
     .fetchIgnoringCacheData,
     .returnCacheAndFetch:
  return true
``` 
etc

flobories: Oh ok, I was actually wondering. Give me a sec!

RolandasRazma:
hmm, not sure if this is correct solution. Wouldn't it be better to change signature to `fetch(cachePolicy: CachePolicy, cancelInFlight: Bool = true)` and do `fetch(cachePolicy: CachePolicy, cancelInFlight: false)` from cache updates?

RolandasRazma:
also this introduces bit uncertainty in `GraphQLQueryWatcher.cancel()` - what are we canceling?

RolandasRazma:
thinking about both cases above I think better way is to figure out solution where canceling fetch won't call `resultHandler` when it's done from cache updates, or maybe not call `resultHandler` at all for cancelations? - what are we gaining there?.

designatednerd:
@RolandasRazma I don't know how we'd infer the correct fetch policy when cancelling from the cache though. What if it's `.returnCacheDataAndFetch` rather than `elseFetch`? 

Note that this also just calls `cancel` straight on the `cancellable` that's tied to the network request - it doesn't have anything to do with the `cancel` on the watcher itself.

Does that make more sense?

RolandasRazma:
the problem is that 
watcher = GraphQLQueryWatcher()
watcher.fetch()
watcher.fetch()
watcher.cancel()

will have one success and one cancelation callback, and that might lead to very confusing results unless you know whole system

designatednerd:
Wait, I don't think I understand why that would happen. Can you explain: 

a) Why this would return one and one?
b) What you expect would happen instead?

RolandasRazma:
a)
watcher = GraphQLQueryWatcher(...) { result: Result<GraphQLResult<Data>, Error>
   print(result)
}
watcher.fetch() // First A
watcher.fetch() // Fetch B
watcher.cancel()

Fetch B is canceled and return error as "canceled"
Fetch A will succeed after some time (if this PR accepted) and will return data

b)
that's a good question "what is expected behaviour?". Reading code above my first thought would be "oh, so both fetched will go in parallel and "cancel" will cancel them both at the same time" (as there no way to specify witch one to cancel)
second thought would be behaviour we have now, where second fetch "replaces" first one

designatednerd:
OK, thank you very much for clarifying. @flobories What's your take on @RolandasRazma's concerns?

flobories:
I see you concern @RolandasRazma about having two different fetches in parallel, only one being cancelable because only one `fetching` reference.

The main concern I'm trying to address is a server fetch being canceled by a cache update, when I'd prefer to server fetch to reach completion uninterrupted as I did not explicitly canceled it.
Do you see a way around this problem without introducing parallel fetches in the `watcher`?

- If we are ok with 2 concurrencies, we could introduce the concept of distinct `storeUpdateFetching` and `refetching` `Cancelable`s instead of a single `fetching`, and have store updates only cancel the `storeUpdateFetching`, similary refetches would only cancel the `refetching`.
The `public cancel()` would cancel both. This could still be tricky because it would return two `canceled` errors I believe?
- If we want to keep only 1, then I believe we need to prioritize cache updates or server fetches either by dropping some or by chaining them. I feel this may be harder to nail? As a user of the library, I would expect cache updates to be "instant", so if a cache updates happen while a server fetch is in flight we'd need to
    1) Cancel the server fetch
    2) Apply the cache fetch
    3) Refetch from server once cache update is done
This would overall take more time? And lead to some cancelation errors?

Does this make sense?

RolandasRazma:
your last scenario is what I was describing, where canceling from cache hides server fetch cancelation and "user of the library" never even knows that his request was canceled and restarted.

Implementation as it stands in this PR will lead to bugs where:
You have watcher on object
You start "fetch A" for that object
You start mutation on server and fetching same object, update it in cache (from mutation)
If your "fetch A" returns later than mutation, your object in cache will be reverted to "incorrect" one (as your fetch started before mutation). It will go "old object" -> "new object" (mutation) -> "old object" ("fetch A" returns)

this requires weird internet condition, but this kind of race conditions impossible to find when you need to fix

designatednerd:
AHA, I think I found a big piece of my confusion, which is partly because I'm noodling with some stuff with a delegate-based URL session: The callback-based APIs will always puke up an error if something is cancelled, but the way I've got it set up in my delegate-based one I'm killing off completion blocks before the cancellation error hits the delegate. 

I feel like maybe we should be performing the fetch directly on the `store` that comes in, and only hitting that `fetch` that creates a `cancellable` if there's actually no results in the cache. Then we could just straight up cancel if something calls `fetch` again. 

What do you guys think about that?

designatednerd:
(FYI, I'm converting this to a draft so I don't brain-dead merge it tomorrow morning before we finish this discussion)

designatednerd:
@RolandasRazma @flobories Thoughts?

RolandasRazma:
sorry I'm not that familiar with codebase to give credible opinion

designatednerd:
@RolandasRazma  You know this part of the codebase at least as well as I do, i've been focused on other stuff 🙃

You can't actually cancel a store load, so there's nothing really *to* track. 

If the entire watcher is cancelled while the store load is happening, then the completion callback should be nilled out in the cancellation and then we wouldn't call it when the store load finishes. 

I *think* you'd only have the cache key create a new fetch if you get that dependent keys changed notification after something got cleared out - then the cache would have nothing and it'd force a re-fetch. 

Like I've said repeatedly, this is extremely complex, to the point that I've been working on this lib for 10 months and still really haven't totally wrapped my head around every possibility here. @flobories the fact that you were able to even identify this problem indicates you're probably in an ok spot to take a stab at it with feedback from me and @RolandasRazma. So yes, POC is welcome!

flobories:
Will do, thanks for the feedback @designatednerd !

flobories:
(actually tied up with other projects at work, will follow up tomorrow 🙃)

designatednerd:
@flobories Were you ever able to get another look at this?

flobories:
I have not 😬 but it’s been on my mind crushing me with guilt 🤣 
I just shipped the feature that has been getting my attention these past couple weeks and this week is my catchup on everything else one!
I’ll prioritize to move this forward before EOW.
I appreciate your patience, and thanks for the nudge!

designatednerd:
Dude, no guilt! Just checking in to see if you were still interested in moving this forward. It's a gnarly problem to solve, so I very much appreciate that you are interested in solving it rather than me 😆

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -66,7 +69,21 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
     if !dependentKeys.isDisjoint(with: changedKeys) {
-      fetch(cachePolicy: .returnCacheDataElseFetch)
+      // First, attempt to reload the query from the cache directly, in order not to interrupt any in-flight server-side fetch.
+      store.load(query: query) { [weak self] result in
+        guard let self = self else { return }
+
+        switch result {
+        case .success(let graphQLresult):
+          self.queue.async {
+            self.dependentKeys = graphQLresult.dependentKeys
```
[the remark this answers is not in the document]
designatednerd: I don't think this should be necessary - you have to call `fetch` on a watcher when you fire it up, and at that point `dependentKeys` would be set by the value returned from the `fetch` method. This would only get called if you initialized the watcher and failed to call the initial `fetch` method, which is not something we want to encourage.

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -35,11 +35,14 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
     fetch(cachePolicy: .fetchIgnoringCacheData)
   }
 
+  // Watchers always call result handlers on the main queue.
+  private let queue: DispatchQueue = .main
```
designatednerd, line 39: maybe `callbackQueue` to disambiguate it from other possible queues

flobories:
Thanks for the review @designatednerd! Just pushed the updates you requested.

designatednerd:
Looks good on my end - @RolandasRazma since you caught the issue with the original, wanna double check that I'm not missing something here?

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -66,7 +69,20 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
 
     if !dependentKeys.isDisjoint(with: changedKeys) {
-      fetch(cachePolicy: .returnCacheDataElseFetch)
+      // First, attempt to reload the query from the cache directly, in order not to interrupt any in-flight server-side fetch.
+      store.load(query: query) { [weak self] result in
+        guard let self = self else { return }
+
+        switch result {
+        case .success:
+          self.callbackQueue.async {
```
RolandasRazma, line 78: isn't this retain cycle? self.callbackQueue -> self.resultHandler

Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -66,7 +69,20 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
+            self.resultHandler(result)
+          }
+        case .failure:
+          // If the cache fetch is not successful, for instance if the data is missing, refresh from the server.
+          self.fetch(cachePolicy: .fetchIgnoringCacheData)
```
RolandasRazma, line 83: is `.fetchIgnoringCacheData` correct here? it was `.returnCacheDataElseFetch` before.
Can we get into some fetch cycle here? `fetch(cachePolicy: CachePolicy)` -> Fail -> `fetch(cachePolicy: CachePolicy)` -> Fail -> ...

designatednerd: Before it was `returnCacheDataElseFetch` because we weren't explicitly trying to load from the cache. By the point this gets hit, we've already checked the cache so `fetchIgnoringCacheData` acknowledges that.

designatednerd: Also the `fail` is for reading directly from the store, rather than for hitting the network, so that cycle should not be an issue.

Sources/Apollo/GraphQLQueryWatcher.swift
designatednerd: ahhh good catch - we're defending against self being released in the `load` closure but not on the `async` closure - @flobories do you mind updating this?

designatednerd:
OK I'm gonna go ahead and merge this - @flobories @RolandasRazma Thanks so much for all your help!

flobories:
Thanks @designatednerd for your help and patience! Feels great to reach the finish line here!

Sources/Apollo/GraphQLQueryWatcher.swift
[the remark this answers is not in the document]
teodorpenkov: Actually this is necessary. If I create a watcher on query `A` and perform another query `B` if I want to update manually `A` in the store with results from `B`  subsequent updates to the updated elements from `B` wouldn't be propagated to `A`'s listeners.

Use case pagination using apollo store.

Watch primary query, perform fetch more and update primary with nodes, page info and has next, then perform mutation in a node that came from fetch more query, primary query's listeners won't retrieve the updated result even though they should.

Issue #1117, opened 2020-04-02
SQLiteNormalizedCache encryption


pgawlowski:
I've got pretty fragile data to cache in a persitant way via grapqhl SQLiteNormalizedCache.

What are the security measures of SQLiteNormalizedCache? Is there any kind of encryption or possibility to inject middleware for ecryption/decryption? 
I was thinking about integrating iOS SQLCipher in between however can not find any information how to handle or how it is being handled right now?

designatednerd:
Anything that would be provided is provided through the [SQLite.swift](https://github.com/stephencelis/SQLite.swift) dependency, which it looks like it has [the ability to integrate with SQLCipher](https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md#using-sqliteswift-with-sqlcipher).

designatednerd:
Since I haven't heard back here in 2 weeks and this functionality is covered by `SQLite.swift`, I'm going to go ahead and close this out. Please let me know if you have further questions. Thank you!

pgawlowski:
I was messing around with SQLCipher however I failed with integration.
Are there any additional suggestions how to handle? 
Is for example `withinReadWriteTransaction` and `withinReadTransaction` valid place to encrypt/decrypt data?

designatednerd:
Unfortunately we don't have any recommendations around this - generally this kind of security is handled at the database level rather than what we're dealing with, which is once stuff has come out of the database (or before it's gone in). 

If you were to do it the way you suggest, you'd have to manually do that in every single read/write transaction, and it would most likely break a bunch of stuff like query watchers. It *could* work if you're OK with those other trade-offs, but I would strongly recommend finding something at the level of the database itself that handles this for you.

pgawlowski:
OK I've figured out sth. 
Integration actually is possible. It requires small dependency modification in podspec - SQLCipher instead of pure SQLite.
I was thinking how we could possibly integrate this in a painless way with current Apollo/SQLite implementation. 

One idea is to introduce additional way of initialization directly with DB instead of filePath.
>public init(db: Connection shouldVacuumOnClear: Bool = false)

Do you think this kind of changes might be possible? I can prepare podspec + SQLiteNormalizedCache changes. 

We can also introduce SQLiteEncryptedNormalizedCache implementation however this will be exactly the same as current implementation with 2 additional lines.
`try db.rekey(secret)`
to set passphrase and 
`try db.key(secret)`
to unlock it :) 

[...]

PR #1162, opened 2020-04-17
SQLCipher support

pgawlowski:
Hello there :) 
Based on this discussion https://github.com/apollographql/apollo-ios/issues/1117
I thought that I will give it a shot and try to introduce SQLCipher integration.

Usage is quite simple. 
`pod 'Apollo/SQLiteCipher'`

      private lazy var store: ApolloStore = {
        do {
            let db = try Connection(.uri(temporarySQLiteFileURL().absoluteString), readonly: false)
            try db.rekey("secret") //To set DB encryption secret
            try db.key("secret") //To open DB
            
            let cache = try SQLiteNormalizedCache(db: db)
            return ApolloStore(cache: cache)
        } catch {
            print(error)
        }
        
        return ApolloStore(cache: InMemoryNormalizedCache())
    }()

Apollo.podspec modification is a result of this issue:
https://github.com/stephencelis/SQLite.swift/issues/570
It is impossible to integrate those two: 
`pod 'Apollo/SQLite'`
`pod 'SQLite.swift/SQLCipher'`
Cause the result is normal SQLiteDB which is unable to use Cipher encryption. Probably it is conflicting somehow with normal SQLite implementation.

I did not introduce SPM packged denepdency - not sure how to handle this part to be honest.

### DIFF
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -35,6 +35,7 @@ Pod::Spec.new do |s|
     ss.dependency 'SQLite.swift', '~>0.12.2'
   end
   
+  # Cache persistance storage mechanism extended with encryption mechanism
   s.subspec 'SQLiteCipher' do |ss|
     ss.source_files = 'Sources/ApolloSQLite/*.swift'
     ss.dependency 'Apollo/Core'
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -34,6 +34,13 @@ Pod::Spec.new do |s|
     ss.dependency 'Apollo/Core'
     ss.dependency 'SQLite.swift', '~>0.12.2'
   end
+  
+  s.subspec 'SQLiteCipher' do |ss|
+    ss.source_files = 'Sources/ApolloSQLite/*.swift'
+    ss.dependency 'Apollo/Core'
+    ss.dependency 'SQLCipher', '~>4.2.0'
+    ss.dependency 'SQLite.swift/SQLCipher', '> 0.12.0'
+  end
 
   # Websocket and subscription support based on Starscream
   s.subspec 'WebSocket' do |ss|
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -31,6 +31,12 @@ public final class SQLiteNormalizedCache {
     try self.createTableIfNeeded()
   }
 
+  public init(db: Connection, shouldVacuumOnClear: Bool = false) throws {
+    self.shouldVacuumOnClear = shouldVacuumOnClear
+    self.db = db
+    try self.createTableIfNeeded()
+  }
+  
   private func recordCacheKey(forFieldCacheKey fieldCacheKey: CacheKey) -> CacheKey {
     let components = fieldCacheKey.components(separatedBy: ".")
     var updatedComponents = [String]()


Apollo.podspec
```diff
@@ -34,6 +34,14 @@ Pod::Spec.new do |s|
...
     ss.dependency 'SQLite.swift', '~>0.12.2'
   end
+  
+  # Cache persistance storage mechanism extended with encryption mechanism
+  s.subspec 'SQLiteCipher' do |ss|
+    ss.source_files = 'Sources/ApolloSQLite/*.swift'
+    ss.dependency 'Apollo/Core'
+    ss.dependency 'SQLCipher', '~>4.2.0'
+    ss.dependency 'SQLite.swift/SQLCipher', '> 0.12.0'
+  end
```
designatednerd, line 44: We're not actually using anything from SQLite cipher inside our library, so I don't think this is necessary (and would add a couple of dependencies that could cause conflicts.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -31,6 +31,12 @@ public final class SQLiteNormalizedCache {
     try self.createTableIfNeeded()
   }
 
+  public init(db: Connection, shouldVacuumOnClear: Bool = false) throws {
+    self.shouldVacuumOnClear = shouldVacuumOnClear
+    self.db = db
+    try self.createTableIfNeeded()
+  }
```
designatednerd, line 38: This seems like a reasonable addition in terms of allowing people to use SQLCipher or other extensions to SQLite.swift, though, so I do think we should keep this. Can you add some documentation to this initializer please?

Apollo.podspec
pgawlowski: The main problem is - it seems to be impossible to use it together with Apollo/SQLite. 
`SQLite.swift/SQLCipher` - requires only this particular SQLite dependency.
Mixing it together with`ss.dependency 'SQLite.swift', '~>0.12.2'` results in not working SQLCipher layer. It is not going to cause any error, crash or compilation warning. Just DB is not going to encrypt at all.

That is why I prepared separate dependency.  
Please check this comment and the answer:
https://github.com/stephencelis/SQLite.swift/issues/570#issuecomment-268530061
This is the exact situation. 

I was already trying to use 
`pod Apollo/SQLite `
`pod SQLCipher`
together with this init
```
    self.shouldVacuumOnClear = shouldVacuumOnClear
    self.db = db
    try self.createTableIfNeeded()
  }
```

and DB is not encrypting. 

Any thoughts how to resolve?

designatednerd: Oof, looking at [the SQLitePodspec](https://github.com/stephencelis/SQLite.swift/blob/master/SQLite.swift.podspec#L70) it looks like there's some additional `xcconfig`s that get set up in order to make that work. Looking at the [`README` for FMDB, another iOS SQLite lib](https://github.com/ccgus/fmdb#cocoapods), I see a similar caveat about how to set up the podspec. 

Looking at the tutorial [on SQLCipher](https://www.zetetic.net/sqlcipher/ios-tutorial/), there's a fairly blunt caveat about only using their SQLite 3: 
 
Screen Shot 2020-04-17 at 3 27 05 PM

Neither SQLite.swift nor FMDB support using this through anything other than CocoaPods, which is a big old red flag for me in terms of maintaining support on the Apollo side, which would be necessary if we were to add this to the library. 

Honestly, at this point, it feels like we're pretzeling ourselves to make this work in a way that just doesn't make sense for the overwhelming majority of users. This is a major reason why `NormalizedCache` is a protocol: so that you can implement your own version of it if the one we provide is insufficient.

designatednerd:
I'm gonna close this one out - I went into detail [in this comment](https://github.com/apollographql/apollo-ios/pull/1162#discussion_r410458686) about some of my concerns but ultimately, this isn't something we have the bandwidth to be supporting in the main library due to the complexity of managing this dependency. 

Probably the easiest way to make this work would be to: 
- Not use the Apollo/SQLite subspec
- Use the SQLite.swift/SQLCipher subspec in your own project
- Create your own `NormalizedCache` implementation that's copy-pasted from what you have here. 

It's not a great answer, but it looks like it's not any easier with FMDB or even just using SQLCipher directly.

mdarnall:
@designatednerd Hi, given the outcome of this I'm wondering if we could still put in the change outlined here to pass in the encrypted DB Connection object (https://github.com/apollographql/apollo-ios/pull/1162#discussion_r410367735)

designatednerd:
Yes if you'd like to open a separate PR with that change, that'd be lovely.

Issue #1170, opened 2020-04-24
Suggestion to refactor `apollo_` prefixed methods in ApolloCodegenLib


yuzushioh:
Hi @designatednerd 

I have been using apollo-ios for about 6 months and I love it. I just migrated the code generation script from apollo-tooling to Swift scripting using ApolloCodegenLib.

While I was doing a work, I noticed some of framework's methods have `apollo_` prefix like `apollo_folderExists`. I think it is there to make it more recognizable for users that those are only used for apollo's usecases. 

RxSwift used to use the same pattern like `rx_text` but it has been changed to `rx.text` by using its own extension. I'm thinking It will be a good change in Apollo too. I can do the work. What do you think?

designatednerd:
Yeah the namespacing is a) To make it clearer that this is something that Apollo added and b) To prevent accidental runtime naming clashes on anything that has to go through the Obj-C runtime (which most stuff descended from `NSObject` has to do). 

I was trying to remember how Rx did this and I found the [`Reactive` type in their codebase](https://github.com/ReactiveX/RxSwift/blob/9b31a15520306b073cb9d46456f64826c1d6dcab/RxSwift/Reactive.swift) - it's not quite as involved as I remember it being. 

I'm a little torn - i think the underscore methods make it easier to discover what `apollo_blah` methods/properties exist in one go for autocomplete, but it does read nicer to have `apollo.blah`, and it seems like the maintenance burden of switching to this style shouldn't be too terrible.

I'd say go ahead and open a PR - I think I'd still like to use `apollo` for the namespace but I think otherwise the `Reactive` stuff seems like it provides a good template to use.

yuzushioh:
@designatednerd Hi, is there any reason why the indentation width is set for 2 tab and 2 space? I personally think it's more common to use 4 for indentation width.

designatednerd:
2-space indentation was set up well before I started on this project. To be honest, I don't really care enough either way to change it, so I'm going with what we were already using.

yuzushioh:
Are you going to use any linting framework like SwiftLint in this project at some point?

designatednerd:
At some point I'd like to, but it's not a super-high priority since at this point I'm the ~benevolent dictator~ primary maintainer of the code. What, in particular, were you wondering about trying to enforce?

yuzushioh:
[...]

PR #1183, opened 2020-05-03
Add apollo custom extension to replace `apollo_` prefix

yuzushioh:
Added `ApolloExtension` and replace `apollo_` prefix by `apollo.` ref: #1170

### DIFF
--- a/Sources/ApolloCodegenLib/OptionalBoolean.swift
+++ b/Sources/ApolloCodegenLib/OptionalBoolean.swift
@@ -1,7 +1,8 @@
 import Foundation
 
 extension Optional where Wrapped == Bool {
-  var valueOrFalseIfNone: Bool {
+  /// It returns false if it is called on `nil`
+  var boolValue: Bool {
     switch self {
     case .none:
       return false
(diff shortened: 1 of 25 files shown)


yuzushioh:
@designatednerd `testWatchedQueryGetsUpdatedWithListReorderingFromOtherQuery ` is failing for some reason. I am not sure if it's because of my change

designatednerd:
@yuzushioh Prooooobably not, that test is super flaky, I'll kick it

Sources/ApolloCodegenLib/OptionalBoolean.swift
```diff
@@ -1,8 +1,7 @@
 import Foundation
 
 extension Optional where Wrapped == Bool {
-  
-  var apollo_boolValue: Bool {
+  var valueOrFalseIfNone: Bool {
```
designatednerd, line 4: I'd still like to keep this as `boolValue` - ultimately it behaves similarly to `[NSNumber boolValue]` in that if you call it on `nil` it returns false.

yuzushioh: OK. I changed it to `boolValue` with a comment on it.

PR #1186, opened 2020-05-04
Allow to clear the cache synchronously

gsabran:
I don't really understand why the existing methods have completion blocks while all the work is done synchronously on the caller thread. This gives the impression that the work is being offloaded to another queue which is not the case. I guess the protocol should not be concerned with the existing implementations. That said I've not replaced the existing methods for backward compatibility.

I had a need for synchronous cache clearance to ensure that everything is done right when the user logs out.

### DIFF
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -32,9 +32,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
 
   public func clear(callbackQueue: DispatchQueue?,
                     completion: ((Result<Void, Error>) -> Void)?) {
-    self.recordsLock.lock()
-    self.records.clear()
-    self.recordsLock.unlock()
+    clearImmediately()
 
     guard let completion = completion else {
       return
@@ -43,5 +41,11 @@ public final class InMemoryNormalizedCache: NormalizedCache {
     DispatchQueue.apollo_returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(()))
+  }
+
+  public func clearImmediately() {
+    self.recordsLock.lock()
+    self.records.clear()
+    self.recordsLock.unlock()
   }
 }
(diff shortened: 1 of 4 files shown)


designatednerd:
@gsabran This is probably a couple things: 

1) Something of a vestigial tail from when this returned a `Promise`
2) An acknowledgement that not all implementations of `NormalizedCache` are going to want to have this operation be synchronous. We are using this asynchronously in our [`BatchedLoadTests`](https://github.com/apollographql/apollo-ios/blob/master/Tests/ApolloTests/BatchedLoadTests.swift), for example, but other client-side implementations may not want this to be sync. 

I'm actually kind of surprised to see we're *not* firing this off to the background for the SQLite implementation, that seems a bit like a better idea, but maybe it's doing that under the hood? I haven't dug deep enough to see quite yet.

gsabran:
Yes I was surprised as well. I was first  synchronously waiting on this call to complete, before realizing that there was no async work done.

designatednerd:
`❌  /Users/distiller/project/Tests/ApolloTests/BatchedLoadTests.swift:6:21: type 'MockBatchedNormalizedCache' does not conform to protocol 'NormalizedCache'` - looks like that needs an update.

gsabran:
Yeah, I'm just confused why that is as I updated this class.

designatednerd:
Looks like you'll need to try again 🙃

gsabran:
Thanks!

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -32,9 +32,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
 
   public func clear(callbackQueue: DispatchQueue?,
```
RolandasRazma, line 33: IMO we should make handler non optional here, otherwise its easy to miss that its asynchronous operation

designatednerd: Eh, I can still see a case for people wanting it to fire off in a non-blocking way but not wanting to pass in a handler, I don't think it's worth it.

PR #1194, opened 2020-05-06
Swift Codegen: Input Objects + Updated Types

designatednerd:
In This PR: 
- Switches `ASTForthcomingVariableType` to be the final version, `ASTVariableType`.
- Updates existing code for changes to the JSON format
- Adds a new `ApolloCore` library for types that need to be shared between the main lib and the codegen lib
- Adds a `GraphQLOptional` type to the core library and uses it within the updated code generation
- Actually generates code for input objects
- Adds documentation to initializer structs


designatednerd:
(Note that a huge chunk of the changeset here is changes to `API.json` caused by updates to the CLI)

Issue #1214, opened 2020-05-18
ApolloCodegenLib doesn't include an equivalent option for customScalarsPrefix


CraigSiemens:
I'm trying out the codeine using SPM and I've found that the flag `--customScalarsPrefix` that we've been using is not available. 

We use the `--passthroughCustomScalars` flag to define out own types for the scalars, and `--customScalarsPrefix="Scalar."` to prevent name conflicts between scalars and Foundation types. For example `Scalar.Date` is using a custom struct for year, month, day and shouldn't conflict with `Foundation.Date`

designatednerd:
Ah yeah, will get that added in

designatednerd:
@CraigSiemens PR #1215 is up

PR #1215, opened 2020-05-19
Add missing custom scalars prefix to Swift codegen

designatednerd:
Addresses #1214.


CraigSiemens:
I took a stab at making this change as well, though I took a slightly different approach of adding an enum for the `CustomScalarFormat`. I should have posted in the issue that I'd try to fix it. I'm not sure wether to make a PR for it now or not. You can take a look at my change if you'd like.

https://github.com/CraigSiemens/apollo-ios/tree/custom-scalars-prefix

designatednerd:
Closing in favor of #1216.

PR #1216, opened 2020-05-19
Added option to generate a custom scalar with a prefix.

### LINKED ISSUE #1214: ApolloCodegenLib doesn't include an equivalent option for customScalarsPrefix

CraigSiemens:
Fixes #1214 

Added an option to `ApolloCodegenOptions` to take a `CustomScalarFormat` enum which can be either `default`, `passthrough` or `passthroughWithPrefix(String)`

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
@@ -55,7 +55,7 @@ public struct ApolloCodegenOptions {
   /// Enum to select how to handle properties using a custom scalar from the schema.
   public enum CustomScalarFormat: Equatable {
     /// Uses a default type instead of a custom scalar.
-    case `default`
+    case none
     /// Use your own types for custom scalars.
     case passthrough
     /// Use your own types for custom scalars with a prefix.
@@ -102,7 +102,7 @@ public struct ApolloCodegenOptions {
               only: URL? = nil,
               operationIDsURL: URL? = nil,
               outputFormat: OutputFormat,
-              customScalarFormat: CustomScalarFormat = .default,
+              customScalarFormat: CustomScalarFormat = .none,
               suppressSwiftMultilineStringLiterals: Bool = false,
               urlToSchemaFile: URL,
               downloadTimeout: Double = 30.0) {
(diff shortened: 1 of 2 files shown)


Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -169,8 +179,13 @@ public struct ApolloCodegenOptions {
...
     }
     
-    if self.passthroughCustomScalars {
+    switch customScalarFormat {
+    case .default:
+      break
+    case .passthrough:
       arguments.append("--passthroughCustomScalars")
+    case .passthroughWithPrefix(let prefix):
+      arguments.append("--customScalarsPrefix='\(prefix)'")
```
designatednerd, line 188: I think this needs to have both `--passthroughCustomScalars` and `--customScalarPrefix`, doesn't it?

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -52,6 +52,16 @@ public struct ApolloCodegenOptions {
     }
   }
   
+  /// Enum to select how to handle properties using a custom scalar from the schema.
+  public enum CustomScalarFormat: Equatable {
+    /// Uses a default type instead of a custom scalar.
+    case `default`
```
designatednerd, line 58: I would call this `none` rather than `default`, since it's applying no formatting to custom scalars.

designatednerd:
The test failure is a known flaky test, FYI

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
CraigSiemens: I took a look at Apoll-tooling and it sets `passthroughCustomScalars` if `customScalarsPrefix` is set.
https://github.com/apollographql/apollo-tooling/blob/aed906a0f29f9e7f801484551c21e1793f6cb0d0/packages/apollo/src/commands/client/codegen.ts#L214

I also ran it manually and the behaviour matched. I can add the flag back in if you want it to be explicit.

designatednerd: [It's not guaranteed in the README](https://github.com/apollographql/apollo-tooling#apollo-clientcodegen-output) and I don't see anyplace where that's being double-checked in the tests - I'd rather go belt-and-suspenders here for now

Issue #1178, opened 2020-05-01
Update auth header on Websocket connection, automatic reconnect of subscription?


gazzer82:
I have this working in a project to a Hasura backend using subscriptions. I am authenticating by adding an ID Token from google auth to the connectingPayload.

The validity of the ID token from google is 1 hour, and I am trying to honour that in Hasura with only allowing the connection to remain valid for the remaining validity of the token when received.

Unfortunately that means I need to change the token in the payload when it has expired, and reconnect the web socket.

I am currently thinking the only way to do this is to create a new WebSocketTransport in

`func webSocketTransport(_ webSocketTransport: WebSocketTransport, didDisconnectWithError error:Error?)`

But the problem is I have some active subscriptions going on when this happens, and they don't reconnect unless manually made to do so.

Am I missing something here, or is their no easy/clean way way to handle the currently in the framework?

Thanks

Gareth

designatednerd:
Yeah - as I understand it, this is a problem with short-lived auth tokens and web sockets in general, since there's not really a way to constantly re-authenticate. 

We do have a `closeConnection()` method that closes out the connection, but that would require you to manually resubscribe to everything, which leaves you where you are now.

I'm going to tag in @fassko here, who's contributed to both `ApolloWebSocket` and the underlying `Starscream` library, and is wayyyy more familiar with WebSockets than I am: Kristaps, have you got any suggestions for Gareth on what the best way to handle this would be? Is there anything Apollo can do to make this easier for our clients or is it more of a conceptual difference between an HTTP connection and a WebSocket connection?

gazzer82:
Thanks @designatednerd for that, I have had a dig through the framework, and adding this to, and calling when the token needs to be refreshed seems to work in initial testing. It seems a bit hack, but by doing it this way calling connect re-starts the subscriptions that where active when disconnect it called it seems.

in WebSocketTransport.swift
```
public func updateHeaderValue(value: String?, for field: String) {
        self.websocket.request.setValue(value ?? "", forHTTPHeaderField: field)
        self.websocket.disconnect()
        self.websocket.connect()
    }
```

Now, I could be missing something very bad I'm doing here, so please let me know if I am?

Cheers

Gareth

designatednerd:
I don't believe there's anything *bad*, but there might be a race condition since the `webSocketDidDisconnect` delegate method automatically reconnects if the option to auto-reconnect is on, since that will automatically call `connect` after whatever the `reconnectInterval` is. 

[...]

PR #1224, opened 2020-05-22
Update WebSockTransport request header and connectingPayload

yootsubo:
Addresses #1178 

In addition, added a similar functionality to update the `connectingPayload`.
This is to handle similar use case to #1178 but for those who use the `connectingPayload` to send the auth token (what my team is currently doing).


designatednerd:
Awesome - any chance you've got a minute to throw some tests on for this?

gazzer82:
Thanks for doing this, got swamped the last few days!

yootsubo:
@designatednerd Tests added! Let me know if theres anything else :)

PR #1228, opened 2020-05-27
Remove unnecessary annotation in tutorial

nugmanoff:


designatednerd:
Hi @nugmanoff - this is actually necessary for [the rendered version of the tutorial](https://www.apollographql.com/docs/ios/tutorial/tutorial-detail-view/) to be able to display an expandable panel for setting up the detail UI - however, it looks like that's borked right now, so I'll ping our docs team and let them know. 

Gonna close this because it needs to stay there but thank you for the heads up that this isn't working!

PR #1242, opened 2020-06-04
Add code generation for interface and union enums

designatednerd:
NOTE: Draft until the JSON here gets finalized.

In Swift Codegen, we're going to be using enums to represent the possible types of unions and interfaces. This PR takes information returned from the server about what interface and union types are used and uses them to generate those enums. 

Also threw in an additional test on regular enums since we didn't have one covering modifiers.

### DIFF
--- a/Sources/ApolloCodegenLib/ASTOutput.swift
+++ b/Sources/ApolloCodegenLib/ASTOutput.swift
@@ -8,6 +8,12 @@ struct ASTOutput: Codable, Equatable {
   /// An array of all fragments to generate code for.
   let fragments: [ASTFragment]
   
-  /// An array of "all" types used <-- TODO: Figure out why some are not in here
+  /// An array of all input types used
   let typesUsed: [ASTTypeUsed]
+  
+  /// An array of Union types used
+  let unionTypes: [ASTUnionType]
+  
+  /// An array of Interface types used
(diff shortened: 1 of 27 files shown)


Sources/ApolloCodegenLib/ASTOutput.swift
```diff
@@ -8,6 +8,12 @@ struct ASTOutput: Codable, Equatable {
   /// An array of all fragments to generate code for.
   let fragments: [ASTFragment]
   
-  /// An array of "all" types used <-- TODO: Figure out why some are not in here
```
amadeu01, line 11: Sorry the ignorance, but have you figured it out why some of the types weren't there before?

designatednerd: Yes, excellent question! 

The answer is that `typesUsed` was designed to be only for _Input_ types, though it was not really documented as such. Unions and interfaces were not included, and I couldn't figure out why when I initially started working through the JSON, which is why I left myself this `TODO`. 

In an upcoming version of code generation, union and interface types will be included - this PR works off some temporary JSON to support generating code for that.

designatednerd:
Just a quick note that things are still being finalized with this, I'll be picking back up with the codegen once this is merged.

designatednerd:
OK! We're *finally* in business on this. Merging away!

PR #1262, opened 2020-06-17
Avoid crash when evaluating optional variables

### LINKED ISSUE #1255: Fatal error: Optional is only JSONEncodable if Wrapped is
Hey hey. 
Getting this error: 
```
Fatal error: Optional is only JSONEncodable if Wrapped is: file .../apollo-ios/Sources/Apollo/JSONStandardTypeConversions.swift, line 109
```
Query which creates the situation:
```
query UncategorizedTransactions($fromDate: Timestamp!, $toDate: Timestamp!) {
  me {
    transactionsConnection(first: 0, filters: {
      fromDate: $fromDate,
      toDate: $toDate,
      categoryId: null
    }) {
      totalCount
    }
  }
}
```
Example of playground response
```
{
  "data": {
    "me": {
      "transactionsConnection": {
        "totalCount": 0
      }
    }
  },
  "extensions": {
    "cacheControl": {
      "version": 1,
      "hints": [
        {
          "path": [
            "me"
          ],
          "maxAge": 0
        },
        {
          "path": [
            "me",
            "transactionsConnection"
          ],
          "maxAge": 0
        }
      ]
    }
  }
}
```
It doesn't crash if the query is like so:
```
query UncategorizedTransactions {
  me {
    transactionsConnection(first: 0, filters: {
      fromDate: 1524379940,
      toDate: 1524379940,
      categoryId: null
    }) {
      totalCount
    }
  }
}
```
Also it doesn't crash if the query is this:
```
query UncategorizedTransactions($dateFrom: Timestamp!, $dateTo: Timestamp!) {
  me {
    transactionsConnection(first: 0, filters: {
      fromDate: $dateFrom,
      toDate: $dateTo,
      categoryId: "someStringValue"
    }) {
      totalCount
    }
  }
}
```

martijnwalraven:
This fixes #1255 by adding an extension method to `Optional` that implements `GraphQLInputValue.evaluate(with:)` by invoking that same method on the wrapped value. Without it, we'd be treating an optional `GraphQLVariable` as an optional `JSONEncodable`, which leads to a runtime crash because `GraphQLVariable` does not in fact conform to `JSONEncodable`.

All this should really be cleaned up by adopting conditional conformances (which weren't available when this was originally written), but that turns out to lead to quite the rabbit hole of issues so that'll have to wait.

### DIFF
--- a/Sources/Apollo/GraphQLInputValue.swift
+++ b/Sources/Apollo/GraphQLInputValue.swift
@@ -27,6 +27,19 @@ extension JSONEncodable {
   }
 }
 
+extension Optional: GraphQLInputValue {
+  public func evaluate(with variables: [String: JSONEncodable]?) throws -> JSONValue {
+    switch self {
+    case .none:
+      return NSNull()
+    case .some(let wrapped as GraphQLInputValue):
+      return try wrapped.evaluate(with: variables)
+    default:
+      fatalError("Optional is only GraphQLInputValue if Wrapped is")
+    }
+  }
+}
+
 extension Dictionary: GraphQLInputValue {
   public func evaluate(with variables: [String: JSONEncodable]?) throws -> JSONValue {
     return try evaluate(with: variables) as JSONObject


designatednerd:
You're the best @martijnwalraven!

PR #1263, opened 2020-06-17
Bring Swift Scripting out of Beta

designatednerd:
In this PR:
- Promoted swift scripting to the final version
- Made some edits to make the whole Swift Scripting doc more readable and put some conceptual stuff up front
- Updated a bunch of instructions to use more user-friendly extensions so users can tell what's happening more easily
- Added a full example script developers can use as a reference.
- Updated the warning on the bash script - this probably will be deprecated in the next version or two. 

Would LOVE feedback on this - I think the scripting stuff is now to a point of maturity that I can rip the beta tag off it, and I hope the changes to the scripting doc make things clearer.

### DIFF
--- a/SwiftScripts/Sources/Codegen/main.swift
+++ b/SwiftScripts/Sources/Codegen/main.swift
@@ -8,8 +8,8 @@ let parentFolderOfScriptFile = FileFinder.findParentFolder()
 // Use that to calculate the source root
 let sourceRootURL = parentFolderOfScriptFile
     .apollo.parentFolderURL() // Sources
-    .apollo.parentFolderURL()  // SwiftScripts
-    .apollo.parentFolderURL()  // apollo-ios
+    .apollo.parentFolderURL() // SwiftScripts
+    .apollo.parentFolderURL() // apollo-ios
 
 // In a typical app, you'll only need to do this for one target, so you'd
 // set these up directly within this file. Here, we're using more than one
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -224,8 +224,8 @@ One of the convenience wrappers available to you in the target is `ApolloSchemaD
 3. Set up your `ApolloSchemaOptions` object. In this case, we'll use the [default arguments for all the constructor parameters that take them](./api/ApolloCodegenLib/structs/ApolloSchemaOptions#methods), and only pass in the endpoint to download from and the folder to put the downloaded file into: 
 
     ```swift:title=main.swift
-    let options = ApolloSchemaOptions(endpointURL: endpoint,
-                                      outputFolderURL: output)
+    let schemaDownloadOptions = ApolloSchemaOptions(endpointURL: endpoint,
+                                                    outputFolderURL: output)
     ```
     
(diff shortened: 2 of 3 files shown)


joekryzak:
I'll walk through this tonight. Is this still the correct link:

https://www.apollographql.com/docs/ios/swift-scripting/

lastneutrino:
I'll take a look at this tonight, is this still the correct link:

https://www.apollographql.com/docs/ios/swift-scripting/

designatednerd:
Overall yes, but here is the preview of the deploy off of this branch: https://deploy-preview-1263--apollo-ios-docs.netlify.app/docs/ios/swift-scripting/

lastneutrino:
Three questions:

1 - I'm assuming if I build 29.0 (Package.swift wouldn't find it) that this issue would go away:

```
/Test1263/Codegen/Sources/Codegen/main.swift:9:6: error: value of type 'URL' has no member 'apollo'
    .apollo.parentFolderURL() // Sources
```

2 - If I build with 28.0, I get the following at the end using the previous instructions

```
Generating query files with 'swift' target [title changed]
Generating query files with 'swift' target [failed]
→ Apollo does not support anonymous operations
GraphQLError: Apollo does not support anonymous operations
    at GraphQLClientProject.get operations [as operations] (/Test1263/Codegen/ApolloCLI/apollo/node_modules/apollo-language-server/lib/project/client.js:266:31)
    at write (/Test1263/Codegen/ApolloCLI/apollo/lib/commands/client/codegen.js:69:75)
    at Task.task (/Test1263/Codegen/ApolloCLI/apollo/lib/commands/client/codegen.js:98:46)
```

3 - The individual commands don't quite match the composite at the end - there are two 'let options' in the step by step - they were renamed to schemaDownloadOptions and codegenOptions in the full composite.

Do you recommend loading 29.0 and re-trying with the composite?

designatednerd:
@lastneutrino 1 - `0.29.0` will be the next release, which will come out after I merge this PR. I would recommend building off the `update/debeta` branch of git that this PR is from to try these out. 

2 - Anonymous operations still won't be supported - basically you _can_ do: 

```graphql
query Something {
   methodName {
       field name
   }
}
```

but you _can't_ do: 

```graphql
query {
   methodName {
       field name
   }
}
```

I would recommend checking your operations to make sure they all have names.

3 - That is intentional since I wanted to disambiguate between the two `options` bits since it's in one script.

lastneutrino:
1 - ok will try tomorrow.
2 - yes, this worked, I have the API.swift file
3 - Correct - I'm speaking about the other way, i usually build as i go along, not realizing the file is at the bottom. You're disambiguating in the composite file at the end, but in the line by line listing, main.swift (built line by line) would end up having two 'let options' declarations - unless you are intending to do that so they get an error when assembling line by line.

designatednerd:
I'm expecting people to use them in separate files (that bit's spelled out explicitly in a couple places, including at the top of the full example), but I suppose I can update them to use the same names as the full example for clarity.

designatednerd:
OK last call for further comments! Will be merging this right before releasing v0.29.

PR #1273, opened 2020-06-24
make parts of GraphQLGetTransformer and MultipartFormData public

ethansinjin:
This allows creating NetworkTransport subclasses outside of the Apollo module that still make use of GraphQLGETTransformer and MultipartFormData.
A use case for doing this (and my actual use case) is replacing the use of URLSessionTask in HTTPNetworkTransport.swift with another networking library, such as Alamofire (ex. AlamofireNetworkTransport.swift)

### DIFF
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-struct GraphQLGETTransformer {
+public struct GraphQLGETTransformer {
 
   let body: GraphQLMap
   let url: URL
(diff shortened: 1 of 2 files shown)


designatednerd:
Hi! A couple questions: 

- What's your reason for wanting to subclass rather than making your own implementation of the `NetworkTransport` protocol? 
- Wouldn't we need to make `HTTPNetworkTransport` have an `open` annotation for this to work? Or is that implicit since it's not `final`?

ethansinjin:
Hey, thanks for the response! Sorry, I misspoke in my original message. I am indeed implementing the `NetworkTransport` protocol, not subclassing HTTPNetworkTransport. However, my class is effectively mirroring `HTTPNetworkTransport`, and only changing parts related to creating a `URLSessionClient`, `URLSessionTask`, and `URLRequest`. Because only those parts are changing, it would be helpful to use `GraphQLGETTransformer` and `MultipartFormData` in the same way that `HTTPNetworkTransport`'s implementation does, to avoid having to reimplement those classes as well.

PR #1279, opened 2020-06-29
Fix for multi-variable uploads

designatednerd:
Thanks to @vandyshev for pointing this out in #1081. The solution there unfortunately wound up hammering arrays of files, but it gave me a place to start for this bit which works with both multiple files with the same field name and multiple files with different field names. 

Previously, we were appending an index to every field name when uploading multiple files. This worked if the files all had the same name, but did not if they had different names. This PR makes updates to ensure that if you are uploading multiple files with different variable names, the correct variable name is included. 

The `testMultipleFilesWithMultipleFieldsWithApolloRequestCreator` test case will show the expected output.

### DIFF
--- a/Sources/Apollo/RequestCreator.swift
+++ b/Sources/Apollo/RequestCreator.swift
@@ -107,15 +107,15 @@ extension RequestCreator {
 
     // Make sure all fields for files are set to null, or the server won't look
     // for the files in the rest of the form data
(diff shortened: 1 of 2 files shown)


designatednerd:
Gonna go ahead and merge this one and put it out with the next version of the lib.

PR #1290, opened 2020-07-06
Update swift-scripting.md

bclubb:
Updated to parentFolderURL instead of parentFolder, which may have been an old version.

PR #1294, opened 2020-07-07
Update swift-scripting.md

ashakirzianov:
Seems to be a typo here, should be 'parentFolderURL' instead of 'parentFolder'


designatednerd:
This is identical to #1290, but since you've already signed the CLA and the submitter of that PR hasn't, I'll merge this once it passes CI.

Issue #1305, opened 2020-07-08
Fatal error ("Optional is only JSONEncodable if Wrapped is") when trying to update SqlNormalizedCache manually after a successful mutation operation


nateirwin:
While using Apollo iOS 0.29.1, I appear to be running into an issue similar to the one reported [here](https://github.com/apollographql/apollo-ios/issues/1271). I asked for help with this issue in [Spectrum](https://spectrum.chat/apollo/apollo-ios/updating-cache-after-insert-mutation~71008cf0-6ccf-4474-95bb-ac22c83e9616), and @designatednerd asked me to create a new issue.

I'm trying to update the SqlNormalizedCache (which is setup and working properly) using the result of a successful mutation operation:

```
apollo.perform(mutation: InsertTaskResponseMutation(taskId: 1, response: "")) { result in
    guard
        let resultData = try? result.get().data,
        let taskResponseDetails = resultData.insertTaskResponses?.returning.first?.fragments.taskResponseDetails
    else {
        return
    }

    apollo.store.withinReadWriteTransaction({ transaction in
        try! transaction.write(object: taskResponseDetails, withKey: "task_responses-\(taskResponseDetails.id)")
    })
}
```

This is throwing a fatal error in "JSONStandardTypeConversions.swift" at line 109:

<img width="502" alt="Screen Shot 2020-07-08 at 12 01 31 PM" src="https://user-images.githubusercontent.com/10443/86954195-c95b5700-c112-11ea-9ea0-7a32a9b95a7a.png">

Printing the description of `self`:

```
Printing description of self:
▿ Optional<Any>
  - some : 07/08/2020
```

In this case, `self` is a property on the `response` property, which is a custom `jsonb` type. Here's the type alias:

```
public typealias jsonb = [String : Any?]

extension Dictionary: JSONDecodable {
    public init(jsonValue value: JSONValue) throws {
        guard let dictionary = value as? Dictionary else {
            throw JSONDecodingError.couldNotConvert(value: value, to: Dictionary.self)
        }
        
        self = dictionary
    }
}
```

And here's the query:

```
query ChallengeTaskResponse($taskResponseId: Int) {
    challenge_responses(where: {id: {_eq: $taskResponseId}}) {
        ...ChallengeTaskResponseDetails
    }
}
```

The mutation:

```
mutation InsertTaskResponse($response: jsonb) {
    insert_task_responses(objects: {response: $response}) {
        returning {
            ...TaskResponseDetails
        }
    }
[...]

PR #1317, opened 2020-07-17
Add workaround for Optional JSONEncoding errors

designatednerd:
This is basically a band-aid on #1305 and similar bugs, which could occur when using a custom JSON scalar and trying to persist it to the cache. If the type of the typealiased JSON was `[String: Any?]`, which is expected, the `as JSONEncodable` would fail to recognize several types that have that conformance in an extension, and those types need to be unwrapped manually. 

This is a band-aid rather than a permanent fix because once I can get the Swift codegen stuff going again, I'm going to rip all this `JSONEncodable` stuff out in favor of a `Codable` solution (you can see a preview of the more flexible part of this in [the JSON type I'm using in the Codegen lib](https://github.com/apollographql/apollo-ios/blob/main/Sources/ApolloCodegenLib/JSON.swift)).  

But that's gonna be a hot minute, so band-aid it is!

### DIFF
--- a/Sources/Apollo/JSONStandardTypeConversions.swift
+++ b/Sources/Apollo/JSONStandardTypeConversions.swift
@@ -105,6 +105,22 @@ extension Optional: JSONEncodable {
       return NSNull()
     case .some(let wrapped as JSONEncodable):
       return wrapped.jsonValue
+      
+    // WORKAROUND: For reasons I don't totally understand, when the underlying type is `Any`,
+    // even though all of these conform to `JSONEncodable`, the `as JSONEncodable` above
+    //  fails, and we need to handle them individually.
+    case .some(let wrapped as String):
+      return wrapped.jsonValue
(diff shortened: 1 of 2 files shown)


Sources/Apollo/JSONStandardTypeConversions.swift
```diff
@@ -105,6 +105,22 @@ extension Optional: JSONEncodable {
       return NSNull()
     case .some(let wrapped as JSONEncodable):
       return wrapped.jsonValue
+      
+    // WORKAROUND: For reasons I don't totally understand, when the underlying type is `Any`,
+    // even though all of these conform to `JSONEncodable`, the `as JSONEncodable` above
+    //  fails, and we need to handle them individually.
```
designatednerd, line 111: If anyone's got any bright ideas why this fails, I'm all ears.

PR #1327, opened 2020-07-24
Expose received message to WebSocketTransportDelegate

ketenshi:
An example use case would be if the backend sends "UNAUTHENTICATED" messages through the websocket: this can be handled directly by the delegate. 
The alternative would be for consumers of the subscriptions to be the starting point of handling that event, which can be extremely cumbersome.

### DIFF
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -11,12 +11,14 @@ public protocol WebSocketTransportDelegate: class {
   func webSocketTransportDidConnect(_ webSocketTransport: WebSocketTransport)
   func webSocketTransportDidReconnect(_ webSocketTransport: WebSocketTransport)
   func webSocketTransport(_ webSocketTransport: WebSocketTransport, didDisconnectWithError error:Error?)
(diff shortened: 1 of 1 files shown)


designatednerd:
Hi! All the websocket tests are failing here. 

Additionally, is there a reason your backend is not sending this back as a GraphQL error rather than just a message?

ketenshi:
Hello, sorry about the failing tests earlier. My changes were ported from an older version. There's some remaining failing tests although I'm not sure how my changes would affect them, I'll see if I can find anything.

Regarding the reasoning for not sending a GraphQL error in place of a message, the solution that was used for the backend didn't include support for surfacing errors.

I know these changes may seem very special case, so if there's no value to the general public I'm happy to closing this.

designatednerd:
The two tests that are failing now are unrelated and flaky as hell, so I wouldn't worry about those.

Unfortunately, I don't think this is something that makes sense to add to the main fork given that the more standard way to do this is to send errors in an appropriate GraphQL container (or at least send a JSON dictionary that's compatible with graphql parsing).  

Anyone else besides @ketenshi having this issue, let me know, but this is the first time we've even had a request for this feature, which is a good indication that your setup is fairly unusual.

ketenshi:
Yep I understand, thanks for taking the time to look at this!

designatednerd:
OK - mind if we close this out?

PR #1330, opened 2020-07-27
Add initializer that takes database Connection

mdarnall:
### Updated SQLiteNormalizedCache
Because:
- It will allow consumers to further configure the Connection in their
application and pass it in
- See: https://github.com/apollographql/apollo-ios/pull/1162/files#r410367735

This commit:
- Adds a new initializer to SQLiteNormalizedCache

### DIFF
--- a/Tests/ApolloSQLiteTests/CachePersistenceTests.swift
+++ b/Tests/ApolloSQLiteTests/CachePersistenceTests.swift
@@ -4,6 +4,7 @@ import XCTest
 import ApolloTestSupport
 import ApolloSQLiteTestSupport
 import StarWarsAPI
+import SQLite
 
 class CachePersistenceTests: XCTestCase {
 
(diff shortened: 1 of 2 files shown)


Tests/ApolloSQLiteTests/CachePersistenceTests.swift
```diff
@@ -59,6 +60,12 @@ class CachePersistenceTests: XCTestCase {
     }
   }
 
+  func testPassInConnection() {
+    let connection = try? Connection()
+    XCTAssertNotNil(connection)
```
designatednerd, line 65: I would recommend making the test a `throws` test so you don't have to mess around with unwrapping:

```swift
func testPassInConnection() throws {
  let connection = try Connection()
  let cache = try SQLiteNormalizedCache(db: connection)
  // Something that tests that the cache exists
}

mdarnall: Oh, sure I can update that.. thanks

mdarnall: @designatednerd I updated this to hopefully simplify, but didn't know if there was really a good way to "tests that the cache exists". Instead, I'm simply asserting the creation does not throw.

designatednerd: That works too!

Issue #1340, opened 2020-07-31
RFC: Networking Updates


designatednerd:
This is the technical outline of a proposal to make major changes to our primary networking interface from the current `HTTPNetworkTransport` to a `RequestChainNetworkTransport` which uses a chain of interceptor objects to set up and process the results of network requests.

Note that this is going to be a **🎉 Spectacularly 🎉** breaking change - while very surface level APIs will remain basically the same, if you're doing anything remotely advanced, this will necessitate some changes, but the idea is to break it now so we don't have to break it way worse later.

I would **REALLY** love feedback on this before I start working towards making this the default option. You can see the code changes in-place in [this PR](https://github.com/apollographql/apollo-ios/pull/1341). I will be updating this RFC with feedback as it is received. 

## Why The Change?

`HTTPNetworkTransport` allows you to hook into various delegates to accomplish various things. There are several limitations to this approach: 

- Users can only do things that are specifically supported by delegates.
- Asynchronous use of delegates without callbacks is basically impossible.
- Any time we want to add a new feature, we need to add a new delegate method and handle it, creating additional complexity.
- There is no flexibility in terms of order of operations, particularly around whether data should be returned to the UI before being written to the cache.

The other major issue driving this update is that the current networking stack is *deeply* tied to the current cache architecture. This isn't ideal for many reasons, the biggest of which is that the cache likely to change in relation to the [Swift Codegen Rewrite](https://github.com/apollographql/apollo-ios/issues/939). 

## What is proposed? 

The proposed new architecture uses the Interceptor pattern to create a customizable request chain. This means users can hook into the system at any point during the request creation or data processing process. 

This also means that the pieces which will need to be swapped out for the Swift Codegen Rewrite are more clearly defined, and less tied to the actual parsing operation. 

Finally, this also opens the opportunity for different patterns than we already support, such as writing to the cache *after* returning data to the UI instead of before, or creating an array of interceptors which hit the network first, then hit the cache if nothing was returned. 

### New Protocols

- **`FlexibleDecoder`**: This is mostly going to be helpful for the `Codable` implementation down the line, but this will allow anything conforming to `Decoder` to be used to decode data.
- **`Parseable`**: This is a wrapper that allows us to continue to support non-`Codable` parsing alongside `Codable` parsing, while keeping us able to constrain and construct things generically. A default implementation for `Codable` will be provided.
[...]

PR #1341, opened 2020-07-31
Network Rearchitecture

designatednerd:
This PR relates to the [RFC for the new networking stack](https://github.com/apollographql/apollo-ios/issues/1340) - this branch is basically where I've been doing explorations thus far.


Sources/Apollo/ApolloClient.swift
```diff
@@ -133,23 +127,15 @@ extension ApolloClient: ApolloClientProtocol {
...
-    // and we can return a network task directly
-    if cachePolicy == .fetchIgnoringCacheData || cachePolicy == .fetchIgnoringCacheCompletely {
-      return self.send(operation: query, shouldPublishResultToStore: cachePolicy != .fetchIgnoringCacheCompletely, context: context, resultHandler: resultHandler)
-    } else {
-      let operation = FetchQueryOperation(client: self, query: query, cachePolicy: cachePolicy, context: context, resultHandler: resultHandler)
-      self.operationQueue.addOperation(operation)
-      return operation
-    }
+                                                                     cachePolicy: CachePolicy = .returnCacheDataElseFetch,
+                                                                     context: UnsafeMutableRawPointer? = nil,
```
amadeu01, line 133: I'm not sure if that is related to this PR, but in the `perform(mutation:)` function, the `context: UnsafeMutableRawPointer? = nil,` was removed, is there a reason why it is kept in the fetch function?

Sources/Apollo/ApolloErrorInterceptor.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// An error interceptor called to allow further examination of error data when an error occurs in the chain.
+public protocol ApolloErrorInterceptor {
```
amadeu01, line 4: I'm not sure if it would work, but what do you think of instead using protocol for this interceptor, actually, use just a function. For instance, 

`typealias ApolloErrorInterceptor =  (Error,
       RequestChain,
       HTTPRequest<Operation>,
       HTTPResponse<Operation>?,
       @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void)) -> Void`

Sources/Apollo/GraphQLResponse.swift
```diff
@@ -1,15 +1,26 @@
 /// Represents a GraphQL response received from a server.
-public final class GraphQLResponse<Data: GraphQLSelectionSet> {
+public final class GraphQLResponse<Data: GraphQLSelectionSet>: Parseable{
```
amadeu01, line 2: You might have missed that space `Parseable{`,  `Parseable {`

designatednerd: 🤦‍♀️ thank you!

Sources/Apollo/ApolloErrorInterceptor.swift
designatednerd: I would wayyyyyy rather have this as a protocol, particularly since it gives longer term flexibility to add more methods if needed.

Sources/Apollo/ApolloClient.swift
designatednerd: nope, good catch - it should stay in the store for direct interaction with the store, but otherwise you're correct.

Sources/Apollo/GraphQLResult.swift
```diff
@@ -29,3 +34,16 @@ public struct GraphQLResult<Data> {
     self.dependentKeys = dependentKeys
   }
 }
+
+extension GraphQLResult where Data: Decodable {
+  
+  public init<T: FlexibleDecoder>(from data: Foundation.Data, decoder: T) throws {
+    // SWIFT CODEGEN: fix this to handle codable better
```
designatednerd, line 41: Thoroughly punting on this for now.

designatednerd:
A preview of the updated networking setup instructions can be found [here](https://deploy-preview-1341--apollo-ios-docs.netlify.app/docs/ios/initialization/). 

Going to be updating the tutorial portion dealing with these changes before cutting a beta. Would love feedback!

designatednerd:
OK, pointing this at `betas/networking-stack` - once tests pass here I'll be cutting a beta from that branch and further PRs will go to that branch until the networking stack gets merged to `main`.

PR #1351, opened 2020-08-06
Add local upload server + tests

designatednerd:
I realized that in order to test something I'm trying to do with uploads in #1341, I need to actually test what we're _already_ doing with uploads a lot more thoroughly. 

In this PR: 
- Yoinked @jbaxleyiii's [simple upload server](https://github.com/jbaxleyiii/upload) to run locally. 
- Added an `UploadAPI` framework that takes the schema from that server and makes requests with it.
- Updated `RequestCreatorTests` to actually use a real request that uses the `Upload` type
- Added `UploadTests` to validate that requests succeed and the appropriate data is uploaded to the local server. 

NOTE: This is gonna be a draft until I can bribe @hwillson into helping me figure out how to add another mutation that has multiple parameters that accept an `Upload` 😇


SimpleUploadServer/index.js
```diff
@@ -0,0 +1,55 @@
...
+    path: String!
+    filename: String!
+    mimetype: String!
+  }
+  type Query {
+    uploads: [File]
+  }
+  type Mutation {
+    singleUpload(file: Upload!): File!
+    multipleUpload(files: [Upload!]!): [File!]!
```
designatednerd, line 17: @hwillson OK so I think I can see how to add this to the typedefs but I want to do something like `multipleParameterUpload(singleFile: Upload!, multipleFiles: [Upload!]!): [File!]!` - I'm not sure how to add that to the resolvers though.

designatednerd:
Ergh, I don't know what the hell is going on with Circle - running `xcodebuild clean build-for-testing -project "Apollo.xcodeproj" -scheme "Apollo" -sdk iphonesimulator13.6 -destination "platform=iOS Simulator,OS=13.6,name=iPhone 11" | xcpretty` locally works just dandy.

SimpleUploadServer/index.js
hwillson: @designatednerd You could do something like:

```js
async multipleParameterUpload(obj, { singleFile, multipleFiles }) {
  const { resolve, reject } = await promisesAll.all(
    [singleFile, ...multipleFiles].map(processUpload)
  );

  if (reject.length)
    reject.forEach(({ name, message }) =>
      // eslint-disable-next-line no-console
      console.error(`${name}: ${message}`)
    );

  return resolve;
}
```

designatednerd: Mostly working, but I think there's something up with this: 

```js
[singleFile, ...multipleFiles].map(processUpload)
```

The files I'm getting back when it resolves are only the ones uploaded via `multipleFiles`

hwillson: @designatednerd That's odd! If you drop a `console.log(singleFile)` in here:

```js
async multipleParameterUpload(obj, { singleFile, multipleFiles }) {
  console.log(singleFile);
  ...
```

Is a `singleFile` value logged when the resolver is hit?

designatednerd: Yes, though there's another error in there:

```
a.txt
TypeError: createReadStream is not a function
```

hwillson: @designatednerd it looks like the server side resolver is getting passed a string instead of an `Upload`. That's why the resolver is dumping out `a.txt` as the value of `singleFile`, instead of something like:

```js
Promise {
  {
    filename: 'a.txt',
    mimetype: 'application/octet-stream',
    encoding: '7bit',
    createReadStream: [Function: createReadStream]
  }
}
```

Since it's getting a string, when we try to access the `createReadStream` function on the upload, we can't find it. Is it possible that something in `UploadMultipleFilesToDifferentParametersMutation` is preventing the full `Upload` type from being sent over properly, and is instead sending a string?

designatednerd: Definitely possible. Will look at this, thank you!

designatednerd:
@hwillson is the 💣!

Issue #1348, opened 2020-08-04
extensions property on GraphQLResult


paulkite:
Hello! Curious if there are any plans to implement the `extensions` hash map support on `GraphQLResult`. I see the Android implementation has this and would love to be able to have the same ability to use this as well.

Thanks!

designatednerd:
This [is implemented on `GraphQLError`](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/GraphQLError.swift#L33) on our end. Are you using it in a non-error capacity?

paulkite:
Yes. Our team sends non-error information through `extensions`.

designatednerd:
OK - looking at [the spec](https://spec.graphql.org/June2018/#sec-Response-Format) it does appear `extensions` is allowed at the root of the response. 

I'm somewhat swamped at the moment - if you want to make a PR you're more than welcome to (just make sure to add tests!) but I'll try to get to it as I can if not.

paulkite:
Thank you so much for the response!

paulkite:
@designatednerd Just curious, what's the usual turnaround on PRs from the community to getting into a release?

designatednerd:
Usually fairly quick if they're small. It'd almost certainly be faster than waiting for me to have bandwidth 😛

designatednerd:
This shipped as part of 0.32.0!

PR #1370, opened 2020-08-27
Add extensions property to GraphQLResult.

paulkite:
- Added extensions property to GraphQLResult.
- Updated references.
- Implemented tests in ParseQueryResponseTests.

Addresses issue: [1348](https://github.com/apollographql/apollo-ios/issues/1348)

### DIFF
--- a/Sources/Apollo/GraphQLResponse.swift
+++ b/Sources/Apollo/GraphQLResponse.swift
@@ -20,6 +20,8 @@ public final class GraphQLResponse<Data: GraphQLSelectionSet> {
       errors = nil
     }
 
+    let extensions = body["extensions"] as? JSONObject
+
     if let dataEntry = body["data"] as? JSONObject {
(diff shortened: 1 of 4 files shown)


Sources/Apollo/GraphQLResponse.swift
```diff
@@ -72,13 +75,16 @@ public final class GraphQLResponse<Data: GraphQLSelectionSet> {
       let data = try decode(selectionSet: Data.self,
                             from: dataEntry,
                             variables: variables)
+      let extensions = body["extensions"] as? JSONObject
```
designatednerd, line 78: This should probably be outside the `if let` statement since theoretically someone could return stuff in `extensions` even if `data` is `nil`.

paulkite: Ah. Yeah, I was wondering if I was making the right choice there or not. I'll make an update ASAP. Thank you!

PR #1372, opened 2020-08-29
Make response parsing helpers public

kevinmbeaulieu:
When extending the functionality of Apollo to perform custom logging or other behavior on requests, these functions would be useful to have access to. In particular, making `parseResultFast()` public enables `decode` to remain internal while still giving users the ability to parse GraphQL responses. Making `parseErrorsOnlyFast()` is less significant, however it still would be helpful to centralize the hardcoding of the "errors" key to this one function rather than hardcoding it in both the Apollo codebase and Apollo users' codebases.

### DIFF
--- a/Sources/Apollo/GraphQLResponse.swift
+++ b/Sources/Apollo/GraphQLResponse.swift
@@ -57,7 +57,7 @@ public final class GraphQLResponse<Data: GraphQLSelectionSet> {
     }
   }
 
-  func parseErrorsOnlyFast() -> [GraphQLError]? {
+  public func parseErrorsOnlyFast() -> [GraphQLError]? {
     guard let errorsEntry = self.body["errors"] as? [JSONObject] else {
       return nil
     }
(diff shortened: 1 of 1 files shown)


designatednerd:
Definitely making `parseResultFast` public in the new networking stack - is this something you need ASAP?

kevinmbeaulieu:
Would be nice, but no not urgent if a new networking stack is coming soon

designatednerd:
Check out #1340 and #1341

designatednerd:
Actually, I'm going to go ahead and merge this - there's no reason this has to wait on the new stack.

kevinmbeaulieu:
@designatednerd That new networking stack PR draft is looking good to me btw! One of the main reasons we've needed a custom `NetworkTransport` has been to do some additional logging of each operation response, but looks like with the new stack we'd be able to just define our own logging interceptor and add it to the chain, using the standard interceptors for everything else

designatednerd:
Sweet, yeah that's the idea.

Issue #1365, opened 2020-08-20
GraphQLQueryWatcher doesn't update its dependent keys


teodorpenkov:
With the `0.23.0` release Apollo changed the way it stores keys in the `QueryWatcher` https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/GraphQLQueryWatcher.swift#L73-L83

The problem with this is that the watcher doesn't update its `dependentKeys` like it should. There is a commit that does that but I couldn't follow why it was changed:
https://github.com/apollographql/apollo-ios/commit/2b07e69e71a0621795b7089fda84c0ca744ba202

The problem that we are experiencing is that we are updating a query in the store with `ReadWriteTransaction` and in the previous versions the updated query captured the newly dependent keys, therefore when a change occured it was propagated to the result handler and now it's not.

What's the reason behind this change and how can we workaround this? Was there a problem with the previous implementation in 0.22.0 https://github.com/apollographql/apollo-ios/blob/0.22.0/Sources/Apollo/GraphQLQueryWatcher.swift#L66-L68

designatednerd:
Hi, the commit you referenced was part of #1156 which went out with `0.28.0`, not `0.23.0`. You can see a detailed explanation of the bug in #1155. 

What version are you actually on right now?

teodorpenkov:
I'd either suggest reverting to `0.22.0`'s implementation https://github.com/apollographql/apollo-ios/blob/0.22.0/Sources/Apollo/GraphQLQueryWatcher.swift#L66-L68

or adding `self.dependentKeys = graphQLresult.dependentKeys` before this line https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/GraphQLQueryWatcher.swift#L79

If you don't see problem with that I'm happy to open a PR.

teodorpenkov:
Hmm, actually I think the problem is somewhere else, I'll spend more time this week to figure it out.

teodorpenkov:
@designatednerd thanks for the links that you shared.  I've check therm and I found the root cause of the problem.

<img width="917" alt="Screen Shot 2020-08-25 at 17 05 17" src="https://user-images.githubusercontent.com/5622288/91184760-a5b99380-e6f5-11ea-8611-75a66b6313ee.png">

I'm pretty sure that if we bring back those lines it wouldn't break anything.

designatednerd:
Hi! Sorry, I was on vacation last week. 

[...]

PR #1375, opened 2020-08-31
Update dependent keys after a successful load

designatednerd:
Addresses #1365 - @teodorpenkov, can you confirm this fixes your issue? I'm still futzing around on how to test this.

### DIFF
--- a/Sources/Apollo/GraphQLQueryWatcher.swift
+++ b/Sources/Apollo/GraphQLQueryWatcher.swift
@@ -74,9 +74,14 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
         guard let self = self else { return }
 
         switch result {
-        case .success:
+        case .success(let graphQLResult):
           self.callbackQueue.async { [weak self] in
-            self?.resultHandler(result)
+            guard let self = self else {
+              return
+            }
+            
+            self.dependentKeys = graphQLResult.dependentKeys
+            self.resultHandler(result)
           }
         case .failure:
           // If the cache fetch is not successful, for instance if the data is missing, refresh from the server.
--- a/Sources/ApolloTestSupport/MockNetworkTransport.swift
+++ b/Sources/ApolloTestSupport/MockNetworkTransport.swift
@@ -2,7 +2,7 @@ import Apollo
 import Dispatch
 
 public final class MockNetworkTransport: NetworkTransport {
-  let body: JSONObject
+  public var body: JSONObject
 
   public var clientName = "MockNetworkTransport"
   public var clientVersion = "mock_version"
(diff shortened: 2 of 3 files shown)


teodorpenkov:
yes, this fixes the issue, I will think about the test today.

 Thank you

teodorpenkov:
@designatednerd here is a test that verifies it's working. I wouldn't add as it is, because it's messy, maybe if we enhance the Star Wars API with mutations about characters it will allow us to build better test.

If I have the time in the next week or so I will try to come up with something better because this is a bit hidden and it's very easy to miss.

```
withCache(initialRecords: [:]) { (cache) in
      let networkTransport = MockNetworkTransport(body: [
        "data": [
          "hero": [
            "id": "0",
            "name": "Artoo",
            "__typename": "Droid",
            "friends": [
              ["id": "10", "__typename": "Human", "name": "Luke Skywalker"]
            ]
          ]
        ]
      ])

      let client = ApolloClient(networkTransport: networkTransport, store: store)
      client.store.cacheKeyForObject = { $0["id"] }

      let query = HeroAndFriendsNamesWithIDsQuery()
      let expectation = self.expectation(description: "Has friend named Han Solo")
      let initialFetch = self.expectation(description: "Initial fetch")
      initialFetch.assertForOverFulfill = false

      _ = client.watch(query: query) { result in
        initialFetch.fulfill()
        if case let Result.success(result) = result {
          if result.data.flatMap({ $0.hero?.friends?.contains(where: { $0?.name == "Han Solo" })}) == true {
            expectation.fulfill()
          }
        }
      }
      wait(for: [initialFetch], timeout: 1)

      let encahnceInitialQuery = self.expectation(description: "Update initial query")
      store.withinReadWriteTransaction({ transaction in
        try transaction.update(query: query) { (data: inout HeroAndFriendsNamesWithIDsQuery.Data) in
          data.hero?.friends?.append(try .init(jsonObject: ["id": "11", "__typename": "Human", "name": "Not Han Solo"]))
          encahnceInitialQuery.fulfill()
        }
      })
      wait(for: [encahnceInitialQuery], timeout: 1)

      networkTransport.body = [
        "data": [
          "hero": [
            "id": "2",
            "name": "R2-D2",
            "__typename": "Droid",
            "friends": [
              ["id": "11", "__typename": "Human", "name": "Han Solo"]
            ]
          ]
        ]
      ]

      client.fetch(query: HeroAndFriendsNamesWithIdQuery(episode: .newhope), cachePolicy: .fetchIgnoringCacheData)

      waitForExpectations(timeout: 5, handler: nil)
    }
```

designatednerd:
@teodorpenkov Thanks for your help with the idea - I've taken the test and fleshed it out a bit, would love your feedback, but it passes when resetting the dependent keys is there and fails when it's commented out, so I think it should be an effective safety net.

teodorpenkov (approved):
LGTM

designatednerd:
@teodorpenkov thanks for the inspo on the test, it already caught a massive bug with query watchers on the networking rewrite branch 🙃

Issue #1376, opened 2020-09-02
No data found for task x, cannot append received data


RolandasRazma:
One of our tests hitting assert in `URLSessionClient.swift:230` in on CI after updating to latest version (0.31.0).
It's not networking stack test so don't think we doing anything specific to hit that assert. Looking at code that would indicate some kind race condition...

designatednerd:
Weird! Is it possible it's something leaking over from an older test?

RolandasRazma:
can't see anything relevant

RolandasRazma:
```swift
public func invalidate() {
    self.clearAllTasks()
    self.session?.invalidateAndCancel()
    self.session = nil
  }
```

shouldn't `self.session?.invalidateAndCancel()` be before `self.clearAllTasks()`

designatednerd:
Hi! Yeah flipping the cleanup order likely wouldn't help since some tasks are still in-flight, I suspect I need to actually cancel all the in-flight tasks. Thanks for the heads up on this!

designatednerd:
Haha damn it, looks like there's a key piece of info about `invalidateAndCancel` that's only in the docs docs and not in the headerdoc: 

<img width="779" alt="Screen Shot 2020-09-08 at 1 58 23 PM" src="https://user-images.githubusercontent.com/1976498/92516873-6e5de300-f1db-11ea-9b93-4437ef49a216.png">

designatednerd:
Shipped with version `0.32.1`!

philfi:
Hi @designatednerd,

I am encountering this issue in `0.33.0`, looking at #1383 I'm not quite certain it addressed the root issue given the shared `URLSession` isn't being used in `URLSessionClient`? I have a fairly reproducible case where this `invalidate` logic is getting hit but I'm still encountering this assertion error afterwards, confirmed with some debugging its definitely the same instance floating around. 

That all said, it does seem that the issue is resolved by calling `invalidateAndCancel` prior to calling the `clear` block (as is currently done in the else branch). I've tested this quite a bit and have been unable to trigger the assertion failure anymore. I'll go ahead and open a PR for you to take a look, hopefully it helps!

RolandasRazma:
Can confirm that it's not solved in 0.33.0. Just got assert while running tests

<img width="1024" alt="Screenshot 2020-10-14 at 11 44 45" src="https://user-images.githubusercontent.com/1027187/95979215-37e73980-0e13-11eb-9976-22d5d1f32349.png">

[...]

PR #1382, opened 2020-09-06
flip cleanup order to avoid hitting assert in race condition

RolandasRazma:
I think this fixes https://github.com/apollographql/apollo-ios/issues/1376


RolandasRazma:
it didn't fix

RolandasRazma:
it might be part of solution, without this change we hit that assert every second run, with it we hit it once in 20 runs. Difficult to say...

designatednerd:
I think the issue is that no matter the order the actual tasks are still in flight and need to be cancelled.

PR #1383, opened 2020-09-08
Manual cancellation of in-flight operations

designatednerd:
This should address #1376 - when a session was being invalidated, its in-flight tasks weren't being cancelled before being removed from the list of active tasks on `URLSessionClient`. Why not, since we were calling `invalidateAndCancel`, which is supposed to cancel active tasks? Turns out there's one verrrry important piece of information that only lives on the web docs: 

Screen Shot 2020-09-08 at 1 58 23 PM

Now, we're going through and actually cancelling all the tasks that are from the `URLSessionClient` before clearing out the array of active tasks. That should prevent tasks from calling back at a point where the assertion failure being hit in #1376 gets hit.

### DIFF
--- a/Sources/Apollo/URLSessionClient.swift
+++ b/Sources/Apollo/URLSessionClient.swift
@@ -74,10 +74,10 @@ open class URLSessionClient: NSObject, URLSessionDelegate, URLSessionTaskDelegat
     
     let currentTaskIDs = self.tasks.value.keys
     if #available(OSX 10.11, *) {
-      session.getAllTasks { tasks in
+      session.getAllTasks { [weak self] tasks in
         for task in tasks {
           if currentTaskIDs.contains(task.taskIdentifier) {
-            task.cancel()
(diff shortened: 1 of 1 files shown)


designatednerd:
@RolandasRazma You wanna give this a shot and see if it resolves your issues?

RolandasRazma:
wasn't it simpler to remove assert? :)
sure will check, thanks for looking into it

designatednerd:
I definitely considered sending that to a `debugPrint` rather than an `assert`, but the assert is there to tell me things ain't working the way they're supposed to (while allowing things to fail more gracefully in prod).

designatednerd:
@RolandasRazma any improvement?

RolandasRazma:
So far so good.
Wasn’t happening before all the time as well, but haven’t seen any on this branch

designatednerd:
Cool, I'm gonna merge this and do a patch release then

PR #1394, opened 2020-09-16
Upgrade Starscream library

designatednerd:
Upgraded the Starscream library to use the newer 4.x version of Starscream, which also gets rid of a transient dependency.


designatednerd:
May the websocket gods have mercy on us all

designatednerd:
LOL FFS I forgot to change it in the `xcodeproj` so it never picked up the change, which is why nothing changed.

Issue #1391, opened 2020-09-16
Apollo 0.33.0 + Xcode 11.7. Use of unknown directive '#filePath'.


Hazeaze:
Hi.
I use SPM for the third-party libraries, updated Apollo from 0.32.1 to 0.33.0, and on the app build, I faced with the problem in the `FileFinder.swift`, where received an error `Use of unknown directive '#filePath'`.

![image](https://user-images.githubusercontent.com/8351874/93308812-91fcdb00-f824-11ea-9050-275be2b39282.png)

Parilar:
I got the same error

designatednerd:
`#filePath` is new in Swift 5.3, which ships with Xcode 12. You will need to use Xcode 12 for versions `0.33.0` and higher.

designatednerd:
I'm going to close this out since this is intended behavior. Enjoy Xcode 12!

dalmatele:
@designatednerd I don't think so. I used Xcode 11.7 and got the same error.

PR #1402, opened 2020-09-19
Fix Xcode version on README.md

manicmaniac:
Since 0.33.0, support for Xcode 11 has been dropped.

https://github.com/apollographql/apollo-ios/issues/1391#issuecomment-693608876
> You will need to use Xcode 12 for versions 0.33.0 and higher.


designatednerd:
Good catch, thank you!

designatednerd (approved):

Issue #1404, opened 2020-09-21
Networking Beta: Convenience constructors for HTTPRequest and JSONRequest


knox:
In my use case i have to append to the `URLRequest.query` before a request is actually sent and that was fairly straight forward with the old network stack using `HTTPNetworkTransportPreflightDelegate.networkTransport(_:willSend:)` 

With the new network stack things have gotten a bit more complex but i was able to come up with a solution involving a custom `ApolloInterceptor` that turns a given `JSONRequest` into a custom subclass which overrides `toURLRequest()`.

So to make implementation of custom HTTPRequests more easy i suggest you to add some convenience constructors for `HTTPRequest` and `JSONRequest` that take a single argument of their own type. This would let me get rid of this bunch of constructor arguments like this:

```
        let customRequest = CustomJSONRequest<Operation>(jsonRequest)
```

instead of this:
```        
        let customRequest = CustomJSONRequest<Operation>(
                operation: jsonRequest.operation,
                graphQLEndpoint: jsonRequest.graphQLEndpoint,
                contextIdentifier: jsonRequest.contextIdentifier,
                clientName: jsonRequest.additionalHeaders["apollographql-client-name"] ?? "",
                clientVersion: jsonRequest.additionalHeaders["apollographql-client-version"] ?? "",
                additionalHeaders: jsonRequest.additionalHeaders,
                cachePolicy: jsonRequest.cachePolicy,
                autoPersistQueries: jsonRequest.autoPersistQueries,
                useGETForQueries: jsonRequest.useGETForQueries,
                useGETForPersistedQueryRetry: jsonRequest.useGETForPersistedQueryRetry,
                requestCreator: jsonRequest.requestCreator
                )
```

designatednerd:
The override is the correct way of doing this - I'm not quite clear why the `super init`, which has a bunch of default parameters, isn't the way you're doing this, though. Can you go into a bit more detail on that?

knox:
When looking for a way to get my `CustomHTTPRequest` with its overriden `toURLRequest()` into place i learned about the concept of `ApolloInterceptor` in the new network stack architecture.

Therefore i have implemented a custom interceptor to replace the original `HTTPRequest` instance with one of `CustomHTTPRequest` as a substitution. Basically what i'm trying to do here is to create a copy of the request _with whatever particular values its properties have_ but extend its original functionality.

If there is something i have misunderstood or a better approach to achive this i'm glad to get enlighted.

[...]

PR #1405, opened 2020-09-21
Update how requests are created so custom subclasses of HTTP requests can be used

designatednerd:
In this PR: 

- Made `RequestChainNetworkTransport` subclassable and changed two methods to be open so they can be subclassed in order to facilitate using subclasses of `HTTPRequest` when needed.  (thanks for the suggestion, #1404) 
- Realized after doing this that uploading setup could be moved to `UploadRequest`, which meant: 
  - `UploadRequest` is now `open` so it can be subclassed.
  - The default creation of a multipart form request now lives in `UploadRequest` instead of `RequestCreator`, and non-default creation should be moved from an implementation of `RequestCreator` to a subclass of `UploadRequest`.
  - `RequestCreator` protocol has now been renamed `RequestBodyCreator`, since it only creates the request body and not a full `HTTPRequest` (which didn't exist in the old networking stack, so the name made more sense then). Associated implementations and property names have been changed accordingly. 
  -  The protocol requirement for multipart form data has consequently been removed from the `RequestBodyCreator` protocol.


Tests/ApolloTests/MultipartFormDataTests.swift
```diff
@@ -0,0 +1,133 @@
...
+//  Apollo
+//
+//  Created by Ellen Shapiro on 9/21/20.
+//  Copyright © 2020 Apollo GraphQL. All rights reserved.
+//
+
+import XCTest
+import Apollo
+
+class MultipartFormDataTests: XCTestCase {
```
designatednerd, line 12: Moved this out of the request creator tests to disambiguate what's actually being tested here

Tests/ApolloTests/TestCustomRequestCreator.swift
```diff
@@ -27,40 +27,4 @@ struct TestCustomRequestCreator: RequestCreator {
 
     return body
   }
-
-  public func requestMultipartFormData<Operation: GraphQLOperation>(for operation: Operation,
```
designatednerd, line 31: This is no longer a requirement of the `RequestBodyCreator` protocol, so it's been removed

Tests/ApolloTests/UploadTests.swift
```diff
@@ -198,6 +199,225 @@ class UploadTests: XCTestCase {
     }
     
     self.wait(for: [expectation], timeout: 10)
+  }
+  
+  // MARK: - UploadRequest
```
designatednerd, line 204: These tests were moved over from the `RequestCreator` tests since they no longer have anything to do with the request creator

designatednerd:
These are slightly larger changes than I'd normally like to make at the RC stage, but I've been itching to get rid of the request creator and this gave me an opening to get rid of its most convoluted bits.

knox (approved):
Looks good to me.

PR #1408, opened 2020-09-25
[Networking Stack] `additionalErrorInterceptor(for:)` doesn't work when subclassing `LegacyInterceptorProvider`

TizianoCoroneo:
## The problem

The problem is: the `additionalErrorInterceptor(for:)` function is never called, if the `InterceptorProvider` class in use is a subclass of any `InterceptorProvider` that *does not* declare the `additionalErrorInterceptor` function.

For example, consider this `CustomInterceptor` type:
```swift
class CustomInterceptor: LegacyInterceptorProvider {
  // Note that we don't use `override` here.
  func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
    ErrorInterceptor()
  }
}
```

Its superclass, `LegacyInterceptorProvider`, does not provide an implementation of `additionalErrorInterceptor(for:)`, relying instead on the default implementation from the protocol, that returns `nil`.
Since its superclass does not implement this function, the implementation in `CustomInterceptor` is not overriding it, but declaring it anew. For some 🤯 reason, this means that the implementation that provides the `additionalErrorInterceptor` is never called. The default implementation in the extension of the `InterceptorProvider` protocol is called instead, returning `nil`.

## The solution

I suggest 2 small changes to solve this issue, and remove this trap.
- Remove the default implementation of `additionalErrorInterceptor(for:)` from `extension InterceptorProvider`. This will force base classes to implement it; the library expects the user to subclass one of the provided InterceptorProviders anyway, so I think it's acceptable.
- Implement `additionalErrorInterceptor(for:)` in both `LegacyInterceptorProvider` and `CodableInterceptorProvider`, so that subclasses will have something to override, but keeping the nice `nil` default.

These changes make for a pretty smooth solution, with only one small breaking change: `InterceptorProvider`s that did implement `additionalErrorInterceptor(for:)` before this change will have to add the `override` keyword, if they inherited from one of the provided `InterceptorProvider`s.

## Double checking

You can check out the first commit of this PR to see that the Unit test I added fails, reproducing the issue. Add the other commit, and everything is green again ✅

### DIFF
--- a/Sources/Apollo/InterceptorProvider.swift
+++ b/Sources/Apollo/InterceptorProvider.swift
@@ -65,6 +65,10 @@ open class LegacyInterceptorProvider: InterceptorProvider {
         LegacyCacheWriteInterceptor(store: self.store),
     ]
   }
+  
+  open func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
+    return nil
+  }
 }
 
 // MARK: - Default implementation for swift codegen
@@ -109,4 +113,8 @@ open class CodableInterceptorProvider<FlexDecoder: FlexibleDecoder>: Interceptor
          // Swift codegen Phase 2: Add Cache Write interceptor
      ]
    }
+  
+  open func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
+    return nil
+  }
 }
--- a/Sources/Apollo/InterceptorProvider.swift
+++ b/Sources/Apollo/InterceptorProvider.swift
@@ -66,7 +66,7 @@ open class LegacyInterceptorProvider: InterceptorProvider {
     ]
   }
   
-  open func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
+  open func additionalErrorInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> ApolloErrorInterceptor? {
     return nil
   }
 }
@@ -114,7 +114,7 @@ open class CodableInterceptorProvider<FlexDecoder: FlexibleDecoder>: Interceptor
      ]
    }
   
-  open func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
+  open func additionalErrorInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> ApolloErrorInterceptor? {
     return nil
   }
 }
--- a/Tests/ApolloTests/RequestChainTests.swift
+++ b/Tests/ApolloTests/RequestChainTests.swift
@@ -229,7 +229,7 @@ class RequestChainTests: XCTestCase {
         ]
       }
       
-      func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
+      override func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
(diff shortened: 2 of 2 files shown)


Sources/Apollo/InterceptorProvider.swift
```diff
@@ -65,6 +56,10 @@ open class LegacyInterceptorProvider: InterceptorProvider {
         LegacyCacheWriteInterceptor(store: self.store),
     ]
   }
+  
+  open func additionalErrorInterceptor<Operation>(for operation: Operation) -> ApolloErrorInterceptor? where Operation : GraphQLOperation {
```
designatednerd, line 60: Do you mind changing this and below to the `<Operation: GraphQLOperation>` instead of the `where` clause to match format in the rest of the code?

Sources/Apollo/InterceptorProvider.swift
```diff
@@ -16,15 +16,6 @@ public protocol InterceptorProvider {
...
 }
 
-/// MARK: - Default Implementation
-
-public extension InterceptorProvider {
-  
-  func additionalErrorInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> ApolloErrorInterceptor? {
-    return nil
-  }
-}
```
designatednerd, line 26: I think you can add the open funcs to the default interceptor providers below without deleting this default implementation - that way if someone makes their own interceptor provider they don't have to implement this, but if they want to override they can since it's implemented in those implementations.

Tests/ApolloTests/RequestChainTests.swift
```diff
@@ -202,4 +210,75 @@ class RequestChainTests: XCTestCase {
       XCTFail("Error interceptor did not receive an error!")
     }
   }
+  
+  func testErrorInterceptorGetsCalledInLegacyInterceptorProviderSubclass() {
```
designatednerd, line 214: 😍 Thank you for adding a test for this 😍

designatednerd (commented):
Hi! Good point, though I think we can make this work without removing the default implementation.

designatednerd:
(current test failure is an intermittent issue and unrelated, when you push up new commits I'll kick the CI server if needed)

TizianoCoroneo:
> Hi! Good point, though I think we can make this work without removing the default implementation. 

I thought that removing the default implementation from the protocol it would create a footgun for users of the library that create their own `InterceptorProvider` without subclassing one of the main ones:
If the base class does not implement the function, then other custom subclasses will incur in the same issue.

```swift
import Apollo

class BaseIP: InterceptorProvider {
    // no implementation for additionalErrorInterceptor
}

class SubIP: BaseIP {
    // implements the function, but incurs in the same issue.
}
```

It's kind of a corner case 🤷‍♂️ we'll do fine either way.

I spent a couple days rewriting a network layer with the new networking stack and it's 🔥🔥🔥 keep up the good work! 👏

TizianoCoroneo:
I've found why this specific issue happens: the subclass cannot override the default implementation of the protocol because... it needs to be implemented in the compiler 😅 it's one of the features mentioned in Swift's [Generics Manifesto](https://github.com/apple/swift/blob/master/docs/GenericsManifesto.md), under "Allowing subclasses to override requirements satisfied by defaults".

designatednerd:
If someone's creating their own `InterceptorProvider` I'd be really surprised if they're creating something with subclasses. 

This is probably worth reminding people of in documentation rather than not including the default implementation, IMO.

TizianoCoroneo:
I'll put back the default implementation in the protocol extension, and remove the implementation from the various `TestProvider`s in the tests.

I'll try to update the docs as well, but one warning: I'm a non-native speaker 😅

Sources/Apollo/InterceptorProvider.swift
TizianoCoroneo: done! 👍

designatednerd (approved):

PR #1409, opened 2020-09-25
Adding the operation type to the request headers

heyzooi:
Adding the operation type to the request headers


designatednerd:
Do you mind trying to make a PR into the `beta/networking-stack` branch for this, since that's going to be released shortly? The place you'd probably want to put this is in `HTTPRequest` in the new stack.

heyzooi:
Awesome, let me rebase against this branch

designatednerd (approved):

PR #1386, opened 2020-09-11
Networking Beta

designatednerd:
Future changes to the networking stack will get merged into this until it comes out of beta (hopefully soon). Opening a PR to get Netlify to auto-generate updated docs that can be referenced here. 

Please note: 

- If you're opening a bug about the networking beta, please prefix the bug title with `Networking Beta:` - this will help me narrow down where things are going wrong more quickly.  
- If you're using the beta, please check the `CHANGELOG` from this branch for updates.
- Updated documentation and code are available here: https://deploy-preview-1386--apollo-ios-docs.netlify.app/docs/ios/
- The code changes to the tutorial are available in the [Tutorial Repo](https://github.com/apollographql/iOSTutorial) on the `new-network-beta` branch.

PR #1410, opened 2020-09-25
Allow custom headers per operation

heyzooi:


designatednerd:
Hi! This is possible in the new networking stack in #1386, each `HTTPRequest` or subclass thereof can get its own set of additional headers as it's constructed. 

I'm going to close this out since this is already implemented in the new networking  stack, but thank you for your work!

heyzooi:
thanks for the explanation

PR #1411, opened 2020-09-25
Retry depending on the operation

heyzooi:


designatednerd:
Hi! We're about to release a new networking stack through #1386, which is currently at release canddiate 2 - you should be able to use the `InterceptorProvider` protocol in that to add or remove a `RetryInterceptor` as needed per operation. 

I'm going to close this PR since it goes to the old networking stack that's about to be removed, but thank you for your interest!

heyzooi:
thanks for the explanation

PR #1420, opened 2020-09-29
Prevent request headers from getting lost

### LINKED ISSUE #1419: JSONRequest losing headers when useGETForQueries is true
## Bug report

After migrating my project to the new version, I'm not able to use a service that has authentication via header.

## Versions

- `apollo-ios` SDK version: 0.34.0
- Xcode version: 12.0.1
- Swift version: 5.3

## Steps to reproduce

Create an interceptor that adds a header to the request, the NetworkTransporter must have `useGETForQueries=true`.

## Further details

When the method `toURLRequest` gets called, the request gets recreated and all headers are lost, including the defaults added by apollo. After digging into the code, I found that the library is instantiating a new `URLRequest` instead of just changing the URL.

https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/JSONRequest.swift#L101

dchohfi:
I believe that this fixes #1419

### DIFF
--- a/Sources/Apollo/JSONRequest.swift
+++ b/Sources/Apollo/JSONRequest.swift
@@ -98,7 +98,7 @@ open class JSONRequest<Operation: GraphQLOperation>: HTTPRequest<Operation> {
     case .GET:
       let transformer = GraphQLGETTransformer(body: body, url: self.graphQLEndpoint)
       if let urlForGet = transformer.createGetURL() {
-        request = URLRequest(url: urlForGet)
+        request.url = urlForGet
         request.httpMethod = GraphQLHTTPMethod.GET.rawValue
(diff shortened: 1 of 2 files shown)


gabrielrinaldi (approved):

designatednerd (approved):
Wow, whoops. 🙃

Thanks for adding tests!

designatednerd:
Test failure is a known flaky test, rerunning then I'll get this merged

dchohfi:
Hey, thank you for merging this, can we get a new tag out? This is breaking my migration.

designatednerd:
Yeah will try to get this out today - you can also point at the commit on `main` if you need it right away.

PR #1432, opened 2020-10-04
Cast Double as String if the jsonValue is of type Double.

thanhdolong:
Similar problem with [this](https://github.com/apollographql/apollo-ios/pull/402)

This PR is trying solve problem with casting scalar as Double to String. Here is an error:
`Apollo.GraphQLResultError(path: filterMoneyTransactions.nodes.0.amount, underlying: Apollo.JSONDecodingError.couldNotConvert(value: 0.24, to: Swift.String))`

Scheme for this part looks like this:
```
type MoneyTransaction {
   ref: EntityRef (scalar)
   amount: MoneyAmount (scalar)
}
```

### DIFF
--- a/Sources/Apollo/JSONStandardTypeConversions.swift
+++ b/Sources/Apollo/JSONStandardTypeConversions.swift
@@ -7,6 +7,8 @@ extension String: JSONDecodable, JSONEncodable {
         self = string
     case let int as Int:
         self = String(int)
+    case let double as Double:
+        self = String(double)
     default:
         throw JSONDecodingError.couldNotConvert(value: value, to: String.self)
     }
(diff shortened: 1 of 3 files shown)


designatednerd (changes requested):
Hmm...This change  seems like it could break some existing validation assumptions. 

The main reason we added the Int change was to deal with some issues around the fact that `ID` can be defined as either an `Int` or a `String` (although it's usually defined as a `String`), and making sure that can be handled when the type generated is a string is more critical. 

Can you talk a little bit about why you aren't able to get your code working by adding typealiases (outside the generated `API.swift`) for your two custom scalars to `Double`?

designatednerd:
@thanhdolong Hi, can you elaborate a bit on this one? If not, I'll probably wind up closing it for now.

designatednerd:
Going to close this since I haven't heard back. Please feel free to reopen with answers to my questions. Thank you!

Issue #1425, opened 2020-09-30
Obsolete types GraphQLHTTPResponseError et. al.


knox:
## Bug report

With the new network stack (yeah!) a few types such as `GraphQLHTTPResponseError` became obsolte but did not get removed from code base yet.

## Versions

- `apollo-ios` SDK version: 0.34.0

## Steps to reproduce

"Find usages" of `GraphQLHTTPResponseError`, `GraphQLHTTPRequestError` and probably others.

designatednerd:
`GraphQLHTTPRequestError` is still used, but good catch on the response error. 

Please note that "find usages" isn't the be-all, end-all, especially for enum types - always worth doing a text search through the full codebase.

knox:
The problem here for me was that by not beeing broken because of a removed class my validations got semantically broken because they where still checking for that type.

Of course you're right about `GraphQLHTTPRequestError`. I've mistaken that.

But still, regarding your advise to check for "usages" with full text search i have to strongly disagree. In fact i heavily rely on this kind of type based analysis and navigating through code and therefore i prefer AppCode over Xcode a lot because of it outstanding capabilities in that. You should give it a try.

designatednerd:
I agree that it's helpful, but particularly when it comes to removing stuff, I've found it better to belt-and-suspenders things (even in AppCode, which I have used before).

Max-Job:
how I can handle server error?

knox:
Try something like this
```
        if case .invalidResponseCode(let response, _) = error as? ResponseCodeInterceptor.ResponseCodeError,
              let httpResponse = response {
            print(httpResponse.statusCode)
        }
```

Max-Job:
@knox oh! thank you so much! it works for me)
but now I have another problem:
if environment (API URL) doesn't exist [it's](https://github.com/apollographql/apollo-ios/issues/1425#issuecomment-734286610) not working

Max-Job:
if I write incorrect url here, I cannot handle error and get status code
```
let transport = RequestChainNetworkTransport(interceptorProvider: provider,
                                                     endpointURL: url)
        return ApolloClient(networkTransport: transport, store: store)
```

designatednerd:
@Max-Job Could you please open this as a different issue and provide more details like: 
- What type of error are you expecting to receive in this instance? 
- What is the type of error which you do receive?

PR #1437, opened 2020-10-05
Remove unused type

designatednerd:
Addresses #1425 - @knox, a brief look at other things in the networking stack didn't turn up anything we're not still using, but if you see anything else, shout it out. Thanks!


taher-mosbah:
Hello @designatednerd, in the changelog I see:
BREAKING: Removed the now-unused-in-the-SDK GraphQLHTTPResponseError type. If you were relying on this class, please copy it out of v0.34.1. (#1437)
but doing that I got: `'textEncoding' is inaccessible due to 'internal' protection level` and `'statusCodeDescription' is inaccessible due to 'internal' protection level` 
is it possible to change that please ?

designatednerd:
You will want to copy the [corresponding extension code](https://github.com/apollographql/apollo-ios/blob/246a2761fa48370a892c2055f245ed5f117fc9f1/Sources/Apollo/HTTPURLResponse%2BHelpers.swift) - you probably want to delete `apollo` and just make it a simple extension on `HTTPURLResponse` rather than the more complex `ApolloExtension` stuff that's in there.

vincent-paing:
What should I use instead of `GraphQLHTTPResponseError`?

knox:
@vincent-paing it's been replaced by `ResponseCodeInterceptor.ResponseCodeError`

PR #1441, opened 2020-10-07
Replace undefined types and variables in doc

shingt:
Hi, noticed some of the documents contain deprecated types/variables after the networking stack change, and so replaced:

* `HTTPNetworkTransport` => either `RequestChainNetworkTransport` / `LegacyInterceptorProvider` / `CodableInterceptorProvider`
* `enableAutoPersistedQueries` => `autoPersistQueries`

Similarly, replaced sample codes in [doc for subscriptions](https://www.apollographql.com/docs/ios/subscriptions/). Mostly they're simply from [playground](https://github.com/apollographql/apollo-ios/blob/1ef93aca33f5223255278e7a030b76de46374351/Playgrounds/ApolloMacPlayground.playground/Pages/Subscriptions.xcplaygroundpage/Contents.swift) though. Let me know if they're not needed!

### DIFF
--- a/docs/source/fetching-queries.md
+++ b/docs/source/fetching-queries.md
@@ -125,9 +125,9 @@ Because the above query won't fetch `appearsIn`, this property is not part of th
 
 By default, Apollo constructs queries and sends them to your graphql endpoint using `POST` with the JSON generated. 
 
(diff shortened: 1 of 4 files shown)


designatednerd (approved):
Awesome, thank you so much for catching this!

PR #1442, opened 2020-10-08
Add helper to deal with parsing and caching raw data in custom `NetworkTransport` instances

designatednerd:
So as part of the network refactor, I moved the logic for caching out of `ApolloClient` and into interceptors, with the logic that making this much more granular would allow people to order things in whatever order they'd like and/or do whatever pre-cache or post-cache manipulation they wanted. 

Turns out I missed one critical use case: Custom `NetworkTransport` implementations that can't be converted into interceptors. In this case, we'd removed all the cache fetch and store logic with nothing to replace it, so essentially every call became `fetchIgnoringCacheCompletely`. 🤦‍♀️

To fix that, I've added a `RawCacheDataHelper`, which allows you to pass in the requisite items as well as an object satisfying the `RawNetworkFetcher` protocol. The helper will use the passed-in `CachePolicy` to determine the order and necessity of hitting the store read -> fetch from network -> store write circuit, and if hitting the network is necessary, it will call `fetchData` with the operation to be fetched so that your network layer can do what's necessary to actually go out and fetch that data. Some examples of how this would work are in the `RawCacheDataHelper` tests. 

I also made a couple things in the `ApolloStore` switch to using completion closures in their internal APIs rather than the `Promise` implementation that's really been more of a source of confusion than anything. I'd eventually like to rip that out and this seemed like a good place to keep that logic from propagating.


designatednerd:
> I also made a couple things in the ApolloStore switch to using completion closures in their internal APIs rather than the Promise implementation that's really been more of a source of confusion than anything. I'd eventually like to rip that out and this seemed like a good place to keep that logic from propagating.

And in the department of Famous Last Words, it looks like I've more thoroughly broken our notoriously flaky threaded cache test in the process - I'll see if I can figure this out tomorrow - worst comes to worst I'll revert to using the promises implementation and deal with that in a separate PR.

designatednerd:
Ugh, I screwed up a rebase here. I think I'm going to use this as an excuse to close this out though - this is very much a band-aid on a (self-inflicted) bullet wound, and I'm going to work on a more permanent fix for this.

Issue #1444, opened 2020-10-09
Can't update the RequestBodyCreator for WebSocketTransport


lowki93:
## Bug report

We use `WebSocketTransport` and we need to override the `requestBodyCreator` parameters.
We define `CustomBodyCreator`:
```
struct CustomBodyCreator: RequestBodyCreator {
  
  public func requestBody<Operation: GraphQLOperation>(
    for operation: Operation,
    sendOperationIdentifiers: Bool
  ) -> GraphQLMap {
    .....
  }
  
}

```

we put it when we initialize the `WebSocketTransport`: 
```
 WebSocketTransport(request: request, requestBodyCreator: CustomBodyCreator())
```

But `requestBody` from our struct is never call.

There is a an extension for `RequestBodyCreator` that define the function in the protocol : 
```
extension RequestBodyCreator {
  /// Creates a `GraphQLMap` out of the passed-in operation
  ///
  /// - Parameters:
  ///   - operation: The operation to use
  ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Defaults to false.
  ///   - sendQueryDocument: Whether or not to send the full query document. Defaults to true.
  ///   - autoPersistQuery: Whether to use auto-persisted query information. Defaults to false.
  /// - Returns: The created `GraphQLMap`
  public func requestBody<Operation: GraphQLOperation>(for operation: Operation,
                                                       sendOperationIdentifiers: Bool = false,
                                                       sendQueryDocument: Bool = true,
                                                       autoPersistQuery: Bool = false) -> GraphQLMap {
.....
}
```

Why can put this default implementation in `ApolloRequestBodyCreator` directly ?
```
// Helper struct to create requests independently of HTTP operations.
public struct ApolloRequestBodyCreator: RequestBodyCreator {
  // Internal init methods cannot be used in public methods
  public init() { }
}
```

Why we can't override this ?

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.34.1
- Xcode version: 12.0.1
- Swift version: 5.2 

## Further details

We need to override this to make `WebSocketTransport`with `AppSync`from AWS

designatednerd:
Hi, apologies for the delay, I was under the weather. 

You can't override `ApolloRequestBodyCreator` because it's not intended to be overridden - that's why `RequestBodyCreator` is a protocol with a default implementation: That way your custom implementation can use the default implementation without having to implement it yourself if you don't need to. 

[...]

PR #1448, opened 2020-10-12
Move default implementation of RequestBodyCreator

lowki93:
To fix this [issues](https://github.com/apollographql/apollo-ios/issues/1444).

I propose to move default implementation of the `RequestBodyCreator` protocol in `ApolloRequestBodyCreator `


designatednerd:
Closing this in favor of #1450 for the extremely long reasons explained there. Thank you very much for taking a stab at this though!

PR #1450, opened 2020-10-15
Fix request body creator non-default implementations getting hammered by compiler

designatednerd:
Addresses #1444. Note that this will be a **breaking change**.

So I seriously had to [rewatch this bit of _my own talk_ about this](https://youtu.be/a0m0acpM8II?t=812) to figure out why this hadn't twigged on me earlier: I was expecting this to be like the `sayHi` method in that example, where because it's declared in the protocol, the compiler uses the lowest level of implementation rather than going to to the default implementation. Instead, it's behaving like the `askHowAreYou` method, which is not defined in the implementation, and which is defaulting to the protocol's default implementation.

The test we had previously validated that things worked when called _directly_, but not when called _indirectly_: When I made the changes to the test in this PR, it immediately started failing, so that definitely meant that we were closer to the `askHowAreYou` situation.

As far as I can tell, this is some weird combination of circumstances in terms of what we were doing with default values on the default argument vs. the implementation in the test. I suspect that because it *looked* like we were making the same call, we thought we were. But due to the default implementation's default values, we were actually calling two totally different methods on the two tests, and that only revealed itself when called indirectly. 

SO. Because I know there are existing implementations that rely on just passing things straight through to what was the default implementation, I've updated that to just be called `defaultRequestBody` so it can be called from other implementations of `RequestBodyCreator`(you can see an example in how I set up the updated `ApolloRequestCreator`). This also allows you to grab the default and modify it instead of having to redo the whole thing yourself.

### DIFF
--- a/Sources/Apollo/RequestBodyCreator.swift
+++ b/Sources/Apollo/RequestBodyCreator.swift
@@ -25,10 +25,10 @@ extension RequestBodyCreator {
   ///   - sendQueryDocument: Whether or not to send the full query document. Defaults to true.
   ///   - autoPersistQuery: Whether to use auto-persisted query information. Defaults to false.
   /// - Returns: The created `GraphQLMap`
-  public func requestBody<Operation: GraphQLOperation>(for operation: Operation,
-                                                       sendOperationIdentifiers: Bool = false,
-                                                       sendQueryDocument: Bool = true,
-                                                       autoPersistQuery: Bool = false) -> GraphQLMap {
+  public func defaultRequestBody<Operation: GraphQLOperation>(for operation: Operation,
+                                                              sendOperationIdentifiers: Bool,
+                                                              sendQueryDocument: Bool,
+                                                              autoPersistQuery: Bool) -> GraphQLMap {
     var body: GraphQLMap = [
       "variables": operation.variables,
       "operationName": operation.operationName,
@@ -64,4 +64,16 @@ extension RequestBodyCreator {
 public struct ApolloRequestBodyCreator: RequestBodyCreator {
   // Internal init methods cannot be used in public methods
   public init() { }
+  
+  public func requestBody<Operation: GraphQLOperation>(
+    for operation: Operation,
+    sendOperationIdentifiers: Bool = false,
+    sendQueryDocument: Bool = true,
+    autoPersistQuery: Bool = false) -> GraphQLMap {
+    
+    return self.defaultRequestBody(for: operation,
+                                   sendOperationIdentifiers: sendOperationIdentifiers,
+                                   sendQueryDocument: sendQueryDocument,
+                                   autoPersistQuery: autoPersistQuery)
+  }
 }
--- a/Sources/Apollo/RequestBodyCreator.swift
+++ b/Sources/Apollo/RequestBodyCreator.swift
@@ -8,7 +8,9 @@ public protocol RequestBodyCreator {
   ///
   /// - Parameters:
   ///   - operation: The operation to use
-  ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Defaults to false.
+  ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Should default to `false`.
+  ///   - sendQueryDocument: Whether or not to send the full query document. Should default to `true`.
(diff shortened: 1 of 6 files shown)


Sources/Apollo/RequestBodyCreator.swift
```diff
@@ -25,10 +25,10 @@ extension RequestBodyCreator {
   ///   - sendQueryDocument: Whether or not to send the full query document. Defaults to true.
   ///   - autoPersistQuery: Whether to use auto-persisted query information. Defaults to false.
   /// - Returns: The created `GraphQLMap`
-  public func requestBody<Operation: GraphQLOperation>(for operation: Operation,
-                                                       sendOperationIdentifiers: Bool = false,
-                                                       sendQueryDocument: Bool = true,
-                                                       autoPersistQuery: Bool = false) -> GraphQLMap {
+  public func defaultRequestBody<Operation: GraphQLOperation>(for operation: Operation,
```
wangjiejacques, line 28: why you rename this method?

Sources/Apollo/RequestBodyCreator.swift
```diff
@@ -64,4 +64,16 @@ extension RequestBodyCreator {
 public struct ApolloRequestBodyCreator: RequestBodyCreator {
   // Internal init methods cannot be used in public methods
   public init() { }
+  
+  public func requestBody<Operation: GraphQLOperation>(
```
wangjiejacques, line 68: put this method in the extension of the protocol, so it won't break the change.

Sources/Apollo/UploadRequest.swift
```diff
@@ -69,7 +69,10 @@ open class UploadRequest<Operation: GraphQLOperation>: HTTPRequest<Operation> {
     // Make sure all fields for files are set to null, or the server won't look
     // for the files in the rest of the form data
     let fieldsForFiles = Set(files.map { $0.fieldName }).sorted()
-    var fields = self.requestBodyCreator.requestBody(for: operation, sendOperationIdentifiers: shouldSendOperationID)
+    var fields = self.requestBodyCreator.requestBody(for: operation,
```
wangjiejacques, line 72: better to not change this.

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -270,7 +270,10 @@ public class WebSocketTransport {
   }
 
   func sendHelper<Operation: GraphQLOperation>(operation: Operation, resultHandler: @escaping (_ result: Result<JSONObject, Error>) -> Void) -> String? {
-    let body = requestBodyCreator.requestBody(for: operation, sendOperationIdentifiers: self.sendOperationIdentifiers)
+    let body = requestBodyCreator.requestBody(for: operation,
```
wangjiejacques, line 273: do not change this.

Sources/Apollo/RequestBodyCreator.swift
designatednerd: Hi! Please read the PR description for a full description of all these changes and why they've been made.

designatednerd:
Took a second pass after a day and discovered the real best way to handle this was to just ditch the default values in the default implementation. Made a couple tests dumber, but I think from a standpoint of not breaking as hard, this is better.

Still technically breaking since I removed the default values, though.

Sources/Apollo/UploadRequest.swift
designatednerd: This has to change due to the removal of the default values for these parameters

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: This has to change due to the removal of the default values for these parameters

Sources/Apollo/UploadRequest.swift
wangjiejacques: Hi @designatednerd , although the default values for these parameters has been removed, but I think it's a good idea to add a new extension method like this:
```
extension RequestBodyCreator {
    public func requestBody<Operation: GraphQLOperation>(for operation: Operation) {
        requestBody(operation, false, true, false)
    }
}
```

designatednerd: I'll consider that, but I am concerned that would run a significant risk of reintroducing this same problem.

Issue #1460, opened 2020-10-18
Initializer default values make it very easy to accidentally create 2 distinct ApolloStore's


namolnad:
## Bug report

Currently it's very easy, through default initializer values, to end up with multiple ApolloStore's and the inability to properly invalidate your cache. I just spent a fair amount of time debugging this issue, which I think could be resolved by removing some default values in one or both the `ApolloClient`'s and `LegacyInterceptorProvider`'s initializers. The problem is as follows:

``` swift
let interceptorProvider: LegacyInterceptorProvider = .init() // More likely to use your own subclass for authentication
let transport: RequestChainNetworkTransport = .init(
    interceptorProvider: interceptorProvider,
    endpointURL: baseUrl.appendingPathComponent("graphql")
)
let client: ApolloClient = .init(
    networkTransport: transport
)
```

The above code creates two separate `ApolloStore`'s with default caches of `InMemoryNormalizedCache()` for both the `LegacyInterceptorProvider` and for the `ApolloClient`. When going to invalidate your cache here, you will invalidate one of the caches (the ApolloStore one), but the InterceptorProvider's cache will remain untouched, which is where results seem to be fetched from, thus making it impossible to actually invalidate the cache. Without these default values, I believe this problem would be obvious to the developer.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.34.0
- Xcode version: 12.0.1
- Swift version: 5.3

## Steps to reproduce

Initialize the ApolloClient using default values, as above.

designatednerd:
Yeah this is basically a dupe of #1438 - I'm going to take a look at your solution, I was still pondering exactly how to get this set up in a friendlier way.

namolnad:
Yup, definitely a dupe — sorry for missing that and thanks for taking a look!

designatednerd:
This has shipped with `0.36.0`!

PR #1461, opened 2020-10-19
Remove default values from initializers

namolnad:
Improves clarity and prevents mistakes showcased in #1460 by removing default values for initializers


designatednerd (approved):
I think this is generally a good start - I want to do some more rethinking before an official release is cut, since I think there's still some pretty significant opportunities for developers to accidentally have different stores. 

May wait a day or two to merge this until I have that bit ready to go, but overall, this is a great start 👍

designatednerd:
I poked at what my idea was and it was WAY over-complex. This is much more straightfoward, so let's go with this approach. I'll update some docs post-merge.

namolnad:
Sounds great — glad it was helpful. Thanks for looking at it!

PR #1466, opened 2020-10-22
Fix typos in error messages

MaxDesiatov:
`Recieved` -> `Received`


designatednerd (approved):
I think I have spelled that correctly on the first try like 4 times in my entire life. Thanks for catching!

PR #1468, opened 2020-10-23
RequestChain error handling callback queue

ejensen:
Ensures the `additionalErrorHandler`'s completion is called on the correct `callbackQueue`.

This fixes a threading error where the RequestChain does not complete on the specified `callbackQueue` due to the queue not being used in the case that there's an `additionalErrorHandler` specified.

### DIFF
--- a/Sources/Apollo/RequestChain.swift
+++ b/Sources/Apollo/RequestChain.swift
@@ -161,13 +161,12 @@ public class RequestChain: Cancellable {
       }
       return
     }
-    
-    
-    additionalHandler.handleErrorAsync(error: error,
-                                       chain: self,
-                                       request: request,
-                                       response: response,
-                                       completion: completion)
+
+    additionalHandler.handleErrorAsync(error: error, chain: self, request: request, response: response) { [weak self] result in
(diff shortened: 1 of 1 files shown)


Sources/Apollo/RequestChain.swift
```diff
@@ -162,12 +162,12 @@ public class RequestChain: Cancellable {
       return
     }
     
-    
-    additionalHandler.handleErrorAsync(error: error,
-                                       chain: self,
-                                       request: request,
-                                       response: response,
-                                       completion: completion)
+    let callbackQueue = self.callbackQueue
```
designatednerd, line 165: Why pull the callback queue out here rather than using `weak self`?

designatednerd (commented):
Ooh, good eye. Question on the implementation for ya.

Sources/Apollo/RequestChain.swift
ejensen: No real reason. It was just limiting the capture variables. Commit 2c275c1 switched the closure variable capturing to`[weak self]`

PR #1476, opened 2020-10-27
cancel task before reassigning and don't retain interceptor

RolandasRazma:
What do you think about such change? 
From previous interactions I know you might not like canceling bit, but if we don't cancel `currentTask` before reassign `public func cancel()` becomes meaningless

### DIFF
--- a/Sources/Apollo/NetworkFetchInterceptor.swift
+++ b/Sources/Apollo/NetworkFetchInterceptor.swift
@@ -29,7 +29,13 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    self.currentTask = self.client.sendRequest(urlRequest) { result in
+    self.currentTask?.cancel()
+    
+    self.currentTask = self.client.sendRequest(urlRequest) { [weak self] result in
+      guard let self = self else {
+        return
+	  }
+      
       defer {
         self.currentTask = nil
       }


Sources/Apollo/NetworkFetchInterceptor.swift
```diff
@@ -29,7 +29,13 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    self.currentTask = self.client.sendRequest(urlRequest) { result in
+    self.currentTask?.cancel()
+    
+    self.currentTask = self.client.sendRequest(urlRequest) { [weak self] result in
```
designatednerd, line 34: Good catch on this, I do think this should probably be a `[weak self]`.

Sources/Apollo/NetworkFetchInterceptor.swift
```diff
@@ -29,7 +29,13 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    self.currentTask = self.client.sendRequest(urlRequest) { result in
+    self.currentTask?.cancel()
```
designatednerd, line 32: Theoretically this should not be necessary since a new `NetworkFetchInterceptor` is created for each request chain.

designatednerd (commented):
To be clear, it's not that I'm philosophically opposed to cancellation, it's that I think it can sometimes obscure other issues when you do it preemptively when it's not necessary. In this case, it shouldn't really be reassigning, because the interceptor shouldn't be getting reused.

Sources/Apollo/NetworkFetchInterceptor.swift
RolandasRazma: > since a new NetworkFetchInterceptor is created for each request chain.
This one is interesting, I haven't seen anywhere that this is requirement. When I was migrating at first I just cached all interceptors but then thought that it might be the case that they "have to be new"

designatednerd: I added this to the docs on #1484

designatednerd:
I added the `weak self` to #1489  - going to close this one out in favor of that. Thank you!

PR #1478, opened 2020-10-27
Ironing JSONSerialization quirks

TizianoCoroneo:
Removed a workaround in the `jsonValue` property of Optional: the solution was to make `NSDictionary` conform to `JSONEncodable`.

Also fixed an issue where trying to serialize `.some(NSNull())` would result in a crash.

**BREAKING CHANGE** (?): This slightly changes the result of the JSON serialization, because instead of taking a `NSDictionary`, casting to `Dictionary` and then calling `jsonValue` on it, it directly uses the `NSDictionary` as the `JSONValue` itself.
Previously, this resulted in having `Bool` values encoded as `NSNumber`s, while now they are encoded as a `true` or `false` JSON constants. IMHO, this is slightly more correct.

### DIFF
--- a/Tests/ApolloTests/JSONTests.swift
+++ b/Tests/ApolloTests/JSONTests.swift
@@ -74,6 +74,14 @@ class JSONTests: XCTestCase {
 """)
   }
 
+  func testEncodingNSNullDoesNotCrash() throws {
+    let nsNull = ["aWeirdNull": NSNull()]
+    let serialized = try JSONSerializationFormat.serialize(value: nsNull)
+    let stringFromSerialized = try XCTUnwrap(String(data: serialized, encoding: .utf8))
+
+    XCTAssertEqual(stringFromSerialized, #"{"aWeirdNull":null}"#)
+  }
+
   func testEncodingOptionalNSNullDoesNotCrash() throws {
     let optionalNSNull = ["aWeirdNull": Optional.some(NSNull())]
     let serialized = try JSONSerializationFormat.serialize(value: optionalNSNull)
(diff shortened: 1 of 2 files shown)


Tests/ApolloTests/JSONTests.swift
```diff
@@ -70,7 +70,23 @@ class JSONTests: XCTestCase {
...
+    XCTAssertEqual(stringFromSerialized, #"{"aWeirdNull":null}"#)
+  }
+
+  func testEncodingDoubleOptionalsDoesNotCrash() throws {
+    let doubleOptional = ["aWeirdNull": Optional.some(Optional<Int>.none)]
+    let serialized = try JSONSerializationFormat.serialize(value: doubleOptional)
+    let stringFromSerialized = try XCTUnwrap(String(data: serialized, encoding: .utf8))
+
+    XCTAssertEqual(stringFromSerialized, #"{"aWeirdNull":null}"#)
+  }
```
designatednerd, line 91: Can you add a test for when someone passes just a straight up `NSNull()` instance in there please? So like 
`["aWeirdNull": NSNull()]`. I'd just like to make sure that won't explode.

designatednerd (commented):
I agree having `true`/`false` is more correct, so I'm fine with that change. 

In theory this shouldn't be breaking, since it shouldn't produce different results under the hood with the exception of `true`/`false` vs `1`/`0`.

Tests/ApolloTests/JSONTests.swift
TizianoCoroneo: done!

designatednerd (approved):

Issue #1473, opened 2020-10-26
Crash in 0.36.0


RolandasRazma:
## Bug report

Crash using interceptors

<img width="1024" alt="Screenshot 2020-10-26 at 18 37 05" src="https://user-images.githubusercontent.com/1027187/97213977-4857c680-17ba-11eb-902d-90d617d3ef97.png">

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.36.0
- Swift version: 5.3

## Steps to reproduce

I have TokenAddingInterceptor that retrieves token from server and asynchronously calls `chain.proceedAsync`. I have check for TokenAddingInterceptor being released to not call up the chain on dealloc, so that's not the case.

designatednerd:
The crash is happening in the `URLSessionClient` rather than in any of the interceptors - it looks like the client either a) Hasn't had a chance to set up its `URLSession` or b) has had the session torn down. 

Can you share how you're setting up your `TokenAddingService`?

RolandasRazma:
There is nothing interesting to see there. It just check for token validity if it it expired updates it from server and calls back. I will try to trace/fix crash as it's blocking us, but as its another race condition its not going to be simple as I can't even replicate it reliably. Sometimes it crashes every run (tests) sometimes I hit https://github.com/apollographql/apollo-ios/issues/1376 and sometimes all tests passes without any problem :)

RolandasRazma:
I think its same bug as https://github.com/apollographql/apollo-ios/issues/1376

I can replicate it with example from https://www.apollographql.com/docs/ios/tutorial/tutorial-mutations/

```swift
class TokenAddingInterceptor: ApolloInterceptor {
    func interceptAsync<Operation: GraphQLOperation>(chain: RequestChain, request: HTTPRequest<Operation>, response: HTTPResponse<Operation>?, completion: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void) {
        DispatchQueue.main.async { [weak self] in
            guard self != nil else { return }
            chain.proceedAsync(request: request, response: response, completion: completion)
        }
    }
    
}
```

RolandasRazma:
This is how memory graph looks for crashing client

<img width="445" alt="Screenshot 2020-10-27 at 10 39 03" src="https://user-images.githubusercontent.com/1027187/97290611-b0e98680-1840-11eb-8bf9-e3690946ecf1.png">

RolandasRazma:
[...]

PR #1480, opened 2020-10-28
URLSessionClientLiveTests testCleanup

RolandasRazma:
Test for crash described in https://github.com/apollographql/apollo-ios/issues/1473

### DIFF
--- a/Tests/ApolloTests/URLSessionClientTests.swift
+++ b/Tests/ApolloTests/URLSessionClientTests.swift
@@ -214,4 +214,27 @@ class URLSessionClientLiveTests: XCTestCase {
     let set = Set(taskIDs.value)
     XCTAssertEqual(set.count, iterations)
   }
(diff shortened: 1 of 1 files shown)


designatednerd:
@RolandasRazma Is this designed to be a failing test? It appears to be failing across the board.

RolandasRazma:
yes. It crashes everything - it shouldn't - Test to illustrate problem

designatednerd:
OK. Wasn't clear from your description if this was to reproduce the problem or act as a solution. Thanks for clarifying!

RolandasRazma:
we could fix crash by returning error if you try calling `sendRequest` after URLSessionClient is invalidated

designatednerd:
This was helpful for #1489, that should now validate that invalidating the session returns an error. Going to close this PR out in favor of that one.

PR #1483, opened 2020-10-28
Add dynamic library product to Package.swift

pwc3:
We have two products that depend upon Apollo for iOS:
- Our core iOS app
- A first-party framework we use in other iOS apps

We recently migrated to use Swift Package Manager in both the core app and in the framework and are in the process of integrating the framework into the core app. This causes an error to occur:

> Swift package product 'Apollo' is linked as a static library by '&lt;core app&gt;' and '&lt;framework&gt;'. This will result in duplication of library code.

This is a known issue when integrating Swift packages in Xcode. The [Swift Package Manager documentation](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md) states that:

> A library's product can either be statically or dynamically linked. If possible, don't declare the type of library explicitly to let the Swift Package Manager choose between static or dynamic linking based on the preference of the package's consumer.

However, Xcode links Swift packages as static libraries unless they are specified as dynamic libraries in their respective `Package.swift` files. ([Reference 1](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062), [Reference 2](https://forums.swift.org/t/migrating-to-spm-from-mix-of-embedded-frameworks-and-static-libraries/34253), [Reference 3](https://developer.apple.com/forums/thread/128806).)

This PR proposes a way to work around this issue: It provides an additional `Apollo-Dynamic` product that is simply the `Apollo` library with the `.dynamic` type specifier. This is similar approach to `Paper`/`PaperStatic`/`PaperDynamic` sample code provided in the [Swift Package Manager documentation](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#product).

This PR allows our core app and framework (and any others encountering this issue) to include the `Apollo-Dynamic` library and resolve the compiler/linker error. The `Apollo` product continues to act as a statically linked library.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,6 +13,10 @@ let package = Package(
       name: "Apollo",
       targets: ["Apollo"]),
     .library(
+        name: "Apollo-Dynamic",
+        type: .dynamic,
+        targets: ["Apollo"]),
+    .library(
       name: "ApolloCodegenLib",
       targets: ["ApolloCodegenLib"]),
     .library(


designatednerd:
Interesting! Thanks for the *very* thorough write-up. 

I'm curious: Why not have an intermediate framework that both `app` and `first-party-framework` depend on, and have that be where you put Apollo into the mix? That's definitely been a workaround to this issue I've seen repeatedly. 

This seems fairly reasonable, though I'm curious what would happen if you tried to add both `Apollo` and `Apollo-Dynamic`. Also, are you able to make this work with `ApolloCore` not being dynamic? There are bits of `Apollo` which depend on it.

pwc3:
I didn't want to go down the intermediate framework path for a couple of reasons. First, the core app and framework live in two different repositories. This would add a third repo into the mix. That's a solvable problem, though. I was more concerned about maintainability. Integrating with Apollo directly via SPM (or previously via CocoaPods), it's relatively simple to stay up to date with the latest releases. Having an intermediate framework in a separate repo would make staying up to date a bit more complicated.

This approach works without making `ApolloCore` dynamic. As I understand it, `ApolloCore` is statically linked to `Apollo` or `Apollo-Dynamic`. Having the core app and the framework depend on `Apollo-Dynamic` ultimately results in only a single copy of `Apollo-Dynamic` in the bundle.

I did an experiment to see what happens if I try to add both `Apollo` and `Apollo-Dynamic` to an app. This resulted in two immediate build errors:

* Swift package target 'ApolloCore' is linked as a static library by '<AppName>' and 'Apollo-Dynamic'. This will result in duplication of library code.
* Swift package target 'Apollo' is linked as a static library by '<AppName' and 'Apollo-Dynamic'. This will result in duplication of library code.

I considered proposing the following products:

```
.library(name: "Apollo", targets: ["Apollo"]),
.library(name: "Apollo-Static", type: .static, targets: ["Apollo"]),
.library(name: "Apollo-Dynamic", type: .dynamic, targets: ["Apollo"]),
```

This approach, [similar to an example in Apple's documentation](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#product), would leave `Apollo` without a specified linkage, allowing SPM to choose the linking. Then the other two products are explicitly statically or dynamically linked. I don't know if that makes things any clearer.

PR #1494, opened 2020-11-03
Fix mac playground compilation

dchohfi:


designatednerd (approved):

designatednerd:
@dchohfi thank you! Ugh I need to figure out a way to test those

PR #1498, opened 2020-11-06
Retain age of cache and result data

Mordil:
Thanks to @gsabran for doing practically all of this in #971!

## Changes
- Add: `RecordRow` struct that becomes the underlying storage base type for `RecordSet` that contains the `Record` and `lastReceivedAt` date
- Add: Mechanism to `SQLIteNormalizedCache` to handle schema migrations
- Add: `GraphQLResultContext` object for holding result metadata, such as the `resultAge`
- Add: `context` property to `GraphQLResult` that allows access to the result metadata
- Change: `NormalizedCache.loadRecords` completion to receive `RecordRow?` instead of `Record?`
- Change: `RecordSet` to use `RecordRow` rather than `Record`
- Change: `SQLiteNormalizedCache.init(fileURL:shouldVacuumOnClear:)` to be a convenience initializer
- Change: `SQLIteNormalizedCache` initializers to accept an optional `RecordSet`

## Motivation

> From @gsabran
>
> Reading from cache is a great way to reduce the number of network requests emitted, and cloud hosting costs 🤑. However developers usually want to make sure that the cache result is not too old, and without this information they are likely to fetch a fresh result just in case. This change makes it possible to know how old the returned result is.

```swift
apolloClient.fetch(
  query: query,
  cachePolicy: .returnCacheDataDontFetch,
  context: nil,
  queue: .main
) { result in
  switch result {
    case let .success(result):
      if result.context.resultAge > Date().milisecondsSince1970 - 3600 {
        // use fresh enough data
      } else {
        // fetch
      }
    case let .failure(error):
      // handle error
  }
}
```

## Tests

1. Database schema migration
1. Age of result validation

## Comments

The biggest difference between the implementations is that I went ahead and made the `GraphQLResultContext` a property of `GraphQLResult` which removes the need to change the handlers in `ApolloClient` and making this a larger breaking change


Mordil:
@designatednerd I _think_ I addressed all of the feedback you had left on #971

Mordil:
@gsabran If there's a different email you want associated with the commit author attribution - let me know and I'll update the commits

designatednerd:
I'm going to ask @martijnwalraven to also take a look at this when he's back on Tuesday - he's been neck-deep in cache stuff so I think he will have more opinions than I do.

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
```diff
@@ -8,6 +8,36 @@ import SQLite
...
+    CREATE TABLE IF NOT EXISTS "records" ("_id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "key" TEXT NOT NULL UNIQUE, "record" TEXT NOT NULL)
+    """)
+    try db.run("""
+    CREATE UNIQUE INDEX IF NOT EXISTS "index_records_on_key" ON "records" ("key")
+    """)
+
+    // Add en entry
+    try db.run("""
+    INSERT OR REPLACE INTO "records" ("key", "record") VALUES ("QUERY_ROOT.hero", "{""name"":""Luke Skywalker"",""__typename"":""Human""}")
+    """)
```
designatednerd, line 25: One thing I've found in the past that can be somewhat less error prone than doing this setup in a test yourself is to provide an actual `.sqlite` file from an old version of the database that you can then make a copy of and test migrations on. That way you guarantee you're using the format that is on older versions.

Tests/ApolloTests/BatchedLoadTests.swift
```diff
@@ -53,16 +53,21 @@ private final class MockBatchedNormalizedCache: NormalizedCache {
 class BatchedLoadTests: XCTestCase {  
   func testListsAreLoadedInASingleBatch() {
     var records = RecordSet()
-    let drones = (1...100).map { number in
-      Record(key: "Drone_\(number)", ["__typename": "Droid", "name": "Droid #\(number)"])
-    }
+    let drones = (1...100)
+      .map { Record(key: "Drone_\($0)", ["__typename": "Droid", "name": "Droid #\($0)"]) }
```
designatednerd, line 57: I think it would help to leave the named parameter on this for clarity, especially since you're also referring to a `$0` parameter in the 2nd map here.

designatednerd (commented):
My comments are mostly about formatting and dates. I have gone down way, way, way too many date rabbit holes. 

I definitely think @martijnwalraven will have more concrete feedback on the actual functionality here.

Sources/Apollo/GraphQLFirstReceivedAtTracker.swift
```diff
@@ -0,0 +1,27 @@
+import struct Foundation.Date
```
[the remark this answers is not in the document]
Mordil: Sorry, this is a habit of mine

Sources/Apollo/GraphQLFirstReceivedAtTracker.swift
```diff
@@ -0,0 +1,27 @@
...
+  func accept(fieldEntry: Date, info: GraphQLResolveInfo) throws -> Date {
+    return fieldEntry
+  }
+
+  func accept(fieldEntries: [Date], info: GraphQLResolveInfo) throws -> Date {
+    return fieldEntries.min() ?? Date(timeIntervalSince1970: 0)
+  }
+
+  func finish(rootValue: Date, info: GraphQLResolveInfo) throws -> GraphQLResultContext {
+    return .init(resultAge: rootValue)
```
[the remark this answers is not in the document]
Mordil: Yes, sorry.

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -359,4 +400,41 @@ class LoadQueryFromStoreTests: XCTestCase, CacheTesting {
...
+      self.store = .init(cache: $0)
+
+      self.load(query: HeroNameQuery()) {
+        switch $0 {
+        case let .success(result):
+          XCTAssertNil(result.errors, file: file, line: line)
+          XCTAssertEqual(result.data?.hero?.name, "R2-D2", file: file, line: line)
+          XCTAssertEqual(
+            expectedResultAge.timeIntervalSince1970,
+            result.context.resultAge.timeIntervalSince1970,
```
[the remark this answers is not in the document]
Mordil: Comparing dates directly doesn't support the `accuracy` parameter in `XCTAssertEqual`, unless I create an overload specifically for Date - which would you prefer?

Sources/Apollo/Math+Helpers.swift
```diff
@@ -0,0 +1,3 @@
+func min<T: Comparable>(_ a: T, _ b: T?) -> T {
+  return min(a, b ?? a)
```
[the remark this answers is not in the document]
Mordil: Sine this is a free-floating function, what would be your preferred approach? The use sites don't make a ton of sense of creating an array just to get the `min` function, but the extension property needs a type to be attached to.

Would you want an empty struct like `ApolloHelpers` with this being a static method?

gsabran (commented):
Wao this is exciting!! To avoid issues with the user changing their device clock, we can use something like:

```swift
/// Return a timestamp in ms from an arbitrary origin that is consistent for a given device.
func getCurrentTimestamp() -> Int {
  Int(clock_gettime_nsec_np(CLOCK_MONOTONIC) / 1000000)
}
```

Sources/ApolloTestSupport/TimeInterval+Helpers.swift
```diff
@@ -0,0 +1,7 @@
+import struct Foundation.TimeInterval
+
+extension TimeInterval {
+  public static var oneHour: Self { 3600 }
+  public static var oneDay: Self { 86400 }
+  public static var oneYear: Self { Self.oneDay * 365 }
```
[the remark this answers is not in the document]
Mordil: Yup, you're definitely right that this is naïve. I only contributed it because of it being in tests - but having these as extensions for all tests leaves a loaded gun for a future developer writing unit tests that are needing to be more accurate.

I've updated references to use `Calendar`

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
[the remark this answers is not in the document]
Mordil: I've updated to use the Calendar API per your suggestion

Tests/ApolloTests/BatchedLoadTests.swift
Mordil: Cleaned this up

Sources/Apollo/GraphQLFirstReceivedAtTracker.swift
```diff
@@ -0,0 +1,27 @@
...
+  func accept(scalar: JSONValue, firstReceivedAt: Date, info: GraphQLResolveInfo) throws -> Date {
+    return firstReceivedAt
+  }
+
+  func acceptNullValue(firstReceivedAt: Date, info: GraphQLResolveInfo) throws -> Date {
+    return firstReceivedAt
+  }
+
+  func accept(list: [Date], info: GraphQLResolveInfo) throws -> Date {
+    return list.min() ?? Date(timeIntervalSince1970: 0)
```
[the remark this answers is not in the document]
Mordil: Having `Date?` be the return type causes the need for verbose code at the use site for converting the result to `Date`, in type annotation and doing the logic that is done here anyway.

I've updated this to just use `.distantPast`

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
Mordil: Done, let me know what you think

Sources/Apollo/Math+Helpers.swift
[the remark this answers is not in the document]
designatednerd: Yeah maybe `ApolloMath` or something like that - I missed that this was a free-floating function 🤦‍♀️, but I suppose `min` is generally one. I just want to make sure it doesn't collide with future `min`s added to the stdlib

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
```diff
@@ -361,9 +361,9 @@ class LoadQueryFromStoreTests: XCTestCase, CacheTesting {
   }
 
   func testResultContextWithDataFromMixedDates() throws {
-    let yesterday = Date().addingTimeInterval(-.oneDay)
-    let oneHourAgo = Date().addingTimeInterval(-.oneHour)
-    let aYearAgo = Date().addingTimeInterval(-.oneYear)
+    let oneHourAgo = Calendar.current.date(byAdding: .hour, value: -1, to: Date())!
```
designatednerd, line 364: I would recommend puling `Date()` out at the top here just so these are all guaranteed to be making adjustments to the same date - in theory the `granularity` should help with that but I do feel like only creating one `Date` is probably less error-prone in general

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
```diff
@@ -160,3 +114,60 @@ class CachePersistenceTests: XCTestCase {
...
+                let newClient = ApolloClient(networkTransport: networkTransport, store: newStore)
+                newClient.fetch(query: query, cachePolicy: .returnCacheDataDontFetch) { newClientResult in
+                  defer { newCacheExpectation.fulfill() }
+                  switch newClientResult {
+                  case .success(let newClientGraphQLResult):
+                    XCTAssertEqual(newClientGraphQLResult.data?.hero?.name, "Luke Skywalker", file: file, line: line)
+                  case .failure(let error):
+                    XCTFail("Unexpected error with new client: \(error)", file: file, line: line)
+                  }
+                  _ = newClient // Workaround for a bug - ensure that newClient is retained until this block is run
```
designatednerd, line 163: I don't necessarily think this is a bug - but I would leave the note about ensuring it's retained until the block is run.

designatednerd (commented):
Awesome, only a couple comments - will wait for Martijn to get a look Tuesday before proceeding. Thanks for the quick turnaround!

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
designatednerd: I absolutely love it. Thank you!

designatednerd:
@gsabran What's the timestamp for?

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
Mordil: This was existing code that I just moved into a shared function so that I could pass the parameter but also make sure that the unit test is picked up by the XCTest sweep.

Do you still want me to update the comment?

Tests/ApolloCacheDependentTests/LoadQueryFromStoreTests.swift
Mordil: Done! This is definitely smarter to do as it would make it less brittle

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
designatednerd: Yes please update the comment, I don't want me or anyone else looking for the bug that doesn't exist :)

Mordil: Done!

Sources/Apollo/GraphQLResultAccumulator.swift
```diff
@@ -22,6 +24,10 @@ func zip<Accumulator1: GraphQLResultAccumulator, Accumulator2: GraphQLResultAccu
   return Zip3Accumulator(accumulator1, accumulator2, accumulator3)
 }
 
+func zip<Accumulator1: GraphQLResultAccumulator, Accumulator2: GraphQLResultAccumulator, Accumulator3: GraphQLResultAccumulator, Accumulator4: GraphQLResultAccumulator>(_ accumulator1: Accumulator1, _ accumulator2: Accumulator2, _ accumulator3: Accumulator3, _ accumulator4: Accumulator4) -> Zip4Accumulator<Accumulator1, Accumulator2, Accumulator3, Accumulator4> {
+  return Zip4Accumulator(accumulator1, accumulator2, accumulator3, accumulator4)
+}
+
```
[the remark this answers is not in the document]
TizianoCoroneo: I think this is still needed, if `Accumulator1`, `Accumulator2`, ... are different concrete types. There might be some changes to Swift generics in the future to remove this code, specifically ["Variadic Generics"](https://github.com/apple/swift/blob/main/docs/GenericsManifesto.md#variadic-generics), but that's still a long way to go.

Sources/Apollo/ApolloStore.swift
```diff
@@ -206,74 +208,86 @@ public final class ApolloStore {
...
-    fileprivate lazy var loader: DataLoader<CacheKey, Record?> = DataLoader(self.cache.loadRecordsPromise)
+    fileprivate lazy var loader: DataLoader<CacheKey, RecordRow?> = DataLoader(self.cache.loadRecordsPromise)
 
     init(cache: NormalizedCache, cacheKeyForObject: CacheKeyForObject?) {
       self.cache = cache
       self.cacheKeyForObject = cacheKeyForObject
     }
 
-    public func read<Query: GraphQLQuery>(query: Query) throws -> Query.Data {
+    public func read<Query: GraphQLQuery>(query: Query) throws -> (Query.Data, GraphQLResultContext) {
```
martijnwalraven, line 218: I feel returning a tuple makes the use of the store read API a bit awkward, because you're often just interested in the data, and now you have to destructure the returned tuple explicitly everywhere. Not sure if the alternative is any better though. I was thinking of returning `GraphQLResult` from all store read methods instead, but that means you have to go through `data` even if you're doing a `readObject` for an individual fragment.

martijnwalraven: Maybe a better question to ask is: do we need the public read store API to provide access to metadata at all? You can always use a `fetch` with `.returnCacheDataDontFetch` to read a result from the cache. I think the main purpose of the store read API is for manual store manipulation within writes, so that may not require the metadata?

martijnwalraven (commented):
Thanks for this PR! This is definitely a worthwhile feature to add. My comments are mainly about the way we want to extend the existing API in a way that sets us up for future extensions. 

The TL;DR is that I'm wary of adding specific pieces of metadata like `firstReceivedAt` to the core of the execution pipeline, because that has repercussions everywhere (it changes the resolver return value, needs to be passed around in all execution methods, the accumulator, etc.). Instead, I think we should take this opportunity to come up with a general metadata mechanism that we can take advantage of for other use cases.

A few additional pieces of feedback that didn't have a clear place to leave them as comments:

- The term `firstReceivedAt` seems a bit confusing to me. Maybe `lastUpdateFromServer` would be a better description?
- Although I think it makes sense to give people access to `firstReceivedAt` from a result when they have a need for custom logic, the example code listed under Motivation seems like the main use case and I think we should have explicit support for it. Maybe some kind of `maxAge` condition on `.returnCacheDataDontFetch`, or a general way to pass a predicate for a conditional fetch (a la HTTP Cache Control)? It could also be something you configure (overridable) defaults for on `ApolloClient`. Besides avoiding repetitive code everywhere a result is received, that would also avoid an unnecessary switch to the main queue and back in case the data is stale and we need to go to the network after all.

Sources/Apollo/RecordSet.swift
```diff
@@ -1,28 +1,37 @@
+import Foundation
+
+/// A row of data that contains a `Record` and some associated metadata.
+public struct RecordRow {
+  public internal(set) var record: Record
+  public internal(set) var lastReceivedAt: Date
```
martijnwalraven, line 6: Is it fine grained enough to keep this per record or do we need the ability to update this on a per field basis? If a query only fetches a subset of the fields, we still need to know that the fields that weren't fetched can be stale (their `lastUpdated` hasn't changed).

Sources/Apollo/ApolloStore.swift
```diff
@@ -324,13 +338,14 @@ public final class ApolloStore {
                        variables: GraphQLMap?) throws {
       let normalizer = GraphQLResultNormalizer()
       let executor = GraphQLExecutor { object, info in
-        return .result(.success(object[info.responseKeyForField]))
+        return .result(.success((object[info.responseKeyForField], Date())))
```
martijnwalraven, line 341: I don't think we always want to update the `lastReceivedAt`with the current date for every write, because there's often a need to differentiate between server updates and (optimistic) client updates.

designatednerd:
So, with this PR, we've run into a bit of an existential dilemma, in that adding a timestamp at the record level isn't really doable, since there may be fields in one operation that get updated as part of a different operation. Ultimately, to make any kind of TTL mechanism work properly, what we need is field-level metadata (and we probably want it to be opt-in to reduce potential unnecessary overhead).

There's also some stuff I know @martijnwalraven wants to address with how normalization works overall, and depending on which direction we ultimately decide with it, there are pretty significant changes 

Ultimately, this PR isn't going to be it, and I don't want to get anyone's hopes up by leaving this open. I'm going to open a new issue on the repo so we can track this as a feature request - I'll link it back to here. 

I really want to thank @gsabran and @Mordil for getting this going - I know it's frustrating that this isn't going to get merged, but we think we've got a line on a much more flexible (and accurate) way of doing this for the future.

designatednerd:
Please follow #1568 for further updates. Thank you!

PR #1508, opened 2020-11-10
Remove redundant nil check in InMemoryNormalizedCache's clear()

gsbernstein:
`returnResultAsyncIfNeeded` already checks for `nil`. In all the other methods in this class, we rely on that check instead of doing it ourselves beforehand, so this is also more consistent.


designatednerd (approved):

designatednerd:
Good catch!

PR #1512, opened 2020-11-12
Add 'CacheClearingPolicy' for more control over how the cache is cleared

Mordil:
Motivation:

It is common to want finer control over which records are removed from the cache without just deleting all records.

This modifies the framework to support different policies that a developer can specify when calling `clear` and `clearImmediately`.

Modifications:

- Rename: `shouldVacuumOnClear` argument label in `SQLiteNormalizedCache` initializers to `compactFileOnClear` to make it clearer for those unfamiliar with SQLite's terminology
- Add: `CacheClearingPolicy` with policies for removing the first/last K records, key pattern matching, and the current behavior of removing all records
- Add: `ApolloStore.clearCache` overload that accepts a `CacheClearingPolicy`
- Change: `RecordSet.clear` to accept a `CacheClearingPolicy`
- Change: `NormalizedCache.clear` and `NormalizedCache.clearImmediately` to accept a `CacheClearingPolicy`
  - The current methods are available as overloads
- Change: `ApolloClientProtocol.clearCache` to accept a `CacheClearingPolicy`
  - The current method is available as an overload

Result:

Developers can now have more control over which records to clear from their cache:

```swift
// current behavior
client.clearCache(usingPolicy: .allRecords)

// delete the first 10 records
client.clearCache(usingPolicy: .first(10))

// delete all hero records
client.clearCache(usingPolicy: .allMatchingKeyPattern("*hero*"))
```

### DIFF
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -81,16 +81,3 @@ public protocol ApolloClientProtocol: class {
                                                     queue: DispatchQueue,
                                                     resultHandler: @escaping GraphQLResultHandler<Subscription.Data>) -> Cancellable
 }
-
-// MARK: conveniences
-
-extension ApolloClientProtocol {
-  /// Clears all records from the cache store.
-  /// - Warning: The cache may be used by other clients. Calling this method will affect all clients using the same cache!
-  /// - Parameters:
-  ///   - callbackQueue: An optional queue to execute the completion handler on. The default is `.main`.
-  ///   - completion: An optional completion closure to execute when the cache has been cleared. The default is `nil`.
-  func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, Error>) -> Void)? = nil) {
-    self.clearCache(usingPolicy: .allRecords, callbackQueue: callbackQueue, completion: completion)
-  }
-}
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -9,13 +9,15 @@ public protocol ApolloClientProtocol: class {
   /// A function that returns a cache key for a particular result object. If it returns `nil`, a default cache key based on the field path will be used.
   var cacheKeyForObject: CacheKeyForObject? { get set }
 
-  /// Clears the underlying cache.
-  /// Be aware: In more complex setups, the same underlying cache can be used across multiple instances, so if you call this on one instance, it'll clear that cache across all instances which share that cache.
-  ///
+  /// Clears the cache store according to the specified policy.
+  /// - Warning: The cache may be used by other clients. Calling this method will affect all clients using the same cache!
   /// - Parameters:
-  ///   - callbackQueue: The queue to fall back on. Should default to the main queue.
-  ///   - completion: [optional] A completion closure to execute when clearing has completed. Should default to nil.
-  func clearCache(callbackQueue: DispatchQueue, completion: ((Result<Void, Error>) -> Void)?)
+  ///   - policy: The cache cleaning policy to use.
+  ///   - callbackQueue: An optional queue to execute the completion handler on. Should default to the `.main` queue.
+  ///   - completion: An optional completion closure to execute when the cache has been cleared. Should default to `nil`.
+  func clearCache(usingPolicy policy: CacheClearingPolicy,
+                  callbackQueue: DispatchQueue,
+                  completion: ((Result<Void, Error>) -> Void)?)
 
   /// Fetches a query from the server or from the local cache, depending on the current contents of the cache and the specified cache policy.
   ///
@@ -78,4 +80,17 @@ public protocol ApolloClientProtocol: class {
   func subscribe<Subscription: GraphQLSubscription>(subscription: Subscription,
                                                     queue: DispatchQueue,
                                                     resultHandler: @escaping GraphQLResultHandler<Subscription.Data>) -> Cancellable
+}
+
+// MARK: conveniences
+
+extension ApolloClientProtocol {
+  /// Clears all records from the cache store.
+  /// - Warning: The cache may be used by other clients. Calling this method will affect all clients using the same cache!
+  /// - Parameters:
+  ///   - callbackQueue: An optional queue to execute the completion handler on. The default is `.main`.
+  ///   - completion: An optional completion closure to execute when the cache has been cleared. The default is `nil`.
+  func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, Error>) -> Void)? = nil) {
+    self.clearCache(usingPolicy: .allRecords, callbackQueue: callbackQueue, completion: completion)
+  }
 }
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -1,10 +1,19 @@
+/// Enumeration of the possible GraphQL operations that can be executed against an API endpoint.
 public enum GraphQLOperationType {
   case query
   case mutation
   case subscription
+
+  var cacheKey: String {
+    switch self {
+    case .query: return "QUERY_ROOT"
+    case .mutation: return "MUTATION_ROOT"
+    case .subscription: return "SUBSCRIPTION_ROOT"
+    }
+  }
 }
 
-public protocol GraphQLOperation: class {
+public protocol GraphQLOperation: AnyObject {
   var operationType: GraphQLOperationType { get }
 
   var operationDefinition: String { get }
@@ -32,6 +41,8 @@ public extension GraphQLOperation {
   }
 }
 
+// - MARK: Conformances
+
 public protocol GraphQLQuery: GraphQLOperation {}
 public extension GraphQLQuery {
   var operationType: GraphQLOperationType { return .query }
@@ -45,9 +56,4 @@ public extension GraphQLMutation {
 public protocol GraphQLSubscription: GraphQLOperation {}
 public extension GraphQLSubscription {
   var operationType: GraphQLOperationType { return .subscription }
-}
-
-public protocol GraphQLFragment: GraphQLSelectionSet {
-  static var fragmentDefinition: String { get }
-  static var possibleTypes: [String] { get }
 }
--- a/Sources/Apollo/GraphQLOperation.swift
+++ b/Sources/Apollo/GraphQLOperation.swift
@@ -4,7 +4,7 @@ public enum GraphQLOperationType {
   case mutation
   case subscription
 
-  var cacheKey: String {
+  var rootCacheKey: String {
     switch self {
     case .query: return "QUERY_ROOT"
     case .mutation: return "MUTATION_ROOT"
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -4,16 +4,14 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
   private let recordsLock = NSRecursiveLock()
 
-  public init(records: RecordSet = RecordSet()) {
+  public init(records: RecordSet = .init()) {
     self.records = records
   }
 
   public func loadRecords(forKeys keys: [CacheKey],
                           callbackQueue: DispatchQueue?,
                           completion: @escaping (Result<[Record?], Error>) -> Void) {
-    self.recordsLock.lock()
-    let records = keys.map { self.records[$0] }
-    self.recordsLock.unlock()
+    let records = self.threadSafe { keys.map{ self.records[$0] } }
     DispatchQueue.apollo.returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(records))
@@ -22,25 +20,31 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   public func merge(records: RecordSet,
                     callbackQueue: DispatchQueue?,
                     completion: @escaping (Result<Set<CacheKey>, Error>) -> Void) {
-    self.recordsLock.lock()
-    let cacheKeys = self.records.merge(records: records)
-    self.recordsLock.unlock()
+    let cacheKeys = self.threadSafe { self.records.merge(records: records) }
     DispatchQueue.apollo.returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(cacheKeys))
   }
 
-  public func clear(callbackQueue: DispatchQueue?,
-                    completion: ((Result<Void, Error>) -> Void)?) {
-    clearImmediately()
+  public func clear(
+    _ clearingPolicy: CacheClearingPolicy,
+    callbackQueue: DispatchQueue?,
+    completion: ((Result<Void, Error>) -> Void)?
+  ) {
+    self.clearImmediately(clearingPolicy)
     DispatchQueue.apollo.returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(()))
   }
 
-  public func clearImmediately() {
+  public func clearImmediately(_ clearingPolicy: CacheClearingPolicy) {
+    self.threadSafe { self.records.clear(clearingPolicy) }
+  }
+
+  private func threadSafe<Result>(_ operation: () -> Result) -> Result {
     self.recordsLock.lock()
-    self.records.clear()
+    let result = operation()
     self.recordsLock.unlock()
+    return result
   }
 }
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -4,7 +4,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
   private let recordsLock = NSRecursiveLock()
 
-  public init(records: RecordSet = .init()) {
+  public init(records: RecordSet = RecordSet()) {
     self.records = records
   }
 
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -2,7 +2,6 @@ import Foundation
 
 public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
-  private let recordsLock = NSRecursiveLock()
 
   public init(records: RecordSet = RecordSet()) {
     self.records = records
@@ -11,7 +10,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   public func loadRecords(forKeys keys: [CacheKey],
                           callbackQueue: DispatchQueue?,
                           completion: @escaping (Result<[Record?], Error>) -> Void) {
-    let records = self.threadSafe { keys.map{ self.records[$0] } }
+    let records = keys.map { self.records[$0] }
     DispatchQueue.apollo.returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(records))
@@ -20,7 +19,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   public func merge(records: RecordSet,
                     callbackQueue: DispatchQueue?,
                     completion: @escaping (Result<Set<CacheKey>, Error>) -> Void) {
-    let cacheKeys = self.threadSafe { self.records.merge(records: records) }
+    let cacheKeys = self.records.merge(records: records)
     DispatchQueue.apollo.returnResultAsyncIfNeeded(on: callbackQueue,
                                                    action: completion,
                                                    result: .success(cacheKeys))
@@ -38,13 +37,6 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   }
 
   public func clearImmediately(_ clearingPolicy: CacheClearingPolicy) {
-    self.threadSafe { self.records.clear(clearingPolicy) }
-  }
-
-  private func threadSafe<Result>(_ operation: () -> Result) -> Result {
-    self.recordsLock.lock()
-    let result = operation()
-    self.recordsLock.unlock()
-    return result
+    self.records.clear(clearingPolicy)
   }
 }
--- a/Sources/Apollo/NormalizedCache.swift
+++ b/Sources/Apollo/NormalizedCache.swift
@@ -1,5 +1,46 @@
 import Foundation
 
+/// An object that drives behavior of clear operations on a given cache.
+public struct CacheClearingPolicy: Equatable {
+  /// Clears all records in the cache.
+  public static let allRecords = CacheClearingPolicy(.allRecords)
+  /// Clears all records whose key matches the provided glob pattern.
+  /// - Parameter pattern: The glob pattern to use for key matching. For example `*pollo` will match both `Apollo` and `pollo`.
+  public static func allMatchingKeyPattern(_ pattern: String) -> CacheClearingPolicy { .init(.allMatchingKeyPattern(pattern)) }
+  /// Clears the the first (oldest) records in the cache up to the given limit.
+  /// - Parameter k: The number of records to remove from the cache.
+  public static func first(_ k: Int) -> CacheClearingPolicy { .init(.first(k)) }
+  /// Clears the the last (most recent) records in the cache up to the given limit.
+  /// - Parameter k: The number of records to remove from the cache.
+  public static func last(_ k: Int) -> CacheClearingPolicy { .init(.last(k)) }
+
+  // actual policy storage
+
+  /// This is public due to language requirements for cross-target access.
+  /// Do not use this value, nor its type as they are an implementation detail.
+  public let _value: _Policy
+  private init(_ policy: _Policy) { self._value = policy }
+
+  /// This is public due to language requirements for cross-target access.
+  /// Do not use this type, nor any of its values, as they are an implementation detail.
+  public enum _Policy: Equatable {
+    case allRecords
+    case allMatchingKeyPattern(String)
+    case first(Int)
+    case last(Int)
+
+    public static func ==(lhs: _Policy, rhs: _Policy) -> Bool {
+      switch (lhs, rhs) {
+      case (.allRecords, .allRecords): return true
+      case let (.allMatchingKeyPattern(left), .allMatchingKeyPattern(right)): return left == right
+      case let (.first(left), .first(right)): return left == right
+      case let (.last(left), .last(right)): return left == right
+      default: return false
+      }
+    }
+  }
+}
+
 public protocol NormalizedCache {
 
   /// Loads records corresponding to the given keys.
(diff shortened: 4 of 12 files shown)


Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -79,3 +81,16 @@ public protocol ApolloClientProtocol: class {
...
+
+// MARK: conveniences
+
+extension ApolloClientProtocol {
+  /// Clears all records from the cache store.
+  /// - Warning: The cache may be used by other clients. Calling this method will affect all clients using the same cache!
+  /// - Parameters:
+  ///   - callbackQueue: An optional queue to execute the completion handler on. The default is `.main`.
+  ///   - completion: An optional completion closure to execute when the cache has been cleared. The default is `nil`.
+  func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, Error>) -> Void)? = nil) {
```
designatednerd, line 93: One thing to watch out for with this that's bitten me on other "convenience" implementations of protocols where i'm setting up default parameters is that if you add more parameters to `clearCache` in the future, this may wind up calling a different method than you're expecting. [Here's an example of where that bit me recently](https://github.com/apollographql/apollo-ios/pull/1450).

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -1,10 +1,19 @@
+/// Enumeration of the possible GraphQL operations that can be executed against an API endpoint.
 public enum GraphQLOperationType {
   case query
   case mutation
   case subscription
+
+  var cacheKey: String {
```
designatednerd, line 7: i would call this `rootCacheKey` to more clearly emphasize that this is not the whole cache key.

Sources/Apollo/GraphQLOperation.swift
```diff
@@ -1,10 +1,19 @@
...
+    switch self {
+    case .query: return "QUERY_ROOT"
+    case .mutation: return "MUTATION_ROOT"
+    case .subscription: return "SUBSCRIPTION_ROOT"
+    }
+  }
 }
 
-public protocol GraphQLOperation: class {
+public protocol GraphQLOperation: AnyObject {
```
designatednerd, line 16: Why this change?

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -4,16 +4,14 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   private var records: RecordSet
   private let recordsLock = NSRecursiveLock()
 
-  public init(records: RecordSet = RecordSet()) {
+  public init(records: RecordSet = .init()) {
```
designatednerd, line 7: reminder: Please do not use `.init`, please use `RecordSet()`

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -4,16 +4,14 @@ public final class InMemoryNormalizedCache: NormalizedCache {
...
     self.records = records
   }
 
   public func loadRecords(forKeys keys: [CacheKey],
                           callbackQueue: DispatchQueue?,
                           completion: @escaping (Result<[Record?], Error>) -> Void) {
-    self.recordsLock.lock()
-    let records = keys.map { self.records[$0] }
-    self.recordsLock.unlock()
+    let records = self.threadSafe { keys.map{ self.records[$0] } }
```
designatednerd, line 14: @martijnwalraven I think there was some point at which you were planning on removing the locks here - what's your take on whether we should keep them or not?

Sources/Apollo/NormalizedCache.swift
```diff
@@ -1,5 +1,46 @@
...
 
+/// An object that drives behavior of clear operations on a given cache.
+public struct CacheClearingPolicy: Equatable {
+  /// Clears all records in the cache.
+  public static let allRecords = CacheClearingPolicy(.allRecords)
+  /// Clears all records whose key matches the provided glob pattern.
+  /// - Parameter pattern: The glob pattern to use for key matching. For example `*pollo` will match both `Apollo` and `pollo`.
+  public static func allMatchingKeyPattern(_ pattern: String) -> CacheClearingPolicy { .init(.allMatchingKeyPattern(pattern)) }
+  /// Clears the the first (oldest) records in the cache up to the given limit.
+  /// - Parameter k: The number of records to remove from the cache.
```
designatednerd, line 11: i would strongly recommend making this something like `numberOfRecords` instead of `k`, particularly since you're using the `_` for the external parameter.

Sources/Apollo/RecordSet.swift
```diff
@@ -10,8 +10,17 @@ public struct RecordSet {
     storage[record.key] = record
   }
 
-  public mutating func clear() {
-    storage.removeAll()
+  public mutating func clear(_ clearingPolicy: CacheClearingPolicy = .allRecords) {
+    switch clearingPolicy._value {
+    case let .first(count): self.storage = .init(self.storage.dropFirst(count))
```
designatednerd, line 15: Please don't keep these on the same line, it makes it a lot harder to read.

Sources/Apollo/RecordSet.swift
```diff
@@ -10,8 +10,17 @@ public struct RecordSet {
...
+
+    case let .last(count): self.storage = .init(self.storage.dropLast(count))
+
+    case let .allMatchingKeyPattern(pattern): self.storage = self.storage.filter { !$0.key.contains(pattern) }
```
designatednerd, line 19: Does `contains` automatically handle globs?

Sources/Apollo/NormalizedCache.swift
```diff
@@ -1,5 +1,46 @@
 import Foundation
 
+/// An object that drives behavior of clear operations on a given cache.
+public struct CacheClearingPolicy: Equatable {
```
designatednerd, line 4: Can you explain why this struct exists rather than just exposing the `Policy` enum directly, since they seem almost identical?

Sources/Apollo/RecordSet.swift
```diff
@@ -10,8 +10,17 @@ public struct RecordSet {
...
+
+    case .allRecords: fallthrough
+    default: self.storage.removeAll()
```
designatednerd, line 22: If you exposed the enum directly, you'd be able to remove this `default` on the switch statement.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -19,28 +19,29 @@ public final class SQLiteNormalizedCache {
...
   /// - Parameters:
-  ///   - db: The database Connection to use
-  ///   - shouldVacuumOnClear: If the database should also be `VACCUM`ed on clear to remove all traces of info. Defaults to `false` since this involves a performance hit, but this should be used if you are storing any Personally Identifiable Information in the cache.
-  /// - Throws: Any errors attempting to access the database
-  public init(db: Connection, shouldVacuumOnClear: Bool = false) throws {
-    self.shouldVacuumOnClear = shouldVacuumOnClear
-    self.db = db
-    try self.createTableIfNeeded()
+  ///   - fileURL: The file URL to use for your database.
+  ///   - shouldVacuumOnClear: If the database file should compact "VACUUM" its storage when clearing records. This can be a potentially long operation.
```
designatednerd, line 37: the name of this parameter should be updated to reflect its external value

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -95,11 +96,30 @@ public final class SQLiteNormalizedCache {
...
   }
 
-  private func clearRecords() throws {
-    try self.db.run(records.delete())
-    if self.shouldVacuumOnClear {
-      try self.db.prepare("VACUUM;").run()
+  private func clearRecords(accordingTo policy: CacheClearingPolicy) throws {
+    switch policy._value {
+    case let .first(count):
+      let firstKRecords = records.select(self.id).order(self.id.asc).limit(count)
```
designatednerd, line 102: a) Mostly out of curiosity, why `K` rather than `N`?
b) for what it's worth I find these types  of things a lot easier to read if they're on separate lines so that you can see each of the steps more granularly

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -95,11 +96,30 @@ public final class SQLiteNormalizedCache {
     return try self.db.prepare(query).map { try parse(row: $0) }
   }
 
-  private func clearRecords() throws {
-    try self.db.run(records.delete())
-    if self.shouldVacuumOnClear {
-      try self.db.prepare("VACUUM;").run()
+  private func clearRecords(accordingTo policy: CacheClearingPolicy) throws {
```
designatednerd, line 99: Optional: I think `with` instead of `accordingTo` would be more in line with other stuff in the lib already, but definitely not a hill i want to die on.

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
```diff
@@ -63,70 +63,105 @@ class CachePersistenceTests: XCTestCase {
...
+    let notFoundExpectation = self.expectation(description: "records should not exist")
+
+    self.testCacheClearing(withPolicy: .allMatchingKeyPattern("*hero*")) { client, _ in
+      client.store.withinReadTransaction {
+        $0.loadRecords(forKeys: ["QUERY_ROOT.hero", "QUERY_ROOT.hero(episode:EMPIRE)"]) { result in
+          defer { notFoundExpectation.fulfill() }
+
+          do {
+            let results = try result.get()
+            XCTAssertTrue(results.allSatisfy({ $0 == nil }))
```
designatednerd, line 81: why should these ones be `nil` instead of getting a `.missingValue` error?

designatednerd (commented):
Again definitely recommend talking through this with @martijnwalraven, but this seems like a reasonable complement to what you're working on in #1498

Sources/Apollo/GraphQLOperation.swift
Mordil: I was in this file at some previous iteration, but here's the rationale: `protocol P: class` is considered deprecated without a warning

https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507/4

Sources/Apollo/InMemoryNormalizedCache.swift
Mordil: Sorry! style habit...

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
Mordil: Is that the way you prefer?

In the SSS ecosystem (and if you look at a lot of Apple code as well) they use the parameter label rather than the argument label and the Xcode Quick Help properly labels the documentation with the argument label.

I can make this change, just wanted to double check 😄

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
Mordil: `Collection.dropFirst` uses the `k` parameter label terminology because AFAIK the "term of art" in the functional world of a "count"

My first instinct is to follow the Swift Standard Library for naming things.

I'm not strongly invested to any of this naming - if you want `N` to be used instead I'll make the change

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
Mordil: Either reads fine, but `with` is shorter so I can make the quick update.

In my work's codebase we use `with` a lot that is overloaded so I have a slight aversion to that argument label :sweat_smile:

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
Mordil: I wasn't getting `.missingValue` I was getting `[nil, nil]`, and I was coding at 02:00 my time so admittedly this was a bit rushed and not written to a high quality 😅

Sources/Apollo/RecordSet.swift
Mordil: 🤦 No, this is my bad. I'll need to remove the globs and then this will work.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
designatednerd: oh! I've been updating it manually like a sucker 🙃

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
designatednerd: fair enough on it matching the standard lib, don't need to change the name.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
designatednerd: Yeah we definitely use it a lot as well, but I think if you're not just going to say `policy` for both the internal and external param names, `with` is probably closer to how we want it to read at the call site

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
designatednerd: so that indicates that the values stored for each key are `nil`, rather than the whole key and value being removed, which is what I believe should happen. Otherwise you could potentially get a `nil` value for something that's non-nullable when you're trying to pull it out of the cache.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
Mordil: I'm a liar. I just tested and I was mistaken - Quick Help does _not_ use the argument label and instead uses the parameter label in documentation.

_However_, usually the parameter label tends to be more descriptive in a documentation context on how the property will be used, while the argument label would not.

For example, from [RediStack](https://gitlab.com/Mordil/RediStack/-/blob/833fc338816746d5b532a66a3a3129649ef64f9e/Sources/RediStack/Commands/BasicCommands.swift#L85):
```swift
public func authorize(with password: String) -> EventLoopFuture<Void>

// call site
client.authorize(with: myPassword)

// documentation
password
The password to authenticate with.
```

I guess it's definitely up to personal preference, and I'm happy to update these to match Apollo's preference

designatednerd: Ahhh, yeah I suppose that's fair. Yeah just leave it. IGNORE ME.

Sources/Apollo/GraphQLOperation.swift
designatednerd: Hm...I'll ping Slava about that on twitter, that post is from 2018 and I wonder why, if that's the case, they haven't added the warning

designatednerd: Sayeth the oracle: [Still the same](https://twitter.com/slava_pestov/status/1326976867627380744)

Sources/Apollo/NormalizedCache.swift
Mordil: I knew this would come up, and was preparing for this. We can discussion synchronously outside of this PR if you want about it 😄 

I admit that this isn't the most elegant implementation, as I've seen some other, but it is the simplest for what I was hoping to help achieve with it.

This is a pattern that I've seen widely accepted and adopted in the Server-side Swift realm that assists in giving library evolution to SwiftPM static libraries, as adding new cases are technically breaking changes.

When you use a enum-like-struct new cases can arise and be added without much worry for your users. Dave DeLong tweeted about it a few weeks ago: https://twitter.com/davedelong/status/1319368474444005376

A blog post on the anti-pattern nature of raw enums: https://www.jessesquires.com/blog/2016/07/31/enums-as-configs/

A lot of this can also be found in this forum discussion: https://forums.swift.org/t/extensible-enumerations-for-non-resilient-libraries/35900

Tests/ApolloSQLiteTests/CachePersistenceTests.swift
Mordil: The difference in the two unit tests is that one is going through the whole GraphQL executor by calling `fetch` on the client, while the other (the one this comment was left on) was accessing the cache direction in a read transaction.

I've re-written this to be more consistent and more direct at what's being looked for.

Sources/Apollo/NormalizedCache.swift
designatednerd: Super-hard disagree on this one. I would _much_ rather people have to consciously handle any additional cases (especially when most of the handling is going to be internal to the library) than have default behavior which doesn't throw a warning when a new case is added. In my opinion the breaking behavior of this is a feature, not a bug. 

Enums certainly aren't the solution for everything, but in this case where we have: 

- A fixed set of cases we're handling
- Not allowing the user to create subclasses which can do their own handling
- Not adding any additional properties that can't be set up as an associated object
- Using this as a way to tell the internals of our library what to do rather than as a way to say "Here are the options of what this could be" to an external library

an `enum` is absolutely what we want here.

Sources/Apollo/ApolloClientProtocol.swift
designatednerd: I'd say default in the Apollo conformances - that's what most people are using rather than creating their own client.

Mordil: 👍 Done!

Sources/Apollo/NormalizedCache.swift
Mordil: This is fair, and just wanted to do it this way first to see the discussion 👍 I've made the change to a plain enum

Sources/Apollo/InMemoryNormalizedCache.swift
martijnwalraven: We should definitely remove those locks (and from `InMemoryNormalizedCache` as well). They serve no real useful purpose, because transactions only exist at the level of the store, and that's the proper level of isolation. You should be able to share a store between clients, but a cache should never be shared.

Sources/Apollo/NormalizedCache.swift
```diff
@@ -1,5 +1,19 @@
...
+/// An enumeration of behaviors for clear operations on a given cache.
+public enum CacheClearingPolicy: Equatable {
+  /// Clears all records in the cache.
+  case allRecords
+  /// Clears all records whose key matches the provided glob pattern.
+  ///
+  /// For example `*pollo` will match both `Apollo` and `pollo`.
+  case allMatchingKeyPattern(String)
+  /// Clears the the first (oldest) records in the cache up to the given limit.
+  case first(Int)
```
martijnwalraven, line 12: Hmmm, I don't think we currently have a way to support the `first(Int)` and `last(Int)` policies. The implementations below use an unspecified order (for the in memory cache) or orthographic order of the key (for the SQLite cache), and neither maps to oldest or most recent. This probably has to wait until we have resolved https://github.com/apollographql/apollo-ios/pull/1498.

Sources/Apollo/RecordSet.swift
```diff
@@ -10,8 +10,20 @@ public struct RecordSet {
     storage[record.key] = record
   }
 
-  public mutating func clear() {
-    storage.removeAll()
+  public mutating func clear(_ clearingPolicy: CacheClearingPolicy = .allRecords) {
+    switch clearingPolicy {
+    case let .first(count):
+      self.storage = Dictionary(self.storage.dropFirst(count))
```
martijnwalraven, line 16: As far as I know, `Dictionary` has no guaranteed order, so I don't think we can use `dropFirst`/`dropLast` here.

Sources/Apollo/NormalizedCache.swift
```diff
@@ -1,5 +1,19 @@
 import Foundation
 
+/// An enumeration of behaviors for clear operations on a given cache.
+public enum CacheClearingPolicy: Equatable {
+  /// Clears all records in the cache.
+  case allRecords
+  /// Clears all records whose key matches the provided glob pattern.
+  ///
+  /// For example `*pollo` will match both `Apollo` and `pollo`.
+  case allMatchingKeyPattern(String)
```
martijnwalraven, line 10: I wonder how useful a regex really is here, because real keys (based on `id` or other key fields) are usually pretty opaque, and we're actually planning to move away from using path-based keys as a fallback completely.

So it might make more sense to follow Apollo Client web here and just offer a way to clear/evict records by key, perhaps allowing multiple keys to be passed in.

martijnwalraven: Until we have removed path-based keys, maybe it would make sense to have `evict` remove all records that have the specified key as a prefix?

Sources/Apollo/NormalizedCache.swift
Mordil: Ah, you're right between this feedback and the one on Dictionary key ordering.

I'm just going to remove these policies. They can be re-added later if devs need them

Sources/Apollo/NormalizedCache.swift
Mordil: I can see that possible, but what if I want to evict all query responses?  `.allMatchingKeyPattern("*search*")`

With a prefix that requires me knowing the implementation detail that Apollo uses the `QUERY_ROOT` prefix in order to evict query responses.

`.anyKeyWithPrefix("QUERY_ROOT")` is effectively the same as `.allMatchingKeyPattern("QUERY_ROOT*")`

Sources/Apollo/InMemoryNormalizedCache.swift
Mordil: Removed

Sources/Apollo/NormalizedCache.swift
martijnwalraven: What would you expect `.allMatchingKeyPattern("*search*")` to do, and what is the use case for that?

We definitely have to do a better job explaining how the normalized cache works. My hope is that removing path-based keys will make things easier, but I also want to make sure our APIs don't complicate matters further.

For now, the problem is that `.allMatchingKeyPattern("QUERY_ROOT*")` likely won't do what people think it does. It would only remove records that do not have their own key configured (through `cacheKeyForObject()`). Those objects won't have an `id` or other key field that is used to uniquely identify them, so we currently use a fake path-based key, which means the results of different queries can't be merged.

Most of the time, that is not what you want, because that means missing out on all the benefits of normalization (while still paying the performance costs of storing objects as separate records). We will likely follow the example of Apollo Client web v3, and simply nest these "value objects" within the same record.

At the same time, we should make it easier for people to configure keys (likely with sensible defaults that look at `__typename` and `id`). What does your `cacheKeyForObject()` look like right now?

If the result of a query that requests `user(id: 123)` is stored in the cache for example, `QUERY_ROOT` would only contain a reference to the `User:123` record. So you wouldn't remove that record by removing `QUERY_ROOT`, just the mapping between `user(id: 123)` and that record.

What scenarios do you see in your app? How fine grained are the cache eviction operations you have found a need for?

Mordil: Our `cacheKeyForObject` is looking at just `__typename` and `id`, otherwise returning `nil`.

We don't have a particular use case for this clearing policy in question - we're actually after an age-based policy, but since that is some longer-term work, we have the policy implemented only in our fork for now.

These other policies were ones that I brainstormed in order to make this a valuable change to the codebase, rather than adding an enum with a single case.

If you think there's value in just making the change to add a single case (which is the default value used everywhere) as a foundation to build off of, and we later add policies on a singular case-by-case basis, then I'm happy to remove this policy.

---
My example is coming from the GitHub GraphQL API with a very minimal setup of Apollo that uses the same `__typename` and `id` cache key generation.

An example of a repository search would provide the following record:
| _id | key | record |
|---|---|---|
| 247393 | `QUERY_ROOT.search(after:Y3Vyc29yOjc2MA==,first:30,query:Design,type:REPOSITORY).edges.29` | `{"node":{"$reference":"Repository:MDEwOlJlcG9zaXRvcnkyMjMyOTQzMjA="},"__typename":"SearchResultItemEdge"}` |

I would expect `*search*` to remove any key that contains the phrase `search`. I wouldn't necessarily expect the referenced records to be deleted, just the _query_ response record(s) themselves.

Doing it by prefix provides two issues:
1. It forces me to understand that the default key prefix is `QUERY_ROOT`, which is an implementation detail that could, and you allude to it will, change
1. If I want to delete all searches, I need to have the longer `"QUERY_ROOT.search"` as the value, which adds noise to the intent of the cache clear: I just want to clear up any key that is a `search` related record.

Taking away the situation where Apollo's defaults are used partially or wholly - having a prefix only policy pushes effort on the cache key algorithm the developers establishes to be mindful of possibly easier cleanup later because they would then have to specify the prefix in order to match when it's a glob pattern, they can be much more liberal with their key generation.

Mordil: And if I want to remove any search that matches a particular query I could use the pattern `*search*query:Design*`

martijnwalraven: I think what may be confusing about the current situation is that the path-based fallback blurs the distinction between object keys and field keys. An object key is the result of `cacheKeyForObject`, so something like `User:123`. A field key is generated by combining the field name with argument values. 

Because we currently generate separate records for every level of a query result, regardless of whether an object has a real cache key, you end up with lots of useless records like `QUERY_ROOT.search(after:Y3Vyc29yOjc2MA==,first:30,query:Design,type:REPOSITORY).edges.29`. 

Note that `search` is not the cache key of an object here, it is a field name. It's an implementation detail that we currently store every edge of a connection as a separate record, using the complete path as a key, and that is something we plan on fixing. But that's exactly the reason the API shouldn't rely on this.

The only real cache keys are things like `User:123`, and `Repository:MDEwOlJlcG9zaXRvcnkyMjMyOTQzMjA=`. The query root is a special case however. Queries start by selecting fields on the query root, which is a singleton object for which we use the cache key `QUERY_ROOT`. As soon as you encounter an object with an identity of its own, your cache key will no longer have that prefix.

We can't support `.allMatchingKeyPattern("*search*")`, because it isn't clear what that would mean. The problem is that even though `search` currently occurs in the keys of records, that will change, because it does not identify an object.

What you might want to say instead is something like `store.evict("QUERY_ROOT", "search")`, which would evict results for the `search` field from the `QUERY_ROOT` record. Specifying `QUERY_ROOT` is necessary here, because your schema may have other types that have a `search` field defined. `store.evict("Company:123", "search")` would evict results for a `search` field defined on a `Company` type for example. (We could define a constant to avoid using a literal `QUERY_ROOT` here, or perhaps some kind of default, but you still need to be able to differentiate between evicting a root field or a field on another object.)

You always have to start out by identifying an object by key. We have no way of efficiently selecting all records by the fields they contain, because records are only indexed by cache key. And even if we did find a way to do that, I suspect the semantics would be more confusing than helpful.

That doesn't mean there is no place for regexes. I can imagine it being useful to be able to do `store.evict(`User:*`) for example, which would evict all users. Or even `store.evict(`QUERY_ROOT`, `*search*Design*`) if we are ok relying on how field keys are encoded.

Mordil: That's only for `InMemoryNormalizedCache`. `SQLiteNormalizedCache` replaces the `*` with `%` to be able to use the `LIKE` operator

designatednerd: That seems like a fairly significant difference based on which cache you're using, which IMO seems like it could be really confusing for users (and would also make things like a hybrid in-memory and SQLite cache much more difficult to implement in the future)

Mordil: That's a good point. This is my bad as a contributor, as I was focused primarily on SQLite and didn't invest a lot of time about fully considering InMemory.

The InMemory cache should probably actually use a regex to compare, but it sounds like we're moving towards explicit keys or at least a prefix.

designatednerd:
After discussion between @martijnwalraven and myself, we've decided on a two-part approach for proceeding with cache clearing stuff: 

1. Implement #1555 to allow for deleting the object for a particular key. It won't be as flexible as this, but it does at least give _some_ ability to get rid of things  in the cache, which we do lack at the moment. This will be happening within the next few weeks.
2. Work on a longer-term solution to cache eviction and invalidation. On web, we've made some solid progress with this, but there's some significant changes we'd need to make to make that work on iOS. This is likely going to be part of our work in Q1 of 2021. 

Unfortunately, this PR isn't quite what we're looking for longer-term, and it may make some of the things we want to do longer-term harder, so while we really appreciate the effort and your time bouncing ideas here, I'm gonna go ahead and close this one out.

PR #1514, opened 2020-11-13
Fix auth header in `UserManagementInterceptor` example

MaxDesiatov:
The correct name of the header is `Authorization`, not `Authentication`. For example, [the "Downloading Schema" section](https://www.apollographql.com/docs/ios/downloading-schema/) uses `Authorization` to download the schema. Additionally, there should be no colon between `Bearer` and the actual token.


designatednerd (approved):
Technically both of these are server-specific implementation details, but this does line up with more common usages, so I'll merge it.

MaxDesiatov:
This is aligned with what [Apollo Server uses](https://www.apollographql.com/docs/apollo-server/security/authentication/). We had it deployed and, and copy pasting `UserManagementInterceptor` from the docs to our Apollo iOS code led to this hard to diagnose issue. I hope that making it consistent with Apollo Server and the common use is better on balance and helps to avoid confusion.

designatednerd:
That's correct for getting a schema, though many people wind up using middleware to handle authentication and authorization on actual queries and requests. Nonetheless, yes, I do agree this is the more common usage. And it's already merged 🙃

PR #1520, opened 2020-11-16
Add cache fallback CachePolicy

Mordil:
Motivation:

It is sometimes acceptable to fallback to whatever cache data might be available when a particular fetch request fails either due to network or server problems.

This adds a missing 'CachePolicy' that falls back to use the cache after first attempting a fetch request.

Modifications:

- Add: `fetchReturningCacheDataOnError` case to `CachePolicy`
- Add: `CacheError` type to `LegacyCacheReadInterceptor` that is returned if both the network and cache fails to provide a successful result
- Change: `LegacyCacheReadInterceptor` to handle new policy

Result:

Developers should now have the ability to rely on their cache in more scenarios, such as when the server or network is having issues.

### DIFF
--- a/Sources/Apollo/LegacyCacheReadInterceptor.swift
+++ b/Sources/Apollo/LegacyCacheReadInterceptor.swift
@@ -2,6 +2,9 @@ import Foundation
 
 /// An interceptor that reads data from the legacy cache for queries, following the `HTTPRequest`'s `cachePolicy`.
 public class LegacyCacheReadInterceptor: ApolloInterceptor {
+  public enum CacheError: Error {
+    case fetchAndCacheFailed(fetchError: Error, cacheError: Error)
+  }
     
   private let store: ApolloStore
   
@@ -27,6 +30,35 @@ public class LegacyCacheReadInterceptor: ApolloInterceptor {
                          completion: completion)
     case .query:
       switch request.cachePolicy {
+      case .fetchReturningCacheDataOnError:
+        chain.proceedAsync(
+          request: request,
+          response: response,
+          completion: { fetchResult in
+            // if the fetch request failed, try to read from the cache
+            switch fetchResult {
+            case let .success(fetchData):
+              chain.returnValueAsync(for: request, value: fetchData, completion: completion)
+
+            case let .failure(fetchError):
+              self.fetchFromCache(for: request, chain: chain) { cacheResult in
+                switch cacheResult {
+                case let .success(cacheData):
+                  chain.returnValueAsync(for: request, value: cacheData, completion: completion)
+
+                case let .failure(cacheError):
+                  chain.handleErrorAsync(
+                    CacheError.fetchAndCacheFailed(fetchError: fetchError, cacheError: cacheError),
+                    request: request,
+                    response: response,
+                    completion: completion
+                  )
+                }
+              }
+            }
+          }
+        )
+
       case .fetchIgnoringCacheCompletely,
            .fetchIgnoringCacheData:
         // Don't bother with the cache, just keep going
--- a/Sources/ApolloTestSupport/MockGraphQLServer.swift
+++ b/Sources/ApolloTestSupport/MockGraphQLServer.swift
@@ -40,7 +40,7 @@ public class MockGraphQLServer {
     }
   }
   
-  public typealias RequestHandler<Operation: GraphQLOperation> = (HTTPRequest<Operation>) -> Result<JSONObject, Error>
+  public typealias RequestHandler<Operation: GraphQLOperation> = (HTTPRequest<Operation>) throws -> JSONObject
   
   private class RequestExpectation<Operation: GraphQLOperation>: XCTestExpectation {
     let file: StaticString
@@ -79,7 +79,7 @@ public class MockGraphQLServer {
     _ operationType: Operation.Type,
     file: StaticString = #filePath,
     line: UInt = #line,
-    requestHandler: @escaping (HTTPRequest<Operation>) -> Result<JSONObject, Error>
+    requestHandler: @escaping RequestHandler<Operation>
   ) -> XCTestExpectation {
     return queue.sync {
       let expectation = RequestExpectation<Operation>(description: "Served request for \(String(describing: operationType))", file: file, line: line, handler: requestHandler)
@@ -90,18 +90,6 @@ public class MockGraphQLServer {
       return expectation
     }
   }
-
-  public func expect<Operation: GraphQLOperation>(
-    _ operationType: Operation.Type,
-    file: StaticString = #filePath,
-    line: UInt = #line,
-    requestHandler: @escaping (HTTPRequest<Operation>) -> JSONObject
-  ) -> XCTestExpectation {
-    return self.expect(operationType, file: file, line: line) { request -> Result<JSONObject, Error> in
-      let result = requestHandler(request)
-      return .success(result)
-    }
-  }
   
   func serve<Operation>(request: HTTPRequest<Operation>, completionHandler: @escaping (Result<JSONObject, Error>) -> Void) where Operation: GraphQLOperation {
     // Dispatch after a small random delay to spread out concurrent requests and simulate somewhat real-world conditions.
@@ -109,7 +97,7 @@ public class MockGraphQLServer {
       let operationType = type(of: request.operation)
       
       if let expectation = self[operationType] {
-        completionHandler(expectation.handler(request))
+        completionHandler(Result { try expectation.handler(request) })
         expectation.fulfill()
       } else {
         completionHandler(.failure(ServerError.unexpectedRequest(String(describing: operationType))))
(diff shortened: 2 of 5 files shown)


Sources/ApolloTestSupport/MockGraphQLServer.swift
```diff
@@ -85,14 +90,26 @@ public class MockGraphQLServer {
...
+  }
   
   func serve<Operation>(request: HTTPRequest<Operation>, completionHandler: @escaping (Result<JSONObject, Error>) -> Void) where Operation: GraphQLOperation {
     // Dispatch after a small random delay to spread out concurrent requests and simulate somewhat real-world conditions.
     queue.asyncAfter(deadline: .now() + .milliseconds(Int.random(in: 10...50))) {
       let operationType = type(of: request.operation)
       
       if let expectation = self[operationType] {
-        completionHandler(.success(expectation.handler(request)))
+        completionHandler(expectation.handler(request))
```
designatednerd, line 112: This seems like an odd change - why does this no longer need to be wrapped in `success`?

Sources/ApolloTestSupport/MockGraphQLServer.swift
```diff
@@ -85,14 +90,26 @@ public class MockGraphQLServer {
...
+    _ operationType: Operation.Type,
+    file: StaticString = #filePath,
+    line: UInt = #line,
+    requestHandler: @escaping (HTTPRequest<Operation>) -> JSONObject
+  ) -> XCTestExpectation {
+    return self.expect(operationType, file: file, line: line) { request -> Result<JSONObject, Error> in
+      let result = requestHandler(request)
+      return .success(result)
+    }
+  }
```
designatednerd, line 104: What's the thought on adding a synchronous version of this instead of just using the async version?

Tests/ApolloCacheDependentTests/FetchQueryTests.swift
```diff
@@ -73,6 +73,65 @@ class FetchQueryTests: XCTestCase, CacheDependentTesting {
     
     wait(for: [serverRequestExpectation, fetchResultFromServerExpectation], timeout: defaultWaitTimeout)
   }
+
+  func testFetchReturningCacheDataOnErrorReturnsDataOnFetchError() throws {
```
designatednerd, line 77: I would also recommend adding a test here for what error you get when you get an error from the server and there's nothing in the cache so we can keep that behavior consistent

Sources/Apollo/LegacyCacheReadInterceptor.swift
```diff
@@ -27,6 +30,35 @@ public class LegacyCacheReadInterceptor: ApolloInterceptor {
                          completion: completion)
     case .query:
       switch request.cachePolicy {
+      case .fetchReturningCacheDataOnError:
+        chain.proceedAsync(
```
designatednerd, line 34: One thing I worry about here is that this sort of messes up the idea that interceptors should be called in the order they're provided, but I suppose this isn't _that_ different from a retry. 

I wonder if there's a way to accomplish this where if an error comes in from the network stack and this is the cache policy, we catch that and then set up the retry with a `returnCacheDataDontFetch` policy. Then we can just proceed as we do with the `fetchIgnoringCacheData` etc here, and then when it comes again as `returnCacheDataDon'tFetch` we've already got handling for it. 

What do you think?

Tests/ApolloCacheDependentTests/WatchQueryTests.swift
```diff
@@ -269,7 +269,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
     }
     
     runActivity("Fetch same query from server with different argument but returning same object with changed data") { _ in
-      let serverRequestExpectation = server.expect(HeroNameWithIdQuery.self) { request in
+      let serverRequestExpectation = server.expect(HeroNameWithIdQuery.self) { request -> JSONObject in
```
designatednerd, line 272: I think this stuff was async to help test our async loading - @martijnwalraven care to comment here?

Sources/ApolloTestSupport/MockGraphQLServer.swift
```diff
@@ -40,7 +40,7 @@ public class MockGraphQLServer {
     }
   }
   
-  public typealias RequestHandler<Operation: GraphQLOperation> = (HTTPRequest<Operation>) -> JSONObject
+  public typealias RequestHandler<Operation: GraphQLOperation> = (HTTPRequest<Operation>) -> Result<JSONObject, Error>
```
martijnwalraven, line 43: I'm not sure we need to return `Result` here. As the handler is synchronous, it might be easier to make it throwing instead.

Sources/ApolloTestSupport/MockGraphQLServer.swift
martijnwalraven: Note that neither versions are asynchronous. The other one has been changed to return `Result`, but it's still synchronous.

Sources/ApolloTestSupport/MockGraphQLServer.swift
martijnwalraven: I think this makes sense if request handlers are changed to return `Result` (so we don't need to wrap the return value again). But I think it would be easier to make result handlers throwing. That way, we could simply do `completionHandler(Result { try expectation.handler(request) })` here.

Tests/ApolloCacheDependentTests/WatchQueryTests.swift
martijnwalraven: Not entirely sure what you mean. This just seems to add a type annotation?

Sources/Apollo/LegacyCacheReadInterceptor.swift
martijnwalraven: One other thing to keep in mind is that there might be more people want to customize when it comes to error handling, and we need to consider GraphQL errors in addition to network errors.

Apollo Client web has a configurable [error policy](https://www.apollographql.com/docs/react/data/error-handling/) to specify how GraphQL errors (as opposed to network errors) are handled for example. We don't really take these into account at all right now. So we store incomplete data into the cache even if a response contains errors for example. That should probably be configurable. And it will also interact with `fetchReturningCacheDataOnError`, because returning data from the cache might be preferable to returning incomplete data.

I think having sensible policies for common situations is important, but ideally these would be built on composable primitives. There is a lot to consider here, especially once we have TTL added to the mix, so I don't think we can cover all cases with predefined policies.

Tests/ApolloCacheDependentTests/WatchQueryTests.swift
Mordil: This is a type annotation to avoid ambiguity in the type-checker due to the overload. Changing to the throwing form as suggested by Martijn will absolve the need of this change.

Tests/ApolloCacheDependentTests/FetchQueryTests.swift
Mordil: Isn't this covered by the `testFetchReturningCacheDataOnErrorReturnsBothErrors` test case?

```swift
      // starting at line 115
      try XCTAssertFailureResult(result) {
        let cacheError = try XCTUnwrap($0 as? LegacyCacheReadInterceptor.CacheError)

        switch cacheError {
        case let .fetchAndCacheFailed(fetchError, cacheError):
          let fetchError = try XCTUnwrap(fetchError as? TestError)
          let cacheError = try XCTUnwrap(cacheError as? JSONDecodingError)
          guard case .missingValue = cacheError else {
            XCTFail("Unexpected json decoding error: \(cacheError)")
            return
          }
          XCTAssertEqual(fetchError, .expectation)
        }
      }
```

Sources/Apollo/LegacyCacheReadInterceptor.swift
Mordil: While the error policy is definitely something great to evolve towards, I think it's a little orthogonal to this capability.

What if instead this becomes an optional configuration on the `NetworkFetchInterceptor` where the developer can determine what do to on an error, and it kicks off a retry like @designatednerd suggested?

designatednerd: well, in theory it would likely need to be an interceptor (possibly even an `additionalErrorInterceptor`), since for instance something that comes through with like a 400 error would not be caught by the `NetworkFetchInterceptor`, but by the `ResponseCodeInterceptor`.

designatednerd:
@martijnwalraven and I had some long discussions on the cache, and we've come to some conclusions around this, #1512, and #1498. I'll detail thoughts on the other two in those tickets. 

For this one, I think a better solution is to use `additionalErrorInterceptor` on the request chain as the place where you're deciding whether or not to fetch from the cache. This would be the only kind of cache policy that dealt with any kind of error handling *after* a network call, and often people are going to want to have some errors fall into getting cache data to display (eg, server is down) vs some errors that just need to be handled more directly (eg, user has been logged out at the server level).

I'll need to turn `cachePolicy` into a var on `HTTPRequest` for this to work properly, but the idea is that you can look at the error received in the `additionalErrorInterceptor and then decide if you want to retry while changing the `cachePolicy` to `returnCacheDataDontFetch`. 

I'll add an example with the change to the Request object, but otherwise, I'm going to close this issue out.

PR #1521, opened 2020-11-16
Allow toggle of cache policy used in 'perform' method

### LINKED ISSUE #777: Mutations always publish result to store
When performing a mutation, the result always gets published to the store. 

```
  public func perform<Mutation: GraphQLMutation>(mutation: Mutation,
                                                 context: UnsafeMutableRawPointer? = nil,
                                                 queue: DispatchQueue = DispatchQueue.main,
                                                 resultHandler: GraphQLResultHandler<Mutation.Data>? = nil) -> Cancellable {
    return self.send(operation: mutation,
                     shouldPublishResultToStore: true,
                     context: context,
                     resultHandler: wrapResultHandler(resultHandler, queue: queue))
  }
```

I see why we might want to publish the result if the changes affect any watch query. Does it make sense to also be able to specify the cache policy for mutations as we do for queries?

Versions
Apollo-iOS 0.15.3

Mordil:
Motivation:

In certain circumstances it makes sense for developers to have full control over their cache data, and when performing mutations they do not want Apollo to automatically publish the results to the store.

`LegacyCacheWriteInterceptor` properly handles this `CachePolicy` case, but `ApolloClient.perform` conformance uses `CachePolicy.default` with no configuration possibility.

Modifications:

- Change: `ApolloClientProtocol.perform` to accept a new `publishResultToStore` parameter
- Change: `ApolloClient` to inspect `publishResultToStore` to determine which `CachePolicy` to use.

Result:

Developers should now have the ability to toggle the cache behavior of 'perform' to ignore the cache completely.

This resolves #777

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -110,14 +110,18 @@ extension ApolloClient: ApolloClientProtocol {
 
   @discardableResult
   public func perform<Mutation: GraphQLMutation>(mutation: Mutation,
+                                                 publishResultToStore: Bool = true,
                                                  queue: DispatchQueue = .main,
                                                  resultHandler: GraphQLResultHandler<Mutation.Data>? = nil) -> Cancellable {
-    return self.networkTransport.send(operation: mutation,
-                                      cachePolicy: .default,
-                                      contextIdentifier: nil,
-                                      callbackQueue: queue) { result in
-       resultHandler?(result)
-    }
+    return self.networkTransport.send(
+      operation: mutation,
+      cachePolicy: publishResultToStore ? .fetchIgnoringCacheData : .fetchIgnoringCacheCompletely,
+      contextIdentifier: nil,
+      callbackQueue: queue,
+      completionHandler: { result in
+        resultHandler?(result)
+      }
+    )
   }
 
   @discardableResult
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -115,7 +115,7 @@ extension ApolloClient: ApolloClientProtocol {
                                                  resultHandler: GraphQLResultHandler<Mutation.Data>? = nil) -> Cancellable {
     return self.networkTransport.send(
       operation: mutation,
-      cachePolicy: publishResultToStore ? .fetchIgnoringCacheData : .fetchIgnoringCacheCompletely,
+      cachePolicy: publishResultToStore ? .default : .fetchIgnoringCacheCompletely,
       contextIdentifier: nil,
       callbackQueue: queue,
       completionHandler: { result in
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -257,6 +257,7 @@
 		C35D43C622DDE28D00BCBABE /* a.txt in Resources */ = {isa = PBXBuildFile; fileRef = C304EBD322DDC7B200748F72 /* a.txt */; };
 		C377CCA922D798BD00572E03 /* GraphQLFile.swift in Sources */ = {isa = PBXBuildFile; fileRef = C377CCA822D798BD00572E03 /* GraphQLFile.swift */; };
 		C377CCAB22D7992E00572E03 /* MultipartFormData.swift in Sources */ = {isa = PBXBuildFile; fileRef = C377CCAA22D7992E00572E03 /* MultipartFormData.swift */; };
+		D87AC09F2564D60B0079FAA5 /* ApolloClientOperationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D87AC09E2564D60B0079FAA5 /* ApolloClientOperationTests.swift */; };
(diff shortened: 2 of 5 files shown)


Sources/Apollo/ApolloClient.swift
```diff
@@ -110,14 +110,18 @@ extension ApolloClient: ApolloClientProtocol {
...
                                                  resultHandler: GraphQLResultHandler<Mutation.Data>? = nil) -> Cancellable {
-    return self.networkTransport.send(operation: mutation,
-                                      cachePolicy: .default,
-                                      contextIdentifier: nil,
-                                      callbackQueue: queue) { result in
-       resultHandler?(result)
-    }
+    return self.networkTransport.send(
+      operation: mutation,
+      cachePolicy: publishResultToStore ? .fetchIgnoringCacheData : .fetchIgnoringCacheCompletely,
```
Mordil, line 118: @designatednerd I went back and forth on this, as I can see both `.returnCacheAndFetch` and `.fetchIgnoringCacheData` being expected behaviors by "default".

I'm willing to revert this to `.default`, but I wanted your take on this.

designatednerd: I would say `.default` since for mutations this doesn't actually hit the cache on the way out.

designatednerd (commented):
Exposing this as a boolean seems reasonable to me - would be great to have some tests on this to make sure I don't break it in the future 😇

Sources/Apollo/ApolloClient.swift
Mordil: Oh right, I glossed over that part and didn't piece it together in my head.

Though, that still doesn't seem right to rely on the interceptor chain setup given that devs might be using their own caching implementation rather than the LegacyCacheRead that Apollo provides.

Then _again_... that's up to them to properly handle everything that they want to do anyway...

I'll just revert this portion to use `.default` like we agreed 😅

designatednerd: Ahhh fair point - I think at some point I'd had the `load` on `ApolloStore` set up to throw an error if you tried to call it on anything other than a query (it was previously restricted to things that conformed to the `GraphQLQuery` protocol but that wound up being unworkable with the current setup), but it looks like that got moved up to the intereceptor. I'll chat with @martijnwalraven on that one.

Mordil:
@designatednerd I added a unit test, and also a query for `Review` objects by episode.

When writing the unit test I thought I'd need it, but it turns out I didn't.

I figured I'd leave it in anyway since I did the work to add it

designatednerd:
@Mordil If you don't mind removing that query that'd be great - trying to keep that API file from having too much stuff we're not actually using.

Mordil:
@designatednerd Done!

Tests/ApolloTests/ApolloClientTests.swift
```diff
@@ -0,0 +1,73 @@
+import Apollo
+import ApolloTestSupport
+import StarWarsAPI
+import XCTest
+
+final class ApolloClientTests: XCTestCase, CacheDependentTesting {
```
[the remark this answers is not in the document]
Mordil: `ApolloClientConfigurationTests`?

Tests/ApolloTests/ApolloClientTests.swift
```diff
@@ -0,0 +1,73 @@
...
+        ]
+      ]
+    }
+    let performResultFromServerExpectation = resultObserver.expectation(description: "Mutation was successful") { _ in }
+
+    self.client.perform(mutation: mutation, publishResultToStore: false, resultHandler: resultObserver.handler)
+
+    self.wait(for: [serverRequestExpectation, performResultFromServerExpectation], timeout: self.defaultWaitTimeout)
+
+    let cacheExpectation = self.expectation(description: "Cache returns nil data for review mutation")
```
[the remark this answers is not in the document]
Mordil: Ah, I had misread the generic constraint on `store.load` and thought I'd have to construct a `GraphQLQuery` type.

I've simplified this

Tests/ApolloTests/ApolloClientTests.swift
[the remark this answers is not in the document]
Mordil: Eh, I'll be more specific and name this `ApolloClientOperationTests`

Sources/Apollo/ApolloClient.swift
martijnwalraven: Yeah, `cachePolicy` was originally meant to only apply to queries, because the semantics don't really make sense for mutations (or subscriptions for that matter). I think we'll want to revisit cache policies once we have a better idea about the interaction between the request chain and caching, maybe that will allow us to differentiate between operation types again. That would also be the time to think about more flexible error handling, taking TTL into account, etc. So for now, I don't think what we do here really matters, as long as it means the default chain respects `publishResultToStore`.

Mordil:
So what's left on this PR? It looks like from my perspective it's in a good enough shape?

designatednerd (approved):
LGTM!

PR #1524, opened 2020-11-17
Update JSON v2 to have expressible by _ literal

designatednerd:
From a discussion I was having last week in #1482, here's some literal conformances in my codable-compatible version of arbitrary JSON for use in future tests.

### DIFF
--- a/Sources/ApolloCodegenLib/JSON.swift
+++ b/Sources/ApolloCodegenLib/JSON.swift
@@ -31,60 +31,6 @@ public enum JSONValue: Codable, Equatable {
     }
   }
   
-  public init(_ array: [Any?]) throws {
-    var jsonArray = [JSONValue]()
-    for item in array {
-      if let unwrappedValue = item {
-        if let boolValue = unwrappedValue as? Bool {
-          jsonArray.append(.bool(boolValue))
-        } else if let intValue = unwrappedValue as? Int {
-          jsonArray.append(.int(intValue))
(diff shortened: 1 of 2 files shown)


martijnwalraven (commented):
I may be missing something, but looking at the [Swagger `JSONValue` implementation](https://github.com/swagger-api/swagger-codegen/blob/0ceba0f33f442f474cad24596c543abe1930532e/modules/swagger-codegen/src/main/resources/swift5/JSONValue.mustache), I think it should be possible to simplify this by getting rid of the `Array` and `Dictionary` initializers that take `Any?`, and to change the `arrayLiteral` and `dictionaryLiteral` initializers to take `JSONValue` instead. Since the other literals are already convertible to `JSONValue`, that should give you a type safe way to initialize nested structures.

Removing the other initializers and replacing the literal ones with this makes everything compile and pass, except for `testAddingNonJSONElementWillThrowError` (which makes sense because that check is now performed at compile time):

```swift
extension JSONValue: ExpressibleByDictionaryLiteral {
    public init(dictionaryLiteral elements: (String, JSONValue)...) {
        self = .object([String: JSONValue](uniqueKeysWithValues: elements))
    }
}
extension JSONValue: ExpressibleByArrayLiteral {
    public init(arrayLiteral elements: JSONValue...) {
        self = .array(elements)
    }
}
```

designatednerd:
Ooh! Neato. I'm all in favor of simplifying.

martijnwalraven (approved):

Issue #1506, opened 2020-11-10
Compatibility issues on Linux with CommonCrypto


abdimaye:
## Feature request

The CommonCrypto module is only available in Xcode 10+. It would be nice to have this replaced with an open source module such that the apollo client can run on other platforms.

## Motivation

We are currently trying to run parts of our codebase on Linux.

## Proposed solution

Replace CommonCrypto with an open source library. There are only two files in ApolloCore that use it and the diff wouldn't be that big. 

I was looking at `swift-crypto` https://swift.org/blog/crypto/ but there are compatibility issues where it requires mocos 10.15 and apollo requires 10.10.

designatednerd:
We don't *officially* support Linux, but I would very much like to. My thought is that we could make some kind of protocol that uses `swift-crypto` for Linux band still uses `CommonCrypto` on iOS etc. Would be more than open to a PR if you have some time, I'm buried at the moment.

abdimaye:
I started on these changes a couple of weeks ago but got sidetracked on other things... I'll create a draft PR now for visibility as there are more changes required than expected 😣.

AnthonyMDev:
Linux support is something we are still considering for some time in the future. I'm going to close this inactive outdated issue though.

PR #1534, opened 2020-11-23
Fix Linux compatibility issues

abdimaye:
[Issue #1506](https://github.com/apollographql/apollo-ios/issues/1506)

List of changes:
- Swift-tools-version:5.3
- Optionally import swift-crypto for Linux
- Create Crypt wrapper for crypto functions
- Import `FoundationNetworking` - required due to new recent move of network-related objects https://stackoverflow.com/questions/58592508/how-can-i-use-foundations-urlrequest-type-on-linux
- ~~Create StringEncoding wrapper and map encoding for Linux~~


designatednerd:
@abdimaye can you elaborate a bit on what's still not working?

abdimaye:
> @abdimaye can you elaborate a bit on what's still not working?

Planning to add the details to the list of outstanding errors later today 👍

designatednerd:
Some thoughts that might help looking over those errors: 

- I'm pretty sure for `Bundle+Helpers` you can just throw an `if not linux` around the contents of that file since it deals with `Bundle` stuff that I'm almost positive doesn't exist in the Linux version of Foundation. 
- Is `URLRequest` not one of the things that's in `FoundationNetworking`? That's a significant surprise if not.
- I think for the `URLSessionTaskMetrics` and `URLSessionStreamTask` you should be able to use an `unavailable` marker for linux on the enclosing methods. 
- Not sure what Foundation for Linux has got to figure out the string encoding, but you'd probably need to do something similar to what you're doing with `Crypt` to create a protocol to handle those
- Not sure what the hell is going on with the JSON error, tbh.

Sources/Apollo/StringEncoding.swift
```diff
@@ -0,0 +1,64 @@
...
+            return .japaneseEUC
+        case .macOSRoman:
+            return .macOSRoman
+        case .utf8:
+            return .utf8
+        case .utf16:
+            return .utf16
+        case .utf32:
+            return .utf32
+        }
```
abdimaye, line 44: Don't really know how else to do this. This is also no exhaustive as it's quite tricky to figure out the encoding properties in swift https://developer.apple.com/documentation/swift/string/encoding and the actual names https://www.iana.org/assignments/character-sets/character-sets.xhtml 😫 

Any help/pointers would be much appreciated!

abdimaye:
@designatednerd sorry I've been ghost! I think the encoding issue was the last. Looks like there are some merge conflicts also which I'll take a peek at later.

Sources/Apollo/StringEncoding.swift
designatednerd: Actually, I just looked to see where we're using this and...we aren't. I think I can delete this method, which will remove the need for all of this.

designatednerd: Just opened #1579, once that gets merged I think you should be able to look at the merge conflicts.

designatednerd:
Awesome work - though I'm a bit confused as to why the tests aren't running on this at all. I'll poke at it this afternoon.

designatednerd:
I'd also like to know if you have suggestions for at least making sure things continue to build on linux - what linux setup should we be using for tests, etc

designatednerd:
I wonder if it wasn't building because this was a draft from a forked repo - can you make another commit please?

abdimaye:
> I'd also like to know if you have suggestions for at least making sure things continue to build on linux - what linux setup should we be using for tests, etc

I've now added a Dockerfile and build script for linux. Not entirely sure about the testing as I'm still trying to figure this out...

designatednerd:
Taking a stab at adding a linux build to the matrix...apologies for pushing to your branch but it's not going to work on any other branch at the moment 😛

designatednerd:
Woof. Running into some serious issues around the fact that even though we're specifically saying only import `Crypto` on linux, it's still saying it's a dependency on macOSX. I have no idea why this wasn't failing the build before - I'm going to try to do some swift 5.3 stuff on a separate branch to try and disentangle a bunch of changes I need to make to file structure because of resources from this PR

designatednerd:
Looks like the SQlite sub-lib is not compiling on linux since it uses a bunch of stuff which is iOS specific. Ugh, I wish platform restrictions per target were a thing. 

I'll have to reach out to the swift forums about why things are trying to import `Crypto` lib even though it specifically says it should only be brought in on linux

abdimaye:
> Looks like the SQlite sub-lib is not compiling on linux since it uses a bunch of stuff which is iOS specific. Ugh, I wish platform restrictions per target were a thing.
> 
> I'll have to reach out to the swift forums about why things are trying to import `Crypto` lib even though it specifically says it should only be brought in on linux

😕 Would it be reasonable to skip any tests which have platform specific dependencies for this first iteration?

designatednerd:
Unfortunately the problem is not that the tests won't run, it's that the libraries won't even build. Again, I'm not sure why this wasn't showing up as a problem earlier. 😭

abdimaye:
Have you tried xcodebuild? Something like `xcodebuild -scheme Apollo`?

designatednerd:
That's [what we're doing on the iOS and mac tests already](https://github.com/apollographql/apollo-ios/blob/5d9022f6bd871b13b5e7a78c2d69ca08dc586b7a/.circleci/config.yml#L63), unfortuantely. And it's still giving us problems.

abdimaye:
Found another issue that's kind of related to this - type checking `queryDocument` would timeout on linux. Here's the PR from my colleague: https://github.com/apollographql/apollo-tooling/pull/2198

abdimaye:
Completely forgot about this! I won't be looking into this further so closing PR.

PR #1538, opened 2020-11-24
Remove setter from `Atomic` to prevent API misuse, fix TSAN error in network interceptor

designatednerd:
@martijnwalraven pointed out in #1531 that it's a little too easy to accidentally misuse the `Atomic` API as it stands now, particularly in terms of dealing with things like toggling a boolean or making changes to an array. This PR removes the setter wrapper for `value`, so that if you need to make changes, you can only change things by using `mutate`, which executes the changes in a block for safety. 

I also fixed a thread sanitizer error in the Network interceptor around the `cancellable` that it holds onto. Once this and #1531 are merged, that should allow us to turn on thread sanitizer in tests permanently. 🤞

### DIFF
--- a/Sources/Apollo/NetworkFetchInterceptor.swift
+++ b/Sources/Apollo/NetworkFetchInterceptor.swift
@@ -1,9 +1,12 @@
 import Foundation
+#if !COCOAPODS
+import ApolloCore
+#endif
 
 /// An interceptor which actually fetches data from the network.
 public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
   let client: URLSessionClient
-  private var currentTask: URLSessionTask?
+  private var currentTask: Atomic<URLSessionTask?> = Atomic(nil)
   
   /// Designated initializer.
   ///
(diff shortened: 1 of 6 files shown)


Sources/Apollo/NetworkFetchInterceptor.swift
```diff
@@ -57,9 +60,15 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
                            completion: completion)
       }
     }
+    
+    self.currentTask.mutate { $0 = task }
   }
   
   public func cancel() {
-    self.currentTask?.cancel()
+    guard let task = self.currentTask.value else {
```
martijnwalraven, line 68: It seems this could just be `self.currentTask.value?.cancel()`, but I assume you prefer this because you find it more readable?

martijnwalraven (approved):
Great, looks good to me!

Sources/Apollo/NetworkFetchInterceptor.swift
designatednerd: somewhat, yep.

PR #1552, opened 2020-12-01
Add pluralization and singularization support for new codegen

designatednerd:
This PR brings in [`InflectorKit`](https://github.com/mattt/InflectorKit) from [a fork](https://github.com/apollographql/InflectorKit) where I've added SPM support. Probably going to PR that back to the main repo as well. 

I've also created a wrapper around it to keep it abstract, and added an `inflectionRule` enum to allow easy passing of multiple inflection rules without having to worry about using `InflectorKit` in the consuming code.


martijnwalraven (approved):
This looks great to me!

Issue #1544, opened 2020-11-27
chain.retry not call completion to fetch method


vfumin:
Hi, i try to retry request from errorinterceptor, after refreshtoken, request sends, but i can't get completion event in fetch method

```

class RefreshTokenInterceptor: ApolloErrorInterceptor {

    private let requestManager: RequestManager
    private let bag = DisposeBag()

    init(requestManager: RequestManager) {
        self.requestManager = requestManager
    }

    /// Asynchronously handles the receipt of an error at any point in the chain.
    ///
    /// - Parameters:
    ///   - error: The received error
    ///   - chain: The chain the error was received on
    ///   - request: The request, as far as it was constructed
    ///   - response: [optional] The response, if one was received
    ///   - completion: The completion closure to fire when the operation has completed. Note that if you call `retry` on the chain, you will not want to call the completion block in this method.
    func handleErrorAsync<Operation: GraphQLOperation>(
        error: Error,
        chain: RequestChain,
        request: HTTPRequest<Operation>,
        response: HTTPResponse<Operation>?,
        completion: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void) {
        guard response?.httpResponse.statusCode == 401 else {
            completion(.failure(error))
            return
        }
        requestManager.refreshToken()
            .subscribe(
                onSuccess: { [chain] in
                    chain.retry(request: request, completion: completion)
                },
                onError: { error in
                    completion(.failure(error))
                }
            )
            .disposed(by: bag)
    }
}

```

_Originally posted by @vfumin in https://github.com/apollographql/apollo-ios/issues/1530#issuecomment-734241073_

designatednerd:
Do you hit `onSuccess` or `onError` in that code? I have a sneaking suspicion your `disposeBag` is getting emptied before your subscription hits either of those.

designatednerd:
Also, you may want to consider making this an `ApolloInterceptor` and putting it in the chain right after the `NetworkFetchInterceptor` rather than routing every single error through it, which is what you're doing with an `ApolloErrorInterceptor`.

vfumin:
chain.retry is working, but after complete retry request, not calling resultHandler in fetch(query:, resultHandler:).

i try debugging and not calling  completion(.success(value)) in retrychain, but getting     self.callbackQueue.async
```
public func returnValueAsync<Operation: GraphQLOperation>(
    for request: HTTPRequest<Operation>,
    value: GraphQLResult<Operation.Data>,
[...]

PR #1563, opened 2020-12-08
Fix deallocation issue on `additionalErrorHandler` completion when retry is called.

designatednerd:
Looks like the fact that we're wrapping `additionalErrorHandler`'s callback with a `weak self` was causing a failure when a `retry` was done from within `additionalErrorHandler`, because `self`, and in turn the callback queue it was holding onto, got deallocated.

This fix pulls out the callback queue to prevent a retain cycle but hang on to the queue so the completion closure can be properly executed.


designatednerd:
Addresses #1544.

PR #1569, opened 2020-12-10
Make `cachePolicy` a var on `HTTPRequest`

designatednerd:
This PR comes out of discussion around #1520, which was considering adding a blanket cache fallback policy. 

Ultimately, most of the time whether you want something to display information from the cache or an error message is often a factor of what, exactly the error is. By making `cachePolicy` mutable on `HTTPRequest`, we open the door to the `additionalErrorInterceptor` being able to examine the error which came through and be able to decide whether the error should be passed on to the caller, or the request chain should be retried with a different cache policy. 

An example is provided in the test I've added: I created an additional error interceptor which resets the `cachePolicy` to hit the cache, then fed it to a network request designed to error out. When the error for the network request is received, the additional interceptor is hit, and the request is retried with the updated cache policy. 

I think this gives us more flexibility to achieve the same goal: Being able to say "If we try to hit the network and it doesn't work, show results from the cache."

Would be interested in feedback particularly from @Mordil but from anyone else who was interested in this feature - particularly about how I should document this. It seems like it might be overkill to write a docs section about it, maybe I can add some interceptor playground examples.


martijnwalraven (approved):
This looks like a sensible way to solve this within the current request chain API. As discussed, we may want to explore other ways of structuring the request chain, and separating caching from networking. I suspect other solutions for this may fall out of that, but for now this seems like a useful pattern.

PR #1576, opened 2020-12-15
Use storage for locks

makadaw:
& operator in swift not equal to C's operator. Passing lock with inout
ref may cause race and potential crashes.

More details http://www.russbishop.net/the-law
Apple example https://fuchsia.googlesource.com/third_party/swift-corelibs-foundation/+/upstream/fuchsia_release/Foundation/NSLock.swift

### DIFF
--- a/Sources/Apollo/Locking.swift
+++ b/Sources/Apollo/Locking.swift
@@ -1,25 +1,27 @@
 import Foundation
 
 final class Mutex {
-  private var _lock = pthread_mutex_t()
+  private var _lockPtr = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
 
   init() {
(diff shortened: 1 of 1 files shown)


designatednerd:
@martijnwalraven Are we even still using `Mutex` at this point? It looks like no from a quick search but wanted to confirm

martijnwalraven:
> @martijnwalraven Are we even still using `Mutex` at this point? It looks like no from a quick search but wanted to confirm

No, we're not using either the `Mutex` or the `ReadWriteLock` at this point. So I think we should remove the whole file.

designatednerd:
@martijnwalraven 👍
@makadaw - Thanks for pointing this out! I'm going to close this PR out but I'll remove the file in another PR later this afternoon.

Issue #1555, opened 2020-12-03
Add `deleteObject(key)` to `ReadWriteTransaction`


designatednerd:
As raised by #1551, when you remove something from a list its reference is deleted, but the object the reference was referring to may become orphaned if there are no other references to it. 

Long-term we should be looking into something similar to what [the web client is doing with garbage collection](https://www.apollographql.com/docs/react/caching/garbage-collection/), but as a short-term workaround, we can add (with sufficient warnings about potentially creating invalid references) an option to delete the value for a particular key if the developer is able to know for certain that something should be completely removed from the DB.

designatednerd:
This shipped with 0.39.0!

PR #1583, opened 2020-12-22
Add `deleteRecord(for key:)` to `ReadWriteTransaction`

designatednerd:
This comes with a ton of caveats, but I tried to document them in both code and in the docs. 

Note that we're going to be looking at a much more robust system for cache eviction and possibly garbage collection in the future, but this is a stopgap until that's ready to go live.

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -242,7 +242,13 @@ public final class ApolloStore {
       try write(object: object, withKey: key, variables: variables)
     }
     
-    public func removeRecord(for key: CacheKey) throws {
+    /// Removes the object for the specified cache key. Does not cascade
+    /// or allow removal of only certain fields. Does nothing if an object
+    /// does not exist for the given key.
+    ///
+    /// - Parameters:
+    ///   - key: The cache key to remove the object for
+    public func removeObject(for key: CacheKey) throws {
       try self.cache.removeRecord(for: key)
     }
 
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -241,6 +241,10 @@ public final class ApolloStore {
       try body(&object)
       try write(object: object, withKey: key, variables: variables)
     }
+    
+    public func removeRecord(for key: CacheKey) throws {
+      try self.cache.removeRecord(for: key)
+    }
 
     public func write<Query: GraphQLQuery>(data: Query.Data, forQuery query: Query) throws {
       try write(object: data,
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -101,6 +101,12 @@ public final class SQLiteNormalizedCache {
       try self.db.prepare("VACUUM;").run()
     }
   }
+  
+  private func removeSQLiteRecord(for cacheKey: CacheKey) throws {
+    let recordKey = self.recordCacheKey(forFieldCacheKey: cacheKey)
+    let query = self.records.filter(key == recordKey)
+    try self.db.run(query.delete())
+  }
 
   private func parse(row: Row) throws -> Record {
     let record = row[self.record]
@@ -125,6 +131,10 @@ extension SQLiteNormalizedCache: NormalizedCache {
   
   public func merge(records: RecordSet) throws -> Set<CacheKey> {
     return try mergeRecords(records: records)
+  }
+  
+  public func removeRecord(for key: CacheKey) throws {
+    try removeSQLiteRecord(for: key)
   }
   
   public func clear() throws {
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -103,8 +103,7 @@ public final class SQLiteNormalizedCache {
   }
   
   private func removeSQLiteRecord(for cacheKey: CacheKey) throws {
-    let recordKey = self.recordCacheKey(forFieldCacheKey: cacheKey)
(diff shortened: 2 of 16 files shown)


designatednerd:
Addresses #1555

designatednerd:
Gonna let @martijnwalraven take a look at this before I merge it to make sure I'm not doing anything _too_ dumb here

Sources/Apollo/ApolloStore.swift
```diff
@@ -241,6 +241,10 @@ public final class ApolloStore {
       try body(&object)
       try write(object: object, withKey: key, variables: variables)
     }
+    
+    public func removeRecord(for key: CacheKey) throws {
```
martijnwalraven, line 245: For consistency with the other methods, I would make this `removeObject(withKey:)`. (I don't think we should use record as part of the public API in general, only for cache implementors.)

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -127,6 +132,10 @@ extension SQLiteNormalizedCache: NormalizedCache {
     return try mergeRecords(records: records)
   }
   
+  public func removeRecord(for key: CacheKey) throws {
+    try removeSQLiteRecord(for: key)
```
martijnwalraven, line 136: Why the separate `removeSQLiteRecord()`? Wouldn't `removeRecord()` on the SQLite cache already imply that?

docs/source/caching.mdx
```diff
@@ -177,3 +177,38 @@ store.withinReadWriteTransaction({ transaction in
   }
 })
 ```
+
+### delete
+
+Delete functionality is limited at this time. There are presently two options available on `NormalizedCache` which are supported by the included implementations of it:
```
martijnwalraven, line 183: > There are presently two options available on `NormalizedCache` which are supported by the included implementations of it
This sentence seems a bit hard to understand, and I would avoid `NormalizedCache`. Maybe something like:
> There are presently two deletion methods available on `ApolloStore`, which are supported by both the in memory and the SQLite cache.

docs/source/caching.mdx
```diff
@@ -177,3 +177,38 @@ store.withinReadWriteTransaction({ transaction in
...
 ```
+
+### delete
+
+Delete functionality is limited at this time. There are presently two options available on `NormalizedCache` which are supported by the included implementations of it: 
+
+1. `clear` - Removes everything from the cache immediately. This is basically the "Nuke it from orbit, it's the only way to be sure" option. 
+2. `removeRecord(for:)` Removes a single record for the given `CacheKey`. 
+
+`removeRecord(for:)` will only remove things at the level of a record - that is, it cannot remove individual fields from a record, and it cannot remove outside references **to** a record.
```
martijnwalraven, line 188: I wouldn't use the term record here, but talk about removing objects because records are not part of the public API.

docs/source/caching.mdx
```diff
@@ -177,3 +177,38 @@ store.withinReadWriteTransaction({ transaction in
...
+
+Delete functionality is limited at this time. There are presently two options available on `NormalizedCache` which are supported by the included implementations of it: 
+
+1. `clear` - Removes everything from the cache immediately. This is basically the "Nuke it from orbit, it's the only way to be sure" option. 
+2. `removeRecord(for:)` Removes a single record for the given `CacheKey`. 
+
+`removeRecord(for:)` will only remove things at the level of a record - that is, it cannot remove individual fields from a record, and it cannot remove outside references **to** a record. 
+
+```swift
+// Assuming we start with records of the following format:
```
martijnwalraven, line 191: I think in order to be understandable, this needs a much more detailed explanation of cache normalization and the structure of records. It may make sense to move the implementation details to a separate section and focus on the API and observable behavior here.

martijnwalraven (changes requested):

Sources/Apollo/ApolloStore.swift
designatednerd: Fair

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
designatednerd: to match the other things in this file that have a wrapper method

docs/source/caching.mdx
designatednerd: Gonna work on this bit tomorrow when I have a couple functioning brain cells

docs/source/caching.mdx
```diff
@@ -177,3 +177,31 @@ store.withinReadWriteTransaction({ transaction in
...
+
+```swift
+apollo.cacheKeyForObject = { $0["id"] }
+```
+
+This would indicate that for every object which has an `id` property, it will be cached with that `id` as the key. This would be ideal for an API which uses globally unique identifiers, as our Star Wars API does. 
+
+This means that if you have previously fetched an object with the ID `"2001"`, you can call `transaction.removeObject(for: "2001")` on a `ReadWriteTransaction` to remove any object cached with that key, along with all of its associated scalar properties and the references to other objects it stores. 
+
+`removeObject` does _not_ cascade its removals - if you remove an object which has reference to another object, the reference will be removed, but that other object will not be removed and will remain in the cache. Likewise, if you delete an object, references _to_ that object will not be deleted, they will simply fail when you attempt to load the object again.
```
martijnwalraven, line 202: Maybe mention that depending on the cache policy, failure because of dangling references doesn't necessary imply an error but will just result in a cache miss and refetching the data from the server.

martijnwalraven (approved):

docs/source/caching.mdx
designatednerd: yeah I think I can rephrase this to "to fail, causing a cache miss," which should theoretically clarify that a bit more.

PR #1584, opened 2020-12-22
Swift Package Manager 5.3 compatibility

designatednerd:
In this PR: 

- Added a `Package@swift-5.2.swift` to preserve backwards compatibility
- Explicitly made text files in test package resources
- Excluded ALL the Info.plists from libs
- Moved graphql files in `StarWarsAPI`, `GitHubAPI`, and `UploadAPI` into excluded `graphql` folders
- Cleaned up a whole mess of file redirecting nonsense from doing all that.
- Added something that should eventually be removed (and will eventually fail tests once we switch codegen engines) if the codegen engine is the WIP engine


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -48,9 +48,11 @@ public class ApolloCodegen {
...
       }
     case .singleFile(let fileURL):
-      guard fileURL.apollo.isSwiftFileURL else {
-        throw CodegenError.singleFileButNotSwiftFileURL(fileURL)
-      }
+      if options.codegenEngine == .typescript {
+        guard fileURL.apollo.isSwiftFileURL else {
+         throw CodegenError.singleFileButNotSwiftFileURL(fileURL)
+        }
+      } // else we're fine with JSON at this point for intermediate purposes.
```
designatednerd, line 55: Once default codegen becomes swift codegen, this will fail the tests that check that a folder URL can't be passed in here.

Sources/GitHubAPI/operationIdsPath.json
```diff
@@ -1,10 +0,0 @@
-{
-  "63e25c339275a65f43b847e692e42caed8c06e25fbfb3dc8db6d4897b180c9ef": {
```
designatednerd, line 2: This was deleted because we switched to "operationIDs.json"

designatednerd:
May the ghost of Steve Jobs have mercy on us all.

PR #1605, opened 2021-01-04
Drop iOS/tvOS 9-11, watchOS 3-4, macOS 10.10-10.13 support

designatednerd:
One of the clearest results of the 2020 developer survey is that our users aren't supporting iOS 9 and 10 anymore. Dropping these two operating systems and their associated versions of tvOS, macOS, and watchOS allowed us to get rid of basically every availability check we currently have. 🎉

BUT, I think we can go one better: [iOS 12 supports all the same devices iOS 11 does](https://blakespot.com/ios_device_specifications_grid.html), so I think we can also drop iOS 11 support. We only had one person who mentioned having to support 11, and their stated reason was to support older devices - but all older devices that support 11 also support 12. 

This also lines up with Apple's "Hey there, you sure you don't want to make 12.0 your min SDK?" prompt given every time Xcode 12 is updated. Personally, I'd love to drop 12 too to use Combine within the SDK, but over 25% of our users still have iOS 12 as their min SDK, so I don't think that's realistic for at least another year. 

This would leave our min SDKs as: 
iOS/tvOS: 12.0
watchOS: 5.0
macOS: 10.14

**If you have strong objections to dropping any of these operating systems, please register them in the comments below!**


martijnwalraven (approved):
That cleans up nicely! 

From what I can see this looks good. I wish we didn't have to duplicate configuration across Xcode project, Cocoapods and Swift package manager. But unfortunately that's the reality of the ecosystem right now.

designatednerd:
Yeah, the duplication is annoying, but given how infrequently this changes, I'm kinda fine with it.

designatednerd:
OK i'm gonna leave this up until Thursday morning my time for objections, otherwise we'll merge it in!

Issue #636, opened 2019-07-15
Generate .xcfilelist from codegen


designatednerd:
Great suggestion from @zeveisenberg in #369: 

>It would be nice if Apollo generated a .xcfilelist file, which could be used to declare all the .swift files it generates. (This is in cases where you’re generating a folder of files, rather than just a single API.swift file.)

designatednerd:
I took a stab at this in our current typescript setup and it was NOT straightforward. Probably going to punt on this to until I implement Swift codegen (#682).

designatednerd:
[Good workaround to this from an issue on the tooling repo](https://github.com/apollographql/apollo-tooling/issues/1573#issuecomment-539754801)

cltnschlosser:
If you are using cocoapods and multiple file generation (not just API.swift) then you can do something like this:
```ruby
# Generate xcfilelists on pod install
inputs = File.open('codegen-inputs.xcfilelist', 'w')
inputs.puts "$(PODS_TARGET_SRCROOT)/codegen.sh"
inputs.puts "$(PODS_TARGET_SRCROOT)/schema.json"
inputs.puts "$(PODS_TARGET_SRCROOT)/Operations/"
Dir.glob("Operations/**/*/").each do | path |
  inputs.puts "$(PODS_TARGET_SRCROOT)/Operations/" + Pathname.new(path).basename.to_s + "/"
end
inputs.close

outputs = File.open('codegen-outputs.xcfilelist', 'w')
Dir.glob("Sources/Generated/*.graphql.swift").each do | path |
  outputs.puts "$(PODS_TARGET_SRCROOT)/Sources/Generated/" + Pathname.new(path).basename.to_s
end
outputs.close

Pod::Spec.new do |s|
  # Local lib info redacted

  s.dependency 'Apollo'

  s.source_files = 'Sources/Generated/**/*.swift'
  s.script_phase = {
      :name => 'Apollo Code Gen',
      :script => %{
cd "${PODS_TARGET_SRCROOT}"
./codegen.sh
      },
      :input_file_lists => ['$(PODS_TARGET_SRCROOT)/codegen-inputs.xcfilelist'],
      :output_file_lists => ['$(PODS_TARGET_SRCROOT)/codegen-outputs.xcfilelist'],
      :execution_position => :before_compile
    }
  end
end
```

[...]

PR #1621, opened 2021-01-14
Revamp Swift Scripting doc

designatednerd:
As much as I love Swift Scripting, it's clear that our existing setup required too much work to set up to actually be easier than bash once all was said and done. I still wholeheartedly believe in the benefits of having this be a) In Swift to make dealing with the filesystem more type safe and less annoying and b) thoroughly tested, so I've revamped how we recommend the user set things up. 

To start, I've created an [iOS Codegen Template](https://github.com/apollographql/iOSCodegenTemplate) that uses the Swift Argument Parser to determine what to run, and then has two commands that users can configure to work as they wish with their system. I've then gone through the docs with a sledgehammer to try and make it work more clearly with the new setup. 

Would love to hear feedback on this from folks who have been using swift scripting already, and also from those who wanted to use it but found it too cumbersome to set up.

### DIFF
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -147,29 +147,13 @@ MyProject // SourceRoot
   | - ApolloCodegen // Contains Swift Scripting files
 ```
 
-Next, now that you have a schema, you need to create a GraphQL file with an operation in order to generate code. 
+Next, now that you have a schema, you a GraphQL file with an operation in order to generate code. 
 
-## Creating a `.graphql` file with an operation
+## Adding a `.graphql` file with an operation
 
-The easiest and most common type of operation to create is a Query. Identify where your server's [GraphiQL](https://github.com/graphql/graphiql) instance lives. 
+If you're not familiar with creating an operation in graphQL, please check out the [portion of our tutorial on executing your first query](https://www.apollographql.com/docs/ios/tutorial/tutorial-execute-query/). You can stop after the section about adding your query to Xcode.
 
-GraphiQL is a helpful web interface for interacting with and testing out a GraphQL server. This can generally be accessed by going to the same URL as your GraphQL endpoint in a web browser, but you might need to talk to your backend team if they host it in a different place.
-
-You'll see something that looks like this: 
-
-<img alt="GraphiQL empty" src="screenshot/graphiql_empty.png" class="screenshot"/>
-
-In the "Docs" tab on the right-hand side, you should be able to access a list of the various queries you can make to your server: 
-
-<img alt="Docs tab" src="screenshot/graphiql_docs_tab.png" class="screenshot"/>
-
-You can then type out a GraphQL query on the left-hand side and have it give you auto-completion for your queries and the properties you can ask for on the returned data. Clicking the play button will execute the query, so you can validate that the query works:
-
-<img alt="Completed query" src="screenshot/graphiql_query.png" class="screenshot"/>
-
-Now you can create a new empty `.graphql` file in your Xcode project, give it the same name as your query,  and paste in the query. 
-
-You'll want to add it to the project files, ideally at or above the level of the `schema.json` (Otherwise, you'll need to manually pass the URL of your GraphQL files to your code generation step):
+Make sure you've added the operation file to the project files, ideally at or above the level of the `schema.json` (Otherwise, you'll need to manually pass the URL of your GraphQL files to your code generation step):
 
 ```txt:title=Sample%20Project%20Structure
 MyProject // SourceRoot
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -147,7 +147,7 @@ MyProject // SourceRoot
   | - ApolloCodegen // Contains Swift Scripting files
 ```
 
-Next, now that you have a schema, you a GraphQL file with an operation in order to generate code. 
+Next, now that you have a schema, you need a GraphQL file with an operation in order to generate code. 
 
 ## Adding a `.graphql` file with an operation
 
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -170,7 +170,7 @@ Here, for example, is what this looks like in a file for one of the queries in o
 
 <img alt="Launch list file" src="screenshot/graphql_file_launchlist.png" class="screenshot"/>
 
->**Before you start**: Remember, you need to have a locally downloaded copy of your schema and at least one `.graphql` file containing an operation in your file tree. If you don't have **both** of these, code generation will fail. [Read the section above](#creating-a-graphql-file-with-an-operation) if you don't have an operation set up!
+**Note:** You do **not** need to add this file to your target in Xcode. Only the generated Swift code needs to be included in your target for it to work. 
 
 ## Running code generation from your main project
 
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -187,7 +187,7 @@ This is best achieved with a Run Script Build Phase.
 
     ```
     cd "${SRCROOT}"/Codegen
-    xcrun -sdk macosx swift run
+    xcrun -sdk macosx swift run generate
     ```
 
     >**Note**: If your package ever seems to have problems with caching, run `swift package clean` before `swift run` for a totally clean build. It is not recommended to do this by default, because it substantially increases build time.
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -3,7 +3,9 @@ title: Swift scripting
 sidebar_title: Swift scripting
 ---
 
-Apollo Client for iOS enables you to use Swift scripting to perform certain operations that otherwise require the command line. This document guides you through setting up a Swift Package Manager executable and then using it to: 
+Apollo Client for iOS enables you to use Swift scripting to perform certain operations that otherwise require the command line. 
+
+This document guides you through setting up a Swift Package Manager executable project using our template, and then using that project to: 
 
 - Download a schema
 - Generate Swift code for your model object based on your schema and operations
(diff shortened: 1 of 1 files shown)


raymondk-nf:
I think the docs are very comprehensive. The challenges I ran into were when things did not work and working through the error output. The command line errors were often times ambiguous and can be cumbersome to debug.

designatednerd:
@raymondk-nf can you go into some detail about what was hard there? Some of the errors are coming from the underlying typescript library but if it's anything we can handle I'd love to try.

raymondk-nf:
@designatednerd It was months ago when I ran into those command line errors. Once I figured it out, I never touched it again.  The error messages are getting better after the last few updates.  This is one I ran into recently:

`
It looks like there are 0 files associated with this Apollo Project. This may be because you don't have any files yet, or your includes/excludes fields are configured incorrectly, and Apollo can't find your files. For help configuring Apollo projects, see this guide: https://go.apollo.dev/t/config
Loading Apollo Project [started]
{ Error: Error in "Loading schema for Unnamed Project": Error: Introspection is not allowed for this request
    at Object.error...}
Loading Apollo Project [failed]
→ Error initializing Apollo GraphQL project "Unnamed Project": Error: Error in "Loading schema for Unnamed Project": Error: Introspection is not allowed for this request
 ›   Error: Error initializing Apollo GraphQL project "Unnamed Project": Error:
 ›    Error in "Loading schema for Unnamed Project": Error: Introspection is 
 ›   not allowed for this request

Program ended with exit code: 1
`

Initially, it didn't dawn on me what `Introspection is not allowed for this request` meant. Turns out I wasn't on VPN. 😆

docs/source/swift-scripting.md
```diff
@@ -3,7 +3,9 @@ title: Swift scripting
 sidebar_title: Swift scripting
 ---
 
-Apollo Client for iOS enables you to use Swift scripting to perform certain operations that otherwise require the command line. This document guides you through setting up a Swift Package Manager executable and then using it to: 
+Apollo Client for iOS enables you to use Swift scripting to perform certain operations that otherwise require the command line.
```
AnthonyMDev, line 6: I think this could be clearer about the use cases. I was under the impression that this was just a replacement for the schema download and code generations commands. But only after talking to @designatednerd offline found out that it actually enables you to do a lot more complex custom scripting.

Doesn't have to be in the PR, but maybe an example of how you would write a custom script for a contrived use case would help illustrate this point!

AnthonyMDev (commented):
This overall looks great! Very thorough and easy to understand.

GitHub is stupid and won't let me comment on unchanged lines (can't believe they haven't changed that yet). So I'll just ask this here:

On line 206 doesn't it need to say something after `swift run`? Like `ApolloCodegen generate` or something like that?

designatednerd:
@raymondk-nf Unfortunately, all that is coming from the underlying typescript codegen at the moment.  I agree that error message could be significantly clearer.

docs/source/swift-scripting.md
```diff
@@ -35,230 +37,123 @@ The rest of this guide will help you set up a Swift Package Manager executable t
...
+```txt:title=Sample%20Project%20Structure
+MyProject // Source root
+  | MyProject.xcodeproj
+  | - MyProject // Contains app target source files
+  | - MyLibraryTarget // Contains lib target source files
+  | - MyProjectTests // Contains test files
+```
 
-    **ALSO NOTE**: Having to specify the name is a workaround for [SR-12110](https://bugs.swift.org/browse/SR-12210). Hopefully once that's fixed, SPM should pick up the name automatically. 
+Drag the `ApolloCodegen` folder in at the same level as your other targets (in Finder, not in Xcode):
```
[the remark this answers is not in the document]
designatednerd: Not that i've found that makes any sense with how we're trying to make this flexible to use.

The problem is that you can only distribute a library as a package, you can't distribute an executable. This is why `ApolloCodegenLib` is a library, but this executable is in its own project. 

We also want to make it possible for people to use with whatever their setup is, and not have to worry about conforming their project to our assumptions about where things should be. To do that, we'd either have to make parameters for everything supported by the TS codegen (which to me, defeats the point of the whole wrapper since it puts you back into "let's figure out where things are in the filesystem using bash and environment variables!" mode, which is my least favorite mode), or we'd only be able to support a really limited subset of options. 

The other problem is if we shipped this as a compiled binary tied to a version of the library, there's not a great way to control where that gets checked out since that checkout is handled by SPM. We'd be right back where we are with the bash script in terms of having to go folder-diving through derived data to figure out where this is checked out, and that's something I'm explicitly trying to avoid here.

docs/source/swift-scripting.md
```diff
@@ -35,230 +37,123 @@ The rest of this guide will help you set up a Swift Package Manager executable t
...
-    ```
-    
-6. In `main.swift`, import the Codegen lib and Apple's Foundation library at the top of the file:
+Double-click `Package.swift` in the `ApolloCodegen` folder to open the executable's package in Xcode. 
 
-    ```swift:title=main.swift
-    import Foundation
-    import ApolloCodegenLib
-    ```
+**Important!** Since a particular version of code generation is tied to a particular version of the SDK, you need to make sure that the `dependencies` section of `Package.swift` is set to grab the same version of the `apollo-ios` library you're using in your main application:
```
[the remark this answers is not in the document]
designatednerd: I wish to hell this were possible, I agree that this is not great. My thought would be mayyyybe having people add a macOS command line executable target to their Xcode project, but I'm not sure how well that would work with scripting (it's a lot easier to do `swift run` on a command line executable in a particular folder than it is to run another target in the same Xcodeproj)

docs/source/swift-scripting.md
```diff
@@ -35,230 +37,123 @@ The rest of this guide will help you set up a Swift Package Manager executable t
...
+  | - MyProjectTests // Contains test files
+  | - ApolloCodegen // Contains Swift Scripting files
+```
+
+Next, now that you have a schema, you need to create a GraphQL file with an operation in order to generate code. 
 
 ## Creating a `.graphql` file with an operation
 
-Because you've already [downloaded a schema](#downloading-a-schema), you can now proceed to creating an operation. The easiest and most common type of operation to create is a Query. 
+The easiest and most common type of operation to create is a Query. Identify where your server's [GraphiQL](https://github.com/graphql/graphiql) instance lives.
```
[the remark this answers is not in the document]
designatednerd: Yeah that's fair - maybe I can link to the instructions in the tutorial here, that'd probably work.

docs/source/swift-scripting.md
```diff
@@ -284,73 +179,20 @@ MyProject // SourceRoot
...
-
-    Again, you might want to make sure the folder exists before proceeding:
+## Running code generation from your main project
 
-    ```swift:title=main.swift 
-    try FileManager
-          .default
-          .apollo.createFolderIfNeeded(at: targetURL)
-    ```
+Codegen should be set up to run from your main project when you build - this will allow any changes you've made to your graphQL files to be picked up and the code to be regenerated.
```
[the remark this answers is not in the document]
designatednerd: Ideally we'd be able to create an `xcfilelist` from what we're getting back from codegen IR, which allows Xcode to say "don't run this script unless something in this files list changed." [I've got an open issue for that](https://github.com/apollographql/apollo-ios/issues/636).

docs/source/swift-scripting.md
designatednerd: 🤔 I mean, we are essentially using an example of this in the `SwiftScripts` in this repo in order to generate our docs. Is that what you're thinking of, or are you thinking of something else?

docs/source/swift-scripting.md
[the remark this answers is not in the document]
AnthonyMDev: That `xcfilelist` thing is a dirty dirty lie that Apple has been telling for years. Have you ever gotten it to actually work successfully? I've tried with so many different run scripts and it is so flaky, my scripts run when they shouldn't all the time even with input files... :/

docs/source/swift-scripting.md
AnthonyMDev: Kind of! Looking at that example, the only part of the Codegen lib we are using is the `CodegenLogger`, so not a really great example of the power that this provides, but it shows the basic concept.

AnthonyMDev: I also imagine that the scripting tools could get more powerful in the future if we provide a way to introspect on the compiled IR.

docs/source/swift-scripting.md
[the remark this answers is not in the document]
designatednerd: Yeah I've not had good luck with it with folders, but theoretically it's supposed to work when you explicitly name the files. 

We also can look at whether there's a way to tell this from our CLI.

docs/source/swift-scripting.md
designatednerd: ah i mean that as more of an example of "here are other things you can do with Swift scripting" than what you can do specifically with the codegen lib

docs/source/swift-scripting.md
```diff
@@ -147,29 +147,13 @@ MyProject // SourceRoot
   | - ApolloCodegen // Contains Swift Scripting files
 ```
 
-Next, now that you have a schema, you need to create a GraphQL file with an operation in order to generate code. 
+Next, now that you have a schema, you a GraphQL file with an operation in order to generate code.
```
AnthonyMDev, line 150:
```suggestion
Next, now that you have a schema, you need a GraphQL file with an operation in order to generate code. 
```

docs/source/swift-scripting.md
```diff
@@ -203,10 +187,10 @@ This is best achieved with a Run Script Build Phase.
 
     ```
     cd "${SRCROOT}"/Codegen
-    xcrun -sdk macosx swift run
+    xcrun -sdk macosx swift run generate
     ```
 
-    >**Note**: If your package ever seems to have problems with caching, run `swift package clean` before `swift run` for a totally clean build. It is not recommended to do this by default, because it substantially increases build time.
+    >**Note**: If your package ever seems to have problems with caching, run `swift package clean` before `swift run` for a totally clean build. Do not this by default, because it substantially increases build time.
```
AnthonyMDev, line 193: "Do not this by default"? Getting to the end of the day? ;)

AnthonyMDev (approved):
Two typos, but otherwise this looks awesome!

docs/source/swift-scripting.md
designatednerd: i english gud

PR #1624, opened 2021-01-15
I ended up using these commands

LucasVanDongen:
I followed the tutorial and up to this point it was pretty complete, but I had to edit the script phase to make it work.

### DIFF
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -186,8 +186,8 @@ This is best achieved with a Run Script Build Phase.
 3. Update the build phase run script to `cd` into the folder where your executable's code lives, then run `swift run` (using `xcrun` so that you can ensure it runs with the correct SDK, no matter what type of project you're building): 
 
     ```
-    cd "${SRCROOT}"/Codegen
-    xcrun -sdk macosx swift run generate
+    cd "${SRCROOT}"/ApolloCodegen/Sources/ApolloCodegen
+    xcrun -sdk macosx swift run ApolloCodegen generate
     ```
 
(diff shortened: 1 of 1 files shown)


docs/source/swift-scripting.md
```diff
@@ -186,8 +186,8 @@ This is best achieved with a Run Script Build Phase.
 3. Update the build phase run script to `cd` into the folder where your executable's code lives, then run `swift run` (using `xcrun` so that you can ensure it runs with the correct SDK, no matter what type of project you're building): 
 
     ```
-    cd "${SRCROOT}"/Codegen
-    xcrun -sdk macosx swift run generate
```
designatednerd, line 190: Gah, good catch on this second one but I think the first one you shouldn't need to `cd` into the `Sources`, you should only need to `cd` to `ApolloCodegen`. Mind double-checking me on that and then updating this?

LucasVanDongen: I remembered you can run the script from both Source and Source/ApolloCodegen and only the second one actually worked because of path resolving in the FileManager

designatednerd: In theory that shouldn't be the case, because `FileFinder` [here](https://github.com/apollographql/iOSCodegenTemplate/blob/main/ApolloCodegen/Sources/ApolloCodegen/FileStructure.swift#L12) uses the location of the file rather than the location where the script is run as where it needs to start finding things. Where you run it from should not make a difference.

designatednerd:
Closing in favor of #1629 which also fixed a couple other things I missed 🙃

PR #1644, opened 2021-01-28
Fix uploads not getting `additionalHeaders`

designatednerd:
In this PR: 
- Fix for issue raised in #1634 where an auth token set in `additionalHeaders` was not getting passed to an upload request
- Updated testing to send request creation through the `RequestChainNetworkTransport`'s creator method to be able to test that the additional headers were actually getting set.


designatednerd:
I think unfortunately because of the changes I had to make to make this testable I'll have to do this as a minor instead of a patch release, but I'm OK with that.

AnthonyMDev (approved):
Looks good, thanks for the quick fix!

PR #1653, opened 2021-02-09
Fix encoding of + in parameters for GET urls

### LINKED ISSUE #1652: iOS client not properly escaping "+" sign when using GET for queries
## Bug report

When strings are an input variable to a query, that input is not correctly sent to the server. In particular, "+" characters are changed into " " characters when converting into the GET query.

See: https://stackoverflow.com/a/24888789/7612797
> while it's not contemplated in the aforementioned RFC, section 5.2, URL-encoded form data, of the W3C HTML spec says that application/x-www-form-urlencoded requests should also replace space characters with + characters (and includes the asterisk in the characters that should not be escaped). And, unfortunately, `URLComponents` won't properly percent escape this, so Apple advises that you manually percent escape it before retrieving the url property of the `URLComponents` object.

We had a very simple query our graphQL server was supporting.
```
query PhoneCheck($phone: String!) {
    checkPhone(phone: $phone)
}
```

Often, the `$phone` variable may include a "+" sign in front (e.g. "+1-555-554-0909").

When querying the web console, this input variable is correctly received by the server. But when querying from the iOS client, the server somehow does not get the "+" sign. (e.g. "+1-555-554-0909" becomes " 1-555-554-0909" ).

## Versions

- `apollo-ios` SDK version: 0.41.0
- Xcode version: 11.4
- Swift version: Latest
- Package manager: Cocoapods

## Steps to reproduce

See description

## Further info
Since the iOS client is using GET queries, strings should therefore be URL-encoded.

Generated code is as follows:
```swift
  public final class PhoneCheckQuery: GraphQLQuery {
    /// The raw GraphQL definition of this operation.
    public let operationDefinition: String =
      """
      query PhoneCheck($phone: String!) {
        checkPhone(phone: $phone)
      }
      """

    public let operationName: String = "PhoneCheck"

    public let operationIdentifier: String? = "dabcc241b73fedbe393a0e90c8855802a94b82ad04cd5e9825c514b359e0ce78"

    public var phone: String

    public init(phone: String) {
      self.phone = phone
    }

    public var variables: GraphQLMap? {
      return ["phone": phone]
    }

    public struct Data: GraphQLSelectionSet {
      public static let possibleTypes: [String] = ["Query"]

      public static var selections: [GraphQLSelection] {
        return [
          GraphQLField("checkPhone", arguments: ["phone": GraphQLVariable("phone")], type: .scalar(PhoneNumberCheckResult.self)),
        ]
      }

      public private(set) var resultMap: ResultMap

      public init(unsafeResultMap: ResultMap) {
        self.resultMap = unsafeResultMap
      }

      public init(checkPhone: PhoneNumberCheckResult? = nil) {
        self.init(unsafeResultMap: ["__typename": "Query", "checkPhone": checkPhone])
      }

      /// Returns a check of the phone number.
      public var checkPhone: PhoneNumberCheckResult? {
        get {
          return resultMap["checkPhone"] as? PhoneNumberCheckResult
        }
        set {
          resultMap.updateValue(newValue, forKey: "checkPhone")
        }
      }
    }
  }
  ```

### PR DESCRIPTION
**Pull request by AnthonyMDev**
Fixes #1652

### DIFF
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -42,17 +42,9 @@ public struct GraphQLGETTransformer {
       return nil
     }
 
-    components.queryItems = queryItems.map { item in
-      guard let value = item.value else {
-        return item
-      }
-      
-      guard value.contains("+") else {
-        return item
-      }
-
-      return URLQueryItem(name: item.name, value: value.replacingOccurrences(of: "+", with: "%2B"))
-    }
+    components.queryItems = queryItems
+    components.percentEncodedQuery =
+      components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
 
     return components.url
   }
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -43,6 +43,8 @@ public struct GraphQLGETTransformer {
     }
 
     components.queryItems = queryItems
+    components.percentEncodedQuery =
+      components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
 
     return components.url
   }
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -42,9 +42,17 @@ public struct GraphQLGETTransformer {
       return nil
     }
 
-    components.queryItems = queryItems
-    components.percentEncodedQuery =
-      components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
+    components.queryItems = queryItems.map { item in
+      guard let value = item.value else {
+        return item
+      }
+      
+      guard value.contains("+") else {
+        return item
+      }
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Sources/Apollo/GraphQLGETTransformer.swift [hunk 1]
```diff
@@ -43,6 +43,8 @@ public struct GraphQLGETTransformer {
     }
 
     components.queryItems = queryItems
+    components.percentEncodedQuery =
+      components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
```
designatednerd, line 47: I wonder if this needs to actually be replaced in the query items rather than in the percent encoded query for it to essentially get re-percent-encoded so it's not read by the server as a raw string.

designatednerd (commented):
Tests look reasonable but I think mayyybe there's a different solution here?

designatednerd:
Ran this through [a couple](https://meyerweb.com/eric/tools/dencoder/) [different](https://www.urldecoder.org/) URL decoders and discovered that @AnthonyMDev's fix is the correct one - my fix would get decoded as `%2B`, which makes sense since it's double-encoded.

PR #1659, opened 2021-02-16
Update Starscream, for real this time

designatednerd:
So I tried this in #1394 and messed it up something awful, so I took a step back and tried again. It's marginally less awful this time!

There's one place where a test on reconnection is failing because if you voluntarily call `disconnect`, it appears you never get any kind of callback that you've disconnected due to [this issue with Starscream](https://github.com/daltoniam/Starscream/issues/869). I'm gonna see if anyone knows what's going on with that before giving up and updating the code in the tests that's waiting for the confirmation that the disconnection happened before trying to reconnect.

### DIFF
--- a/Sources/ApolloWebSocket/ApolloWebSocket.swift
+++ b/Sources/ApolloWebSocket/ApolloWebSocket.swift
@@ -17,30 +17,15 @@ public protocol ApolloWebSocketClient: WebSocketClient {
 
   /// Queue where the callbacks are executed
   var callbackQueue: DispatchQueue { get set }
-}
-
-public protocol SOCKSProxyable {
   
-  /// Determines whether a SOCKS proxy is enabled on the underlying request.
-  /// Mostly useful for debugging with tools like Charles Proxy.
-  var enableSOCKSProxy: Bool { get set }
 }
 
 // MARK: - WebSocket
 
 /// Included implementation of an `ApolloWebSocketClient`, based on `Starscream`'s `WebSocket`.
-public class ApolloWebSocket: WebSocket, ApolloWebSocketClient, SOCKSProxyable {
+public class ApolloWebSocket: WebSocket, ApolloWebSocketClient {
   
   private var stream: FoundationStream!
-  
-  public var enableSOCKSProxy: Bool {
-    get {
-      return self.stream.enableSOCKSProxy
-    }
-    set {
-      self.stream.enableSOCKSProxy = newValue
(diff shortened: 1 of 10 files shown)


Sources/ApolloWebSocket/ApolloWebSocket.swift
```diff
@@ -5,48 +5,40 @@ import Foundation
...
 
   /// Queue where the callbacks are executed
   var callbackQueue: DispatchQueue { get set }
-}
-
-public protocol SOCKSProxyable {
   
-  /// Determines whether a SOCKS proxy is enabled on the underlying request.
-  /// Mostly useful for debugging with tools like Charles Proxy.
-  var enableSOCKSProxy: Bool { get set }
```
designatednerd, line 26: unfortunately v4 of Starscream doesn't support proxying at this time, so I had to remove this.

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -413,35 +442,23 @@ extension WebSocketTransport: WebSocketDelegate {
...
-
-  public func websocketDidReceiveData(socket: WebSocketClient, data: Data) {
-    processMessage(socket: socket, data: data)
+  
+  private func attemptReconnectionIfDesired() {
+    guard self.reconnect.value else {
+      return
+    }
+    
+    DispatchQueue.main.asyncAfter(deadline: .now() + reconnectionInterval) {
```
AnthonyMDev, line 460: Is it right to always be doing this on the main queue? Do we have to use main for some reason here?

AnthonyMDev (approved):
This is super clean and straightforward. I'm not a web socket pro, but this seems easy enough. Would like to figure out whats up with that bug for sure, but I don't want to let it block us from moving forward here.

Sources/ApolloWebSocket/WebSocketTransport.swift
designatednerd: That's a reasonable question - this bit was basically moving around some existing code. 

I suspect part of the reason it's on `main` is because we do the initial connection on `main` by default, but I suspect the bigger reason is because you can't just grab the current queue and dispatch on it the way you could in ObjC.

designatednerd:
I've poked around a bit on the bug and gotten no response, so for now I'm going to disable the check for the disconnection notification so we can get this out the door.

PR #1673, opened 2021-02-19
Update Star Wars Example Paths

Sfresneda:
The current paths to Star Wars example are broken, so I did update that links


designatednerd (approved):
Good catch, thank you!

PR #1674, opened 2021-02-19
don't retain WebSocketTransport during reconnect

RolandasRazma:
don't retain WebSocketTransport during reconnect


designatednerd:
I'm curious: Is this a defensive `weak self` or have you seen this be a problem?

RolandasRazma:
I did not seen any issues. I was looking into doing exponential backoff and retaining to reconnect if you reconnect time is big made no sense

designatednerd:
Makes a lot more sense in the context of exponential backoff, thank you!

designatednerd (approved):

PR #1679, opened 2021-02-22
Don't allow multiple cancels on RequestChain

RolandasRazma:
I'm seeing some crashes in canceling interceptors. It's probably not `RequestChain` problem, but I see no reason why you should be able to cancel it multiple times.

Screenshot 2021-02-22 at 10 59 07


designatednerd (approved):
Seems like a reasonable change!

PR #1687, opened 2021-02-26
Fixes an issue where existing query params on graphQLEndpoint are ignored

min:
## What

Currently there's no way to add additional query parameters to the GET graph requests as it appears the `GraphQLGETTransformer` is creating a new array of `queryItems` and ignoring any parameters that's already on the `graphQLEndpoint`


designatednerd (approved):
Thank you for the clear explanation and for adding a test!

Issue #1690, opened 2021-03-01
Downloading schema from Apollo Studio without introspection is not supported for Swift Scripting setup


kimroen:
## Bug report

It is currently not possible to download a schema from Apollo Studio using the Swift Scripting package solution. You can pass in an api key, but there is no way to specify a variant, and the `endpointURL` argument is required.

## Versions

- `apollo-ios` SDK version: 0.41
- Xcode version: N/A
- Swift version: N/A
- Package manager: N/A

_Originally posted by @kimroen in https://github.com/apollographql/apollo-ios/discussions/1689_

designatednerd:
Yep, this is my bad - there's also no way to supply a graph ID.

kimroen:
> Yep, this is my bad - there's also no way to supply a graph ID.

That's fair! It's in the token, so I just assumed it would figure it out I suppose.

designatednerd:
This will go out later this week with our next release but you can pull from `main` if you need it right away

kimroen:
It works 🎉

PR #1691, opened 2021-03-02
Add handling for downloading schema from registry to Swift Scripting

designatednerd:
Our wrapper for downloading a schema has been limited to downloading from introspection endpoints without me realizing it. This PR updates our `SchemaOptions` to allow setup of download from the registry, which is supported by the underlying CLI.

This is a **breaking change** since I've broken up introspection vs registry download into an enum (since data required for each is mutually exclusive), and changed the name of the parameter that information must be handed to.

Addresses #1690.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloSchemaOptions.swift
+++ b/Sources/ApolloCodegenLib/ApolloSchemaOptions.swift
@@ -11,24 +11,38 @@ public struct ApolloSchemaOptions {
   
   /// How to attempt to download your schema
   public enum DownloadMethod: Equatable {
-    ///   - apiKey: The API key to use when retrieving your schema.
-    ///   - graphID: The identifier of the graph to fetch. Can be found in Apollo Studio.
-    ///   - variant: [Optional] The variant of the graph to fetch. Defaults to nil, which will return whatever is set to the current variant.
-    case registry(apiKey: String,
-                  graphID: String,
-                  variant: String?)
+
+    case registry(_ settings: RegistrySettings)
     ///   - endpointURL: The endpoint to hit to download your schema.
     case introspection(endpointURL: URL)
+    
+    public struct RegistrySettings: Equatable {
+      public let apiKey: String
+      public let graphID: String
+      public let variant: String?
+      
+      /// Designated initializer
+      ///
+      /// - Parameters:
+      ///   - apiKey: The API key to use when retrieving your schema.
+      ///   - graphID: The identifier of the graph to fetch. Can be found in Apollo Studio.
+      ///   - variant: [Optional] The variant of the graph to fetch. Defaults to nil, which will return whatever is set to the current variant.
+      public init(apiKey: String,
+                  graphID: String,
+                  variant: String? = nil) {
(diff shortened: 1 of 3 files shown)


Sources/ApolloCodegenLib/ApolloSchemaOptions.swift
```diff
@@ -9,8 +9,48 @@ public struct ApolloSchemaOptions {
...
-  let apiKey: String?
-  let endpointURL: URL
+  /// How to attempt to download your schema
+  public enum DownloadMethod: Equatable {
+
+    case registry(_ settings: RegistrySettings)
+    ///   - endpointURL: The endpoint to hit to download your schema.
+    case introspection(endpointURL: URL)
+    
+    public struct RegistrySettings: Equatable {
```
designatednerd, line 19: This is a separate object so that it can have new parameters added without breaking the enum.

SwiftScripts/Sources/SchemaDownload/main.swift
```diff
@@ -20,10 +20,20 @@ let output = sourceRootURL
     .appendingPathComponent("Sources")
     .appendingPathComponent("UploadAPI")
 
+// Introspection download:
 let options = ApolloSchemaOptions(schemaFileName: "schema",
-                                  endpointURL: endpoint,
+                                  downloadMethod: .introspection(endpointURL: endpoint),
                                   outputFolderURL: output)
 
+// Registry download:
```
designatednerd, line 28: This is left commented out for end to end testing in the future - not a great way to end to end test this without setting up a whole way of getting secrets into the repo.

AnthonyMDev (approved):
LGTM!

Sources/ApolloCodegenLib/ApolloSchemaOptions.swift
```diff
@@ -9,8 +9,48 @@ public struct ApolloSchemaOptions {
...
+      public let apiKey: String
+      public let graphID: String
+      public let variant: String?
+      
+      /// Designated initializer
+      ///
+      /// - Parameters:
+      ///   - apiKey: The API key to use when retrieving your schema.
+      ///   - graphID: The identifier of the graph to fetch. Can be found in Apollo Studio.
```
kimroen, line 28: There's nothing explicitly named "graph id" in Apollo Studio that I can see, but I'm guessing it's the dynamic part of the URL (after "/graph/"). If so, this is also the part between "service:" and the key itself in the token.

I've changed the name of our graph in Apollo Studio, but the name in the URL and token retains the old name so I'm guessing the "graph id" is the full thing and not just the id-looking part after the old name. I'll give it a shot and see what works!

kimroen: My thoughts above were accurate, and it works 👍 An internet-savvy person can probably figure it out, but it's not entirely obvious what to put as the `graphID`.

designatednerd: Yep, that's being addressed on the Studio end shortly.

PR #1692, opened 2021-03-02
Remove mention of mutation from useGETForQueries option description

kimroen:
I don't know for sure that this is an accurate change, but it sure seems like it is based on my understanding of the feature.


designatednerd (approved):
Hm, I know there was a point where we were doing this, but looking at our current setup, you're correct that we're not using it for mutations at the moment. Thanks!

kimroen:
> Hm, I know there was a point where we were doing this, but looking at our current setup, you're correct that we're not using it for mutations at the moment. Thanks!

Wonderful! The name of the option strongly suggest that it's for queries and not mutations, but I could be reading it wrong :)

Also, I'm not sure GET requests make sense for mutations based on what I know, so I'd be curious to read more about the intention there if that's the case!

designatednerd:
There's a common problem with describing GraphQL things where people often refer to "queries" when they mean "operations". 

A query is a specific _type_ of operation, but not all operations are queries. In practice though, the vast majority of operations are queries, so sometimes people just say "queries" when they mean operations. This most definitely extends to naming conventions. 

You can't do subscriptions with a `GET` request since the way those are presently architected requires a web socket, but in theory you could do a mutation with it, since the only real difference there is the contents of the serialized dictionary.

kimroen:
Huh! Thanks, I appreciate you taking the time to break that down ✨

PR #1710, opened 2021-03-11
Cache clarification

morgz:
I'm just spit balling the kind of explanation I'd have found useful. Feel free to use or discard and please let me know if my understanding is incorrect. Thanks!


designatednerd:
This is helpful - I need to do some noodling on this but I will probably use this as a jumping off point. Thank you!

Nickersoft:
Just found this PR and I think this is definitely something that needs to be clarified in the documentation :) Unlike the React docs, which strongly encourages manually updating the cache after mutations/subscriptions to propagate to the rest of the app, the current iOS docs only show examples in which the result of subscription or mutation are used to manually update the UI (such as appending to an array and refreshing a table view). Given the fact that Swift is strongly typed, unlike JS objects, figuring out how to manually update the cache becomes more ambiguous. 

For the time being, I've gone back to simply refetching queries after subscriptions, but I would love to see this PR rounded out and pushed through, because constantly refetching impacts performance. Just thought I'd give my two cents!

AnthonyMDev (approved):

Issue #1700, opened 2021-03-04
[Mega-Issue] New Codegen


designatednerd:
This Mega-Issue helps summarize and link to all the various things we're working on around New Codegen for high-level planning purposes. This issue will be updated as things change. Please open any comments on more specific issues, or file a new issue. Thank you!

### Previously on New Codegen
#939 was an RFC proposing a new format for generated code, along with the steps to get us off having code generated by our typescript CLI. 

### The Good News
The process #939 outlined is going to be basically the same: Get codegen working without caching, add immutable caching, add mutable caching, remove old caching. 

### The Medium News
We were initially anticipating either continuing to do parsing and validation with the existing JS CLI or with our Rust CLI. We're going to be deprecating the first and the second has a different focus, but we have found a workaround: Using GraphQL-JS wrapped in [`JavaScriptCore`](https://nshipster.com/javascriptcore/), which means no dependencies, and more importantly, no NPM. 

### The Bad News 
After over a year of various people banging their heads on it, the expected code presented in #939 is ultimately unworkable. The biggest issue, sadly, is that Fragments As Protocols is basically impossible to make usable with _any_ nested types given Swift's lack of support for covariance. 

You know how you hate the warning that says that you can't use something as a generic constraint because it has self or associated type requirements? Believe me, no matter how much you hate it, I hate it more. 

Once we realized we couldn't represent fragments as protocols, this broke a bunch of other assumptions, particularly around trying not to duplicate data. @AnthonyMDev has been heads down working on a new solution and is making progress, but there's still a lot of things up in the air. There will be an updated RFC once we've got something we're internally happy with, but this has very much been a two steps forward, two steps back problem. 

### The plan!

Here is the rough outline of what we're going to be working on: 

- [ ] Create new desired generated code
- [x] Use `JavascriptCore` to take advantage of GraphQL-JS's parsing and validation without the need for NPM.
- [ ] Use `JavascriptCore` it to generate an intermediate representation we can use to generate swift code
- [ ] Consume intermediate representation from JS to actually generate Swift code using Swift code. 
- [ ] Proceed with the rest of the plan in #939. 

### Bugs in the existing implementation we want to fix

- [x] #1133, #955 Pluralization makes some weird choices that can't be overridden
- [ ] #126, #184 Better handling of fragment and its field having the same name.
- [ ] #72, #463 Capitalization is not respected
- [ ] #843, #1222 Occasionally parameter or enum case order is nondeterministic
[...]

PR #1716, opened 2021-03-14
parentType in SelectionSet of ApolloCodegenLib should not be array

liuliu:
From the ir.ts: https://github.com/apollographql/apollo-ios/blob/main/Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/ir.ts

This is not an array. Fixed this problem such that we can query
parentType properly now from SelectionSet.

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
+++ b/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
@@ -68,7 +68,7 @@ public class CompilationResult: JavaScriptObject {
   }
   
   public class SelectionSet: JavaScriptObject {
-    private(set) lazy var parentType: [GraphQLCompositeType] = self["parentType"]
+    private(set) lazy var parentType: GraphQLCompositeType = self["parentType"]
     
     private(set) lazy var selections: [Selection] = self["selections"]
   }


AnthonyMDev:
Thank you for this fix! @martijnwalraven does this make sense to you?

@liuliu Are you currently utilizing the JS compiler here? I'm working on rebuilding our CodeGeneration to use it, but I'm not aware of anyone else using this yet. I'd love to hear about what you are doing with this and if there is any cross over!

liuliu:
@AnthonyMDev Yes, currently, I am trying to build interoperability layer between Dflat's flatbuffers schema and the Apollo GraphQL generated code. There ain't much done yet, just some initial probing: https://github.com/liuliu/dflat/blob/unstable/src/graphql/codegen.swift

The idea is a bit different from Apollo already supported normalized cache protocol hence I didn't try to fit in existing normalized cache, but rather, to generate proper flatbuffers schemas, helper initializers etc so the Dflat objects can be init'ed properly from GraphQL objects.

AnthonyMDev:
Ah very cool! Thanks for sharing! Just a heads up that we are currently planning on re-working our entire codegen models pretty significantly. See #1700 . Not sure if that is going to be relevant to your work.

AnthonyMDev:
I'm going to go ahead and merge this, as I'm 99% sure this is correct. If @martijnwalraven has an issue with it, we can talk about reverting.

PR #1720, opened 2021-03-16
Add section on defining custom scalars for dynamic types (ie, JSON)

designatednerd:
We get this question A LOT, so it seemed like we should make this answer easier to surface.

### DIFF
--- a/docs/source/fetching-queries.md
+++ b/docs/source/fetching-queries.md
@@ -125,6 +125,35 @@ If you want to use the custom scalars within your code, you must set `passthroug
 
 Once you've done that, you can either create your own type locally or use a `typealias` to declare an equivilent. This is very, very frequently used with `Date` types. Please see the [Custom Scalar Playground Page](https://github.com/apollographql/apollo-ios/blob/main/Playgrounds/ApolloMacPlayground.playground/Pages/CustomScalars.xcplaygroundpage/Contents.swift) available within the `apollo-iOS` repo for a full example using a custom date type.
 
+#### JSON Custom Scalars
+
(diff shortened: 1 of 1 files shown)


docs/source/fetching-queries.md
```diff
@@ -125,6 +125,35 @@ If you want to use the custom scalars within your code, you must set `passthroug
 
 Once you've done that, you can either create your own type locally or use a `typealias` to declare an equivilent. This is very, very frequently used with `Date` types. Please see the [Custom Scalar Playground Page](https://github.com/apollographql/apollo-ios/blob/main/Playgrounds/ApolloMacPlayground.playground/Pages/CustomScalars.xcplaygroundpage/Contents.swift) available within the `apollo-iOS` repo for a full example using a custom date type.
 
+#### JSON Custom Scalars
```
AnthonyMDev, line 128: Love this section, but the heading names are confusing me. The "JSON Custom Scalars" header I think should be for the entire Custom Scalars explanation (including the section just above this one, "Notes on working with Custom Scalars"). Then this section should be titled something like "Custom Scalars With Multiple Return Types" or something like that. 

Unless I'm confused here?

AnthonyMDev (approved):

docs/source/fetching-queries.md
designatednerd: 🤔 I kind of wanted this for SEO since JSON is, by far, the most frequent one of these people ask about, but you're right that this covers more than JSON. Lemme think of a better header.

designatednerd:
Got approval from @AnthonyMDev via slack so 🚀

Issue #1658, opened 2021-02-15
RFC: Swap out SQLite dependency


designatednerd:
`SQLite.swift` does not appear to still be actively developed, as it hasn't [received a new commit for the last year](https://github.com/stephencelis/SQLite.swift). We'd like to make sure we're using something that gets updates reasonably often, so I propose moving us to a different underlying SQLite dependency. 

There are two big ones I can think of: [`FMDB`](https://github.com/ccgus/fmdb) and [`GRDB`](https://github.com/groue/GRDB.swift). Of those two, I'm leaning towards `GRDB` since: 

- It is in Swift, which lines up with our commitment to keep things as Swifty as possible. 
- It also supports `SQLCipher`, which I know some folks are using for db encryption.
- It is **very** actively maintained - as of my most recent look there were zero open issues or PRs and the last commit was two hours ago. 

I'm planning to do some tests to make sure databases created by `SQLite.swift` can be opened by `GRDB`, but I *think* this shouldn't be a big deal. 

Would love to hear from folks using `ApolloSQLite` about how swapping out the underlying library would affect you: 

- Are you 👍  or 👎  on this change, or something more mixed? 
- Are you using `SQLite.swift` outside of Apollo as well, or only with Apollo?
- Is it a serious problem if I can't make the migration from an old database work, and you have to create a new database?

groue:
Hello, I'm the maintainer of GRDB. Thanks for considering using it!

> * Are you using `SQLite.swift` outside of Apollo as well, or only with Apollo?

Applications can depend on both SQLite.swift and a future Apollo+GRDB without any conflict.

Now, mixing both the system SQLite and SQLCipher in a single binary is a recurrent pain point. The problem is that both libraries expose the same C symbols, and this trips up the linker. In an ideal world, Apollo would accept both, so that applications can choose which SQLite flavor they prefer (when they maintain their own database for whatever purpose). This is easy with CocoaPods (with an `Apollo/SQLCipher` subspec), but I have no clue for other package management systems.

Also, SQLite throws errors when two database connections attempt at concurrently write in a given database file (an SQLite database has a single write lock). It also throws errors when reads happen concurrently with writes, unless the database is in the WAL mode. If you intend to expose the Apollo database file to both GRDB and SQLite.swift, then you'll have to prevent or deal with those errors. If the database file is not shared, then you don't have to worry.

> * Is it a serious problem if I can't make the migration from an old database work, and you have to create a new database?

[...]

PR #1722, opened 2021-03-16
SQLite dependency inversion

designatednerd:
In this PR, in preparation for ripping out `SQLite.swift` (see #1658), I've pulled stuff out into a protocol to try to figure out _exactly_ which APIs we're using in `SQLite.swift` so we can make a better determination of how to replace it. 

Doing this as a separate PR since a) it gives me more options when trying to replace it and b) helps me know what I done broke when I do try to replace it 😛

### DIFF
--- a/Sources/ApolloSQLite/SQLiteDatabase.swift
+++ b/Sources/ApolloSQLite/SQLiteDatabase.swift
@@ -7,7 +7,9 @@
 //
 
 import Foundation
+#if !COCOAPODS
 import Apollo
+#endif
 
 public struct DatabaseRow {
   let cacheKey: CacheKey
--- a/Sources/ApolloSQLite/SQLiteDatabase.swift
+++ b/Sources/ApolloSQLite/SQLiteDatabase.swift
@@ -1,11 +1,3 @@
-//
-//  SQLiteDatabase.swift
-//  ApolloSQLite
-//
-//  Created by Ellen Shapiro on 3/16/21.
-//  Copyright © 2021 Apollo GraphQL. All rights reserved.
-//
-
 import Foundation
 #if !COCOAPODS
 import Apollo
--- a/Sources/ApolloSQLite/SQLiteDatabase.swift
+++ b/Sources/ApolloSQLite/SQLiteDatabase.swift
@@ -1,0 +1,50 @@
+//
+//  SQLiteDatabase.swift
+//  ApolloSQLite
+//
+//  Created by Ellen Shapiro on 3/16/21.
+//  Copyright © 2021 Apollo GraphQL. All rights reserved.
+//
+
+import Foundation
+import Apollo
+
+public struct DatabaseRow {
+  let cacheKey: CacheKey
+  let storedInfo: String
+}
+
+public protocol SQLiteDatabase {
+  
+  init(fileURL: URL) throws
+  
+  func createRecordsTableIfNeeded() throws
+  
+  func selectRawRows(forKeys keys: Set<CacheKey>) throws -> [DatabaseRow]
+
+  func addOrUpdateRecordString(_ recordString: String, for cacheKey: CacheKey) throws
+  
+  func deleteRecord(for cacheKey: CacheKey) throws
+  
+  func clearDatabase(shouldVacuumOnClear: Bool) throws
+  
+}
+
+public extension SQLiteDatabase {
+  
+  static var tableName: String {
+    "records"
+  }
+  
+  static var idColumnName: String {
+    "_id"
+  }
+
+  static var keyColumnName: String {
+    "key"
+  }
+
+  static var recordColumName: String {
+    "record"
+  }
+}
--- a/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
+++ b/Sources/ApolloSQLite/SQLiteNormalizedCache.swift
@@ -9,15 +9,69 @@ public enum SQLiteNormalizedCacheError: Error {
   case invalidRecordShape(object: Any)
 }
 
+public final class SQLiteDotSwiftDatabase: SQLiteDatabase {
+  private var db: Connection!
+  
+  private let records: Table
+  private let keyColumn: Expression<CacheKey>
+  private let recordColumn: Expression<String>
+  
+  public init(fileURL: URL) throws {
+    self.records = Table(Self.tableName)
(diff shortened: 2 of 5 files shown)


Sources/ApolloSQLite/SQLiteDatabase.swift
```diff
@@ -0,0 +1,44 @@
...
+#if !COCOAPODS
+import Apollo
+#endif
+
+public struct DatabaseRow {
+  let cacheKey: CacheKey
+  let storedInfo: String
+}
+
+public protocol SQLiteDatabase {
```
AnthonyMDev, line 11: Would like to see the functions on this protocol very well documented. If someone wants to inject a DIFFERENT SQLLite dependency in the future, they should be able to easily understand what their adapter needs to do in each of these functions.

AnthonyMDev (approved):
This is so awesome! Would just like to see documentation of the protocol. Otherwise LGTM!

Sources/ApolloSQLite/SQLiteDatabase.swift
designatednerd: oh for sure - i was thinking i'd do that when i hit the final version but it's probably just as well to do it now given my goldfish brain

AnthonyMDev: As long as we do it at some point. If you want to do it later, go ahead and merge this. Up to you!

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -64,17 +57,8 @@ public final class SQLiteNormalizedCache {
...
-  private func createTableIfNeeded() throws {
-    try self.db.run(self.records.create(ifNotExists: true) { table in
-      table.column(id, primaryKey: .autoincrement)
-      table.column(key, unique: true)
-      table.column(record)
-    })
-    try self.db.run(self.records.createIndex(key, unique: true, ifNotExists: true))
-  }
-
   private func mergeRecords(records: RecordSet) throws -> Set<CacheKey> {
```
groue, line 60: Hello, since I'm a recent Apollo user, I dare to chime in!

This `mergeRecords` method performs several database operations.

If it were wrapped in a transaction, SQLite would be faster. I do not know if it is a concern.

If it were wrapped in a transaction, you'd get never get a half-modified database on disk, should an error happen at some point during the method execution. I also do not know if it is a concern.

If it were wrapped in a transaction, eventual concurrent reads would not see a half-modified database. Maybe a concern, or not.

AnthonyMDev: This is the pinnacle of open source collaboration. <3 
A maintainer of a popular open source repo being a user of another repo _using_ their repo. And making suggestions to improve the usage of their repo using their knowledge of both of them. :) Haha thanks @groue this is a great comment.

AnthonyMDev: I guess technically we aren't using GRDB _yet_ .... but we are getting there. >_<

groue: That's why this pull request is a very good idea :-) I tried to ask questions in a way that is independent from the preferred adapter. I surely don't want to imply anything else or to influence your decision. It's just that defining an adapter protocol is a good opportunity to wonder about how Apollo interacts with its cache, and document the expected behavior, as you said above.

AnthonyMDev: Do you think that the `SQLiteDatabase` protocol should have some concept of a transaction built-in? I imagine that most any SQLLite implementation would likely have transactions and being able to utilize them via the inverted dependency protocol would be valuable.

groue: I don't know for sure. I mentioned `mergeRecords` because it looks like a method that could enjoy a transaction (for speed, or in order to enforce integrity guarantees). Transactions, when needed, are usually defined by the database client (here, Apollo), because only the database client knows how individual database operations should be grouped. Maybe Apollo does not need any transaction! But if it does, then I would indeed ask the adapter to provide transaction boundaries.

Yes, all "serious" SQLite wrappers make it able to perform transactions. APIs fall generally into three groups:

1. Block-based transactions

    ```swift
    // Form 1
    wrapper.transaction {
         wrapper.doStuff()
    }

    // Form 2
    wrapper.transaction { dbAccess in
         dbAccess.doStuff()
    }

    // Form 3
    wrapper.perform {
         wrapper.doStuff()
         wrapper.save()
    }
    ```

    SQLite.swift and Realm are form 1. FMDB and GRDB are form 2. Core Data is form 3.

2. "Free-form" transactions

    ```swift
    wrapper.beginTransaction()
    wrapper.doStuff()
    wrapper.commit()
    ```

    Most SQLite implementations that allow raw SQL are here. Some kind of concurrency control must be added.

3. Transaction objects

    ```swift
    let transaction = wrapper.transaction()
    transaction.doStuff()
    transaction.commit()
    ```

    Core Data can be used like this, by detaching an NSManagedObjectContext. GRDB as well, by creating a new DatabaseQueue. OK, this is really not frequent.

Some wrappers distinguish *read* from *write* transaction. This is an opportunity to improve performance by allowing multiple readers with a single writer.

I think you can accept all forms above with an adapter that looks like:

```swift
adapter.write { transaction in
    transaction.doStuff()
}
```

Something should serialize writes (either the adapter contract, or Apollo).

If Apollo needs to read data, with a concern about cache integrity, and a need for multiple readers and single writer for adapters that support it, then a read transaction is useful as well:

```swift
let data = adapter.read { transaction in
    transaction.loadStuff()
}
```

---

This only applies if you think Apollo needs transactions.

designatednerd: @groue @AnthonyMDev 🎶 It's the ciiiiiiiircle of liiiiiiiife... 🎶

I think i probably want to leave transaction support out for this PR since i'm just trying to do a 1:1 "what are we already doing" conversion before adding any further complexity. 

However, you're absolutely correct that this would be way better, particularly from a standpoint of write-ahead log support (which is better for things with extensions etc). I'll take a look at that in my next PR.

designatednerd:
Merging this bit.

PR #1723, opened 2021-03-17
Allow callbackQueue to be specified for GraphQLWatcher

ketenshi:
As mentioned in https://github.com/apollographql/apollo-ios/discussions/1718 there may be situations where it makes sense to call the result handler for the `GraphQLWatcher` in a background thread to lessen the load of the main thread.

### DIFF
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -22,7 +22,7 @@ public protocol ApolloClientProtocol: class {
   /// - Parameters:
   ///   - query: The query to fetch.
   ///   - cachePolicy: A cache policy that specifies when results should be fetched from the server and when data should be loaded from the local cache.
-  ///   - queue: A dispatch queue on which the result handler will be called. Defaults to the main queue.
+  ///   - queue: A dispatch queue on which the result handler will be called. Should default to the main queue.
   ///   - contextIdentifier: [optional] A unique identifier for this request, to help with deduping cache hits for watchers. Should default to `nil`.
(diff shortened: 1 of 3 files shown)


Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -37,10 +37,12 @@ public protocol ApolloClientProtocol: class {
   /// - Parameters:
   ///   - query: The query to fetch.
   ///   - cachePolicy: A cache policy that specifies when results should be fetched from the server or from the local cache.
+  ///   - callbackQueue: A dispatch queue on which the result handler will be called. Defaults to the main queue.
   ///   - resultHandler: [optional] A closure that is called when query results are available or when an error occurs.
   /// - Returns: A query watcher object that can be used to control the watching behavior.
   func watch<Query: GraphQLQuery>(query: Query,
                                   cachePolicy: CachePolicy,
+                                  callbackQueue: DispatchQueue,
```
ketenshi, line 45: This is a breaking change so I'm not sure what the standard procedure is for this.

We don't use this method in our app and create the watcher directly, but I think this change aligns with all the other network methods in the protocol

designatednerd: Luckily we're still at 0.x so you can sneak this one in. 😇

You should update the docs to `should default to the main queue` since you can't declare a default on a protocol.

designatednerd (commented):
looks reasonable, one minor note on the docs in response to your question

ketenshi:
@designatednerd lmk if there's something else you need from me for this PR

designatednerd (approved):
Thanks for the poke. LGTM!

PR #1727, opened 2021-03-18
#720: Changed fragment spreads to still collect the fields for new types

### LINKED ISSUE #720: Crash when gets new interface type from server
### **Version**
- ApolloClient 0.14.0

### **Problem**

when I have the following types:
```graphql
interface Notification {
  id: ID!
}

type NotificationArticle implements Notification {
  id: ID!
  article: Article!
}

type NotificationComment implements Notification {
  id: ID!
  comment: Comment!
}
```

and my app use query from this fragment.

```graphql
fragment NotificationFragment on Notification {
  id
  ... on NotificationArticle {
    ... ArticleFragment
  }
  ... on NotificationComment{
    ... CommentFragment
  }
}

```
Later, I add another implementation of `Notification` below

```graphql
type NotificationShare implements Notification {
  id: ID!
  share: Share!
}
```

Without updating app with the new schema, my app will crash when I tried to read `id` from  `NotificationShare`. The crash happens in generated `API.swift` below

```swift
public var id: GraphQLID {
 get {
  return resultMap["id"]! as! GraphQLID
 }
 set {
   resultMap.updateValue(newValue, forKey: "id")
 }
}
```

It seems that Apollo Core doesn't read any field from response when `__typename` is not exist in fragment's possible types below (from `GraphQLExecutor.swift`)

```swift
if let runtimeType = runtimeType, fragment.possibleTypes.contains(runtimeType) {
  try collectFields(selections: fragment.selections, forRuntimeType: runtimeType, into: &groupedFields, info: info)
}
```
### **Expected Behavior**
The app should not crash because `id` is a field in `Notification` interface.

CraigSiemens:
Fixes #720 

When collecting fields for a `GraphQLFragmentSpread`, removed the check that the `runtimeType` is one of the `possibleTypes` the fragment contains. My comment on the original issue goes into more details. https://github.com/apollographql/apollo-ios/issues/720#issuecomment-801537410

This allows Fragments on an interface type to support new types returned from the server with out requiring an app update.

## Example
Based the starwars API. Let's say the app has the following query and the codegen used the flag `mergeInFieldsFromFragmentSpreads: false`.
```graphql
query HanSolo {
  character(id: 1002) {
    name
    friends {
      ...Friend
    }
  }
}

fragment Friend on Character {
  name
}
```

Some time later after releasing the app, the backend realizes it's missing Chewbacca. So they add a new type `Wookie` that conforms to `Character` and add the missing friend relationships.

Now when the already released app fetches the `HanSoloQuery`, it will try to parse the response from the backend. The `__typename` coming back on the response for one of the friends is `Wookie`, but the code generated type for the `Friend` fragment doesn't have that listed as one of the possible types so it skips collecting the `name` field. Then when the app tries to read the `name` field from the `Friend` type, it crashes because the getter for that field assumes it will be there.
```swift
return resultMap["name"]! as! String
```

## Additional Notes
I also found that this method isn't called by any of the tests. A future improvement could be to add tests that use the generated code using different flags. It seems like the default value `mergeInFieldsFromFragmentSpreads: true` is only used in the tests currently, and there could be other flags with similar gaps in their testing.

### DIFF
--- a/Sources/Apollo/Decoding.swift
+++ b/Sources/Apollo/Decoding.swift
@@ -59,12 +59,10 @@ private func collectFields(from selections: [GraphQLSelection],
     case let fragmentSpread as GraphQLFragmentSpread:
       let fragment = fragmentSpread.fragment
 
-      if let runtimeType = runtimeType, fragment.possibleTypes.contains(runtimeType) {
-        try collectFields(from: fragment.selections,
-                          forRuntimeType: runtimeType,
-                          into: &groupedFields,
-                          variables: variables)
-      }
+      try collectFields(from: fragment.selections,
+                        forRuntimeType: runtimeType,
+                        into: &groupedFields,
+                        variables: variables)
     case let typeCase as GraphQLTypeCase:
       let selections: [GraphQLSelection]
       if let runtimeType = runtimeType {


designatednerd:
You're correct that not having test coverage on this isn't great - the `StarWarsAPI` does use `mergeFieldsFromFragmentSpreads` so we'd need to have a different API where that isn't set to test this. 

However I think there's a couple rabbit holes this may open up (especially on "things that look like they conform to a protocol but don't, they just have fields with the same names") and I've asked @AnthonyMDev to take a look at this when he has a second since he's deep in the wild world of codegen and typing.

AnthonyMDev:
Thanks for the contribution @CraigSiemens, and for the thorough explanation here! 

I agree that we need more testing around these models in general, but with the new codegen we are creating, we don't currently intend on this being an option. We will _always_ be merging in fragment fields. However the new codegen dramatically decreases the number of lines of generated code in the first place, so we don't anticipate this should be an issue anymore on code size. (We may consider adding that option back in if user feedback expresses a need for it.)

I haven't gotten too deep into the execution for the new codegen. I'm hoping to not have to change the execution too much, but I suspect that a lot of this will change pretty dramatically.

As for the current system, I **_think_** @designatednerd is correct here that this can cause other issues. But I'm not familiar enough with the current system to know for certain. @martijnwalraven might be able to help us reason this through.

Consider the altered scenario here: 

Given the schema:

```graphql
interface Character {
  id: ID!
  name: String!
  friends: [Friend!]!
}

interface Friend {
   ...
}

type Human implements Friend & Character {
  ...
}
```

And the query:
```graphql
query HanSolo {
  character(id: 1002) {
    name
    friends {
      ...Friend
    }
  }
}

fragment Friend on Character {
  name
}
```

Then in the future, as in your example, `Wookie` is added to the schema later.
```graphql

type Wookie implements Friend {
  ...
}
```

Note that:

1. The `friends` field has a type `[Friend!]!`, not `[Character!]!`.
2. `Wookie` implements `Friend`, but not `Character` 

This would mean that:
1. The fragment is being used as a type condition (it only resolves those fields if the runtime type implements the protocol).
2. The server would not return the `name` field for the `Wookie`. (Since it's not a `Character`).

With this PR in place, would this cause the `Wookie` selection set to then try to collect the `name` field? That would then cause a `JSONDecodingError.missingValue` during execution, which would cause the entire query resolution to fail.
Or would this be handled by the `GraphQLTypeCase` and never attempt to collect the fields for the fragment? 


@martijnwalraven , do you remember how this works? If this case isn't an issue, do you see any other edge cases in which removing this check causes issues?

CraigSiemens:
Using `Friend` for both an interface and a fragment confused me a bit. I'll try using `FriendFragment` and `FriendInterface` just so I can keep them straight.

My understanding is in that example, the generated type `HanSoloQuery.Character.Friend` would 
have it `selections` change from
```swift
internal static var selections: [GraphQLSelection] {
  return [
    GraphQLField("__typename", type: .nonNull(.scalar(String.self))),
    GraphQLFragmentSpread(FriendFragment.self),
  ]
}
```
to something like
```swift
internal static var selections: [GraphQLSelection] {
  return [
    GraphQLTypeCase(
      variants: [
        "Human": FriendFragment.selections,
        "Droid": FriendFragment.selections,
        // I'm not 100% sure about this since I cant find an example of this I can run through codegen.
        // I suspect this is the case since `__typename` will only return concrete type names.
        // It would only support variants that it knows about at compile time.
      ],
      default: [
        GraphQLField("__typename", type: .nonNull(.scalar(String.self))),
      ]
    )
  ]
}
```
That's because the `FriendFragment` now only applies to a subset of the types that can be returned from a field of type `FriendInterface`. In this case, if the app was returned a `Wookie` from the API, it would use the default fields due to this case 
https://github.com/apollographql/apollo-ios/blob/f4c058311f460a7729487561436c2a9cba689bc8/Sources/Apollo/Decoding.swift#L68-L79

That would result in a `HanSoloQuery.Character.Friend` instance where `asHuman` and `asDroid` both return `nil`, and since there's no additional fields on it, there's no risk of accessing a property that causes a force unwrap and crashes the app.

This PR should be adding the same kind of default behaviour when the `selections` contains a `GraphQLFragmentSpread`.

AnthonyMDev:
Assuming the generated code came out the way that you describe above, I agree with you. 

The only reason I can come up with for this check to be needed in that case is if the response data for some reason returns an object in the `friends` list that doesn't actually implement the `Character` interface. But that would only happen if someone mutates the cache manually to make some invalid data, or the server itself has some very bad bugs. It would be a response that is not schema compliant.

Since we should be able to make the assumption that response data is going to be compliant with the GraphQL spec for the given schema, I think this check would be unnecessary!

However I just ran a test on this, and that's not how it works currently. You can look at the `720-test` branch that I just pushed up.

In the [first commit](https://github.com/apollographql/apollo-ios/commit/2339e2ed6e0a30fb824b03d81d37e7b598c04a52), I changed the StarWars API to use `mergeInFieldsFromFragmentSpreads` to `false`. You can see the diff to see how the API we regenerated.

Then in the [second commit](https://github.com/apollographql/apollo-ios/commit/faa93b95c75791d1c1ad46b78e53337deed637d3), I manually changed the schema and queries to reflect the scenario I posed above. You can see in the diff there that when the `API.swift` file is regenerated, it does not create the `GraphQLTypeCase` variants as you proposed.

@martijnwalraven would know better, but my guess is that somewhere in the codegen logic, we are looking at the possible types to deduce that, since all known types that conform to the `Friend` interface also conform to the `Character` interface, it assumes that it can just select the fields from the fragment.

# Conclusion

I believe this means that this statement is correct:
> With this PR in place, would this cause the Wookie selection set to then try to collect the name field? That would then cause a JSONDecodingError.missingValue during execution, which would cause the entire query resolution to fail.
Or would this be handled by the GraphQLTypeCase and never attempt to collect the fields for the fragment?

In order to make this work, the code generation library itself would need to be modified pretty heavily. As we are working on replacing CodeGen completely, I don't think that it's a great idea for us to make large modifications to the legacy codegen. Any changes we make could cause bugs that we would then be responsible for fixing in the legacy codegen, and I don't want us dedicated any more hours to that than is necessary. (I'd rather spend those hours working on the new codegen.

@CraigSiemens Thank you for working through this, but I'm going to close this PR, as I believe the work necessary to do this safely is out of scope for us at the moment. If you have any ideas on how to work around this in a more simple manner please feel free to re-open it!

# Looking to the Future

This issue has brought some important concepts to light however. 

### New Codegen

As we continue work on the new version of code generation, specifically the determination of `selections` and the execution pipeline, we should take a different approach.

**The assumption that the downloaded schema contains all possible runtime types for a selection set should be discarded. Rather, we should always assume that new types may be added to the schema after code generation.**

We should be considering how to handle unrecognized types gracefully (or at least as gracefully as is possible).

This will likely make the implementation of the executor more complicated than it already is, but I believe it is necessary to resolve this and other problems like it in the future.

### Type Metadata API

@benjamn brought up something in a conversation this week. He mentioned an idea for the future about a mechanism that would allow a GraphQL server to provide additional schema information for unknown types in the response data. Theoretically, clients could consume this information at runtime to better inform the GraphQL executor about what interfaces an unknown type implements. 

There are no plans yet to support this, and it likely is a huge undertaking. But a mechanism like this could help us to more elegantly resolve this issue in the future.

CraigSiemens:
Thanks for creating the example branch. My example was what I was hoping the generated code would be for it to work correctly. Unfortunately your example shows off two issues with the generated code.

## Issue 1

The first is the issue I was trying to solve. If a `type Wookie implements Friend` is returned from the api and the app's code try's accessing the `id`, the app will crash because the generated code assumes the it will be there, but the decoding is removing it.

Setting `mergeInFieldsFromFragmentSpreads` to `true` does work around the issue since the `GraphQLTypeCase` code path defaults to the same field that we're present when `mergeInFieldsFromFragmentSpreads` was `false`.

My change was just to use the same behaviour.

## Issue 2

The second issue is from the absence of a change to `HeroFriendsOfFriendsNamesQuery.Data.Hero.Friend`.

In your changes, you made both [`Human`](https://github.com/apollographql/apollo-ios/commit/faa93b95c75791d1c1ad46b78e53337deed637d3#diff-07dadc6cfaf84bb0469f17e7e48570f25d7c626e8dfa83e426885e8d33b637cdR924) and [`Droid`](https://github.com/apollographql/apollo-ios/commit/faa93b95c75791d1c1ad46b78e53337deed637d3#diff-07dadc6cfaf84bb0469f17e7e48570f25d7c626e8dfa83e426885e8d33b637cdR1190) conform to `Friend`.

According to the schema, there is no relationship between `Friend` and `Character`. Not all `Friend`s are `Character`s, and not all `Character`s are `Friend`s. The code gen makes the incorrect assumption
- The only types implementing `Friend` in the schema are `Human` and `Droid`
- `Human` and `Droid` both implement `Character`
- therefor all types conforming to `Friend` are a `Character`

So again if the API returns a `type Wookie implements Friend` and the app's code tries to access the `.friends` property, the app will crash.

---

> With this PR in place, would this cause the Wookie selection set to then try to collect the name field? That would then cause a JSONDecodingError.missingValue during execution, which would cause the entire query resolution to fail.
Or would this be handled by the GraphQLTypeCase and never attempt to collect the fields for the fragment?

I took your example and added a `Wookie` in [this commit](https://github.com/apollographql/apollo-ios/commit/67d352473f65c904170cdd5572196de3f3ee211d#diff-07dadc6cfaf84bb0469f17e7e48570f25d7c626e8dfa83e426885e8d33b637cd).

```graphql
type Wookie implements Friend {
  # The ID of the wookie
  id: ID!

  # What this wookie calls themselves
  name: String!
}
```

This is what the selections looks like before. It incorrectly assumed that every object that could component through the `friends` property would have a `friends` field on it. That's incorrect cause the `Friend` interface doesn't have a `friends` field.
```swift
      public struct Friend: GraphQLSelectionSet {
        public static let possibleTypes: [String] = ["Human", "Droid"]

        public static var selections: [GraphQLSelection] {
          return [
            GraphQLField("__typename", type: .nonNull(.scalar(String.self))),
            GraphQLField("id", type: .nonNull(.scalar(GraphQLID.self))),
            GraphQLField("friends", type: .list(.object(Friend.selections))),
          ]
        }
```

Once the backend adds a `Wookie` type, the generated code changes to this. This is more flexible as it now handles `Wookie` and unknown types that the API could add in the future as long as they implement the `Friend` interface. And the api should never return a type that doesn't implement `Friend` through the `friends` field as that would break agreed upon schema.
```swift
      public struct Friend: GraphQLSelectionSet {
        public static let possibleTypes: [String] = ["Human", "Droid", "Wookie"]

        public static var selections: [GraphQLSelection] {
          return [
            GraphQLTypeCase(
              variants: ["Human": AsHumanOrDroid.selections, "Droid": AsHumanOrDroid.selections],
              default: [
                GraphQLField("__typename", type: .nonNull(.scalar(String.self))),
                GraphQLField("id", type: .nonNull(.scalar(GraphQLID.self))),
              ]
            )
          ]
        }
```

# My takeaways
There is a mismatch in some the assumptions made by the codegen and the apollo sdk.

In the case of issue 1, codegen adds fields to the selection to support new types implementing an interface, but the sdk's parsing only supports the types known at build time.

For issue 2, the codegen assumes that if all the types that implement interface `A` also implement interface `B`, then a object from a field of type `A` is always a `B`. When the API returns a new type implementing `A`, that assumption is no longer valid.

Both of these cause the API returning a new type before the app has been updated to be a crashing change. But as far as the API is concerned, it's still following the schema by only returning types implementing the interface.

## Goal
My only goal is for fields whose type is an interface to support any type that conforms to it, not just the ones that were known at build time. That would make the app way more resilient to future API changes. I still haven't seen a case where my solution would fail, especially considering `mergeFieldsFromFragmentSpreads:true` ends up using the same `GraphQLField`s without an issue.

## Workarounds
The workaround we had right now was a hacky server change to hide the newly added type from the app. It's still returning the exact same fields, we just manually changed `__typename` to be one the app knows about. It's now a great solution, but it does avoid us having to force our users to update their app.

We could turn on the `mergeFieldsFromFragmentSpreads` flag, but out git diff after turning on that flag was
```
14 files changed, 119524 insertions(+), 7092 deletions(-)
```
Apart from the one place that's changed to use a `GraphQLTypeCase` that solves our current issue, an additional 100k and the increase to compile times isn't something we want to have to deal with.

CraigSiemens:
Just for completeness I did find another workaround that we're going to be using in the meantime.

Don't create a fragment on an interface type. This causes the selections for the field that used to contain a `GraphQLFragmentSpread`, to be the `GraphQLTypeCase` that used to be in the fragment. Then the decoding can use the default case from the `GraphQLTypeCase` when it gets a new `runtimeType`.

AnthonyMDev:
Thank you for the awesome writeup @CraigSiemens! Though I'm glad you have a workaround, I'm definitely not happy that you have to go this route. This issue has really helped to shape the next iteration of Codegen. I'm confident we will not encounter these same issues when that is released. 

I wish there was an easy way to correct this in the meantime, but I think it's a better use of time to focus on getting Codegen 2.0 released ASAP.

PR #1729, opened 2021-03-18
Avoid having extra ? at the end of request url when queryItems are empty

tranvutuan:
Summary
----
Avoid having extra ? at the end of request url when queryItems are empty

Details
----
 
Currently `GraphQLGETTransformer` is used to create query items for GET request endpoint. If property `body` of this struct is empty ( does not have value for a given key) such as:

> key : "variables"
> value : nil

which leads to queryItems is empty. By doing `components.queryItems = queryItems`, then the url ends up to have extra ? at the end like following: `https://this_is_sample/abcd?` instead of `https://this_is_sample/abcd`

Fix
----
Only assign `queryItems` to `URLComponent` when it is not empty

### DIFF
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -42,7 +42,10 @@ public struct GraphQLGETTransformer {
       return nil
     }
 
-    components.queryItems = queryItems
+    if queryItems.count > 0 {
+      components.queryItems = queryItems
+    }
+
     components.percentEncodedQuery =
       components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
 
--- a/Sources/Apollo/GraphQLGETTransformer.swift
+++ b/Sources/Apollo/GraphQLGETTransformer.swift
@@ -42,7 +42,7 @@ public struct GraphQLGETTransformer {
       return nil
     }
 
-    if queryItems.count > 0 {
+    if queryItems.apollo.isNotEmpty {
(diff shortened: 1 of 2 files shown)


Sources/Apollo/GraphQLGETTransformer.swift
```diff
@@ -42,7 +42,10 @@ public struct GraphQLGETTransformer {
       return nil
     }
 
-    components.queryItems = queryItems
+    if queryItems.count > 0 {
```
designatednerd, line 45: note to myself to use `.apollo.isNotEmpty` here, but there are some shenanigans around the import I won't make you deal with

designatednerd (approved):

designatednerd:
@tranvutuan I can't merge this until you sign the CLA, but this looks good. Mind signing it?

Sources/Apollo/GraphQLGETTransformer.swift
tranvutuan: I wasn't aware of `.apollo.isNotEmpty` until now. Have just updated the PR.

designatednerd: Cool - I'll have to do some silliness to make cocoapods play nice later but i appreciate you doing this bit!

designatednerd:
Cool, thanks for the heads up.

tranvutuan:
> Cool, thanks for the heads up.

Got a green light from our end and I have just signed the CLA.

designatednerd:
Woo hoo, thank you!

PR #1733, opened 2021-03-26
Replace `class` constraint with `AnyObject`

yhkaplan:
Pretty insignificant, but the recommended practice in Swift now is to use AnyObject instead of class, and recently there have been motions toward a new reference type (`actor`) that the `class` naming makes a little unclear, while AnyObject is more of a generic and all-encompassing term for reference types. Also, the class constraint typealias may be removed in future versions of Swift. 

ref: https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject

### DIFF
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// The `ApolloClientProtocol` provides the core API for Apollo. This API provides methods to fetch and watch queries, and to perform mutations.
-public protocol ApolloClientProtocol: class {
+public protocol ApolloClientProtocol: AnyObject {
 
   ///  A store used as a local cache.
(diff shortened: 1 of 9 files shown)


AnthonyMDev:
LGTM! Thanks for the clean up!

cltnschlosser:
With Xcode 12.5 in RC now, it would be nice to get a version released with this warning fixed. Thanks!

designatednerd:
Hey - I'm working on some tests for some web socket stuff that was merged earlier today before I do a release but will try to get this out early next week.

cltnschlosser:
> Hey - I'm working on some tests for some web socket stuff that was merged earlier today before I do a release but will try to get this out early next week.

Great, thanks!

designatednerd:
Guh, apple, why you gotta release Xcode on a Monday? will cut a release tomorrow morning.

PR #1734, opened 2021-03-26
Fix bundling of JS Frontend for Codegen Lib with SPM

AnthonyMDev:
Moved the bundled JS files to a directory that is not being ignored by SPM so they package correctly.


Sources/ApolloCodegenLib/Frontend/ApolloCodegenFrontend.swift
```diff
@@ -2,8 +2,16 @@ import Foundation
 import JavaScriptCore
 
 public final class ApolloCodegenFrontend {
-  static let bundle = Bundle(for: ApolloCodegenFrontend.self)
-  private static let libraryURL = bundle.url(forResource: "ApolloCodegenFrontend.bundle", withExtension: "js")!
+  #if SWIFT_PACKAGE
+  private static let bundle = Bundle.module
+  private static let libraryURL = bundle.url(forResource: "ApolloCodegenFrontend.bundle",
```
designatednerd, line 7: Generally I'm in favor, if there's more than one line in the `#if` statement, of indenting things in `#if/#else` blocks for clarity about what's handled in each scenario. Not a hill I want to die on or anything but I do think it makes things more readable.

designatednerd (approved):
LGTM with one comment

Sources/ApolloCodegenLib/Frontend/ApolloCodegenFrontend.swift
AnthonyMDev: Yeah, I get that reasoning. The only reason i don't is because Xcode's auto indent is going to change it back to this the next time I hit ctrl+i ... haha. I always try to keep things consistent with the Xcode behavior there, even if I don't love the formatting, because it makes it easier to use auto indentation.

designatednerd: fair enough. I wish there was a way to more easily override that ...

PR #1737, opened 2021-03-30
Apply cache policy case .default  in fetch and watch

deepumukundan:
The CachePolicy enum has a default policy but it was only used in subscribe. This updates the other 2 locations


designatednerd (approved):
Good catch!

deepumukundan:
@designatednerd Could you merge this? I don't have merge rights

designatednerd:
Thanks for the poke! I had to rerun one test then promptly forgot about it 🤦‍♀️

PR #1738, opened 2021-03-31
Convert GraphQLInputValue protocol to InputValue enum

AnthonyMDev:
Creates a new `InputValue` enum to replace `GraphQLInputValue`.

This is compatible with the current codegen, as all possible generated argument types can be initialized as `InputValue` using `ExpressibleBy[Type]Literal` initializers.

This removes public protocol extensions on Foundation types that exposed internal implementation details publicly to consumers of Apollo.


Apollo.xcodeproj/xcshareddata/xcschemes/StarWarsAPI.xcscheme
```diff
@@ -0,0 +1,67 @@
...
+      </MacroExpansion>
+   </ProfileAction>
+   <AnalyzeAction
+      buildConfiguration = "Debug">
+   </AnalyzeAction>
+   <ArchiveAction
+      buildConfiguration = "Release"
+      revealArchiveInOrganizer = "YES">
+   </ArchiveAction>
+</Scheme>
```
designatednerd, line 67: In theory this should already be shared - any idea why it's being added here?

Sources/Apollo/GraphQLMap.swift
```diff
@@ -0,0 +1,19 @@
+public typealias GraphQLMap = [String: JSONEncodable?]
+
+fileprivate extension Dictionary where Key == String, Value == JSONEncodable? {
+  var withNilValuesRemoved: Dictionary<String, JSONEncodable> {
+    filter { $0.value != nil }
```
designatednerd, line 5: Any particular reason to use this instead of `compactMap { $0.value }`?

Sources/Apollo/GraphQLMap.swift
```diff
@@ -0,0 +1,19 @@
...
+  var graphQLMap: GraphQLMap { get }
+}
+
+public extension GraphQLMapConvertible {
+  var jsonValue: JSONValue {
+    return graphQLMap.withNilValuesRemoved.jsonValue
+  }
+}
+
+public typealias GraphQLID = String
```
designatednerd, line 19: This seems like something we should probably be defining somewhere else but given that I'm not sure where, here's fine for now.

Sources/Apollo/InputValue.swift
```diff
@@ -0,0 +1,118 @@
...
+}
+
+extension InputValue: ExpressibleByStringLiteral {
+  @inlinable public init(stringLiteral value: StringLiteralType) {
+    self = .scalar(value)
+  }
+}
+
+extension InputValue: ExpressibleByIntegerLiteral {
+  @inlinable public init(integerLiteral value: IntegerLiteralType) {
```
designatednerd, line 68: what's the benefit of explicitly specifying this as `@inlineable`? In theory I thought the compiler inlined things like this anyway.

Sources/ApolloCodegenLib/JSON.swift
```diff
@@ -59,18 +59,18 @@ public enum JSONValue: Codable, Equatable {
...
-            return directValue
-        }
-        
-        return try directValue.valueForKeyPath(remainingKeys)
+      guard let directValue = dictionary[currentKey] else {
+        throw JSONValueError.noValueForKey(currentKey)
+      }
+
+      let remainingKeys = Array(keyPath.dropFirst())
+      guard !remainingKeys.isEmpty else {
```
designatednerd, line 67: you can probably use `remainingKeys.apollo.isNotEmpty` here if you find that more readable (I do, which is why I made it)

Tests/ApolloTests/CacheKeyForFieldTests.swift
```diff
@@ -48,8 +68,8 @@ class CacheKeyForFieldTests: XCTestCase {
   }
   
   func testFieldWithInputObjectArgumentIsOrderIndependent() {
-    let field1 = GraphQLField("hero", arguments: ["episode": Episode.jedi, "nested": ["foo": "a", "bar": "b"]], type: .scalar(String.self))
-    let field2 = GraphQLField("hero", arguments: ["episode": Episode.jedi, "nested": ["bar": "b", "foo": "a"]], type: .scalar(String.self))
+    let field1 = GraphQLField("hero", arguments: ["episode": "JEDI", "nested": ["foo": "a", "bar": "b"]], type: .scalar(String.self))
```
designatednerd, line 71: Why are we not able to use the enum here anymore?

Tests/ApolloTests/GraphQLMapEncodingTests.swift
```diff
@@ -3,9 +3,13 @@ import XCTest
...
 import StarWarsAPI
 
-class InputValueEncodingTests: XCTestCase {
+class GraphQLMapEncodingTests: XCTestCase {
+  private struct MockGraphQLMapConvertible: GraphQLMapConvertible {
+    let graphQLMap: GraphQLMap
+  }
   private func serializeAndDeserialize(_ map: GraphQLMap) -> NSDictionary {
-    let data = try! JSONSerializationFormat.serialize(value: map.withNilValuesRemoved)
+    let input = MockGraphQLMapConvertible(graphQLMap: map)
```
designatednerd, line 11: I think you told me this at some point, but remind me why this wrapper is now necessary.

Apollo.xcodeproj/xcshareddata/xcschemes/StarWarsAPI.xcscheme
AnthonyMDev: Maybe you had it local but not shared? It wasn't in my local project so I added it.

Sources/Apollo/GraphQLMap.swift
AnthonyMDev: Nope, I had that and changed it. Same thing, just what we are doing here is filtering out nils, so I figured this was more expressive of the intent.

Sources/Apollo/GraphQLMap.swift
AnthonyMDev: Yes haha. I 100% agree. It was defined with all of this stuff already, and I just moved the GraphQLMap to its own file, so I just moved it with that.

Sources/Apollo/InputValue.swift
AnthonyMDev: Just for performance. You're right, the compiler will 99% do that on it's own, but this makes it more likely that it will get inlined.

Sources/ApolloCodegenLib/JSON.swift
AnthonyMDev: Haha cool. I just fixed auto-indent on this file, didn't even look to improve it.

Tests/ApolloTests/CacheKeyForFieldTests.swift
AnthonyMDev: Well the reason I caught it is because this is being converted into an `InputValue` and there is no `ExpressibleByEnumWithRawValueLiteral` in the swift language haha. 

But this is actually better because that is what the generated code that the current Codegen outputs. We want these tests to reflect the same thing that the generated code looks like, so this is better anyways.

Apollo.xcodeproj/xcshareddata/xcschemes/StarWarsAPI.xcscheme
designatednerd: AH, right - i do have it local and not shared, but it's not shared for Carthage reasons. If it's shared, Carthage will build the SWAPI framework in addition to everything else in the universe.

Sources/Apollo/GraphQLMap.swift
designatednerd: I'd argue `compactMap` also expresses that intent, but I'm not dying on this hill ☠️

Sources/Apollo/InputValue.swift
designatednerd: Does it sort of save the compiler the trouble of determining whether something can be inlined?

Tests/ApolloTests/GraphQLMapEncodingTests.swift
AnthonyMDev: I did not! haha.

It's not, but this test was testing that serialization happened correctly when the `GraphQLMap.withNilValuesRemoved` was called. But that is actually an implementation detail. The only place this is being used when we call `jsonValue` on a `GraphQLMapConvertible`. So two things

1) `withNilValuesRemoved` should really be `fileprivate` (which is now is in GraphQLMap.swift). The only reason it wasn't private before was so it could be called in this test.

2) And reason 1 should be an indicator that your test is not testing the actual intention. These tests should be testing that `GraphQLMapConvertible.jsonValue` returns the correct value to be serialized. Which includes removing the nil values. Currently, if we removed the `.withNilValuesRemoved` from the `jsonValue` computed property, these tests would still pass, but the production code would be broken. Because this test explicitly calls the same behavior we actually want, rather than verifying the behavior occurs. This PR fixes that. :)

Tests/ApolloTests/CacheKeyForFieldTests.swift
designatednerd: I think you could do an extension on  `RawRepresentable where RawValue == String` or something like that.

I'm also slightly confused about where this gets turned from the enum to the raw string value - we can generate it with the string for things specified at compile time, but not runtime.

Sources/Apollo/GraphQLMap.swift
AnthonyMDev: Yeah, I really truly don't care either way myself either.

Tests/ApolloTests/GraphQLMapEncodingTests.swift
designatednerd: Yeah this is always a fun border issue when writing tests. Good catch.

Sources/Apollo/InputValue.swift
AnthonyMDev: Not 100% sure if it improves the build times or not, but I know it's a hint for the compiler that is should prefer inlining this whenever possible.

Tests/ApolloTests/CacheKeyForFieldTests.swift
AnthonyMDev: For the new version of Codegen, I fully intend on improving this. The current codegen already uses the Raw String here, and to support using an enum, I'd need the codegen to wrap the enum in an `InputValue` case. Something like `"episode": .enum(Episode.jedi)`. I haven't decided the exact format for that yet, but since the old codegen already does it this way, I don't have to worry about that just yet.

Apollo.xcodeproj/xcshareddata/xcschemes/StarWarsAPI.xcscheme
AnthonyMDev: Fixed this and removed other schemas that shouldn't be there.

Tests/ApolloTests/CacheKeyForFieldTests.swift
designatednerd: I'm slightly missing where we're doing the conversion to a string if the enum value is passed in at runtime - the query you added specifies it has to be a particular episode, but what about something like `CreateReviewForEpisodeMutation` where the `variables` is using a `GraphQLMap` with the enum value?

AnthonyMDev: Great question. In `InputValues` we have this case for evaluating variables. 

```swift
case .variable(let name):
      guard let value = variables?[name] else {
        throw GraphQLError("Variable \(name) was not provided.")
      }
      return value.jsonValue
```

The `GraphQLMap` is converted to it's `jsonValue` in that last line. The `jsonValue` is using the `Dictionary` `jsonValue` extension in `JSONStandardTypeConversions` to recursively convert all of its values!

designatednerd: Goooot it, this was happening in `extension GraphQLVariable: GraphQLInputValue {` before. Thank you!

designatednerd (approved):
:shipit:

PR #1740, opened 2021-04-01
Fixed a data race that occasionally causes a crash

STomperi:
if webSocketTransport.closeConnection() and webSocketTransport.initServer() are called in succession. Fix was verified with Xcode Thread sanitizer

### DIFF
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -228,7 +228,7 @@ public class WebSocketTransport {
       self.acked = false
 
       if let str = OperationMessage(payload: self.connectingPayload, type: .connectionInit).rawMessage {
-        write(str, force:true)
+        self.write(str, force:true)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


letrongtriet (approved):

designatednerd:
@STomperi thank you for adding the test! Would you mind checking if making `acked` use `Atomic<Bool>` also fixes the crash? I have some concerns about firing this off to a background queue if we don't have to.

designatednerd:
@STomperi Also, please sign the CLA when you get a chance - I will not be able to merge this without that.

STomperi:
@designatednerd Sorry for the delay. I tested with Atomic<Bool> for acked. It fixes the data race for acked, but I still get data race in queue.keys.max() without firing to background queue:

Screen Shot 2021-04-08 at 15 45 38

designatednerd:
Blech, okay. Please sign the CLA, otherwise I can't merge this. Thank you!

designatednerd:
heads up that i've rebased to fix the merge conflict and am going to force-push to your branch so this becomes mergeable

PR #1761, opened 2021-04-20
Move schema downloader tests to integration test target

AnthonyMDev:
As title. This just moves tests that are dependent upon the mock server being spun up to the integration test target.


Tests/ApolloServerIntegrationTests/StarWarsApolloSchemaDownloaderTests.swift
```diff
@@ -0,0 +1,81 @@
...
+//  Created by Anthony Miller on 4/20/21.
+//  Copyright © 2021 Apollo GraphQL. All rights reserved.
+//
+
+#if os(macOS)
+import XCTest
+import ApolloTestSupport
+@testable import ApolloCodegenLib
+
+class StarWarsApolloSchemaDownloaderTests: XCTestCase {
```
designatednerd, line 14: I think this could be `ApolloSchemaDownloaderIntegrationTests` since there's nothing that's particularly required about the schema being downloaded being the SW schema, but not a hill I wanna die on.

designatednerd (approved):

Tests/ApolloServerIntegrationTests/StarWarsApolloSchemaDownloaderTests.swift
AnthonyMDev: Oh, yeah, just all the other tests that depend on the SW server currently have StarWars at the start of the name.

PR #1762, opened 2021-04-21
Fix for exponential WebSocket reconnections

### LINKED ISSUE #1753: Exponential web socket reconnections leads to exhaustion of file descriptors, device heat, and application crash
## Bug report

`WebSocketTransport` attempts at reconnecting whenever Starscream outputs an error:

- https://github.com/apollographql/apollo-ios/blob/0.42.0/Sources/ApolloWebSocket/WebSocketTransport.swift#L408-L420
- https://github.com/apollographql/apollo-ios/blob/0.42.0/Sources/ApolloWebSocket/WebSocketTransport.swift#L448-L453

On iOS 14.2.2, a device on airplane mode has the underlying Foundation web socket stream output 2 errors NSPOSIXErrorDomain code 50 "Network is down", which are both notified by Starscream. This has Apollo attempts 2 reconnections. Which have Starscream create two new sockets, which generate two errors each, and now we have four reconnections. Then 8, 16, etc. With the default reconnection interval of 0.5, this goes pretty fast. Eventually file descriptors are exhausted, device turns hot, and application crashes.

`WebSocketTransport` is not making sure the number of reconnections is kept under control.

## Versions

- `apollo-ios` SDK version: 0.42.0
- Xcode version: 12.4
- Swift version: 5.3
- Package manager: Xcode's

## Steps to reproduce

```swift
let cache = InMemoryNormalizedCache()
let client = URLSessionClient()
let provider = NetworkInterceptorProvider(store: store, client: client)
let requestChainTransport = RequestChainNetworkTransport(interceptorProvider: provider, endpointURL: httpURL)
let webSocketTransport = WebSocketTransport(request: URLRequest(url: webSocketURL), connectingPayload: [:])
let splitNetworkTransport = SplitNetworkTransport(
    uploadingNetworkTransport: requestChainTransport,
    webSocketNetworkTransport: webSocketTransport
)
let apollo = ApolloClient(networkTransport: splitNetworkTransport, store: store)
```

And then turn airplane mode on.

It's delicate to setup a demo project, since my server is private.

groue:
This pull request fixes #1753.

When `WebSocketTransport` gets two errors in a row from Starscream (https://github.com/daltoniam/Starscream/issues/881), it no longer performs two reconnections.

### DIFF
--- a/Sources/ApolloCore/Atomic.swift
+++ b/Sources/ApolloCore/Atomic.swift
@@ -22,10 +22,11 @@ public class Atomic<T> {
   
   /// Mutates the underlying value within a lock.
   /// - Parameter block: The block to execute to mutate the value.
-  public func mutate(block: (inout T) -> Void) {
+  public func mutate<U>(block: (inout T) -> U) -> U {
     lock.lock()
-    block(&_value)
+    let result = block(&_value)
     lock.unlock()
+    return result
   }
 }
 
--- a/Sources/ApolloCore/Atomic.swift
+++ b/Sources/ApolloCore/Atomic.swift
@@ -22,6 +22,7 @@ public class Atomic<T> {
   
   /// Mutates the underlying value within a lock.
   /// - Parameter block: The block to execute to mutate the value.
+  /// - Returns: The value returned by the block.
   public func mutate<U>(block: (inout T) -> U) -> U {
     lock.lock()
     let result = block(&_value)
--- a/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -40,7 +40,16 @@ public class WebSocketTransport {
   private final let protocols = ["graphql-ws"]
   
   /// non-private for testing - you should not use this directly
-  var isSocketConnected = Atomic<Bool>(false)
+  enum SocketConnectionState {
+    case disconnected
+    case connected
+    case failed
+    
+    var isConnected: Bool {
+      self == .connected
+    }
+  }
+  var socketConnectionState = Atomic<SocketConnectionState>(.disconnected)
 
   private var acked = false
 
@@ -121,7 +130,7 @@ public class WebSocketTransport {
   }
 
   public func isConnected() -> Bool {
-    return self.isSocketConnected.value
+    return self.socketConnectionState.value.isConnected
   }
 
   public func ping(data: Data, completionHandler: (() -> Void)? = nil) {
@@ -240,7 +249,7 @@ public class WebSocketTransport {
   private func write(_ str: String,
                      force forced: Bool = false,
                      id: Int? = nil) {
-    if self.isSocketConnected.value && (acked || forced) {
+    if self.socketConnectionState.value.isConnected && (acked || forced) {
       websocket.write(string: str)
     } else {
       // using sequence number to make sure that the queue is processed correctly
@@ -383,7 +392,7 @@ extension WebSocketTransport: WebSocketDelegate {
       case .connected:
         self.handleConnection()
       case .disconnected(let reason, let code):
-        self.isSocketConnected.mutate { $0 = false }
+        self.socketConnectionState.mutate { $0 = .disconnected }
         self.error.mutate { $0 = nil }
         debugPrint("websocket is disconnected: \(reason) with code: \(code)")
         self.handleDisconnection()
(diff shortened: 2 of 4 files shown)


groue:
Hello @designatednerd,

I do not understand the "This branch has conflicts that must be resolved" : my local git happily merges or rebase the main repo's `main` and this branch. I did not merge or rebase, because I prefer using the latest stable version 0.42 + my fix, rather than the raw `main` (I don't know how stable it is).

Please tell me how you'd like this PR to move forward.

Thank you!

designatednerd:
Main should be stable, I'd recommend merging or rebasing off that. I will get a look at this this afternoon when I get home from 💉.

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -417,13 +431,22 @@ extension WebSocketTransport: WebSocketDelegate {
...
         
-        self.handleDisconnection()
+        switch previousState {
+        case .connected, .disconnected:
+          self.handleDisconnection()
+        case .failed:
+          // Don't attempt at reconnecting if already failed.
+          // Websockets will sometimes notify several errors in a row, and
+          // we don't want to perform disconnection handling multiple times.
+          // This avoids https://github.com/apollographql/apollo-ios/issues/1753
```
designatednerd, line 441: 👍 great comment

Sources/ApolloCore/Atomic.swift
```diff
@@ -22,10 +22,11 @@ public class Atomic<T> {
   
   /// Mutates the underlying value within a lock.
   /// - Parameter block: The block to execute to mutate the value.
-  public func mutate(block: (inout T) -> Void) {
+  public func mutate<U>(block: (inout T) -> U) -> U {
```
designatednerd, line 25: worth making this a `@discardableResult` so if we don't need it, we don't have to explicitly discard it? 

Also can you please update the doc comment to reflect what the return value is? Thank you!

designatednerd (commented):
mostly looks good, one question!

Sources/ApolloCore/Atomic.swift
groue: I did update the documentation.

I would advise against `@discardableResult` unless you really want it. By not making the result discardable, this method behaves the same as standard methods such as [DispatchQueue.sync(_:)](https://developer.apple.com/documentation/dispatch/dispatchqueue/2016081-sync), [autoreleasepool(invoking:)](https://developer.apple.com/documentation/objectivec/2299644-autoreleasepool/), and many others.

Not using `@discardableResult` makes sure Apollo developers get a warning when they write, for example:

```swift
// Expectable compiler warning: the value is unused.
myAtomic.mutate {
    $0 = true
    return <some value>
}
```

Not using `@discardableResult` does not create any friction when `U` is `Void`: the Swift compiler is smart enough to not emit any warning:

```swift
// No warning
myAtomic.mutate { $0 = true }
```

Sometimes, as in 1f355c91ec500f99d1a4ddcdddcfe75c4a82f62c, one will have to insert a `_` in order to silence a warning. One would do the same in the context of `Dispatch.sync` and other similar methods. So not using `@discardableResult` does not add any inconvenience on top of the standard ones.

groue:
> Re the conflict: Looks like that file was moved to `Tests/ApolloTests/WebSocket` during a recent reorganization of the test suite.

I rebased the pull request on top of `main`.

Sources/ApolloCore/Atomic.swift
designatednerd: > Not using @discardableResult does not create any friction when the `U` generic type is `Void`

I think this alleviates most of my concerns since this covers a pretty solid majority of cases.

designatednerd (approved):
I'll add some tests to this before shipping but thank you for the PR!

groue:
You're welcome, @designatednerd! Maybe think about hiring someone who wants to learn about web sockets, since this is a pretty important facet of the Apollo product!

hwillson:
> Maybe think about hiring someone who wants to learn about web sockets, since this is a pretty important facet of the Apollo product!

Thanks for the PR @groue, but please note that making comments like this to the maintainers of OSS projects, just isn't great. We welcome all valid project criticisms and really appreciate the time you and others are willing to put into Apollo iOS to help the project grow, but derogatory comments like this can be hurtful. Just a heads up anyways - you probably didn't mean for this comment to come across the way it did, but we want to ensure all communication on Apollo issue threads is healthy and respectful.

groue:
Be assured on my profound respect, @hwillson. For more context, I let you read the full https://github.com/apollographql/apollo-ios/issues/1753  conversation: you will get a better measure of my involvement, and maybe better understand the origin of the critic. Have a nice week-end!

PR #1763, opened 2021-04-23
Update docs for class to AnyObject change

cltnschlosser:
Docs update for https://github.com/apollographql/apollo-ios/pull/1733


designatednerd:
Hey Colton! Thanks for doing this - these are actually auto-regenerated when we do a new release, so I'd prefer to let our generation tool do that. Is that cool?

cltnschlosser:
Yeah, I’ll close this!

PR #1767, opened 2021-04-28
chore: fixed a typo

TizianoCoroneo:
Hello 👋 

I've found a typo in the CodegenLib: `singluarize` instead of `singularize`.


designatednerd (approved):
Me fail english? That's unpossible!

(Thank you for catching this)

PR #1769, opened 2021-04-29
Fix swift warning when using cocoapods

cltnschlosser:
`File 'GraphQLGETTransformer.swift' is part of module 'Apollo'; ignoring import`


designatednerd (approved):
Thank you!

PR #1790, opened 2021-05-13
Document improvement for README about replace .xcworkspace to .xcodeproj

bannzai:
## What
Replace file extension from `.xcworkspace` to `.xcodeproj` for README.

## Why
Looking into the [past](https://github.com/apollographql/apollo-ios/pull/1029), xcodeproj seems to be the correct.

## Motivation
I was reading the docs hoping to contribute to amazing product of apollo-ios and I found a question for project file extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ This combination of models with value semantics, one way data flow, and automati
 
 This project is being developed using Xcode 12 and Swift 5.3.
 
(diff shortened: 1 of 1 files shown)


bannzai:
I was already agree 👍

designatednerd (approved):

PR #1802, opened 2021-05-20
Allow different types of fetches on a watcher

gsbernstein:
Publicly exposing `GraphQLQueryWatcher`'s fetch() function would allow us to make other kinds of fetches besides the `fetchIgnoringCacheData` exposed by `refetch()`. Our use case is that our custom version of `InMemoryNormalizedCache` has cache expiration, and we want to be able to hit the network only if the cache has become invalid. Alternately, we could manually check the cache, but this seems cleaner.

### DIFF
--- a/Sources/Apollo/GraphQLQueryWatcher.swift
+++ b/Sources/Apollo/GraphQLQueryWatcher.swift
@@ -43,7 +43,7 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
     fetch(cachePolicy: .fetchIgnoringCacheData)
   }
 
-  public func fetch(cachePolicy: CachePolicy) {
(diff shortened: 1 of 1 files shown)


AnthonyMDev:
Thanks for the PR and explanation @gsbernstein. I think it might be cleaner to expose the cachePolicy as a parameter on `refetch()`, since the initial fetch is done for you automatically. So I'd change it to `public func refetch(cachePolicy: CachePolicy = .fetchIgnoringCacheData)`

gsbernstein:
In that case, should I just combine the fetch and refetch functions into a single `refetch(cachePolicy: CachePolicy = .fetchIgnoringCacheData)`?

AnthonyMDev:
Let's leave the private fetch function as is for now and just change the `refetch` function please!

AnthonyMDev:
Thanks for finishing this up!

PR #1817, opened 2021-06-03
Foundational Work For Codegen Rewrite

AnthonyMDev:
There is a lot in this PR, so bear with me. This makes a few changes to start enabling the work on the Codegen rewrite.

- Renames `ApolloCore` to `ApolloUtils`
    - This library will be imported by `Apollo` and `ApolloCodegenLib`. It provides common utility functionality that can be helpful for any of our code.
    - In the future, more of the files in the `Utilities` folder of `Apollo` will likely be moved here.
- Adds the new `ApolloModels` library.
    - This library will be imported by `Apollo` and by any generated models that are generated using the new codegen.
    - Contains protocols, model structs, and enums that enable generated code to access data concisely.
    - **I'm not sure if I'm happy with the name of this library currently. Looking for suggestions.**
- Includes these 2 new libraries for use with SPM and Cocoapods
    - **Need to ensure this works with Carthage currently. (I'll need your help there @designatednerd)**
    - Cocoapods and SPM will exclude the files in the `ApolloModels/CodegenV1` folder. These are needed for future development work, but don't need to be included in distributed library until version 1.0.0 with the new Codegen.
- Renamed `Reference` to `CacheReference`
    - The name was too vague and confused me.
    - **Open to other suggestions here. Maybe `EntityReference`?**
- `InputValue` refactored
    - No longer uses `JSONEncodable` directly.
        - I wanted `JSONEncodable` to be in `Apollo`, but `InputValue` needs to be in `ApolloModels` So the evaluation functions were moved into an extension that is in `Apollo`. 
        - This evaluation now checks if scalar values conform to `JSONEncodable` and if they don't (they always should though), it throws an error.
- **Other Minor Improvements/Clean Up**
    - Removed custom Dictionary initializer helper function in favor of using standard library function `Dictionary(_ elements: uniquingKeysWith:)`
    - `JSONValueMatcher` to wrap the `equals(_: Any, _: Any) -> Bool` function. You now need to explicitly call `JSONValueMatcher.equals(x, y)` and we get that function out of the global namespace.
    - `Matchable` moved into `ApolloTestSupport`
        - It is only being used for `JSONDecodingError` and only for test matchers. Moved the `extension JSONDecodingError: Matchable` into `ApolloTestSupport` as well.
    - Replaced some unneeded code for transforming dictionaries during execution with use of `compactMapValues`.
    - Improved `OperationMessage` by making the `message` property a `let` and computing it once in the initializer.


designatednerd:
Looks like there's a bunch of compilation errors on the codegen tests, I think you can probably find those pretty quick if you compile locally. 

I'll have to take a look at Carthage stuff separately, but as long as it compiles locally since carthage uses the underlying xcode project to compile, theoretically it should work correctly.

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
```diff
@@ -0,0 +1,3 @@
+#include "../Shared/Workspace-Universal-Framework.xcconfig"
+
+INFOPLIST_FILE = Sources/ApolloModels/Info.plist
```
martinbonnin, line 3: My 2 cents: Apollo Android uses `apollo-api` as a name for lib that the generated models uses. I contemplated making this `apollo-runtime`. This is what a lot of other libraries do (antlr, wire, sqldelight, etc...) but there's so much brain muscle around `apollo-api` that I didn't do it eventually.
All in all, I'd personally expect `ApolloModels` to contain models and not "support code for the models". But again, different platform, etc..

Sources/Apollo/CacheReference.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+/// A reference to a cache record.
+public struct CacheReference {
```
martinbonnin, line 4: Apollo Android uses `CacheReference` too. I'd be in for renaming to `EntityReference` as we're still at a moment where we can rename stuff on Android. Feels like that would be more adequate. It's a reference to an Entity, not to a Cache. Or maybe `ObjectReference` ? (We don't have entity anywhere in the Android codebase)

Package.swift
```diff
@@ -6,94 +6,107 @@ import PackageDescription
...
+                "ApolloModels",
+                "ApolloUtils"
+            ],
+            exclude: [
+                "Info.plist"
+            ]),
+        .target(
+            name: "ApolloModels",
+            dependencies: [
+                "ApolloUtils"
```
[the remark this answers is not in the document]
AnthonyMDev: Yeah, I messed that up, it doesn't need to. Will fix it. Though if it needs to in the future, that's not a huge problem. Generated models just don't want to be pulling in all of `Apollo`

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
AnthonyMDev: I like `ApolloRuntime`!

Sources/Apollo/CacheReference.swift
AnthonyMDev: @designatednerd What's your opinion on the naming of this? Entity, Object, something else?

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
designatednerd: i'm not a huge fan of `ApolloRuntime` since I think of the things like networking and dealing with the cache as the runtime, and that's precisely what we're trying **not** to expose here. 

I don't actually have a better suggestion than `ApolloModels`, which has its own problems since not everything in it is precisely a model. 

`ApolloThingsUsedByOurGeneratedCode` seems a hair too long.

Sources/Apollo/CacheReference.swift
designatednerd: @martinbonnin Is your point that it's a reference to an entity _within_ the cache rather than to the cache itself? I think that's valid, but `Entity` by itself is so overloaded I might go with something like `CacheEntityReference`.

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
martinbonnin: > I think of the things like networking and dealing with the cache as the runtime

Fair. This is actually the current state of the Android repo. We have `apollo-api` for supporting classes and `apollo-runtime` for networking. 

What about `apollo-client` for networking code etc though? Looks like this is what the web is doing and also that would match the main entry point of the lib?

Sources/Apollo/CacheReference.swift
martinbonnin: > Is your point that it's a reference to an entity within the cache rather than to the cache itself?

Yes it was! Rereading that, I'm certainly overthinking this. `CacheReference` sounds reasonable. 
And now back to overthinking this... What's the difference between a `CacheReference` and a `CacheKey` ? Could we merge both somehow?

designatednerd: My recollection is that the key is the key used to access stored data vs the reference is the actual stored data - @AnthonyMDev is that about right?

martinbonnin: I got curious and tried to merge `CacheKey` and `CacheReference`. Turns out it's working well: https://github.com/apollographql/apollo-android/pull/3157

The only reason I can think of having two different data structures in Kotlin is if we want to make `CacheKey` a value class that is inlined to a `String` for performance reasons. If we do that, we can't check the type of `CacheKey` anymore which means everything becomes a string and we can't serialize the CacheKey to their special String value. 

But besides that I actually like the reduced mental load and API surface.

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
AnthonyMDev: I really like "runtime", but I understand @designatednerd's point here. I'm good with `ApolloAPI` also or `ApolloModelRuntime`. I spent a lot of time this weekend thinking about it, and couldn't come up with anything better. So I just want to make a decision and move on with it at this point honestly. haha

Other ideas I had:

```
ApolloModelEngine
ApolloModelSupport
ApolloAPISupport
ApolloOperationSupport
ApolloGQLOperationSupport
ApolloGQLOperationEngine
ApolloOperationModels
ApolloResponseData
```

Are any of those more appealing to you two?

Sources/Apollo/CacheReference.swift
AnthonyMDev: Hmmmmm. I'm going to look into this. Thanks for that @martijnwalraven 

> My recollection is that the key is the key used to access stored data vs the reference is the actual stored data - @AnthonyMDev is that about right?

No, the reference isn't the actual stored data. It's just a wrapper type for the CacheKey as a `String`. It indicates that the raw String data isn't just a regular raw a String property on the object, but is a `String` representing the reference to the cache key of another object.

designatednerd: Gooooot it, thank you for clarifying

Configuration/Apollo/Apollo-Target-ApolloModels.xcconfig
designatednerd: `ApolloAPISupport` or `ApolloModelSupport` both seem flexible but close enough to what we're looking for in my book

martinbonnin: I'd lean towards `ApolloApi` to match Android but really not sure how that'd sound on iOS so both `ApolloAPISupport` or `ApolloModelSupport` work for me!

PR #1830, opened 2021-06-14
Fix memory leaks due to xctest

AnthonyMDev:
See this article for explanation: https://qualitycoding.org/xctestcase-teardown/

### DIFF
--- a/Sources/ApolloTestSupport/XCTAssertHelpers.swift
+++ b/Sources/ApolloTestSupport/XCTAssertHelpers.swift
@@ -93,7 +93,7 @@ public func XCTAssertTrueEventually(_ test: @autoclosure () -> Bool, timeout: Ti
     }
     runLoop.run(until: Date(timeIntervalSinceNow: 0.01))
   } while Date().compare(timeoutDate) == .orderedAscending
-  // 4
+  
   XCTFail(message, file: file, line: line)
 }
 
--- a/Sources/ApolloTestSupport/XCTAssertHelpers.swift
+++ b/Sources/ApolloTestSupport/XCTAssertHelpers.swift
@@ -75,6 +75,41 @@ public func XCTAssertFailureResult<Success>(_ expression: @autoclosure () throws
   }
 }
 
+/// Checks that the condition is eventually true with a given timeout (default 1 second).
+///
+/// This assertion runs the run loop for 0.01 second after each time it checks the condition until
+/// the condition is true or the timeout is reached.
+///
+/// - Parameters:
+///   - test: An autoclosure for the condition to test for truthiness.
+///   - timeout: The timeout, at which point the test will fail. Defaults to 1 second.
+///   - message: A message to send on failure.
+public func XCTAssertTrueEventually(_ test: @autoclosure () -> Bool, timeout: TimeInterval = 1.0, message: String = "", file: StaticString = #file, line: UInt = #line) {
+  let runLoop = RunLoop.current
+  let timeoutDate = Date(timeIntervalSinceNow: timeout)
+  repeat {
+    if test() {
+      return
+    }
+    runLoop.run(until: Date(timeIntervalSinceNow: 0.01))
+  } while Date().compare(timeoutDate) == .orderedAscending
+  // 4
+  XCTFail(message, file: file, line: line)
+}
+
+/// Checks that the condition is eventually false with a given timeout (default 1 second).
+///
+/// This assertion runs the run loop for 0.01 second after each time it checks the condition until
+/// the condition is false or the timeout is reached.
+///
+/// - Parameters:
+///   - test: An autoclosure for the condition to test for falsiness.
+///   - timeout: The timeout, at which point the test will fail. Defaults to 1 second.
+///   - message: A message to send on failure.
+public func XCTAssertFalseEventually(_ test: @autoclosure () -> Bool, timeout: TimeInterval = 1.0, message: String = "", file: StaticString = #file, line: UInt = #line) {
+  XCTAssertTrueEventually(!test(), timeout: timeout, message: message, file: file, line: line)
+}
+
 /// Downcast an expression to a specified type.
 ///
 /// Generates a failure when the downcast doesn't succeed.
--- a/Tests/ApolloCodegenTests/ASTParsingTests.swift
+++ b/Tests/ApolloCodegenTests/ASTParsingTests.swift
@@ -12,7 +12,7 @@ import ApolloCodegenTestSupport
 
 class ASTParsingTests: XCTestCase {
   
-  lazy var starWarsJSONURL: URL = {
+  static let starWarsJSONURL: URL = {
     let sourceRoot = CodegenTestHelper.sourceRootURL()
     let starWarsJSONURL = sourceRoot
       .appendingPathComponent("Sources")
@@ -23,7 +23,6 @@ class ASTParsingTests: XCTestCase {
     return starWarsJSONURL
   }()
 
-  
   private func loadAST(from url: URL,
                        file: StaticString = #file,
                        line: UInt = #line) throws -> ASTOutput {
@@ -32,7 +31,7 @@ class ASTParsingTests: XCTestCase {
   
   func testLoadingStarWarsJSON() throws {
     do {
-      let output = try loadAST(from: starWarsJSONURL)
+      let output = try loadAST(from: Self.starWarsJSONURL)
       XCTAssertEqual(output.operations.count, 38)
       XCTAssertEqual(output.fragments.count, 15)
       XCTAssertEqual(output.typesUsed.count, 3)
@@ -46,7 +45,7 @@ class ASTParsingTests: XCTestCase {
   func testParsingASTUnionTypes() throws {
     let output: ASTOutput
     do {
-      output = try loadAST(from: starWarsJSONURL)
+      output = try loadAST(from: Self.starWarsJSONURL)
     } catch {
       CodegenTestHelper.handleFileLoadError(error)
       return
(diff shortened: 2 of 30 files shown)


Tests/ApolloServerIntegrationTests/URLSessionClientTests.swift
```diff
@@ -4,7 +4,19 @@ import ApolloUtils
...
+    super.setUp()
+
+    client = URLSessionClient()
+  }
+
+  override func tearDown() {
+    client = nil
+
+    super.tearDown()
+  }
```
designatednerd, line 19: is this defensive or is there an actual problem here? I ask because I've never seen this be an issue when a `let` of `URLSessionClient`, and this would create a new instance per test run rather than reusing an existing variable.

Tests/ApolloServerIntegrationTests/UploadTests.swift
```diff
@@ -6,21 +6,33 @@ import StarWarsAPI
...
 
-  lazy var client: ApolloClient = {
-    let store = ApolloStore()
-    let provider = LegacyInterceptorProvider(store: store)
-    let transport = RequestChainNetworkTransport(interceptorProvider: provider,
-                                                 endpointURL: self.uploadClientURL,
-                                                 additionalHeaders: ["headerKey": "headerValue"])
-    transport.clientName = "test"
-    transport.clientVersion = "test"
+  var client: ApolloClient!
```
designatednerd, line 11: Same question here re `client`.

Tests/ApolloServerIntegrationTests/UploadTests.swift
```diff
@@ -6,21 +6,33 @@ import StarWarsAPI
 
 class UploadTests: XCTestCase {
 
-  let uploadClientURL = TestServerURL.uploadServer.url
+  static let uploadClientURL = TestServerURL.uploadServer.url
```
designatednerd, line 9: I can at least see the rationale for this change a bit better, since it's utterly unchanging throughout the test suite, but I do think this may be a premature optimization since the instance is created and destroyed with every test since it's a `let`, so it's tied to the lifetime of the test.

Tests/ApolloTests/Cache/LoadQueryFromStoreTests.swift
```diff
@@ -10,9 +10,9 @@ class LoadQueryFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
   var cacheType: TestCacheProvider.Type {
     InMemoryTestCacheProvider.self
   }
-  
-  var defaultWaitTimeout: TimeInterval = 5
-  
+
+  static var defaultWaitTimeout: TimeInterval = 5.0
```
designatednerd, line 14: should this be a `let` since it's not actually changing here?

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -9,7 +9,7 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
     InMemoryTestCacheProvider.self
   }
   
-  var defaultWaitTimeout: TimeInterval = 5
+  static var defaultWaitTimeout: TimeInterval = 5.0
```
designatednerd, line 12: same question on `static var` v `static let`

Sources/ApolloTestSupport/XCTAssertHelpers.swift
```diff
@@ -75,6 +75,41 @@ public func XCTAssertFailureResult<Success>(_ expression: @autoclosure () throws
...
+public func XCTAssertTrueEventually(_ test: @autoclosure () -> Bool, timeout: TimeInterval = 1.0, message: String = "", file: StaticString = #file, line: UInt = #line) {
+  let runLoop = RunLoop.current
+  let timeoutDate = Date(timeIntervalSinceNow: timeout)
+  repeat {
+    if test() {
+      return
+    }
+    runLoop.run(until: Date(timeIntervalSinceNow: 0.01))
+  } while Date().compare(timeoutDate) == .orderedAscending
+  // 4
```
designatednerd, line 96: what's this comment in reference to?

Tests/ApolloServerIntegrationTests/LegacyInterceptorProviderIntegrationTests.swift
```diff
@@ -24,6 +24,8 @@ class LegacyInterceptorProviderIntegrationTests: XCTestCase {
 
   override func tearDown() {
     legacyClient = nil
+    
+    super.tearDown()
```
designatednerd, line 28: 🤦‍♀️ good catch

Tests/ApolloServerIntegrationTests/StarWarsServerTests.swift
```diff
@@ -42,7 +42,7 @@ class StarWarsServerTests: XCTestCase, CacheDependentTesting {
     InMemoryTestCacheProvider.self
   }
   
-  var defaultWaitTimeout: TimeInterval = 5
+  static var defaultWaitTimeout: TimeInterval = 5
```
designatednerd, line 45: Should this be a `static let`?

Tests/ApolloTests/Cache/WatchQueryTests.swift
```diff
@@ -9,7 +9,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
     InMemoryTestCacheProvider.self
   }
   
-  var defaultWaitTimeout: TimeInterval = 1
+  static var defaultWaitTimeout: TimeInterval = 1
```
designatednerd, line 12: again with the `static let`

designatednerd (commented):
I like the idea of auditing this in general but I do worry here there's a bunch of things we're changing because we think MAYBE they might be causing a problem rather than us having any evidence that there's a problem. I'm definitely good with all the setting vars to `nil` in `tearDown` methods, but there's some stuff I'm wondering if we're doing some premature optimization on.

Tests/ApolloCodegenTests/ASTParsingTests.swift
```diff
@@ -12,7 +12,7 @@ import ApolloCodegenTestSupport
 
 class ASTParsingTests: XCTestCase {
   
-  lazy var starWarsJSONURL: URL = {
+  static let starWarsJSONURL: URL = {
```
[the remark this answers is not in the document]
AnthonyMDev: It was leaking. It's small, but it leaks. That article I sent you sums up why. https://qualitycoding.org/xctestcase-teardown/

Happy to explain it on a quick call if you're curious!

Sources/ApolloTestSupport/XCTAssertHelpers.swift
AnthonyMDev: Oops will fix that.

Tests/ApolloServerIntegrationTests/StarWarsServerTests.swift
AnthonyMDev: Yupp good catch

Tests/ApolloServerIntegrationTests/URLSessionClientTests.swift
AnthonyMDev: Nope it's actually leaking! I'll call and explain why! (Each test is already creating a new instance currently!)

Tests/ApolloTests/Cache/WatchQueryTests.swift
AnthonyMDev: Fixing now thanks

Tests/ApolloServerIntegrationTests/URLSessionClientTests.swift
AnthonyMDev: If you wanted to reuse the same client you would need this to be `static let`, just `let` won't do it because each test initializes it's own instance of the `XCTestCase` subclass.

AnthonyMDev:
To summarize a lot of the changes @designatednerd questioned here:

It's not that important that a `URL` is `static`. Yes, each tests is going to create and retain it's own value, but they are small, so unless you had 10s of thousands of tests, it's not going to cause a problem. Regardless, this is technically more performant without (in my opinion) making anything more abstract or hard to understand.

It's always best practices for each tests to be completely isolated. You shouldn't be using shared resources unless they are pure value types that are static, constant, and immutable. Setting up separate `URLSessionClient` objects for each individual test (in the `setUp()` method) is the right way to go. Even though right now it doesn't cause an issue, it's ensuring that no future changes accidentally cause bleed over from one test to another.

I've spent way too many hours debugging way too many very difficult test isolation issues caused by not following these best practices when people thought they wouldn't matter and then a change 6 months down the line caused isolation failures. So following these best practices is just good defensive testing in my mind.

Tests/ApolloCodegenTests/ASTParsingTests.swift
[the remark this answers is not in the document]
designatednerd: I'm fine with this being defensive, appreciate the explanation!

designatednerd (approved):
Fair enough! LGTM

calvincestari (approved):

calvincestari:
It's certainly cleaned up many of the `xctest[39775:3136288] [connection] nw_socket_handle_socket_event [C17.1:1] Socket SO_ERROR [61: Connection refused]` errors, save for `[ApolloTests.StoreConcurrencyTests testConcurrentReadsInitiatedFromBackgroundThreads]` which might need another audit later.

Is there any other 'pass' criteria we should be validating?

AnthonyMDev:
Yeah, I need to see if that one is okay or not. If there is a tests that is testing a failed socket, then it's fine to have that message. As long as it's happening during the test and not after. I plan on looking at that soon. It's on my radar at least.

Issue #1829, opened 2021-06-14
Main branch ScalarType and NSNull not found


willchpar:
## Bug report

With the commit mentioned below on the main branch, InputValue cannot find ScalarType.

We temporarily mitigated this problem by moving ScalarTypes.swift one level up from /Sources/ApolloAPI/CodegenV1.

Then we ran into another problem of "NSNull not found" in GraphQLResponseGenerator.swift, GraphQLResultNormalizer.swift and InputValue+Evaluation.swift, which we mitigated by importing Foundation.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version:  main branch commit 4e2a4950bcd9b0c1477f3b5e39e774e6c2bbc562
- Xcode version: 12.5
- Swift version: 5
- Package manager: SPM

## Steps to reproduce

Checking out commit 4e2a4950bcd9b0c1477f3b5e39e774e6c2bbc562 of the main branch will fail on build

designatednerd:
That's quite odd, given that when I look at that merge commit, I can see it passed on all our continuous integration checks. 

Have you tried [doing the derived data dance](deriveddata.dance) and rebuilding?

willchpar:
I started another Xcode project that integrates the main branch, and it still exhibits the same problem. I would think deleting derived data here wouldn't affect anything, since it's a new project.

willchpar:
It should be pretty easy to reproduce. You can create a new Xcode project, add Apollo as a dependency and then set it to use main as below:

![Screen Shot 2021-06-14 at 2 56 15 PM](https://user-images.githubusercontent.com/81775704/121951557-ae309880-cd20-11eb-88f6-412b575ee875.png)

AnthonyMDev:
> That's quite odd, given that when I look at that merge commit, I can see it passed on all our continuous integration checks.

@designatednerd We don't have any CI checks that are actually building the SPM package from the `Package.swift` file. So just because our Xcode project is working properly doesn't mean that we didn't mess up the Package configuration...

I want to improve that, but there is a lot of stuff that needs to change for that. (This is probably blocked while waiting on SPM  to support plugins for external build tools.)

designatednerd:
Oh woof, that's right. This is why I do the whole rigamarole checking against SPM/Carthage/CocoaPods builds before doing a new release. Will get this squared now though 👍

calvincestari:
[...]

PR #1831, opened 2021-06-14
Fix issues with imports and foldering

designatednerd:
Addresses #1829.


designatednerd:
I confirmed checking out an empty project against the `fix/imports` branch this has fixed imports when building with SPM.

AnthonyMDev (approved):
Thanks for the quick fix!

AnthonyMDev:
Are we going to add the CI job for `swift build` to this PR, or do that separately?

designatednerd:
Separately since this needs to be merged for the version we're working on to go out

PR #1832, opened 2021-06-14
Expose cache policy directive to refetch callers

calvincestari:
### What
This picks up from #1802 to expose the cache policy directive for `refetch` calls while keeping the internal `fetch` method private.

Closes #1802 

### Why
See https://github.com/apollographql/apollo-ios/pull/1802#issue-649544503.

### Tests
_None additional_ - the current `WatchQueryTests` test cases validate cache/request behaviour so defaulting to anything other than `fetchIgnoringCacheData` or `fetchIgnoringCacheCompletely` would fail the tests.


designatednerd (approved):

calvincestari:
I'm opting to close this PR and instead modify the original #1802 PR.
* partly me figuring out how to work with contributor forks
* I think it's a good open-source practice to keep the conversation/work in the original PR since it's better feedback to the originator

PR #1838, opened 2021-06-16
Update docs for new library names

designatednerd:
`ApolloCore` has been split into `ApolloUtils` and `ApolloAPI`. I've updated the docs to reflect this. 

I also added in a CI step to run `swift build` so we find out earlier on when we have something where imports are failing.


AnthonyMDev (approved):
LGTM! Thanks!

PR #1842, opened 2021-06-17
Fix Transient Test Failures

AnthonyMDev:
Just a couple more tests cleaned up.

### DIFF
--- a/Tests/ApolloTests/Cache/WatchQueryTests.swift
+++ b/Tests/ApolloTests/Cache/WatchQueryTests.swift
@@ -669,7 +669,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
       wait(for: [serverRequestExpectation, initialWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
     }
     
-    let numberOfFetches = 1000
+    let numberOfFetches = 10
     
     runActivity("Fetch same query concurrently \(numberOfFetches) times") { _ in
       let serverRequestExpectation = server.expect(HeroNameQuery.self) { request in
@@ -708,7 +708,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
         }
       }
       
-      wait(for: [serverRequestExpectation, otherFetchesCompletedExpectation, updatedWatcherResultExpectation], timeout: 10)
+      wait(for: [serverRequestExpectation, otherFetchesCompletedExpectation, updatedWatcherResultExpectation], timeout: 3)
       
       XCTAssertEqual(updatedWatcherResultExpectation.apollo.numberOfFulfillments, 1)
     }
@@ -749,7 +749,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
       wait(for: [serverRequestExpectation, initialWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
     }
     
-    let numberOfFetches = 1000
+    let numberOfFetches = 10
     
     runActivity("Fetch same query concurrently \(numberOfFetches) times") { _ in
       let serverRequestExpectation = server.expect(HeroNameQuery.self) { request in
@@ -790,7 +790,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
         }
       }
       
-      wait(for: [serverRequestExpectation, otherFetchesCompletedExpectation, updatedWatcherResultExpectation], timeout: 10)
+      wait(for: [serverRequestExpectation, otherFetchesCompletedExpectation, updatedWatcherResultExpectation], timeout: 3)
       
       XCTAssertEqual(updatedWatcherResultExpectation.apollo.numberOfFulfillments, numberOfFetches)
     }
--- a/Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
+++ b/Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
@@ -7,6 +7,17 @@ import Starscream
 class WebSocketTransportTests: XCTestCase {
 
   private var webSocketTransport: WebSocketTransport!
+  static var webSocketProvider: ApolloWebSocketClient.Type = MockWebSocket.self
+
+  override class func setUp() {
+    super.setUp()
+    swap(&webSocketProvider, &WebSocketTransport.provider)
+  }
+
+  override class func tearDown() {
+    super.tearDown()
+    swap(&webSocketProvider, &WebSocketTransport.provider)
+  }
 
   override func tearDown() {
     webSocketTransport = nil
@@ -26,8 +37,6 @@ class WebSocketTransportTests: XCTestCase {
   }
 
   func testUpdateConnectingPayload() {
-    WebSocketTransport.provider = MockWebSocket.self
-
     self.webSocketTransport = WebSocketTransport(request: URLRequest(url: TestURL.mockServer.url),
                                                  connectingPayload: ["Authorization": "OldToken"])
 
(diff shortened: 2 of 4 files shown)


Tests/ApolloTests/Cache/WatchQueryTests.swift
```diff
@@ -669,7 +669,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
       wait(for: [serverRequestExpectation, initialWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
     }
     
-    let numberOfFetches = 1000
+    let numberOfFetches = 10
```
designatednerd, line 672: Do we have reasonable confidence that only 10 concurrent fetches vs. say 100 (I get that 1000 causes timeouts) is enough to sufficiently push watchers?

Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
```diff
@@ -7,6 +7,17 @@ import Starscream
...
+  static var webSocketProvider: ApolloWebSocketClient.Type = MockWebSocket.self
+
+  override class func setUp() {
+    super.setUp()
+    swap(&webSocketProvider, &WebSocketTransport.provider)
+  }
+
+  override class func tearDown() {
+    super.tearDown()
+    swap(&webSocketProvider, &WebSocketTransport.provider)
```
designatednerd, line 19: I'm not super familiar with `swap` - what's happening here?

designatednerd (commented):
Mostly looks good, just a couple questions!

Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
calvincestari: From what I can understand of it - WebSocketTransport.provider is [defined as `ApolloWebSocket.self`](https://github.com/apollographql/apollo-ios/blob/fix-watchquery-tests/Sources/ApolloWebSocket/WebSocketTransport.swift#L30) but needs to be mocked out for successful control of the `testUpdateConnectingPayload` test. We can't get in to that class though so `swap` is used as a manual memory change to instead force a `MockWebSocket.self` object to be created [here](https://github.com/apollographql/apollo-ios/blob/fix-watchquery-tests/Sources/ApolloWebSocket/WebSocketTransport.swift#L116). Think of it kinda like Obj-C swizzling - in an abstract way.

Is that correct @AnthonyMDev?

calvincestari (approved):
I think this code is good to go based. There are two comments but none of them would prevent merging for me.

Tests/ApolloTests/Cache/WatchQueryTests.swift
designatednerd: Yeah i think the issue is the CI machines are heavily memory-constrained (sometimes more so than actual iOS devices!), so that may be part of what's causing the slowness.

AnthonyMDev: > This test is configured to use InMemoryNormalizedCache too which I'd expect to be quick.

The `SQLiteWatchQueryTests` actually subclass these tests and provide the SQLite cache. So these tests are run both against the InMemory and SQLite caches.

> Do we have reasonable confidence that only 10 concurrent fetches vs. say 100 (I get that 1000 causes timeouts) is enough to sufficiently push watchers?

There is no known race condition encountered in the past that this test is trying to prevent regressing. To me, this test should be testing behavior of the watcher completion block when multiple fetches hit the cache. This should be testing that reproducible behavior. Trying to get a unit test to check for a race condition when there was no known race condition to test seems futile. These could be happening concurrently or serially. Either way, they test that behavior.

The way you test a race condition is different. If some sort of race condition ever shows up, there is no way to have any confidence this test will catch it with 100, 1000, or 10,000 fetches. In the case of a known race condition, we should write a reproducing test; use it to figure out the source of the race condition; come up with a fix; then write a second test that tests the actual behavior written to prevent the race condition; put comments on both of those tests to indicate the race condition that they check for and link it to an issue so we know what to refer back to when we edit it in the future.

I am open to having my opinion changed, but I'm currently philosophically opposed to having "shotgun" tests that just do a bunch of things in the hopes that they may catch a race condition or a bug if they exist -- at least in the unit test target.

If we want to achieve that goal, I think the proper way to do that is to have stress tests that put the system under large load. These should:
  - Be integration tests, **not unit tests**.
  - Check for increased memory load (checking for memory leaks)
  - Check for increased time over iterations (checking for data/object explosion)
  - Be ran periodically on a cron job **not with unit tests**

These shouldn't be run with unit tests because they will inherently be long running, integration style tests, with failures that will likely be transient. That is not conducive to the development process. I run unit tests locally _constantly_ as I am developing (Test Driven Development requires it). Unit tests also hold up PRs from being merged. So slow tests, especially ones with transient failures, cause a huge problem for the development flow. 

I am not against testing the things that you both are concerned about here, but we should be writing a much more robust set of integration/performance tests. The unit test suite is not the appropriate place for these in my opinion.

AnthonyMDev: If you would like me to try re-writing this test to use a serial queue and wait upon each individual fetch so we can make it more clear that this is not a stress tests, I'll do that. 

I was just trying to quickly get our transient test failures to stop. And doing this concurrently ensures this tests still tests the behavior, while preventing of from being a stress test that overloads our CI machines.

Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
AnthonyMDev: Swap just swaps the object at pointer A with the object at Pointer B.

In a few of the tests here we were setting the `WebSocketTransport.provider` to a mock. That property is a `static var`, which means that once you set it, you've changed global state. Unit tests should be isolated and any changes to global state need to be reverted after the tests are done. That wasn't happening.

By swapping them, we store whatever was set in the property before this test suite, and the revert it when the `WebSocketTransportTests` are complete.

Tests/ApolloTests/Cache/WatchQueryTests.swift
calvincestari: > I am open to having my opinion changed, but I'm currently philosophically opposed to having "shotgun" tests that just do a bunch of things in the hopes that they may catch a race condition or a bug if they exist -- at least in the unit test target.

You don't have to convince me of this, I agree with the philosophy.

> If you would like me to try re-writing this test to use a serial queue and wait upon each individual fetch so we can make it more clear that this is not a stress tests, I'll do that.
> 
> I was just trying to quickly get our transient test failures to stop. And doing this concurrently ensures this tests still tests the behavior, while preventing of from being a stress test that overloads our CI machines.

I don't think it's necessary, this change aims for stability and it can do that in keeping with the existing test.

AnthonyMDev: Okay. I was just trying to be clear that I wasn't just trying to prevent timeouts. I think that having this test as a stress test of any sort is not the right direction. Especially since we removed a bunch of the complex async locking mechanisms a while ago.

designatednerd: I think that's worth filing an issue on to come back to - this is definitely something where there's a lot of places in the tests that are checking for things that were mostly caused by our Promises implementation that got ripped out. I think if you don't think we should be doing this as stress tests, we should definitely rip it out. 

However I'm fine with this as an intermediate step as long as we have some kind of actual intention to come back to it.

Tests/ApolloTests/WebSocket/WebSocketTransportTests.swift
designatednerd: Having to manually swap pointers is a bit of a code smell for me - it points out that we probably haven't architected this well for testing. 

Is there a way to set this up with either a protocol or a subclass that provides the appropriate implementation? 

I'm open to saying we should make a separate issue and punt on this for now, but we definitely should address places where our testability is kind of sucky before we hit 1.0.

AnthonyMDev: Yeah, fully agree. I wanna improve on that before 1.0. This was just a bandaid for now.

Issue #1822, opened 2021-06-10
Add ability to delete objects from the cache based on pattern-matching the key


designatednerd:
[placeholder]

calvincestari:
See https://github.com/apollographql/apollo-android/issues/3167 as the Android reference.

PR #1848, opened 2021-06-24
Pattern matching cache delete

calvincestari:
This adds `removeRecords(matching pattern: CacheKey)` to the normalized cache.

It is intended to be used through `ReadWriteTransaction` following the [established pattern for deletion](https://www.apollographql.com/docs/ios/caching/#delete), _see `removeObject(for:)`_.

Closes https://github.com/apollographql/apollo-ios/issues/1822.

#### Caveats
* This not a cascading delete, i.e.: if other records hold a reference to deleted records fetching them from the cache will fail.

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -252,6 +252,16 @@ public final class ApolloStore {
       try self.cache.removeRecord(for: key)
     }
 
+    /// Removes all objects with a key that fully or partially matches the
+    /// supplied pattern. Does not cascade or allow removal of only certain
+    /// fields. Does nothing if an object does not exist for the given key.
+    ///
+    /// - Parameters:
+    ///   - pattern: The pattern that will be used to find matching cache keys
+    public func removeObjects(matching pattern: CacheKey) throws {
+      try self.cache.removeRecords(matching: pattern)
+    }
+
     public func write<Query: GraphQLQuery>(data: Query.Data, forQuery query: Query) throws {
       try write(object: data,
                 withKey: rootCacheKey(for: query),
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -252,12 +252,19 @@ public final class ApolloStore {
       try self.cache.removeRecord(for: key)
     }
 
-    /// Removes all objects with a key that fully or partially matches the
-    /// supplied pattern. Does not cascade or allow removal of only certain
-    /// fields. Does nothing if an object does not exist for the given key.
+    /// Removes records with keys that match the specified pattern. This method will only
+    /// remove whole records, it does not perform cascading deletes. This means only the
+    /// records with matched keys will be removed, and not any references to them. Key
+    /// matching is case-insensitive.
+    ///
+    /// If you attempt to pass a cache key for a single field, this method will do nothing
+    /// since it won't be able to locate a record to remove based on that key.
+    ///
+    /// This method can be very slow depending on the number of records in the cache.
+    /// It is recommended that this method be called in a background queue.
     ///
     /// - Parameters:
-    ///   - pattern: The pattern that will be used to find matching cache keys
+    ///   - pattern: The pattern that will be applied to find matching keys.
     public func removeObjects(matching pattern: CacheKey) throws {
       try self.cache.removeRecords(matching: pattern)
     }
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -257,8 +257,8 @@ public final class ApolloStore {
     /// records with matched keys will be removed, and not any references to them. Key
     /// matching is case-insensitive.
     ///
-    /// If you attempt to pass a cache key for a single field, this method will do nothing
-    /// since it won't be able to locate a record to remove based on that key.
+    /// If you attempt to pass a cache path for a single field, this method will do nothing
+    /// since it won't be able to locate a record to remove based on that path.
     ///
     /// - Note: This method can be very slow depending on the number of records in the cache.
     /// It is recommended that this method be called in a background queue.
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -260,7 +260,7 @@ public final class ApolloStore {
     /// If you attempt to pass a cache key for a single field, this method will do nothing
     /// since it won't be able to locate a record to remove based on that key.
     ///
-    /// This method can be very slow depending on the number of records in the cache.
+    /// - Note: This method can be very slow depending on the number of records in the cache.
     /// It is recommended that this method be called in a background queue.
     ///
     /// - Parameters:
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -22,7 +22,7 @@ public final class InMemoryNormalizedCache: NormalizedCache {
   }
 
   public func removeRecords(matching pattern: CacheKey) throws {
-    return records.removeRecords(matching: pattern)
+    records.removeRecords(matching: pattern)
   }
 
   public func clear() {
--- a/Sources/Apollo/InMemoryNormalizedCache.swift
+++ b/Sources/Apollo/InMemoryNormalizedCache.swift
@@ -21,6 +21,10 @@ public final class InMemoryNormalizedCache: NormalizedCache {
     return records.merge(records: newRecords)
   }
 
+  public func removeRecords(matching pattern: CacheKey) throws {
+    return records.removeRecords(matching: pattern)
+  }
+
   public func clear() {
     records.clear()
   }
--- a/Sources/Apollo/NormalizedCache.swift
+++ b/Sources/Apollo/NormalizedCache.swift
@@ -35,8 +35,8 @@ public protocol NormalizedCache {
   /// records with matched keys will be removed, and not any references to them. Key
   /// matching is case-insensitive.
   ///
-  /// If you attempt to pass a cache key for a single field, this method will do nothing
-  /// since it won't be able to locate a record to remove based on that key.
+  /// If you attempt to pass a cache path for a single field, this method will do nothing
+  /// since it won't be able to locate a record to remove based on that path.
   ///
   /// - Note: This method can be very slow depending on the number of records in the cache.
   /// It is recommended that this method be called in a background queue.
--- a/Sources/Apollo/NormalizedCache.swift
+++ b/Sources/Apollo/NormalizedCache.swift
@@ -30,14 +30,16 @@ public protocol NormalizedCache {
   ///   - key: The cache key to remove the record for
   func removeRecord(for key: CacheKey) throws
 
-  /// Removes records matching the specified pattern. This method will only remove
-  /// whole records, not individual fields.
+  /// Removes records with keys that match the specified pattern. This method will only
+  /// remove whole records, it does not perform cascading deletes. This means only the
+  /// records with matched keys will be removed, and not any references to them. Key
+  /// matching is case-insensitive.
   ///
-  /// If you attempt to pass a cache key for a  single field, this method will do nothing
+  /// If you attempt to pass a cache key for a single field, this method will do nothing
   /// since it won't be able to locate a record to remove based on that key.
   ///
-  /// This method does not support cascading delete - it will only remove the record
-  /// for the specified key, and not any references to it or from it.
+  /// This method can be very slow depending on the number of records in the cache.
+  /// It is recommended that this method be called in a background queue.
   ///
   /// - Parameters:
   ///   - pattern: The pattern that will be applied to find matching keys.
--- a/Sources/Apollo/NormalizedCache.swift
+++ b/Sources/Apollo/NormalizedCache.swift
@@ -38,7 +38,7 @@ public protocol NormalizedCache {
   /// If you attempt to pass a cache key for a single field, this method will do nothing
   /// since it won't be able to locate a record to remove based on that key.
   ///
-  /// This method can be very slow depending on the number of records in the cache.
+  /// - Note: This method can be very slow depending on the number of records in the cache.
   /// It is recommended that this method be called in a background queue.
   ///
   /// - Parameters:
--- a/Sources/Apollo/NormalizedCache.swift
+++ b/Sources/Apollo/NormalizedCache.swift
@@ -29,7 +29,20 @@ public protocol NormalizedCache {
   /// - Parameters:
   ///   - key: The cache key to remove the record for
   func removeRecord(for key: CacheKey) throws
-  
+
+  /// Removes records matching the specified pattern. This method will only remove
+  /// whole records, not individual fields.
+  ///
+  /// If you attempt to pass a cache key for a  single field, this method will do nothing
+  /// since it won't be able to locate a record to remove based on that key.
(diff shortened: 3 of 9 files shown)


Sources/Apollo/RecordSet.swift
```diff
@@ -14,6 +14,14 @@ public struct RecordSet {
     storage.removeValue(forKey: key)
   }
 
+  public mutating func removeRecords(matching pattern: CacheKey) {
+    for (key, _) in storage {
+      if key.contains(pattern) {
```
calvincestari, line 19: `contains` seemed like the most efficient choice here given that deletion can happen within. It's unfortunate that `Dictionary` doesn't support something like `removeAll(where:)` as `Array` does.

Sources/ApolloSQLite/SQLiteDotSwiftDatabase.swift
```diff
@@ -52,6 +52,13 @@ public final class SQLiteDotSwiftDatabase: SQLiteDatabase {
     let query = self.records.filter(keyColumn == cacheKey)
     try self.db.run(query.delete())
   }
+
+  public func deleteRecords(matching pattern: CacheKey) throws {
+    let wildcardPattern = "%\(pattern)%"
+    let query = self.records.filter(keyColumn.like(wildcardPattern))
```
calvincestari, line 58: From what I could find it appears that [sqlite LIKE is case-insensitive by default](https://sqlite.org/lang_expr.html#the_like_glob_regexp_and_match_operators), hence nothing to explicitly state case insensitivity.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -609,4 +609,155 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
       }
     }
   }
+
+  func testRemoveObjectsByPattern() throws {
```
calvincestari, line 613: This is a long test, if you have suggestions on how to trim it down please say so.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -609,4 +609,155 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
...
+    //
+    // 3. Remove all objects matching the pattern `100`
+    // - This will remove `1002` (Han Solo, hero for the .newhope episode)
+    // - This will remove `1003` (Leia Organa, friend of the hero in .empire episode)
+    //
+
+    let removeCompletedExpectation = expectation(description: "Hero objects removed from cache by pattern")
+
+    store.withinReadWriteTransaction({ transaction in
+      try transaction.removeObjects(matching: "100")
```
calvincestari, line 700: I wonder if we should have another test to validate the case-insensitivity-by-default behaviour. 🤔

designatednerd: I think that's probably a good idea if we're going to be saying it's case insensitive, especially to make sure the case insensitivity is preserved across both `InMemoryNormalizedCache` and `SQLiteNormalizedCache`

Sources/Apollo/NormalizedCache.swift
```diff
@@ -29,7 +29,20 @@ public protocol NormalizedCache {
...
+  ///
+  /// If you attempt to pass a cache key for a  single field, this method will do nothing
+  /// since it won't be able to locate a record to remove based on that key.
+  ///
+  /// This method does not support cascading delete - it will only remove the record
+  /// for the specified key, and not any references to it or from it.
+  ///
+  /// - Parameters:
+  ///   - pattern: The pattern that will be applied to find matching keys.
+  func removeRecords(matching pattern: CacheKey) throws
```
designatednerd, line 44: Probably also worth throwing into the docs that this is a _slow_ method and should only ever be called in the background.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
designatednerd: I would suggest breaking each piece up after `mergeRecordsIntoCache` into convenience methods. This will also make it less horrendous to write more tests on this in the future if needed.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -609,4 +609,155 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
...
+
+      }, completion: { jediResult in
+        XCTAssertSuccessResult(jediResult)
+
+        self.store.withinReadTransaction({ transaction in
+          let query = HeroAndFriendsNamesWithIDsQuery(episode: .empire)
+          _ = try transaction.read(query: query)
+
+        }, completion: { empireResult in
+          defer { readAfterRemoveCompletedExpectation.fulfill() }
```
designatednerd, line 747: Instead of having one expectation and three nested reads, it's probably better here to have three separate reads with three separate expectations. a) Way easier to read b) Can execute concurrently

[same thought for step 2 as well]

Sources/ApolloSQLite/SQLiteDotSwiftDatabase.swift
calvincestari: There is now a test to validate this behaviour - https://github.com/apollographql/apollo-ios/pull/1848/files#diff-15a6945fd0d80fdabb91e30e1b1ffb06c396e8a7466602afb574f13b2c5b650fR613

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
calvincestari: See https://github.com/apollographql/apollo-ios/pull/1848/files#diff-15a6945fd0d80fdabb91e30e1b1ffb06c396e8a7466602afb574f13b2c5b650fR613

Sources/Apollo/RecordSet.swift
```diff
@@ -14,6 +14,14 @@ public struct RecordSet {
     storage.removeValue(forKey: key)
   }
 
+  public mutating func removeRecords(matching pattern: CacheKey) {
+    for (key, _) in storage {
+      if key.range(of: pattern, options: .caseInsensitive) != nil {
```
calvincestari, line 19: This comparison
* doesn't use `contains(:)` because that's case sensitive
* doesn't use `localizedCaseInsensitiveContains(:) because the locale of the device is not guaranteed to be the same as any app-specific locale

calvincestari: Right now this is logically equivalent to `key.uppercased().contains(pattern.uppercased())`. I'm undecided whether we should be using the option `.diacriticInsensitive` too?

Sources/Apollo/ApolloStore.swift
```diff
@@ -252,6 +252,23 @@ public final class ApolloStore {
       try self.cache.removeRecord(for: key)
     }
 
+    /// Removes records with keys that match the specified pattern. This method will only
+    /// remove whole records, it does not perform cascading deletes. This means only the
+    /// records with matched keys will be removed, and not any references to them. Key
+    /// matching is case-insensitive.
+    ///
+    /// If you attempt to pass a cache key for a single field, this method will do nothing
```
martinbonnin, line 260: I'd love us to get some terminology to differentiate between object and field keys as I think it's a common source of  confusion. 
Thinking about it, I don't think "field keys" make a lot of sense. It's not something that can be used as a key in a dictionary or a SQL primary key column. We still use them in watchers to get notified when a field changes but I think using the "key" concept there is misleading there. 

What about `CachePath` instead? For the below query:

```graphql
{
  hero {
    name 
    friend {
      email
    }
  }
}
```

We'd have:

* CacheKeys: `hero`, `hero.friends.0`, `hero.friends.1`, etc..
* CachePaths used by watchers: 
  * `hero`, `hero.friends` (do we even need those?)
  * `hero.name`
  * `hero.friends` (in case the list changes, new items moved, added or removed)

This specific API would take a `CacheKey`, making it explicit that it only works at the "object" level.
Thoughts?

Sources/Apollo/RecordSet.swift
designatednerd: my recollection is you can't use diacritics in graphql keys, but that's worth double-checking

Sources/Apollo/ApolloStore.swift
calvincestari: Yes to the common language, à la [Uniquitous Language from Domain Driven Design](https://martinfowler.com/bliki/UbiquitousLanguage.html).

`CachePath` does seem the more correct and approproiate.

@designatednerd, @AnthonyMDev - any comment/preference?

AnthonyMDev: I like cache path. We have too many overloaded and vague terms at apollo. I'm always for being more careful with our terminology.

Sources/Apollo/ApolloStore.swift
```diff
@@ -252,6 +252,23 @@ public final class ApolloStore {
...
+    /// since it won't be able to locate a record to remove based on that key.
+    ///
+    /// This method can be very slow depending on the number of records in the cache.
```
AnthonyMDev, line 263: Adding - Note: will make the generated docs look nicer.

```suggestion
    /// - Note: This method can be very slow depending on the number of records in the cache.
```

Sources/Apollo/InMemoryNormalizedCache.swift
```diff
@@ -21,6 +21,10 @@ public final class InMemoryNormalizedCache: NormalizedCache {
     return records.merge(records: newRecords)
   }
 
+  public func removeRecords(matching pattern: CacheKey) throws {
+    return records.removeRecords(matching: pattern)
```
AnthonyMDev, line 25: Do we need the `return` here?

Sources/Apollo/NormalizedCache.swift
```diff
@@ -29,7 +29,22 @@ public protocol NormalizedCache {
...
+
+  /// Removes records with keys that match the specified pattern. This method will only
+  /// remove whole records, it does not perform cascading deletes. This means only the
+  /// records with matched keys will be removed, and not any references to them. Key
+  /// matching is case-insensitive.
+  ///
+  /// If you attempt to pass a cache key for a single field, this method will do nothing
+  /// since it won't be able to locate a record to remove based on that key.
+  ///
+  /// This method can be very slow depending on the number of records in the cache.
```
AnthonyMDev, line 41:
```suggestion
  /// - Note: This method can be very slow depending on the number of records in the cache.
```

Sources/Apollo/RecordSet.swift
AnthonyMDev: GraphQL spec only allows ASCII characters.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -609,4 +609,213 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
       }
     }
   }
+
+  func testRemoveObjectByPatternCaseInsensitive() throws {
```
AnthonyMDev, line 613: This hasn't been the pattern used in the past, but I'd like to propose an improvement for naming tests functions. This describes vaguely what is being tested, but a more complete test name would be in the pattern:

"test_${thing-being-tested}_given${parameters-or-details}_(will|should|does|doesNot)${expected-behavior}"

So this would become `func test_removeObjectMatchingPattern_givenPatternNotMatchingCase_deletesCaseInsensitiveMatchingRecords()`

We don't need to go back and change all the test names now, but as we edit them or add new tests, I'd like us to follow this pattern. It makes for a much better understanding of what's being tested and what the actual expected behavior is. With the test above, I don't know if we **want** it to remove the record when it's case insensitive or **not** remove the record.

AnthonyMDev: @calvincestari @designatednerd Thoughts?

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
AnthonyMDev: Yeah helper functions would be nice, but in general tests are often longer than is acceptable for production code functions. I'm okay with this as is.

AnthonyMDev: `func test_removeObjectsMatchingPattern_givenMultipleObjectsMatchingPattern_deletesAllMatchingRecords()`

AnthonyMDev (approved):
Couple of minor suggestions, but overall this looks great! Thanks @calvincestari!

Sources/Apollo/RecordSet.swift
calvincestari: OK - then I'm going to leave it as is with the `range(of:options:)` implementation. I think it's more explicit in what it's doing vs. `contains(:)` with both `uppercased()`.

Sources/Apollo/InMemoryNormalizedCache.swift
calvincestari: No, it's more out of habit - is there a specific style guide we adhere to?

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
designatednerd: I'm fine with this general pattern but if we're gonna do this we should go through and update the rest of the tests soon. 

I would say that regardless of whether we go with that pattern or not right now, @AnthonyMDev is correct that the test name is a little vague. 

Channel your Obj-C experience when naming tests and go stupidly specific (ex, the `testUploadingSingleFileInAnArrayWithAnotherFileForAnotherField` test I was showing y'all yesterday)

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
designatednerd: > but in general tests are often longer than is acceptable for production code functions. I'm okay with this as is.

I generally agree with this but there's a number of things here that are getting repeated both within the test and in another test. That's usually the point at which I start looking to consolidate functions for readability.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
designatednerd: ^still would love to see this

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
calvincestari: I'll look for ways to consolidate functions in another PR which can then be applied to that whole test file.

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
calvincestari: This is done in 9cbd23ab

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
calvincestari: I updated the test names in 9cbd23ab and created https://github.com/apollographql/apollo-ios/issues/1849 for us to track updating test names across all tests; we can pick at it slowly.

PR #1857, opened 2021-06-30
Replace spectrum with Discourse

martinbonnin:


calvincestari (approved):
Thanks @martinbonnin!

calvincestari:
Worth noting that this takes them straight into contributing, i.e.: creating a new topic. I think that's OK vs. the alternative of searching for `client` and `mobile` tags.

martinbonnin:
Indeed, I added the query parameters after the fact. It might be worth changing the wording a bit so that it doesn't come as a surprise. Maybe "open a new topic in our community forums" instead of "join our community".

I'll leave it up to you.

designatednerd:
@calvincestari @martinbonnin There was a branch ages ago I started with more stuff on this that should still be floating around if y'all want to update and merge it

Issue #1813, opened 2021-05-31
Wrong type in generated code


conradwt:
## Bug report

In my GraphQL server, I have a `field` defined as follows:

```elixir
field :total_gross, non_null(:decimal)
```

However, I'm seeing the following in the generated code:

```swift
GraphQLField("totalGross", type: .nonNull(.scalar(String.self))),
```

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.42.0
- Xcode version: 12.5
- Swift version: 5.4
- Package manager: 

## Steps to reproduce

1. define a field as a `decimal` using the Absinthe package
2. swift run ApolloCodegen downloadSchema
3. Command + B

## Further details

I'm using SwiftUI and Phoenix/Elixir for the GraphQL server.  If you require any additional information, please do let me know.

designatednerd:
Hi! This is something you can set as an option in code generation - if you're using the Swift codegen you can set `customScalarFormat` to `.passthrough` and if you're still using Bash you can add the `--passthroughCustomScalars` flag. These are not enabled by default. 

Please let us know if that works - if it doesn't, that's a bug.

conradwt:
@designatednerd I'm guessing that I would set it as follows:

In `main.swift`:

```swift
let codegenOptions = ApolloCodegenOptions(targetRootURL: targetRootURL, customScalarFormat: .passThrough)
```

Does the CLI not accept options at this time?

conradwt:
@designatednerd I was unable to do the above because the designated initializer doesn't take the `customScalarFormat` parameter at this time.  Next, I tried to edit `ApolloCodeGenOptions.swift` but this file is set to `read-only` at this time.  Please advise because I'm guessing that I shouldn't be changing permissions on project files.

AnthonyMDev:
You can make a fork or Apollo and contribute a pull request to add that parameter if you'd like. The other initializer does include that parameter. It looks like you are using the convenience initializer.

conradwt:
@designatednerd @AnthonyMDev I'm going to close this issue because you two provided a solution. 🙏🏾🙏🏾

AnthonyMDev:
Thanks. Would be happy to merge a pull request if you add that parameter!

conradwt:
@AnthonyMDev Are you referring to the CLI?  If so, I would definitely like to do something similar to `Bash`.  For example,

```zsh
[...]

PR #1859, opened 2021-07-01
Add customScalarFormat to convenience initializer on ApolloCodegenOptions

giantramen:
Add customScalarFormat to convenience initializer on ApolloCodegenOptions. 
This allows users following the swift scripting guide (https://www.apollographql.com/docs/ios/swift-scripting/) to use a customScalarFormat.

Fixes issue brought up here: https://github.com/apollographql/apollo-ios/issues/1813

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
@@ -156,8 +156,8 @@ public struct ApolloCodegenOptions {
     self.init(codegenEngine: codegenEngine,
               operationIDsURL: operationIDsURL,
               outputFormat: .singleFile(atFileURL: outputFileURL),
-              urlToSchemaFile: schema,
               customScalarFormat: customScalarFormat,
+              urlToSchemaFile: schema,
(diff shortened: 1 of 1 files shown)


designatednerd:
Hey @giantramen, thanks for the PR! Can you explain a bit why it's better to put an additional field on the convenience initializer rather than updating the documentation to direct people to the designated initializer? 

I'm reluctant to overly-bloat the convenience initializer when the designated initializer is available, but I'm certainly open to arguments that this needs to be an exception.

giantramen:
@designatednerd  You're right that someone could just use the designated init. 
I added it because @AnthonyMDev said in the linked issue that he would merge if it was added :) https://github.com/apollographql/apollo-ios/issues/1813#issuecomment-852442957

AnthonyMDev:
Thanks for this @giantramen! @designatednerd and @calvincestari have convinced me that we shouldn't bloat that convenience initializer after some conversation.

I'm thinking of cleaning this up with some sort of file configuration struct that makes it easier to do this, but right now, I recommend using the designated initializer for most use cases.

PR #1861, opened 2021-07-06
WebSocket Fixes - Revert to Starscream 3.x and invert dependency

AnthonyMDev:
## Changelog Additions
- **Breaking - Downgraded from Starscream v4 to v3!** After upgrading to Starscream 4.0, a lot of our users started to experience crashes while using web sockets. We've decided to revert to the more stable Starscream version 3. In order to fix a few known bugs in Starscream 3, we have made a fork of Starscream that Apollo will depend on going forward. 
In preparation for moving to Apple WebSockets in the future, we have also fully inverted the dependency on Starscream. Between these two changes, a lot of breaking changes to our Web Socket API have been made:
  - The `ApolloWebSocketClient` protocol was removed and replaced with `WebSocketClient`.
  - `WebSocketClient` does not rely directly on Starscream anymore and has been streamlined for easier conformance.
  - `ApolloWebSocket`, the default implementation of the `WebSocketClient` has been replaced with `DefaultWebSocket`. This implementation uses Starscream, but implementations using other websocket libraries can now be created and used with no need for Starscream.
  - `WebSocketClientDelegate` replaces direct dependency on `Starscream.WebSocketDelegate` for delegates.


Package.resolved
```diff
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/kylef/Spectre.git",
         "state": {
           "branch": null,
-          "revision": "f14ff47f45642aa5703900980b014c2e9394b6e5",
-          "version": "0.9.0"
+          "revision": "f79d4ecbf8bc4e1579fbd86c3e1d652fb6876c53",
+          "version": "0.9.2"
```
calvincestari, line 28: I generally don't think we should be pulling other dependency updates into focused PRs. If we ever wanted to revert this PR it ends up affecting other things too. _I'm not requiring a change, will approve_

AnthonyMDev:
@designatednerd Do you know if there are any changes to the docs that need to happen for this before we cut a new release? Want to cut a release probably tomorrow, so I'd like to get that squared away.

calvincestari (approved):

Package.swift
```diff
@@ -40,8 +40,8 @@ let package = Package(
       url: "https://github.com/stephencelis/SQLite.swift.git",
       .upToNextMinor(from: "0.12.2")),
     .package(
-      url: "https://github.com/daltoniam/Starscream",
-      .upToNextMinor(from: "4.0.4")),
+      url: "https://github.com/apollographql/Starscream",
```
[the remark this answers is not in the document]
AnthonyMDev: The main repo is not actively maintained as quickly as it needs to be for us anymore. Plus we're working off of the old 3.1.1 version, so it's unlikely to receive much attention as far as getting our fixes in and new minor versions actually pushed on 3.x while they are focusing the little time they have to maintain this project on 4.x

Sources/ApolloWebSocket/DefaultWebSocketClient.swift
```diff
@@ -0,0 +1,70 @@
+import Starscream
+import Foundation
+
+/// Included default implementation of a `WebSocketClient`, based on `Starscream`'s `WebSocket`.
+public class DefaultWebSocket: WebSocketClient, Starscream.WebSocketDelegate  {
```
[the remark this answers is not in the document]
AnthonyMDev: Nope, mistake. I'll fix that. Thanks

Apollo.xcodeproj/project.pbxproj
```diff
@@ -3807,7 +3812,7 @@
 			repositoryURL = "https://github.com/daltoniam/Starscream.git";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 4.0.4;
+				minimumVersion = 3.1.1;
```
designatednerd, line 3815: It looks like there's some kind of reference to the main Starscream repo floating around in here

Apollo.xcodeproj/project.pbxproj
```diff
@@ -3851,6 +3859,16 @@
...
+		DE3A2815268BCE6700A1BDC8 /* Starscream */ = {
+			isa = XCSwiftPackageProductDependency;
+			package = DE8C84F2268BBF8000C54D02 /* XCRemoteSwiftPackageReference "Starscream" */;
+			productName = Starscream;
+		};
+		DE8C84F3268BBF8000C54D02 /* Starscream */ = {
+			isa = XCSwiftPackageProductDependency;
+			package = DE8C84F2268BBF8000C54D02 /* XCRemoteSwiftPackageReference "Starscream" */;
+			productName = Starscream;
+		};
```
designatednerd, line 3871: also not clear on why this has been added like three times

designatednerd (commented):
In the notes: `we have made a private forked of Starscream` - I think that should probably be `we have made a fork of Starscream`, since the fork isn't actually private.

Apollo.xcodeproj/project.pbxproj
AnthonyMDev: Not clear on that either... XCProject files still have some weird bugs with their SPM integrations sometimes. I'm not too worried about it though. Not sure if there is anything we can do to fix that...

mdubs:
Is there any plan to remove the actual Starscream dependency being declared in the main Apollo Package.swift? Currently even though we don't use the ApolloWebsocket target, SPM/Xcode will clone the Starscream dependency (due to https://github.com/apple/swift-evolution/blob/main/proposals/0226-package-manager-target-based-dep-resolution.md not being fully implemented). This means that we are still prevented from using Starscream 4.x (or really any version except for the Apollo forked 3.1.2 version).

AnthonyMDev:
I will fix this tomorrow.

AnthonyMDev:
So l'm starting to look into this. Moving the Starscream dependency into another target (out of `ApolloWebSocket`) makes sense, but moving it out of the `Package.swift` file altogether seems messy and makes integration for others require additional steps. I'm going to talk to the team today about how we want to approach this.

AnthonyMDev:
#1906 should fix this for you @mdubs! Will put a release out next week.

Issue #939, opened 2019-12-09
RFC: Swift Codegen Rewrite


designatednerd:
This issue will serve as the official Request for Comment for the [Swift Codegen](https://github.com/apollographql/apollo-ios/projects/2) project.

This project is to take our existing code generation, which is done in Typescript, and move it to Swift. 

Note that parsing of the Schema and parsing/validation of queries will still be taking place using our Typescript tooling - these are presently too tied to JS to be able to be replaced without a very, very large amount of work. 

Moving the actual generation of code to Swift will allow Swift developers to more easily understand how code is being generated, and more easily contribute fixes and improvements to the code generation process.

This project will take place in several phases, which are outlined below.

## Phase 0: Generate code with swift Script instead of bash

Swift scripts can be run directly from the command line, and these scripts can also import frameworks. 

Everything we are currently using bash for will be updated to use Swift wrappers in a framework which will be included with the core Apollo library. This includes: 

- Downloading and validating the JS tooling (abstracted away from the user, but still tested)
- Running a command with the Apollo CLI
- Downloading a schema
- Running code generation

This will be considerably easier to test than using random bash scripts, and those tests will help ensure that any changes are non-breaking.

The initial framework and tests are already up as a [draft PR](https://github.com/apollographql/apollo-ios/pull/940) if you're interested in looking at the details.

The other thing this will do is make it considerably easier to add more flags in the future, specifically the flag to enable the Swift-based code generation rather than the Typescript-based code generation. 

This phase will require significant updates to documentation since everything assumes the use of bash right now.

## Phase 1: New Codegen

This is going to involve a few sub-phases. 

### Getting the Abstract Syntax Tree out of our existing codegen in JSON format

This is the part that depends most on our tooling repo. This *should* basically be done, but this could potentially be a blocker if there are pieces that we need to generate Swift code that's clearer and harder to fail at.

### Parsing the AST into Swift Objects for code generation

In theory, this part should be the most straightforward, and will mostly involve parsing everything in the JSON into `Codable` objects which will ultimately be used by the code generation engine.

### Generating code

This will involve a number of new goodies: 

- `Codable` conformance instead of custom JSON parsing
- Automatic `Equatable` and `Hashable` conformance
- Addition of an option to automatically add `Identifiable` conformance when a `GraphQLID` is one of the returned fields. 
- Union types will be represented by `String`-backed enums to make it easier to `switch` on the underlying type and to pull out the types being generated
[...]

PR #1873, opened 2021-07-15
Removing Swift codegen (v1)

calvincestari:
This PR removes the initial implementation of Swift codegen from the main branch.
* It's labelled as 'experimental' and 'use at own risk' but it's not functional yet.
* This code represents the implementation of the [RFC: Swift Codegen Rewrite](https://github.com/apollographql/apollo-ios/issues/939) which is changing.
* Gets main back to a state of supported codegen options.

It does this in distinct commits, some of which will be reverted in another branch..
1. Remove the expected code output, and related tests
2. Remove code generators for expected code output, and related tests
3. Remove two targets that were configured to use the Swift codegen
4. Remove 'swiftExperimental' as a codegen configuration

**Swift codegen is not dead**; there will be a new branch where v2 of the RFC will be built until it is ready for an alpha release.


Sources/ApolloCodegenLib/ASTCondition.swift
```diff
@@ -1,18 +0,0 @@
-import Foundation
```
AnthonyMDev, line 1: @designatednerd Are these AST objects being used by the Javascript Frontend? Is there even anything that is creating them currently, other than unit tests for the experimental codegen? I think we can delete these from `main`, but just want to make sure.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -58,59 +58,6 @@ class ApolloCodegenTests: XCTestCase {
     ])
   }
   
-  func testCreatingOptionsWithAllParameters() throws {
```
AnthonyMDev, line 61: I don't think we should kill this test, just change it to use the typescript engine for now.

AnthonyMDev (commented):
Just want to make sure this AST stuff isn't used elsewhere, I'm pretty sure this is good though!

And lets fix that test instead of deleting it?

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: Yes - my bad. The option was incrementally added to the test vs. created the test for it.

Sources/ApolloCodegenLib/ASTCondition.swift
designatednerd: They should not be, this was being created off reading everything in using the TS CLI, you should be able to nuke them from orbit

SwiftScripts/Sources/Codegen/Target.swift
```diff
@@ -4,23 +4,17 @@ import ArgumentParser
...
         case "StarWars":
             self = .starWars
-        case "StarWars-SwiftCodegen":
-            self = .starWarsSwiftCodegen
         case "GitHub":
             self = .gitHub
         case "Upload":
             self = .upload
-        case "AnimalKingdom":
-            self = .animalKingdom
```
designatednerd, line 23: did we mean to delete this?

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -67,7 +67,7 @@ class ApolloCodegenTests: XCTestCase {
     let namespace = "ANameSpace"
     let prefix = "MyPrefix"
     
-    let options = ApolloCodegenOptions(codegenEngine: .swiftExperimental,
```
designatednerd, line 70: for what it's worth I think it's worth keeping the `swiftExperimental` case since it's still eventually going to be used, just for a completely different setup

designatednerd (commented):
@AnthonyMDev I *think* all the issues I tried to address for enums are at least still in the codegen project, but it might be useful to think about leaving a bunch of test names and just `throw XCTSkip` on those tests for now until it's actually implemented

SwiftScripts/Sources/Codegen/Target.swift
calvincestari: I did, it was only there to be used with `swiftExperimental`. It comes back in the other RFC-focused branch.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: It comes back in the other RFC-focused branch. My opinion was that if it stayed in `main` then it should blow up, aka `fatalError`, because it does nothing at this point. Easier to nuke from `main` and bring it back where we actually use it.

designatednerd: eh, fair enough. I'm always more in favor of just leaving it but I see your point too.

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -58,7 +58,6 @@ public struct ApolloCodegenOptions {
   }
   
   let codegenEngine: CodeGenerationEngine
-  let additionalInflectionRules: [InflectionRule]
```
designatednerd, line 66: just please please please remember to re-add this, it solves so many goddamned problems

designatednerd (approved):

AnthonyMDev:
If we are (temporarily) not using some of these dependencies any more, we should remove them from SPM and the podfile.

I'm pretty sure that Stencil, PathKit, and InflectorKit are all not being used after we kill this code. We're going to need to re-add them on the branch for the codegen RFC at some point, but we don't need our consumers to be downloading and including them in their projects right now for the time being.

designatednerd (changes requested):
Revoking my approval: Please don't remove InflectorKit and the related code completely. That can be used for pluralization/depluralization as needed, which we absolutely will need regardless of whether we use stencil or not. I'm fine with removing the code from the initializer for now since it's not being used, but this is a case where I really don't think removing it just to put it back in later exactly as it was is worth it. 

I'm fine with losing Stencil (which IIRC is where the PathKit dependency comes from) since @AnthonyMDev it sounds like we're not going to use it going forward.

calvincestari:
@designatednerd I feel quite strongly that we shouldn't have code in `main` that isn't actively used. I understand your argument re. "I really don't think removing it just to put it back in later exactly as it was is worth it" but we don't yet know if it'll go back in exactly as it was; please note I'm not disputing it's value.

@AnthonyMDev - do you have an opinion on this or the workflow?

AnthonyMDev:
I don’t really feel strongly either way. I see both sides and don’t really have any additional arguments to add to either of them. 

FWIW, I think I lean towards removing them, even if we are going to re-add them later, but not passionate about that. 

Also, the more I think about it and research, the more I think it’s likely we will end up using Stencil.

designatednerd:
In my research `InflectorKit` is about the only thing that is out there for this kind of pluralization/depluralization in Swift (that's why I originally revived it and contributed swift wrappers to it). There is absolutely no question in my mind we will need _something_ for pluralization/depluralization and probably sooner rather than later. 

I'm fine with ripping out the other stuff as there are other ways to get from point a to point b, but to me it feels like a waste to rip out something that's already set up and tested just to put it right back in when we start needing to do basic pluralization/depluralization of types, which we'll need to do quite quickly when we get the codegen going.

AnthonyMDev (approved):

liuliu:
It is pretty hard to navigate, but where is the "legacy" codegen resides? I suppose that is some Typescript code somewhere in some other repo, but have a hard time to find exact where it is and how we packaged up to legacy_cli that we use.

PR #1876, opened 2021-07-20
[RFC] In Progress - Release 1.0 - Swift Codegen

calvincestari:
This is the current working branch for the 1.0 release candidate.

The primary focus of the 1.0 release is a reimagining of the generated GraphQL operation models. This includes a major overhaul of the internal model and execution layer and a re-write of the code generation engine in Swift.

An RFC for the new format of the generated GraphQL operation models is provided as part of this PR.

## Please make comments on this RFC as a review of this PR.
The full rich text of the RFC can be found at [CodegenProposal.md](https://github.com/apollographql/apollo-ios/blob/release/1.0-alpha-incubating/CodegenProposal.md)

### IN PROGRESS - DO NOT MERGE

* This PR serves to collect all changes that we intend to release as an alpha of `1.0.0`.
* We should continue merging smaller changes into `main` and releasing them.
* Occasionally this branch should be brought up-to-date with `main`.
* Keep adding code to this branch via PRs.


CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+# Schema Type Generation
+
+In addition to generating `SelectionSet`s for your `GraphQLOperation`s, types will be generated for each type (object, interface, or union) that is used in any operations across your entire application. These types will include all the fields that may be fetched by any operation used and can include other type metadata. 
+
+The schema types have a number of functions. 
+
+* Include metadata that allows the `GraphQLExecutor` and runtime type checking on `TypeCase`s to operate. 
+* Can be extended to provide cache key computation for types to configure the normalized cache.
+* Used for interacting with the cache for manual read/write functionality.
+* Used to create mock objects for generated `SelectionSet`s to be used in unit tests.
```
martinbonnin, line 316: I'm experimenting with this at the moment and one drawback of using schema types for mock objects generated is that schema types don't know anything about aliases. If an operation uses aliases, creating a mock for it using schema types only is a bit awkward and certainly non type safe as the alias name would have to be passed at runtime.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+
+
+These schema types can be included directly in your application target, or be generated into a separate shared library that can be used across modules in your application.
```
martinbonnin, line 319: I'm contemplating separating the type information needed for cache/validation from the type information needed for building mock objects. Since the later are to be used during tests, it would be nice not to embed them in the final app.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+
+### Merging Fragment Fields Into Parent `SelectionSet`
+
+In the above example you may note that the `species` field is accessible directly on the `Animal` object without having to access the `AnimalDetails` fragment first. This is because fields from fragments that have the same `__parentType` as the enclosing `SelectionSet` are automatically merged into the enclosing `SelectionSet`. 
+
+In previous versions of the Code Generation tool, this functionality was exposed via the mergeInFieldsFromFragmentSpreads option, which defaulted to `false`. Merging in fragment fields provides for easier consumption of generated models, but it also increases the size of the generated code. Because the size of the generated code is being dramatically reduced with the new Code Generation tooling, we have opted to always merge in fragment fields. If generated code size becomes a concern with the new Code Generation, adding an option to disable fragment merging may be considered in the future.
```
martinbonnin, line 696: The main issue with enabling `mergeInFieldsFromFragmentSpreads` is that codegen size grows exponentially with the depth of nested fragments. So even if the new codegen generates 10x less code, a deep query using a lot of fragments might still generate a huge amount of code.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+# `EnumType` Generation
+
+Enums will be generated for each `enum` type in the schema that is used in any of the operations defined in your application. These enums will conform to a simple `EnumType` protocol. When used as the type for a field on a `SelectionSet`, these enums will be wrapped in the generic `GraphQLEnum`.
+
+```swift
+enum RelativeSize: String, EnumType {
+  case LARGE
+  case AVERAGE
+  case SMALL
+}
```
designatednerd, line 350: probably worth clarifying that unlike in current codegen, we'll respect capitalization from the schema.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+struct MyInput: InputObject {
+  public private(set) var dict: InputDict
+
+  init(size: RelativeSize?) { ... }
+  
+  /// If `nil`, defaults to server-provided default value (.SMALL)
+  var size: RelativeSize? { ... }
+}
+```
+> Note that we are not generating these fields with the provided default values. This is to account for default values that may change on the schema in the future. See [Generate Default Parameter Values for `InputObject` Default Values](#generate-default-parameter-values-for-inputobject-default-values) for more discussion.
```
designatednerd, line 477: One other thing worth considering: What happens if there's a default value provided in the schema when it's generated, but then the default value is removed on the server later?

CodegenProposal.md
AnthonyMDev: Great point! I plan on using the schema types to build the objects, and then when you want to convert them to a specific selection set, I could run them through the `GraphQLExecutor` really quickly to map any aliased fields to their correct fields.

If you want a mock of a specific `SelectionSet`, you'll have to make your mock from the schema type and then run a function to convert it to the `SelectionSet`. Still haven't defined the API for that exactly, but I have a decent idea of how to do that.

CodegenProposal.md
AnthonyMDev: Would be interested in hearing more about this, but it sounds like we are taking very different approaches to mock objects. I haven't started writing the code for my approach yet though, so I could be persuaded to take a different approach.

CodegenProposal.md
AnthonyMDev: Yes, agreed. Because we generate dramatically less code, I feel that it's likely going to not be a problem. But if users complain about it, we can add that option in in the future. Codegen is such a massively scoped project, so we are trying to cut down to the MVP as much as possible, and then add more functionality from there.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+
+For this reason, we have opted to not generate default values.
+
+## Concrete Subtypes as Enums
+
+[@designatedNerd’s initial proposal includes enums with associated values for subtypes.](https://github.com/apollographql/StarWarsiOSMk2/blob/master/StarWarsMark2/Queries/HeroTypeDependentAliasedFieldQueryMk2.swift#L43)
+
+These `SubType` enums thing works for concrete types, not interfaces (because a type could conform to multiple interfaces). We don't plan on generating all the Concrete types as data structures unless they are specifically enumerated (`... on Droid`)
```
designatednerd, line 1386: "These subtype enums thing" could use a little cleanup :)

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+
+## Client-Side Directives For Automatic `CacheKeyProvider` Generation
+
+## Custom Field Validation
+
+## Better Support For Types Added To Schema After Code Generation 
+
+In order to cast new concrete types to type conditions, we would need to know the metadata about what interfaces the types implement. We could possibly use a schema introspection query to fetch additional types added to the schema after code generation.
+
+## Generation of Enums Providing All Possible Types for Unions
```
designatednerd, line 1474: all *known* possible types for unions 😇

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
...
+        }
+      }
+    }
+    ```
+* Implement logic so that if your query has _**one or more**_ fragments on a concrete type, then we generate the subtypes (generate the enum with only 1 case + `_other`)
+* Implement logic so that if your query has _**more than one**_ fragment on a concrete type, then we generate the subtypes
+
+# Possible Future Additions
+
+## Client-Side Directives For Automatic `CacheKeyProvider` Generation
```
designatednerd, line 1466: Isn't this covered by the cache key stuff abovve?

designatednerd (commented):
Would definitely help to have a quick summary at the top of all this of what is changing from existing codegen with links to the bits that are changing.

CodegenProposal.md
```diff
@@ -0,0 +1,1474 @@
+# iOS Codegen Proposal
+
+# Overview
+
+This document provides explanation, context, and examples for a proposal for the new code generation for iOS.
+
+
+> **Example code in this document is used only to illustrate the concepts being discussed, as is not comprehensive. Actual generated objects may have additional properties, functions, or nested types to support all functionality. For complete examples of generated code, see the code repository for this proposal..**
```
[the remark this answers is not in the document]
AnthonyMDev: Already done on my current branch this morning, lol. Cleaning that up in my next commit!

CodegenProposal.md
AnthonyMDev: Will do! Thanks

CodegenProposal.md
AnthonyMDev: That would be a breaking change. Since backwards compatibility is enforced as a standard, breaking changes are "supposed" to be avoided and you would make a new field and deprecate the old one if you needed to do this. 

If developers choose to make breaking changes, the Apollo client library behavior is undefined, and we shouldn't be expected to handle those situations gracefully.

In this case, the server would return a graphQL error response.

CodegenProposal.md
AnthonyMDev: lol yes, it could. :)

CodegenProposal.md
AnthonyMDev: It's mentioned briefly above. I just wanted to have a specific section that we compile a list of possible future considerations.

AnthonyMDev: And this isn't being implemented right now. The stuff above is examples of how you can implement `CacheKeyProviders`, but they aren't going to be generated for you. You have to write them manually. 

This refers to using the `@keyFields` directive that @martinbonnin has been working on defining to generate these for you automatically.

CodegenProposal.md
AnthonyMDev: We should confirm that's the case actually. Going to depend on if the `graphql-js` library does that.

CodegenProposal.md
designatednerd: Probably worth noting that explicitly in this proposal

CodegenProposal.md
designatednerd: Gooot it. Might be worth filling in brief details on what these things are before we call in the public.

CodegenProposal.md
AnthonyMDev: Confirmed that is does respect case of the schema. I'll update that.

designatednerd: Yeah my recollection was there was some weird shit in the TS code that I could never track down that did this, it wasn't GQL-js

CodegenProposal.md
AnthonyMDev: Done

AnthonyMDev:
> Probably worth defining "Type case" here - it's not a phrase I've seen used much externally
@designatednerd I thoroughly define "Type Case" in the linked sections of the document. This was meant to be just a summary of key changes. I'd like to leave the full definition where it is. It's linked right there, so if people are confused, they can click into that section of the doc.

AnthonyMDev:
> Worth noting that we're only generating the types and properties you're actually asking for, not like your entire schema (since some schemas are super gargantuan)

@designatednerd Well, currently, we aren't generating _any_ of the fields. We are generating only the types you are using, but they are empty classes at the moment (Just have metadata for implemented interfaces). Generation of the actual fields will be part of the mutable cache, which will be an entire new RFC. 😅

CodegenProposal.md
martinbonnin: > you'll have to make your mock from the schema type and then run a function to convert it to the SelectionSet

I think that part could be the problematic one. If you have 2 field with different values, one of them being aliased, there's no real way to mock this using schema types:


```graphql
query {
  cat:  animal(species: CAT)
  dog: animal(species: DOG)
}
```

I'm still experimenting with this so maybe there's a convenient and idiomatic way to set aliases for schema types but it feels a bit out of place at first sight.

designatednerd:
Ahhh OK I think I didn't put together that's going to be a separate RFC. Gotcha!

CodegenProposal.md
AnthonyMDev: Yeah, this is definitely a rough edge. I think there is a solution to this. We are already ignoring aliases when storing fields in the cache on iOS. The field is stored using a cache key from the actual field name. If the field has arguments, those arguments become part of the cache key. So this seems doable.

twof:
Hi! Apologies if this is a bad place to ask. Feel free to redirect me. A few clarifying questions.

1. Will the new code generator still rely on GraphQL.js? It sounds like the answer is "yes" based on #1700, but I want to confirm.
2. Will the new code generator rely on apollo-tooling, or is that still being deprecated?

I'm looking to find out what additional work will need to be done on Apollo iOS once [Client Controlled Nullability](https://github.com/graphql/graphql-spec/pull/895#issuecomment-988414387) is approved and the GraphQL.js implementation is merged.

calvincestari:
> Hi! Apologies if this is a bad place to ask. Feel free to redirect me. A few clarifying questions.

Hi @twof - yes, we can address questions like that here since they're directly related to what will become the 1.0 release although this PR has a slightly narrower focus of the code part of code generation RFC.

> Will the new code generator still rely on GraphQL.js? It sounds like the answer is "yes" based on [Mega-Issue] New Codegen #1700, but I want to confirm.

Yes. Code generation in 1.0 will still use [graphql-js](https://github.com/graphql/graphql-js) for validation and compilation of a GraphQL schema and operations.

> Will the new code generator rely on apollo-tooling, or is that still being deprecated?

No, the apollo-tooling dependency is removed and the code generation engine in 1.0 is being written in Swift. The [objectives](https://github.com/apollographql/apollo-ios/blob/release/1.0-alpha-incubating/CodegenProposal.md#objectives) of that decision are outlined in the RFC. It's going to be a vast improvement, I'm excited about it!

> I'm looking to find out what additional work will need to be done on Apollo iOS once [Client Controlled Nullability](https://github.com/graphql/graphql-spec/pull/895) is approved and the GraphQL.js implementation is merged.

We have bold plans for future major versions of the SDK, namely 2.0 will rebuild the networking stack and 3.0 will improve caching. Client controlled nullability would be a great contribution although I expect that change would stretch across all three major parts of the SDK - code generation, networking and caching. It'll require a careful approach to the code that exists now but also trying to prepare for what it will look like in the future.

twof:
Thanks @calvincestari ! That's super helpful.

AnthonyMDev:
@twof, Once Client Controlled Nullability is merged in to graphql-js, we should get on a call about this. If you are willing to do the work to support it, I'd like to get it in as soon as possible (maybe even with the 1.0 release).

twof:
@AnthonyMDev Sounds good! I'll let yall know.

lorderster (approved):

calvincestari:
This PR was closed when we deleted the `release/1.0-alpha-incubating` branch. All Alpha tagged releases can be found on `release/1.0`.

Issue #1849, opened 2021-06-25
Update tests to follow naming convention


calvincestari:
This hasn't been the pattern used in the past, but we're suggesting an improvement for naming test functions. It makes for a better understanding of what's being tested and what the expected behavior is.

The naming convention to adopt is `test_${thing-being-tested}_given${parameters-or-details}_(will|should|does|doesNot)${expected-behavior}`. See [here](https://github.com/apollographql/apollo-ios/pull/1848/files#diff-15a6945fd0d80fdabb91e30e1b1ffb06c396e8a7466602afb574f13b2c5b650fR688) as an example.

_Use this issue to track the tests that have been updated and those remaining; we don't need to get it all done in one PR._

#### ApolloCodegenTests
- [ ] ApolloCodegenTests
- [ ] ApolloSchemaTests
- [ ] ASTParsingTests
- [ ] CLIDownloaderTests
- [ ] CLIExtractorTests
- [ ] CodegenExtensionTests
- [ ] CompilationTests
- [ ] DocumentParsingAndValidationTests
- [ ] EnumGenerationTests
- [ ] FileManagerExtensionsTests
- [ ] InputObjectGenerationTests
- [ ] InterfaceEnumGenerationTests
- [ ] JSONTests
- [ ] PluralizerTests
- [ ] SchemaIntrospectionTests
- [ ] SchemaLoadingTests
- [ ] UnionEnumGenerationTests
- [ ] URLExtensionsTests
- [ ] VariableToSwiftTypeTests

#### ApolloPerformanceTests
- [ ] ParsingPerformanceTests

#### ApolloServerIntegrationTests
- [ ] LegacyInterceptorProviderIntegrationTests
- [ ] SQLiteStarWarsServerAPQsGetMethodTests
- [ ] SQLiteStarWarsServerAPQsTests
- [ ] SQLiteStarWarsServerCachingRoundtripTests
- [ ] SQLiteStarWarsServerTests
- [ ] StarWarsApolloSchemaDownloaderTests
- [ ] StarWarsServerAPQsGetMethodTests
- [ ] StarWarsServerAPQsTests
- [ ] StarWarsServerCachingRoundtripTests
- [ ] StarWarsServerTests
- [ ] StarWarsSubscriptionTests
- [ ] StarWarsWebSocketTests
- [ ] UploadTests
- [ ] URLSessionClientLiveTests

#### ApolloTests
- [ ] ApolloClientOperationTests
[...]

PR #1877, opened 2021-07-20
Codegen test updates

calvincestari:
1. Add test for `.swift` code generation engine argument
2. Update test function names to match #1849

### DIFF
--- a/Tests/ApolloCodegenTests/ApolloCodegenTests.swift
+++ b/Tests/ApolloCodegenTests/ApolloCodegenTests.swift
@@ -57,6 +57,23 @@ class ApolloCodegenTests: XCTestCase {
       "'\(output.path)'",
     ])
   }
+
+  func testConvenienceInitializer_givenAllParameters_shouldBuildCorrectArguments() throws {
+    let targetURL = URL(string: "file://could/be/anything/not/important")!
+    let options = ApolloCodegenOptions(targetRootURL: targetURL, codegenEngine: .swift, downloadTimeout: 42.0)
+
+    XCTAssertEqual(options.downloadTimeout, 42.0)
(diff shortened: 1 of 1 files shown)


Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -2,8 +2,7 @@
 //  ApolloCodegenTests.swift
 //  ApolloCodegenTests
 //
-//  Created by Ellen Shapiro on 10/7/19.
-//  Copyright © 2019 Apollo GraphQL. All rights reserved.
+//  Copyright © 2021 Apollo GraphQL. All rights reserved.
```
designatednerd, line 5: If you'd prefer you can just nuke the headers on the tests like we have for everything in the main lib

designatednerd (approved):

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
AnthonyMDev: Yeah, I definitely prefer not having headers on every file.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -22,14 +21,13 @@ class ApolloCodegenTests: XCTestCase {
     super.tearDown()
   }
   
-  func testCreatingOptionsWithDefaultParameters() throws {
+  func testDesignatedInitializer_givenRequiredParameters_shouldBuildDefaultArguments() throws {
```
AnthonyMDev, line 17: Love this naming! :)

AnthonyMDev (approved):

Issue #955, opened 2019-12-18
Naming error when generating mutation in Apollo Swift Codegen 


andre-felipe:
This is the mutation I'm generating:
```
mutation ReportDriverTaskRelativeETA(
    $taskId: ID!
) {
    reportDriverTaskRelativeETA(input:{
        taskId: $taskId
    }) {
        ... on NoData {
            __typename
            description
        }
    }
}
```

Codegen seems to be renaming the "ETA" part in the mutation name to "ETum" for no apparent reason:
 ```
public struct ReportDriverTaskRelativeETum: GraphQLSelectionSet {
      public static let possibleTypes = ["NoData"]

      public static let selections: [GraphQLSelection] = [
        GraphQLField("__typename", type: .nonNull(.scalar(String.self))),
        GraphQLField("description", type: .nonNull(.scalar(String.self))),
      ]
(...)
}
```

That glitch alone would be "fine" (i.e. usable), but then the `get` accessor for the struct var renames the "ETum" part of the name yet again to "Etum":
```
public var reportDriverTaskRelativeEta: ReportDriverTaskRelativeETum {
      get {
        return ReportDriverTaskRelativeEtum(unsafeResultMap: resultMap["reportDriverTaskRelativeETA"]! as! ResultMap)
      }
      set {
        resultMap.updateValue(newValue.resultMap, forKey: "reportDriverTaskRelativeETA")
      }
    }
(...)
```

This "renaming mismatch" causes a build error:
![Screenshot 2019-12-17 at 16 06 22](https://user-images.githubusercontent.com/11345047/71087175-d6771f80-219b-11ea-8836-ba58cfdddbc2.png)

This only happens to this particular mutation, and if I rename it to something else (say "ReportDriverTaskRelativeEstimatedTimeOfArrival"), it's generated properly and works normally.

Anyone ever ran into something like this? Any ideas why this would happen?

designatednerd:
Urgh. This is our singularization lib in the Typescript codegen thinking "ETA" is something like "Data" and trying to singularize it to something like "Datum".  

This behavior will be removed in the Swift Codegen. I'll have to look at our tooling stuff to see if there's a reasonable way to get rid of it in the TS stuff (I'm not sure there is without borking a bunch of other stuff though). 

In the meantime, the recommendation would be to use a different name for that mutation.

designatednerd:
[...]

PR #1879, opened 2021-07-21
Bring back all the pluralization

calvincestari:
bring-back-all-the-things

This is a revert of https://github.com/apollographql/apollo-ios/pull/1873/commits/42a75f2a47c8aa318939b36981d3b012bff2e521 with some changes
* InflectorKit is now sourced from the origin instead of the apollographql fork
* InflectorKit is pinned to the next minor version of `1.0.0`
* `PluralizerTests` was updated to comply with deprecated function names
* Update `PluralizerTests` function names to match #1849

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
@@ -82,6 +82,7 @@ public struct ApolloCodegenOptions {
   ///
   /// - Parameters:
   ///  - codegenEngine: The code generation engine to use. Defaults to `CodeGenerationEngine.default`
+  ///  - additionalInflectionRules: Any non-default rules for pluralization or singularization you wish to include. Defaults to an empty array. Only used by the Swift code generation engine.
   ///  - includes: Glob of files to search for GraphQL operations. This should be used to find queries *and* any client schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
   ///  - mergeInFieldsFromFragmentSpreads: Set true to merge fragment fields onto its enclosing type. Defaults to true.
   ///  - modifier: The access modifier to use on everything created by this tool. Defaults to `.public`. Only used by the Swift code generation engine.
@@ -94,7 +95,6 @@ public struct ApolloCodegenOptions {
   ///  - suppressSwiftMultilineStringLiterals: Don't use multi-line string literals when generating code. Defaults to false.
   ///  - urlToSchemaFile: The URL to your schema file.
   ///  - downloadTimeout: The maximum time to wait before indicating that the download timed out, in seconds. Defaults to 30 seconds.
-  ///  - additionalInflectionRules: Any non-default rules for pluralization or singularization you wish to include. Defaults to an empty array. Only used by the Swift code generation engine.
   public init(codegenEngine: CodeGenerationEngine = .default,
               additionalInflectionRules: [InflectionRule] = [],
               includes: String = "./**/*.graphql",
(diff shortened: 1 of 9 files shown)


Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -83,7 +84,7 @@ public struct ApolloCodegenOptions {
   ///  - codegenEngine: The code generation engine to use. Defaults to `CodeGenerationEngine.default`
   ///  - includes: Glob of files to search for GraphQL operations. This should be used to find queries *and* any client schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
   ///  - mergeInFieldsFromFragmentSpreads: Set true to merge fragment fields onto its enclosing type. Defaults to true.
-  ///  - modifier: [EXPERIMENTAL SWIFT CODEGEN ONLY] - The access modifier to use on everything created by this tool. Defaults to `.public`.
+  ///  - modifier: The access modifier to use on everything created by this tool. Defaults to `.public`. Only used by the Swift code generation engine.
```
AnthonyMDev, line 87: This is fine for now. I think when we release the swift codegen, we will need to keep support for legacy codegen for a while, but that it should probably have its own set of options entirely so we can diverge more easily. But that's not a concern for today.

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -93,7 +94,9 @@ public struct ApolloCodegenOptions {
   ///  - suppressSwiftMultilineStringLiterals: Don't use multi-line string literals when generating code. Defaults to false.
   ///  - urlToSchemaFile: The URL to your schema file.
   ///  - downloadTimeout: The maximum time to wait before indicating that the download timed out, in seconds. Defaults to 30 seconds.
+  ///  - additionalInflectionRules: Any non-default rules for pluralization or singularization you wish to include. Defaults to an empty array. Only used by the Swift code generation engine.
```
AnthonyMDev, line 97: I feel like parameters should be ordered in documentation the same order as they are in the function. So if this is the second parameter, it should be second in the docs too.

AnthonyMDev (approved):
Thanks for doing this @calvincestari!

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
calvincestari: Not a problem - done in https://github.com/apollographql/apollo-ios/pull/1879/commits/d6413e203ab8e851904a7bf7e3d6826073f2685c.

martinbonnin:
Wait, isn't all this about **singularization** 🤔 ?

calvincestari:
[InflectorKit](https://github.com/mattt/InflectorKit) provides both and `Pluralizer` wraps them.

martinbonnin:
But we don't really need to pluralize? Or do we?

calvincestari:
I am not sure - @designatednerd may have a better answer to that question.

martinbonnin:
I think we use it for things like

```graphql
{
  hero {
    # This is a list type
    friends {
      name
    }
  }
}
```
to generate a `Friend` struct/class (and not Friends):

```swift
struct Friend {
  let name: String
}
```

I don't think we need the other way

designatednerd:
We primarily use it for singularization. Somewhere in the back of my mind I remember there was some point where we were also pluralizing but i'm having a memory read error on where.

designatednerd:
Aha, found it: #955, was trying to pluralize `ETA` as `Etum`

calvincestari:
@designatednerd - with this merged we can probably drop [our fork of InflectorKit](https://github.com/apollographql/InflectorKit).

designatednerd:
Yes, though I would not recommend deleting it since then it won't resolve for older versions of the lib

Issue #1422, opened 2020-09-30
Query watcher not being called when cache is updated on an element by another query/subscription/mutation


benoitletondor:
## Bug report

Query watcher not being called when cache is updated on an element of a collection that is added after calling watch

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.34.0
- Xcode version: 12.0.1
- Swift version: 5.3

## Steps to reproduce

I've been trying a lot of things on that one and wasn't able to find a way to fix my problem. So in my app I'm having a list of conversations, so I'm creating a `watch` on a query that returns the list of conversations, this is something like that:

```
user {
    uuid
    firstName
    lastName
    conversations {
        uuid
        unreadMessagesCount
    }
}
```

> In the app we use the `uuid` key to handle cache, so we make sure to always pass `uuid` in our queries to automatically handle cache update.

So the role of the watcher I'm talking about is both to update existing conversations and also be able to catch when a new conversation is created (it can happen and not be initiated by the user, we then trigger an event from the backend that is listen from a subscription in the app). When this event happens it returns something like that

```
event {
     newConversation {
          conversation {
               uuid
               unreadMessagesCount
               // This is the part that adds the new conversation to the existing ones of the users in the cache
               user {
                    uuid
                    conversations {
                          uuid
                          unreadMessagesCount
                    }
               } 
          }
     }
}
```

It works great, meaning that when this event happens, the watcher is being called with the newly created conversation, but the issue is that any new cache update for that specific conversation doesn't trigger the watcher again.

After investigating a bit, I realised that the cache is being updated because if I'm adding 3 new messages into the new conversation (setting the `unreadMessagesCount` to 3), the watcher doesn't get called but then if I add 1 new message into an old one, the watcher is being called with both the new message on the old conversation and the 3 new on the new one. 

So it really seems like `watch` is not being called again for changes on an item that wasn't in a collection when the `watch` was initially made. I've also take a look at https://github.com/apollographql/apollo-ios/issues/281 and making a `fetch` on the same query again after the event doesn't fix the issue.

Let me know if I'm not clear on something as the whole thing is a bit complicated to explain.

benoitletondor:
To add more to this, I feel like this is not related to collections only. If you have 1 query that watches something, then another query/mutation updates this cache indirectly (not using the same query, but another query that contains the same element with the same `uuid`), it doesn't work.

designatednerd:
[...]

PR #1889, opened 2021-08-02
Publish response from the `WebSocketTransport` to the `ApolloStore`

tgyhlsb:
# Pull Request

- Fixes: https://github.com/apollographql/apollo-ios/issues/1422

# Summary

When using the `WebSocketTransport` for subscriptions, we've noticed that our cache is never updated. In fact, there is no code inside `WebSocketTransport` about any store or cache.

This PR injects the `ApolloStore` inside `WebSocketTransport` and calls the same method used within `CacheWriteInterceptor`.

Since `Apollo` and `ApolloWebSocket` are 2 distincts products, I had to mark some methods as `public`.

The solution seem to fix our issue and fit our needs. But I'm happy to take any suggestions to make it part of the original repository too.

### DIFF
--- a/Playgrounds/ApolloMacPlayground.playground/Pages/Subscriptions.xcplaygroundpage/Contents.swift
+++ b/Playgrounds/ApolloMacPlayground.playground/Pages/Subscriptions.xcplaygroundpage/Contents.swift
@@ -31,7 +31,10 @@ let normalTransport = RequestChainNetworkTransport(interceptorProvider: DefaultI
 //: Next, set up the `WebSocketTransport` to talk to the websocket endpoint. Note that this may take a different URL, sometimes with a `ws` prefix, than your normal http endpoint:
 
 let webSocketURL = URL(string: "ws://localhost:8080/websocket")!
-let webSocketTransport = WebSocketTransport(request: URLRequest(url: webSocketURL))
+let webSocketTransport = WebSocketTransport(
+    request: URLRequest(url: webSocketURL),
+    store: store
+)
 
 //: Then, set up the split transport with the two transports you've just created:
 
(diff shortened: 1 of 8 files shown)


AnthonyMDev:
Thanks for the PR! I'll look into this ASAP!

AnthonyMDev:
So, right now the only place we are publishing results to the store is in the `CacheWriteInterceptor` which is part of the request chain. I like keeping the configuration and setup for this in a single place. Having multiple areas that call into the store can make maintenance more difficult.

I know that @designatednerd wanted to find a way to integrate the `WebSocketTransport` into the request chain. However we are actually planning on dramatically re-writing the request chain prior to shipping v1.0. But that's not going to be until later this year.

I think the long term solution is to integrate web sockets into the request chain, but in the interim, I think this solution is probably better than nothing. A lot of people have had a lot of pain points around web sockets not updating the cache.

Until we rework all of this and put a permanent fix into place, I'm inclined to accept this PR. In order to ensure we don't break things too much for people who are already implementing their own workarounds for the web sockets not updating the cache, I'd like to make this optional. (Once we refactor it for 1.0, I don't mind things breaking, as 1.0 is going to have a lot of breaking changes and people will need to migrate their codebase.)

@tgyhlsb would you be willing to update this PR with the following change:

- Make the `ApolloStore` parameter optional on the `WebSocketTransport` initializer, with a default value of `nil`.
- Only publish results to the store if the `ApolloStore` is passed to the `WebSocketTransport`.

I'm also not a huge fan of making the `publish` and `parseResult` functions `public`, but I think that is okay for now. We will address this when we rework the request chain.

Thanks so much for this contribution. This is going to make a LOT of people happy!

tgyhlsb:
Thanks for the feedback. 100% agree with the optional `ApolloStore`, I'll make the change now ;)

AnthonyMDev:
Thanks so much! I like having it in the tests, if it breaks anything (race conditions or something), these might help catch it?

Merging now!

AnthonyMDev:
Going to make some documentation additions here and then we will release a new version with this at some point in the near future.

PR #1892, opened 2021-08-04
Add docs and improve merging of records from WebSockets into cache.

AnthonyMDev:
The previous PR (#1889) has a few issues here. This fixes them.
- We shouldn't parse the result twice here. If the results are parsed with records, we don't need to also call `parseResultsFast()`.
- We should be passing the context identifier to the `publish()` function.
- Completion should wait for the cache merge to occur before being called.


calvincestari (approved):
Good catch on these changes - nice work.

PR #1901, opened 2021-08-11
Add execution tests for ApolloClient clearCache callback queue

calvincestari:
This introduces new tests to verify the callbacks on `clearCache` are being used as expected.

It is difficult to verify which queue is active through the `DispatchQueue` API so the test uses [setSpecific](https://developer.apple.com/documentation/dispatch/dispatchqueue/2883699-setspecific)/[getSpecific](https://developer.apple.com/documentation/dispatch/dispatchqueue/1780751-getspecific) to associate a key/value with the queue and validate in the completion callback - _thanks for the tip @designatednerd_.

PR #1904, opened 2021-08-11
Use the provided callback queue instead of the store's default.

### LINKED ISSUE #1900: ApolloClient doesn't use the provided callbackQueue in clearCache.
## Bug report

The `ApolloClient`'s implementation of `clearCache(callbackQueue:completion:)` from `ApolloClientProtocol` does not use the callbackQueue.

It is currently implemented like this:
```Swift
public func clearCache(callbackQueue: DispatchQueue = .main,
                       completion: ((Result<Void, Error>) -> Void)? = nil) {
  self.store.clearCache(completion: completion)
}
```

Which drops the callbackQueue without ever using it.

I would expect the implementation to be like this:
```Swift
public func clearCache(callbackQueue: DispatchQueue = .main,
                       completion: ((Result<Void, Error>) -> Void)? = nil) {
  self.store.clearCache(callbackQueue: callbackQueue, completion: completion)
}
```

## Versions

- `apollo-ios` SDK version: 0.45.0 (But wasn't changed in 0.46.0 from what I can see)
- Xcode version: All
- Swift version: All
- Package manager: SPM

## Steps to reproduce

Call
```Swift
ApolloStore.clearCache(callbackQueue:completion:)
```
with a callbackQueue that is not `.main`.

## Further details

If this is correct and I'm not missing anything I'm happy to submit a PR with my proposed change (or something different if I'm missing a better solution).

isaacressler:
This resolves #1900 by passing the provided callback queue in `ApolloClient.clearCache(callbackQueue:completion:)` into the `ApolloStore.clearCache(callbackQueue:completion:)` call.

@calvincestari has kindly provided the relevant tests for this in #1901, so I won't duplicate them here.

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -82,7 +82,7 @@ extension ApolloClient: ApolloClientProtocol {
 
   public func clearCache(callbackQueue: DispatchQueue = .main,
                          completion: ((Result<Void, Error>) -> Void)? = nil) {
-    self.store.clearCache(completion: completion)
+    self.store.clearCache(callbackQueue: callbackQueue, completion: completion)
   }
   
   @discardableResult public func fetch<Query: GraphQLQuery>(query: Query,


calvincestari (approved):

calvincestari:
Thanks for the contribution @isaacressler!

PR #1905, opened 2021-08-12
Removes playgrounds from apollo-ios

### LINKED ISSUE #1866: Move playground to its own repo
We want to move the Playground with examples of how to set things up for Apollo-iOS to its own repo for a couple of reasons: 

- Discoverability: Most people have no idea it exists. Having its own repo will help with visibility
- Decoupling: This will help decouple changes to the repo vs. changes to the playground. This will allow us to update the playground when a new version is released more easily, and not have a mix of released and unreleased changes in the playground. 
- Carthage: Carthage automatically builds every shared scheme in a given `xcodeproj` in the root of a repo. There is not a way within Carthage to exclude or include specific schemes. This is a real pain for the playground scheme since it builds all sub-libraries, and vastly bloats Carthage build time for our library.

calvincestari:
* Deletes the two playgrounds
* Deletes the scheme used to run the playgrounds
* Cleans up the release checklist re. playgrounds

_I deleted `ApolloCodegenFrontendPlayground` too because it was for internal purposes and in that case I think a TDD approach is better._

Closes #1866


designatednerd:
Probably worth stating that we're moving the playgrounds to another repo!

calvincestari:
I added a commit which updates the README linking out to the new playground repo.

designatednerd:
I'm *pretty* sure there's at least one reference to the playgrounds in the docs - I think if you search `playground` you should be able to find any refs

calvincestari:
Yup there was one, good catch! See a108c299.

designatednerd (approved):

PR #1906, opened 2021-08-13
Refactor websocket implementation directly into ApolloWebSocket

AnthonyMDev:
Due to dependency management conflicts in #1903, it has become easier for us to maintain our WebSockets as part of the `ApolloWebSockets` target instead of an external dependency on a forked version of Starscream.

This PR moves over the Starscream implementation directly into our target. I did a bit of refactoring and clean up to make the Starscream code make sense as a more integrated part of our websocket stack and to fix warnings for deprecated usage of unsafe Swift APIs.


Sources/ApolloWebSocket/Compression.swift
```diff
@@ -0,0 +1,183 @@
+//  Created by Joseph Ross on 7/16/14.
+//  Copyright © 2017 Joseph Ross.
+//  Modified by Anthony Miller & Apollo GraphQL on 8/12/21
+//
+//  This is a derived work derived from
+//  Starscream(https://github.com/daltoniam/Starscream)
```
designatednerd, line 6: Are you trying to set up markdown here or is there a space missing? 

(also good work playing by [the rules on the apache 2.0 license](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)))

Sources/ApolloWebSocket/WebSocketError.swift
```diff
@@ -24,6 +25,8 @@ public struct WebSocketError: Error, LocalizedError {
         return "Websocket error: Serialized message not found"
       case .neitherErrorNorPayloadReceived:
         return "Websocket error: Did not receive an error or a payload."
+      case .upgradeError:
+        return "Websocket error: Invalid HTTP upgrade."
```
designatednerd, line 29: Since we've got the code as part of the error is it worth at least putting the error code in this message?

Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -46,6 +46,7 @@ public class WebSocketTransport {
   }
   var socketConnectionState = Atomic<SocketConnectionState>(.disconnected)
 
+  /// Indicates if the websocket connection has been acknowledge by the server.
```
designatednerd, line 49:
```suggestion
  /// Indicates if the websocket connection has been acknowledged by the server.
```

Tests/ApolloServerIntegrationTests/StarWarsSubscriptionTests.swift
```diff
@@ -22,7 +21,7 @@ class StarWarsSubscriptionTests: XCTestCase {
     connectionStartedExpectation = self.expectation(description: "Web socket connected")
 
     webSocketTransport = WebSocketTransport(
-      websocket: DefaultWebSocket(
+      websocket: WebSocket(
```
designatednerd, line 24: Do we not still want to abstract this out for when we offer `URLSessionWebSocket` compatibility in the future?

Sources/ApolloWebSocket/WebSocketError.swift
calvincestari: Same could really be said for all the other errors too though - "Invalid HTTP upgrade" is quite specific. Not sold that the additional code is needed.

Sources/ApolloWebSocket/Compression.swift
calvincestari: This is present in all the attribution headers; if fixed here please do the others too.

Sources/ApolloWebSocket/Compression.swift
```diff
@@ -0,0 +1,183 @@
...
+//
+//  Compression implementation is implemented in conformance with RFC 7692 Compression Extensions
+//  for WebSocket: https://tools.ietf.org/html/rfc7692
+
+import Foundation
+import zlib
+
+class Decompressor {
+  enum Error: Swift.Error {
+    case resetFailed
```
calvincestari, line 19: What are your thoughts on distinguishing "decompression reset" from "compression reset"? Both errors are simply "reset". Yes within their own class but that also draws question of whether there should be a single error class. `Decompressor.decompressionFailed` and `Compressor.compressionFailed` state the same thing twice.

calvincestari:
This is nitpicking but it might be better to keep the Starscream files in their own sub-folder. If ApolloWebSocket grows in number of files it would be good to be able to isolate Starscream contributions.

Sources/ApolloWebSocket/Compression.swift
AnthonyMDev: Sure, I can change that.

Sources/ApolloWebSocket/Compression.swift
AnthonyMDev: I like scoping the error into the class that throws it. If I used the same error for both, It would have to be an error enum in the global scope. Only reason I did it this way really. @designatednerd opinions? I don't care strongly either way.

Tests/ApolloServerIntegrationTests/StarWarsSubscriptionTests.swift
AnthonyMDev: It is still injected, you just need something that conforms to the `WebSocketClient` protocol. Just changed the name of the "Default" one to remove "Default". 

I decided to hold off on moving the non-abstract parts of this into `Apollo` target until we do the networking layer refactor, because the API is going to likely change so drastically at that point anyways.

Sources/ApolloWebSocket/Compression.swift
designatednerd: I personally prefer keeping it scoped to the particular class because then your hands aren't tied by naming of other errors related to other classes. I also am strongly in favor of not rewriting the starscream stuff any more than we absolutely have to.

Tests/ApolloServerIntegrationTests/StarWarsSubscriptionTests.swift
designatednerd: OK sounds like a plan

Sources/ApolloWebSocket/WebSocketError.swift
designatednerd: fair enough

designatednerd:
Agreed on throwing the Starscream files into a folder for clarity.

calvincestari (approved):

AnthonyMDev:
I don’t think we should be naming the folder starscream. I don’t know why we are even moving them into a sub folder. I have already modified the starscream code modestly, and it’s now directly integrated as part of Apollo.

designatednerd:
@AnthonyMDev I disagree - we're taking it over, but I think it's really worth keeping "what is stuff we developed in-house" and "what is stuff we adopted from Starscream" very clearly delinated.

AnthonyMDev:
I’m very okay with acknowledging the contributions of Starscream that we are deriving from, but going forward, this is no longer “Starscream” code. We have responsibility and ownership over the future or it. And it’s not included using any abstractions or DI or via a separate package dependency.

AnthonyMDev:
@designatednerd Alright I hear you, but I’m not clear what actual value we get out of doing that.

I guess I expect that as this library evolves, the lines between the two are going to get blurred even more. When we implement native Apple Websockets, we will still likely be using part of the Starscream derived code, but it will become even more heavily modified. 

I just don’t see the benefit of separation here. It feels like we are creating an artificial abstraction.


Once we move the protocols and the WebSocketTransport into the base Apollo target (when we refactor the networking layer prior to 1.0), this Starscream derived code is going to become basically the entire target anyways.

AnthonyMDev:
I’d be more inclined to accept a folder for “DefaultImplementation” or something. But I  don’t think that it should be associated with Starscream in the project structure.

designatednerd:
I'm good with `DefaultImplementation` as well if that works for @calvincestari - much more concerned with separating out this existing implementation from any we create in the future.

PR #1911, opened 2021-08-17
Include missing Swift files for ApolloWebSocket in podspec

ketenshi:
Files in https://github.com/apollographql/apollo-ios/tree/main/Sources/ApolloWebSocket/DefaultImplementation were not part of the framework when used with Cocoapods.


AnthonyMDev:
Ooof good catch! Thank you for this! @calvincestari looks like we'll need to make a patch release anyways, so that Cocoapods CDN issue may be a moot point!

PR #1913, opened 2021-08-18
Remove root redirect and build deploy previews at /

trevorblades:
This branch removes the root-level redirect in the docs and forces deploy previews to be deployed at the root to account for this change.


AnthonyMDev:
Thanks @trevorblades! Anything else we need to do after merging this in?

trevorblades:
Nope, we're all good 👍  thanks for merging!

PR #1917, opened 2021-08-23
Use a failable iterator when selecting records

calvincestari:
The `prepare` function being used in `selectRawRows(forKeys:)` is [changing behaviour](https://github.com/apollographql/apollo-ios/issues/1658#issuecomment-901674890) such that errors during selection would be hidden. I think we can keep the expected error-on-iteration by switching to a row iterator that supports failure. This PR changes the internals of `selectRawRows(forKeys:)` to use `prepareRowIterator`, returning a `RowIterator` which conforms to `FailableIterator` therefore any error thrown would still be propagated out to the caller.


calvincestari:
I don't think there is any test required for this change. I initially wanted to ensure that `RowIterator.map` uses `failableNext` internally which throws errors but that's both the wrong thing to test and this is the wrong place for that test.

PR #1925, opened 2021-08-31
Bump macOS minimum deployment target to 10.15

### LINKED ISSUE #1868: Deprecated initialization of URLSessionTask
On `URLSessionClient():122` we are initializing a dummy `URLSessionTask` using a deprecated initializer. This is only being used when a request fails pre-flight. Should we make this function throwing or return an optional to solve this?

```swift
 @discardableResult
  open func sendRequest(_ request: URLRequest,
                        rawTaskCompletionHandler: RawCompletion? = nil,
                        completion: @escaping Completion) -> URLSessionTask {
    guard self.hasNotBeenInvalidated else {
      completion(.failure(URLSessionClientError.sessionInvalidated))
      return URLSessionTask()
    }
    
    let task = self.session.dataTask(with: request)
    let taskData = TaskData(rawCompletion: rawTaskCompletionHandler,
                            completionBlock: completion)
    
    self.tasks.mutate { $0[task.taskIdentifier] = taskData }
    
    task.resume()
    
    return task
  }
```

calvincestari:
**This PR is a DRAFT:**
* TODO: validation test for nil `URLSessionTask.sendRequest` return

This change is in preparation for being able to target SQLite.swift version 0.13.0 which requires a minimum of macOS 10.15.

Deprecations as a result of dropping 10.14:
* **Breaking:** `URLSessionTask.init` is no longer available so `URLSessionTask.sendRequest` now returns an optional task with `nil` being returned when the request could not be sent.

Closes #1868


Sources/ApolloWebSocket/DefaultImplementation/SSLSecurity.swift
```diff
@@ -157,9 +157,7 @@ open class SSLSecurity : SSLTrustValidator {
                 collect.append(SecCertificateCreateWithData(nil,cert as CFData)!)
             }
             SecTrustSetAnchorCertificates(trust,collect as NSArray)
-            var result: SecTrustResultType = .unspecified
-            SecTrustEvaluate(trust,&result)
-            if result == .unspecified || result == .proceed {
+            if SecTrustEvaluateWithError(trust,nil) {
```
calvincestari, line 160: This logic loses the `.unspecified` (default) behaviour but I think it's better now that we take action based only on the return. We don't gather any error because it was not being logged before and I don't see any good way in this module to output it in a way that is actually useful to a production app.

_`SecTrustEvaluateWithError(:,:)` is the synchronous replacement of `SecTrustEvaluate(:,:)`. It is now a simple boolean return._

Sources/ApolloWebSocket/DefaultImplementation/SSLSecurity.swift
```diff
@@ -205,8 +203,8 @@ open class SSLSecurity : SSLTrustValidator {
         SecTrustCreateWithCertificates(cert, policy, &possibleTrust)
 
         guard let trust = possibleTrust else { return nil }
-        var result: SecTrustResultType = .unspecified
-        SecTrustEvaluate(trust, &result)
+        _ = SecTrustEvaluateWithError(trust, nil)
```
calvincestari, line 206: The result was not being used here so we don't store it either with the new call. Again, no error was being reported perviously.

_`SecTrustEvaluateWithError(:,:)` is the synchronous replacement of the deprecated `SecTrustEvaluate(:,:)`. It is now a simple boolean return._

calvincestari:
This is on hold while #1943 is evaluated.

calvincestari:
I'm closing this PR since #1943 has been accepted. I debated pulling https://github.com/apollographql/apollo-ios/pull/1925/commits/1df73419d92006a4b614776e9a2d2ce9573dfc31 into #1940 but since all apps implementing apollo-ios must support iOS 12 there is no real need to break behaviour that existing users might be relying on; empty task vs. nil task.

OrindaCoder:
@calvincestari My company's app has iOS 14 as its deployment target, so it's getting a deprecation warning on `return URLSessionTask()`. It sounds like you can't remove this because iOS 12, but would it be possible to use alternate code for higher iOS versions? I'd be happy to raise a PR if that would be helpful.

calvincestari:
Hi @OrindaCoder 👋🏻 - correct, we still support iOS 12 which is why the code still uses that deprecated `URLSessionTask` initializer. We intend to keep iOS 12 as the minimum for the upcoming 1.0 release but 2.0 will certainly bump that to iOS 13 or higher; see details in our [Roadmap](https://github.com/apollographql/apollo-ios/blob/main/ROADMAP.md).

The solution here could be to simply merge the changes in 1df7341 into main as I [mentioned above](https://github.com/apollographql/apollo-ios/pull/1925#issuecomment-923278438). We're still in 0.x releases so making that kind of breaking change now is somewhat acceptable. If you have an alternate solution though please feel free to submit a PR and we can certainly review and discuss.

OrindaCoder:
@calvincestari I will wait for a future version of Apollo that fixes this warning. Thanks for all the work you do!

PR #1935, opened 2021-09-10
Remove dependency on apollo CLI for fetching schema

### LINKED ISSUE #1823: Remove dependency on apollo CLI for fetching schema
As part of preparation for ditching the Typescript CLI altogether, we want to move our `ApolloSchemaDownload` tool to using pure swift rather than calling into the Typescript CLI.

calvincestari:
This change removes a dependency on the Apollo CLI for schema downloading. Schema downloads are now Swift-based and maintainable/extensible within apollo-ios.

#### Notable changes:
* Abstracted `URLDownloader` from `CLIDownloader`; used in both CLI and schema downloading. _CLI downloading will be removed along with the Swift codegen implementation_.
* Registry schema requests are done via `UntypedGraphQLRequestBodyCreator` - _a non-type-safe request creator to facilitate sending requests not using code generation_.
* Refactored `ApolloSchemaOptions`
  * Renamed to `ApolloSchemaDownloadConfiguration` for a more specific naming context.
  * Schema output is always SDL.
  * Defaults to the current registry variant.
* Renamed `ApolloSchemaDownloader` APIs
  * `fetch(with:)` is more expressive than `run(options:)`.

Closes #1823


Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
```diff
@@ -13,7 +13,8 @@
     "npm": ">=7"
   },
   "dependencies": {
-    "graphql": "file:./vendor/graphql-16.0.0-pre+modern-js.tgz"
+    "graphql": "file:./vendor/graphql-16.0.0-pre+modern-js.tgz",
```
calvincestari, line 16: Should this and the associated file be updated to the latest 16.0.0 release? Right now that would be `v16.0.0-rc.2`.

Tests/ApolloCodegenTests/CLIDownloaderTests.swift
```diff
@@ -1,56 +1,45 @@
...
   }
   
-  func testTimeoutThrowsCorrectError() throws {
+  func testTimeout_shouldThrowCorrectError() throws {
     let scriptsURL = CodegenTestHelper.cliFolderURL()
     
-    // This file is big enough that unless both you and the server have a terabyte connection, 2 seconds won't be enough time to download it.
     do {
-      try CLIDownloader.forceRedownload(cliFolderURL: scriptsURL, timeout: 2.0)
+      try CLIDownloader.forceRedownload(to: scriptsURL, timeout: 0.5)
```
calvincestari, line 32: Time savings none-the-less.

Apollo.xcodeproj/project.pbxproj
```diff
@@ -248,6 +250,9 @@
 		DED46051261CEAD20086EF63 /* StarWarsAPI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9FCE2CFA1E6C213D00E34457 /* StarWarsAPI.framework */; };
 		E616B6D126C3335600DB049E /* ExecutionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E616B6D026C3335600DB049E /* ExecutionTests.swift */; };
 		E61DD76526D60C1800C41614 /* SQLiteDotSwiftDatabaseBehaviorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E61DD76426D60C1800C41614 /* SQLiteDotSwiftDatabaseBehaviorTests.swift */; };
+		E69C255026EC2A9D00E9495D /* ApolloCodegenLib.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9B7B6F47233C26D100F32205 /* ApolloCodegenLib.framework */; };
```
calvincestari, line 253: `ApolloCodegenLib` was linked to the `ApolloPerformanceTests` target because `MockNetworkSession` is part of the `ApolloTestSupport` bundle which is a dependency of `ApolloPerformanceTests`.

Tests/ApolloCodegenTests/URLDownloaderTests.swift
```diff
@@ -0,0 +1,131 @@
...
+    downloader = URLDownloader()
+    session = nil
+  }
+
+  override func tearDown() {
+    downloader = nil
+    session = nil
+  }
+
+  private func setRequestHandler(statusCode: Int, data: Data? = nil, error: Error? = nil, abandon: Bool = false) {
```
calvincestari, line 23: The unfortunate side-effect is that if you do not call `setRequestHandler` in one of the tests it'll fail with an obscure `Could not connect to the server.` error because `URLDownloader` defaults to using the `URLSession` extension for `NetworkService`. Ideally it would tell you that you had not set a request handler but I couldn't find a great way to do that.

Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
designatednerd: @martijnwalraven probably knows

Tests/ApolloCodegenTests/CLIDownloaderTests.swift
designatednerd: FWIW I think I set this to 2 because sometimes it'd time out on CI otherwise

Apollo.xcodeproj/project.pbxproj
designatednerd: I think it should be in `ApolloTestSupport`.

Tests/ApolloCodegenTests/URLDownloaderTests.swift
designatednerd: I would throw a comment on this method explaining that it needs to be called in each of the test methods then

Tests/ApolloCodegenTests/CLIDownloaderTests.swift
calvincestari: This is testing a timeout error though, so that's exactly what we want. Could even go lower probably, i.e.: fail fast.

Tests/ApolloCodegenTests/URLDownloaderTests.swift
calvincestari: A bit more thinking and I found the solution I was looking for - 3d7efb146. If you forget you'll now get this error message to guide you, `failed - You must call setRequestHandler before using downloader!`.

Sources/ApolloCodegenLib/UntypedGraphQLRequestBodyCreator.swift
```diff
@@ -0,0 +1,54 @@
+import Foundation
+
+class UntypedGraphQLRequestBodyCreator {
+  
+  /// A non-type-safe request creator to facilitate sending requests not using code generation.
+  ///
+  /// Note: This is only public for use in `ApolloCodegenLib`, we do **not** recommend or support direct use of this in client applications.
```
calvincestari, line 7: I question whether we need this comment:
* the class is not public
* developers don't import `ApolloCodegenLib` into their projects and even if they did this class has `internal` access.

If there is concern about this why don't we move the class into `ApolloSchemaDownloader.swift` with `fileprivate` access?

designatednerd: Yep I had to back this out from being in the codegen lib because of cross-compilation issues. This whole comment can go away.

Tests/ApolloCodegenTests/CLIDownloaderTests.swift
designatednerd: 🤔 hmm...I suppose that's fair, but I have a sneaking feeling setting it too fast might hide other errors. I guess we'll jump off that bridge when we get to it.

Tests/ApolloServerIntegrationTests/SchemaRegistryApolloSchemaDownloaderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+import ApolloTestSupport
+import ApolloCodegenTestSupport
+@testable import ApolloCodegenLib
+
+class SchemaRegistryApolloSchemaDownloaderTests: XCTestCase {
+  func testDownloadingSchema_fromSchemaRegistry_shouldOutputSDL() throws {
+    let testOutputFolderURL = CodegenTestHelper.outputFolderURL()
+    XCTAssertFalse(FileManager.default.apollo.fileExists(at: testOutputFolderURL))
+
+    guard let apiKey = ProcessInfo.processInfo.environment["REGISTRY_API_KEY"] else {
```
designatednerd, line 12: What's the difference between this and the other test that uses `REGISTRY_API_KEY`, additional specificity?

calvincestari: I moved the two schema downloading tests out of `ApolloSchemaTests.swift` and the `ApolloCodegenTests` target because:
* The starwars introspection download test would have required the test to spin up the starwars server. Since that was already happening in this target (`ApolloServerIntegrationTests`) I felt it was better suited to live here.
* That made me bring across the registry-based schema download test to group tests that rely on external services - no other reason. So the 'other' registry download test isn't there anymore.
* I added [a schema-based test that uses static JSON](https://github.com/apollographql/apollo-ios/pull/1935/commits/75b72785c477995a3020a3cf9e44c5e6811ac336#diff-76997a5d08e3f423718fcbb13c8baaf45ad038ac8938936781856373b3df5db3R66) instead so it didn't need to actually download anything.

Apollo.xcodeproj/project.pbxproj
calvincestari: It is in the `ApolloCodegenTestSupport` target.

Tests/ApolloServerIntegrationTests/SchemaRegistryApolloSchemaDownloaderTests.swift
designatednerd: ah right that all got moved while I was working on all this. thank you!

Sources/ApolloCodegenLib/ApolloCLI.swift
```diff
@@ -20,11 +20,11 @@ public struct ApolloCLI {
     let lock = try waitForCLIFolderLock(cliFolderURL: cliFolderURL, timeout: timeout)
     defer { lock.unlock() }
 
-    try CLIDownloader.downloadIfNeeded(cliFolderURL: cliFolderURL, timeout: timeout)
+    try CLIDownloader.downloadIfNeeded(to: cliFolderURL, timeout: timeout)
```
AnthonyMDev, line 23: 👍

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
```diff
@@ -0,0 +1,90 @@
...
+
+/// A configuration object that defines behavior for schema download.
+public struct ApolloSchemaDownloadConfiguration {
+  
+  /// How to attempt to download your schema
+  public enum DownloadMethod: Equatable {
+
+    case registry(_ settings: RegistrySettings)
+    ///   - endpointURL: The endpoint to hit to download your schema.
+    case introspection(endpointURL: URL)
```
AnthonyMDev, line 11: Can we document these two options and what they mean. I'm not even sure I know how to use a registry. For new GraphQL users, they may not even know what a registry IS.

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
```diff
@@ -0,0 +1,90 @@
...
+
+  /// Designated Initializer
+  ///
+  /// - Parameters:
+  ///   - downloadMethod: How to download your schema.
+  ///   - downloadTimeout: The maximum time to wait before indicating that the download timed out, in seconds. Defaults to 30 seconds.
+  ///   - headers: [optional] Any additional headers to include when retrieving your schema. Defaults to nil
+  ///   - outputFolderURL: The URL of the folder in which the downloaded schema should be written
+  ///   - schemaFilename: The name, without an extension, for your schema file. Defaults to `"schema"
+  public init(using downloadMethod: DownloadMethod,
```
AnthonyMDev, line 69: For me, I would probably put the documentation above each of the actual properties on lines 56-59. Do we want to duplicate the same docs in both places? Is it better to have it on the initializer than the properties? I feel like I want to be able to see that info on the properties when I look at them also to understand what's going on.

@designatednerd What is the best way for us to handle documentation in situations like this. As far as the generated documentation, is there a preferred method for this?

Sources/ApolloCodegenLib/UntypedGraphQLRequestBodyCreator.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+
+class UntypedGraphQLRequestBodyCreator {
```
AnthonyMDev, line 3: If this is only being used for static functions and is never going to be instantiated, the best way to do that is to make it an enum with no cases. (enum or struct are better for performance than classes also)

I feel like we shouldn't be duplicating this code in both targets. If it ever needs to change for bug fixes, changes should be done in one place. Can we move this object into `ApolloUtils`, then have the "typed" `RequestBodyCreator` in `Apollo` call into this function? Like, if it has the operation, it can just pass all the properties from the operation into this function and it should work just right still, yeah?

AnthonyMDev (approved):
Great job on this! So glad to have this finished, and it looks great. Just a couple of suggestions.

One question: Do you know why the bundled ApolloCodegenFrontend.bundle.js changed? Did we need to make changes to the js library?

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
designatednerd: The problem with doing it above the properties and not here is that then the generated docs don't get info about the properties. It's probably a good, if very duplicative, idea to have the docs on the properties too though

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
designatednerd: The registry is the Apollo registry. This *might* be one case where we could put `ApolloRegistrySettings` in for clarity

Sources/ApolloCodegenLib/UntypedGraphQLRequestBodyCreator.swift
designatednerd: ...maybe. I know I tried something like this like two months ago and for some reason it wasn't working, and now i can't remember why

designatednerd:
> Do you know why the bundled ApolloCodegenFrontend.bundle.js changed? Did we need to make changes to the js library?

Yeah there was one method I had to expose for moving between JSON and SDL

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
AnthonyMDev: Yeah, that's what I thought. I think we should duplicate it in both places then.

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
AnthonyMDev: Yeah, the Registry is a service provided by Apollo as a company, I think it makes sense to refer to it as apollo registry. 

I'm thinking the `RegistrySettings` is fine to stay as it. I'd actually prefer to rename the enum case to `case apolloRegistry(_ settings: RegistrySettings)`

We should still document this stuff a bit better.

calvincestari: Good point on the "Apollo" prefix. I'll add that and add some documentation for the two values.

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
calvincestari: It's needed in both places; generated docs and at declaration site for Xcode. I'll duplicate them.

calvincestari:
> > Do you know why the bundled ApolloCodegenFrontend.bundle.js changed? Did we need to make changes to the js library?
> 
> Yeah there was one method I had to expose for moving between JSON and SDL - I'm not quite seeing where the actual code went, though. Not sure if it got eaten in a merge or what

This is the exact diff - `e.printSchema=function e(t){return e()},`. Goes together with [this](https://github.com/apollographql/apollo-ios/pull/1935/files#diff-5cedbe3ff6ea3fef450107247209068eb8b74b11e56bfdc83f35208686fd1c26R73) and [this](https://github.com/apollographql/apollo-ios/pull/1935/files?file-filters%5B%5D=.js&file-filters%5B%5D=.map&file-filters%5B%5D=.pbxproj&file-filters%5B%5D=.swift&file-filters%5B%5D=.ts&file-filters%5B%5D=dotfile#diff-f494310596d1296d51ccb6dba06303718a5fa041416458babf948e40bdc0c595R11).

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
calvincestari: It should make more sense for new GraphQL users with 7b0c7779.

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
calvincestari: I updated a few other places with documentation too, in 7fffa63f.

Sources/ApolloCodegenLib/UntypedGraphQLRequestBodyCreator.swift
calvincestari: I had a crack at this and after talking with @AnthonyMDev we decided it was not worth it; maybe the same hassles you experienced @designatednerd.

_If you're interested to see what changes were needed, you can [take a look at the diff](https://gist.github.com/calvincestari/b5066f59079762f326b93ec125794e1c)._
* _Having to use mapValue ([here](https://gist.github.com/calvincestari/b5066f59079762f326b93ec125794e1c#file-untypedgraphqlrequestbodycreator-diff-L127) and [here](https://gist.github.com/calvincestari/b5066f59079762f326b93ec125794e1c#file-untypedgraphqlrequestbodycreator-diff-L132)) is not great, particularly since this would be called for each typed request too_
* _I also didn't like having to expose `UntypedGraphQLRequestBodyCreator` as `public`_

AnthonyMDev (approved):
LGTM!

Sources/ApolloCodegenTestSupport/MockNetworkSession.swift
```diff
@@ -0,0 +1,24 @@
+@testable import ApolloCodegenLib
+
+public final class MockNetworkSession: NetworkSession {
+  let statusCode: Int
+  let data: Data?
+  let error: Error?
+  let abandon: Bool
+
+  public init(statusCode: Int, data: Data? = nil, error: Error? = nil, abandon: Bool = false) {
```
designatednerd, line 9: What is the `abandon` parameter here?

designatednerd (approved):
One question re testing, otherwise this is :shipit:

Sources/ApolloCodegenTestSupport/MockNetworkSession.swift
calvincestari: `abandon` is used to mock a timeout error. Idea being to [exit early](https://github.com/apollographql/apollo-ios/pull/1935/files/6216e332256f63d2cf92a40da14480f9b25e8058#diff-aa32e38e100398072d34522c78690f7e3d530e6d3ad6ba6999a4507c1e553ddcR17) and never have the completion block called = response took longer than the timeout value used in the download configuration.

Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
calvincestari: I'm going to leave this as-is for now. I think we should revisit the question once graphql-js 16.0.0 is officially released.

PR #1941, opened 2021-09-16
Add Pull Request templates

calvincestari:
* I'd like to get this in for release PRs so I don't always have to refer back to the release checklist.
* Regular PRs are still an empty canvas.

PR #1942, opened 2021-09-17
Fix release process Pull Request template

calvincestari:
After wondering why the PR template I created earlier in #1941 was not working I discovered PR templates are not as convenient as Issue templates.
* There is no selector list available when creating a new PR.
* Multiple PR templates are only accessible by appending the `template=` query param.
* It is also not possible to create a static link such as https://github.com/apollographql/apollo-ios/compare/?template=pull-request-template-release.md because as soon as the compare branches are selected the `template=` param is discarded.

This PR adds a note to the original release checklist with the query param that can be manually added to the URL when creating a release PR. I still believe there is value in having this as a template even if it requires a manual step at the moment. I'm hoping GitHub improves this experience soon.


AnthonyMDev (commented):
Lame, but okay. lol

Issue #1961, opened 2021-09-23
Struct initialiser for ApolloSchemaDownloadConfiguration.HTTPHeader is internal


nsivertsen:
## Bug report

The initialiser for the struct `ApolloSchemaDownloadConfiguration.HTTPHeader` is internal, which means that custom HTTP headers cannot be added to a `ApolloSchemaDownloadConfiguration`.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.49.0
- Xcode version: Version 13.0 beta 5
- Swift version: 5.5.0
- Package manager: Swift Package Manager

## Steps to reproduce

Add the [iOS Codegen Template](ApolloSchemaDownloadConfiguration.HTTPHeader) to a project and attempt to define and pass custom HTTP headers for the `downloadSchema` command:

```swift
let headers = [
  ApolloSchemaDownloadConfiguration.HTTPHeader(key: "Authorization", value: "Bearer MY_TOKEN"),
]

let schemaDownloadOptions = ApolloSchemaDownloadConfiguration(using: .introspection(endpointURL: endpoint),
                                                              headers: headers,
                                                              outputFolderURL: folderForDownloadedSchema)

```

Running the `downloadSchema` command then gives this error:

```
error: 'ApolloSchemaDownloadConfiguration.HTTPHeader' initializer is inaccessible due to 'internal' protection level
```

calvincestari:
The fix for this will be released in `0.49.1` shortly.

PR #1962, opened 2021-09-23
Make initializer for ApolloSchemaDownloadConfiguration.HTTPHeader public

nsivertsen:
Fixes https://github.com/apollographql/apollo-ios/issues/1961

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
@@ -54,6 +54,11 @@ public struct ApolloSchemaDownloadConfiguration {
     public var debugDescription: String {
       "\(key): \(value)"
     }
+
+    public init(key: String, value: String) {
+      self.key = key
+      self.value = value
+    }
   }
 
   /// How to download your schema. Supports the Apollo Registry and GraphQL Introspection methods.
--- a/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
@@ -62,13 +62,13 @@ public struct ApolloSchemaDownloadConfiguration {
   }
 
   /// How to download your schema. Supports the Apollo Registry and GraphQL Introspection methods.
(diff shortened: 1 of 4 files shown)


designatednerd (approved):
Guh, I always forget that you have to do this even for public structs. Thank you - if you can sign the CLA we can get this merged.

AnthonyMDev:
Thanks so much for this fix @nsivertsen! Excited to merge this in once the CLA is signed!

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
```diff
@@ -54,16 +54,21 @@ public struct ApolloSchemaDownloadConfiguration {
...
+
+    public init(key: String, value: String) {
+      self.key = key
+      self.value = value
+    }
   }
 
   /// How to download your schema. Supports the Apollo Registry and GraphQL Introspection methods.
-  let downloadMethod: DownloadMethod
+  public let downloadMethod: DownloadMethod
```
calvincestari, line 65: There's no reason to keep these private; once you've created the struct you should be able to inspect the property values.

Tests/ApolloCodegenTests/ApolloSchemaPublicTests.swift
```diff
@@ -1,9 +1,9 @@
 import XCTest
 import ApolloTestSupport
 import ApolloCodegenTestSupport
-@testable import ApolloCodegenLib
+import ApolloCodegenLib
```
calvincestari, line 4: Changed how this test imports `ApolloCodegenLib` to ensure `public` access is caught in further tests.

calvincestari (approved):

AnthonyMDev (approved):

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
designatednerd: True for as long as they remain `let`s - If they're `var`s we may or may not want to allow direct access, although we could handle that with a `private(set)`.

designatednerd (approved):
SLGTM

PR #1964, opened 2021-09-23
Refactor GraphQLOperation definition models

AnthonyMDev:
I want use to allows for code gen to generate only the required data when using APQs.

Some users have expressed the desire to use only "allowlisted" operations and they don't want the code generation in that case to expose the entire query definition in the compiled binary. The query document was required before, now you can set it up to use different methods.

### DIFF
--- a/Sources/Apollo/RequestBodyCreator.swift
+++ b/Sources/Apollo/RequestBodyCreator.swift
@@ -9,13 +9,11 @@ public protocol RequestBodyCreator {
   ///
   /// - Parameters:
   ///   - operation: The operation to use
-  ///   - sendOperationIdentifiers: Whether or not to send operation identifiers. Should default to `false`.
   ///   - sendQueryDocument: Whether or not to send the full query document. Should default to `true`.
   ///   - autoPersistQuery: Whether to use auto-persisted query information. Should default to `false`.
   /// - Returns: The created `GraphQLMap`
   func requestBody<Operation: GraphQLOperation>(
     for operation: Operation,
-    sendOperationIdentifiers: Bool,
     sendQueryDocument: Bool,
     autoPersistQuery: Bool
   ) -> JSONEncodableDictionary
@@ -27,7 +25,6 @@ extension RequestBodyCreator {
   
   public func requestBody<Operation: GraphQLOperation>(
     for operation: Operation,
-    sendOperationIdentifiers: Bool,
     sendQueryDocument: Bool,
     autoPersistQuery: Bool
   ) -> JSONEncodableDictionary {
@@ -37,14 +34,6 @@ extension RequestBodyCreator {
 
     if let variables = operation.variables {
       body["variables"] = variables.jsonEncodableObject
-    }
-
-    if sendOperationIdentifiers {
-      guard let operationIdentifier = operation.operationIdentifier else {
-        preconditionFailure("To send operation identifiers, Apollo types must be generated with operationIdentifiers")
-      }
-
-      body["id"] = operationIdentifier
     }
 
     if sendQueryDocument {
--- a/Sources/Apollo/RequestBodyCreator.swift
+++ b/Sources/Apollo/RequestBodyCreator.swift
@@ -5,13 +5,13 @@ import ApolloUtils
 #endif
 
 public protocol RequestBodyCreator {
-  /// Creates a `GraphQLMap` out of the passed-in operation
+  /// Creates a `JSONEncodableDictionary` out of the passed-in operation
   ///
   /// - Parameters:
   ///   - operation: The operation to use
   ///   - sendQueryDocument: Whether or not to send the full query document. Should default to `true`.
   ///   - autoPersistQuery: Whether to use auto-persisted query information. Should default to `false`.
-  /// - Returns: The created `GraphQLMap`
+  /// - Returns: The created `JSONEncodableDictionary`
   func requestBody<Operation: GraphQLOperation>(
     for operation: Operation,
     sendQueryDocument: Bool,
@@ -37,7 +37,10 @@ extension RequestBodyCreator {
     }
 
     if sendQueryDocument {
-      body["query"] = operation.queryDocument
+      guard let document = operation.definition?.queryDocument else {
+        preconditionFailure("To send query documents, Apollo types must be generated with `OperationDefinition`s.")
+      }
+      body["query"] = document
     }
 
     if autoPersistQuery {
--- a/Sources/ApolloAPI/GraphQLNullable.swift
+++ b/Sources/ApolloAPI/GraphQLNullable.swift
@@ -1,5 +1,12 @@
 import Foundation
 
+/// Indicates the presence of a value, supporting both `nil` and `null` values.
+///
+/// In GraphQL, explicitly providing a `null` value for an input value to a field argument is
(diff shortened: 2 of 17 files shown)


Sources/Apollo/RequestBodyCreator.swift
```diff
@@ -39,16 +36,11 @@ extension RequestBodyCreator {
...
-        preconditionFailure("To send operation identifiers, Apollo types must be generated with operationIdentifiers")
-      }
-
-      body["id"] = operationIdentifier
-    }
-
     if sendQueryDocument {
-      body["query"] = operation.queryDocument
+      guard let document = operation.definition?.queryDocument else {
+        preconditionFailure("To send query documents, Apollo types must be generated with `OperationDefinition`s.")
```
designatednerd, line 41: why `preconditionFailure` instead of just `fatalError`, out of curiosity?

Sources/ApolloAPI/GraphQLNullable.swift
```diff
@@ -1,5 +1,12 @@
 import Foundation
 
+/// Indicates the presence of a value, supporting both `nil` and `null` values.
+///
+/// In GraphQL, explicitly providing a `null` value for an input value to a field argument is
+/// semantically different from not providing a value at all (`nil`). This enum allows you to
+/// distinguish your input values between `null` and `nil`.
```
designatednerd, line 7: I would add something specific here about how sending a null value generally clears any existing value but not sending a value at all does nothing

Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -6,33 +6,92 @@ public enum GraphQLOperationType {
...
 
+/// The means of providing the operation document that includes the definition of the operation
+/// over network transport.
+///
+/// This data represents the `Document` as defined in the GraphQL Spec.
+/// - See: [GraphQLSpec - Document](https://spec.graphql.org/draft/#Document)
+///
+/// The Apollo Code Generation Engine will generate the `DocumentType` on each generated
+/// `GraphQLOperation`. You can change the type of `DocumentType` generated in your
+/// [code generation configuration](// TODO: ADD URL TO DOCUMENTATION HERE).
```
designatednerd, line 17: should this be updated to a warning-todo, or are you going to fix this before the PR is merged?

Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -6,33 +6,92 @@ public enum GraphQLOperationType {
...
+      return definition
+    default: return nil
+    }
   }
 
-  var variables: Variables? {
-    return nil
+  var operationIdentifier: String? {
+    switch self.document {
+    case let .automaticallyPersisted(identifier, _), let .persistedOperationsOnly(identifier):
```
designatednerd, line 91: I generally prefer one case per line here since it's easier to parse while reading and also makes changes more obvious in a diff. Not a hill I'm gonna die on though

Sources/ApolloTestSupport/MockOperation.swift
```diff
@@ -5,12 +5,8 @@ open class MockOperation<SelectionSet: RootSelectionSet>: GraphQLOperation {
 
   public let operationType: GraphQLOperationType
 
-  public var operationDefinition: String = "None"
-  public var operationIdentifier: String?
   public var operationName: String = "MockOperationName"
-
-  public var stubbedQueryDocument: String?
-  public final var queryDocument: String { stubbedQueryDocument ?? operationDefinition }
+  public var document: DocumentType = .notPersisted(definition: .init("None"))
```
designatednerd, line 9: I would make this something like "Mock" rather than preserving the existing "None" for clarity

Sources/Apollo/RequestBodyCreator.swift
AnthonyMDev: This is actually a failure based on the precondition of you telling the code gen configuration that they should generate the operation definitions. This isn't a regular run time failure based on bad inputs. The way to fix this is to change your config and re-run Code Generation.

Sources/ApolloAPI/GraphQLNullable.swift
AnthonyMDev: So, that isn't actually enforced or even specifically stated as part of the spec for how `null` values are used. 
In the GraphQL spec that is linked in this documentation all that is guaranteed is that a `null` value "may be interpreted differently" than a `nil`. It also very often is NOT any different. That's up to your server implementation details.

The spec docs do use clearing an existing value as the primary example of how these things "could" be semantically different, but I don't think that it's appropriate to include that here. I don't want any confusion where people think that **IS** the difference here and they can just pass `null` and assume their server will delete fields. 

I generally really like pointing to the GraphQL spec and letting engineers learn how to actually use GraphQL.

AnthonyMDev: @calvincestari @hwillson What are your opinions here? If the team thinks adding that example to the documentation is better, I will.

I generally have a documentation style that is very technical. It is 100% accurate and direct. But it is ONLY accurate and direct to exactly what the code is doing. I try not to make generalizations or assume the reader has any other knowledge or is using my code in any specific use case.

I feel like that kind of commentary is better left to tutorials and usage guides, not the inline-documentation. But I am fine with adding it here if you would like.

Sources/ApolloAPI/GraphQLOperation.swift
AnthonyMDev: I'll fix this before releasing 1.0. My release 1.0 check-list includes running a find for "TODO" and cleaning them all up :)

AnthonyMDev: I use warning todos a lot too, but I won't be able to fix this one until after we have documentation for the Code Gen Engine itself, so that will be a while.

Sources/ApolloAPI/GraphQLOperation.swift
AnthonyMDev: Fixing this now.

Sources/ApolloTestSupport/MockOperation.swift
AnthonyMDev: Fair enough. Will do!

Sources/ApolloAPI/GraphQLNullable.swift
hwillson: Great suggestion - I agree that an example might be better suited in our non-API docs, especially in this case where things aren't enforced by the spec.

designatednerd: It's tricky if it's not part of the spec, but I feel like it's probably worth giving an example of *why* this needs to exist - otherwise people may be confused as to why we're using this instead of optionals

AnthonyMDev: I also added an explanation of this, with the example to the RFC. (I very much expect that large portions of the RFC will be taken whole sale for the docs site for the release.)

calvincestari: Examples certainly help to illustrate concepts and since we're dealing with Swift, where nil and optional are concepts readers will already be familiar with, I think I'd err on the side of caution and go with the example here so the reader doesn't infer their own understanding of what it these may mean.

PR #1965, opened 2021-09-24
Change introspection_response.json back to schema.json

acoomans:
Using `introspection_response.json` instead of `schema.json` broke assumptions in the code base, including the https://github.com/apollographql/iOSCodegenTemplate.

There's only a few instances of `introspection_response.json` in the codebase (https://github.com/apollographql/apollo-ios/search?q=introspection_response) but a lot of instances of `schema.json` (https://github.com/apollographql/apollo-ios/search?q=schema.json), so it makes sense to revert it back to `schema.json`


designatednerd:
Hi! This was changed on purpose since we're trying to move to using SDL for downloaded schemas since it's a) Way smaller, file-size wise, and b) Way easier to read.

I think our preference would be to update the codegen template, but can you be more specific about what broke here for you?

acoomans:
Ok sounds like maybe the template needs some changes instead. Let's discuss over there https://github.com/apollographql/iOSCodegenTemplate/pull/7

PR #1969, opened 2021-09-27
Added InputObject and RFC documentation

AnthonyMDev:


Sources/StarWarsAPI/APIv1.swift
```diff
@@ -104,7 +104,7 @@ public enum SearchResult: UnionType, Equatable {
 }
 
 /// The episodes in the Star Wars trilogy
-public enum Episode: String, CaseIterable, Apollo.JSONDecodable, Apollo.JSONEncodable {
+public enum Episode: String EnumType {
```
designatednerd, line 107: missing a comma here

AnthonyMDev: Thanks!

AnthonyMDev: Will fix in next PR. This code isn't even really compiling anywhere right now, just experimenting with how parts of the StarWars API should look under new code gen

PR #1970, opened 2021-09-27
Added InputObject, GraphQLEnum and EnumType to RFC

AnthonyMDev:


CodegenProposal.md
```diff
@@ -338,13 +338,142 @@ A `SchemaConfiguration` object will also be generated for your schema. This obje
...
 # `InputObject` Generation
 
-TODO
+Input objects will be generated for each `input` type in the schema that is used in an argument for any of the operations defined in your application. Input objects are structs that are backed by a `InputDict` struct that stores the values for the fields on the input object in a dictionary. This allows for `InputObject`s to be treated as values types but use copy-on-write semantics under the hood.
 
-# `EnumType` Generation
+Nullable fields on input objects are represented using `GraphQLNullable` to allow for both `null` and `nil` values. 
 
-TODO
+Following the [Input Coercion rules](https://spec.graphql.org/draft/#sec-Input-Objects.Input-Coercion) from the GraphQL spec, the server defined default value for a field will be used when passing `nil`. For non-nullable fields, if the schema provides a default value, the field will be represented as an optional to allow for this. Nullable fields on input objects are represented using `GraphQLNullable` to allow for both `null` and `nil` values.
```
designatednerd, line 406: One thing probably worth clarifying here: We're not going to send what's in the schema at the time the code was generated, we just won't send anything, which will cause the current default on the server to be used. 

This is useful if the default value on the server side has changed since the code was originally generated

CodegenProposal.md
```diff
@@ -338,13 +338,142 @@ A `SchemaConfiguration` object will also be generated for your schema. This obje
...
+struct MyInput: InputObject {
+  public private(set) var dict: InputDict
+
+  init(size: RelativeSize) { ... }
+  
+  var size: RelativeSize { ... }
+}
+```
+Non-nullable field with a default value:
+```
```
designatednerd, line 462: probably worth tagging some of these operation and schema examples with `graphql` so it renders prettier

AnthonyMDev: I don't think that graphql works in GitHub markdown, does it? I will add that!

CodegenProposal.md
AnthonyMDev: Good point. I'll add this now, thanks.

CodegenProposal.md
designatednerd: I believe that it does

AnthonyMDev: You are correct! This is done and merged into the release 1.0 branch now!

PR #1975, opened 2021-09-29
Refactor codegen configuration

calvincestari:
This change is to reduce the number of parameters for codegen configuration and increase readability of the configuration through enums instead of bools.

Below is a summary of the changes between `ApolloCodegenConfiguration` and `ApolloCodegenOptions` since the diff is confused with the file rename:

**Changed:**
* `includes` - moved to `FileInput.glob`
* `namespace` - moved to `FileOutput.SchemaTypesFileOutput.name`
* `omitDeprecatedEnumCases` - renamed to `deprecatedEnumCases` with value in `CompositionOption`
* `operationIDsURL` - moved to `FileInput.operationIDs`
* `outputFormat` - moved to `FileOutput.schemaTypes` and `FileOutput.operations`. _Single file output is no longer supported._
* `suppressSwiftMultilineStringLiterals` - renamed to `graphqlQueryStringLiteralFormat` with value in `GraphQLQueryStringLiteralFormat`.
* `urlToSchemaFile` - moved to `FileInput.schema`

**Removed:**
* `codegenEngine` - Swift is the only choice now
* `mergeInFieldsFromFragmentSpreads` - fragment spreads will _always_ be merged in
* `modifier` - the introduction of a shared module will require `public` on the schema types and operation objects, if output to the shared module, otherwise `internal` if output to the main target.
* `only` - dropped for now.
* `downloadTimeout` - use `URLSession` defaults (`timeoutIntervalForRequest` / `timeoutIntervalForResource`)

**Added:**
* `schemaDocumentation` - determines whether comments-as-documentation are copied from the schema into the generated schema types files. Value in `CompositionOption`

_There are no tests for this change because there is no logic in the structure and the convenience initializer is already tested in `testConfiguration_usingOnlyInputOutputFolders_shouldGenerateCompatibleFilenames` to ensure compatibility with `ApolloSchemaDownloadConfiguration`._


Apollo.xcodeproj/project.pbxproj
```diff
@@ -46,7 +46,7 @@
 		9B708AAD2305884500604A11 /* ApolloClientProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B708AAC2305884500604A11 /* ApolloClientProtocol.swift */; };
 		9B78C71E2326E86E000C8C32 /* ErrorGenerationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B78C71B2326E859000C8C32 /* ErrorGenerationTests.swift */; };
 		9B7B6F59233C287200F32205 /* ApolloCodegen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B7B6F57233C287100F32205 /* ApolloCodegen.swift */; };
-		9B7B6F5A233C287200F32205 /* ApolloCodegenOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B7B6F58233C287100F32205 /* ApolloCodegenOptions.swift */; };
+		9B7B6F5A233C287200F32205 /* ApolloCodegenConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B7B6F58233C287100F32205 /* ApolloCodegenConfiguration.swift */; };
```
calvincestari, line 49: A more descriptive name for what the structure is, similar to `ApolloSchemaDownloadConfiguration` and `URLSessionConfiguration`.

Apollo.xcodeproj/project.pbxproj
```diff
@@ -963,6 +964,7 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				E6D21E1B2705016000110BBD /* Nimble in Frameworks */,
```
calvincestari, line 967: This was missing from the `ApolloCodegenTests` causing build errors.

Sources/ApolloCodegenTestSupport/CodegenTestHelper.swift
```diff
@@ -11,19 +11,6 @@ import XCTest
 
 public struct CodegenTestHelper {
   
-  public static func dummyOptions() -> ApolloCodegenOptions {
```
calvincestari, line 14: Always good to remove dead code. 💀

Tests/ApolloCodegenTests/ApolloSchemaInternalTests.swift
```diff
@@ -27,13 +27,5 @@ class ApolloSchemaInternalTests: XCTestCase {
     let postType = try schema.getType(named: "Post")
     XCTAssertEqual(postType?.name, "Post")
   }
-
-  func testConfiguration_usingOnlyOutputFolders_shouldGenerateCompatibleFilenames() {
```
calvincestari, line 31: Moved into the public tests file since `ApolloCodegenConfiguration` now publicly exposes more of its properties.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
+import Foundation
+
+/// A configuration object that defines behavior for code generation.
+public struct ApolloCodegenConfiguration {
```
calvincestari, line 4: See the PR description for a full list of property changes.

Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
```diff
@@ -1,219 +0,0 @@
-import Foundation
-
-/// An object to hold all the various options for running codegen
-public struct ApolloCodegenOptions {
```
calvincestari, line 4: See the PR description for a full list of property changes.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+  /// Specify the input files required for code generation.
+  public struct FileInput {
+    /// Path to the GraphQL schema file. Can be in JSON or SDL format.
+    public let schema: URL
+    /// Glob of files to search for GraphQL operations. This should be used to find queries and any client schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
```
AnthonyMDev, line 9: Nit: If documentation lines are longer than 100 character line limit, I've been pulling them to a second line.
```swift
///  Glob of files to search for GraphQL operations. This should be used to find queries and any client schema extensions. Defaults to `./**/*.graphql`,
/// which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
```

What's your opinion on that as a style guideline?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+      case carthage
+      /// Generate a package.swift file.
+      case swiftPackageManager
+    }
+
+    /// The namespace to use for generated operation objects. When combined with `.includedInTarget` this should be the main target name. When used with any other dependency manager this should be the name of the new shared module that will be created.
+    public let name: String
+    /// An absolute location where the schema types files should be generated.
+    public let url: URL
+    /// Automation to ease the integration of the generated schem types file with compatible dependency managers.
```
AnthonyMDev, line 41:
```suggestion
    /// Automation to ease the integration of the generated schema types file with compatible dependency managers.
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+      case includedInTarget
+      /// Generate a podspec file.
+      case cocoaPods
+      /// Generate a cartfile.
+      case carthage
+      /// Generate a package.swift file.
+      case swiftPackageManager
+    }
+
+    /// The namespace to use for generated operation objects. When combined with `.includedInTarget` this should be the main target name. When used with any other dependency manager this should be the name of the new shared module that will be created.
```
AnthonyMDev, line 37: > When combined with `.includedInTarget` this should be the main target name.

That's not right.

When combined with `.includedInTarget` this will be the name of the namespace `enum` that we put the schema types in. It should definitely never be the name of your application target.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// The location structure for the schema types files.
+    public let schemaTypes: SchemaTypesFileOutput
+    /// The location structure for the operation object files.
+    public let operations: OperationsFileOutput
+  }
+
+  /// Defines the location structure for the schema types files.
+  public struct SchemaTypesFileOutput {
+    /// Compatible dependency manager automation.
+    public enum DependencyManagerAutomation {
```
AnthonyMDev, line 26: What about
```swift
  /// Compatible dependency manager automation.
    public enum ModuleType {
      /// No module will be created for the generated schema types. 
      /// Generated files must be manually added to the main application target.
      ///
      /// Because no module is created, the generated files will be namespaced to prevent naming conflicts.
      case manuallyLinked
      /// Generates a module with a podspec file that is suitable for linking to your project using CocoaPods.
      case cocoaPods
      /// Generates a module with a cartfile that is suitable for linking to your project using Carthage.
      case carthage
      /// Generates a module with a package.swift file that is suitable for linking to your project using Swift Package Manager
      case swiftPackageManager
    }
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /// Specify the formatting of the GraphQL query string literal.
+  public enum GraphQLQueryStringLiteralFormat {
+    /// The query string will be copied into the operation object with all line break formatting removed.
+    case singleLine
+    /// The query string will be copied with original formatting into the operation object.
+    case multiline
+    /// The query string will not be copied into the operation object. You must be using preregisterd queries with this configuration.
+    case exclude
```
AnthonyMDev, line 62: I'm not a fan of this here. It means someone could create an invalid configuration if then aren't using APQs and put `exclude`. I think that this should just be part of the APQ configuration, and if it's persisted queries only, then we just ignore this option since we aren't generating the query string literals.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  public enum CompositionOption {
```
AnthonyMDev, line 65: I love this.  Great idea!

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+/// A configuration object that defines behavior for code generation.
+public struct ApolloCodegenConfiguration {
+  /// Specify the input files required for code generation.
+  public struct FileInput {
+    /// Path to the GraphQL schema file. Can be in JSON or SDL format.
+    public let schema: URL
+    /// Glob of files to search for GraphQL operations. This should be used to find queries and any client schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
+    public let glob: String
+    /// [optional] Path to an operation id JSON map file. If specified, also stores the operation IDs (hashes) as properties on operation types. Defaults to `nil`.
+    public let operationIDs: URL?
```
AnthonyMDev, line 12: I'm pretty sure this is actually an OUTPUT not input. We know how to generate the operationIDs, and we output a file that can be sent to the server to register all your operations. @designatednerd correct me if I'm wrong here.

AnthonyMDev (commented):
Great work on this! Lets talk about the names for a few of these options and see what we can do to make them clear. This is confusing stuff for users.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: Personally I'm fine with wrapping docs because they're auto-wrapped by the IDE, but it is a pain to read them in GH so fine if y'all want to do that.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: Yes that is definitely an output not an input.

Apollo.xcodeproj/project.pbxproj
designatednerd: Are we actually using nimble now?

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -34,31 +34,13 @@ public class ApolloCodegen {
...
-        }
-      } // else we're fine with JSON at this point for intermediate purposes.
-      
-      try FileManager.default.apollo.createContainingFolderIfNeeded(for: fileURL)
-    }
 
-    let cli = try ApolloCLI.createCLI(cliFolderURL: cliFolderURL, timeout: options.downloadTimeout)
-    return try cli.runApollo(with: options.arguments, from: folder)
+    #warning("TODO: Build folder structure from configuration")
+    return "Not implemented!"
```
designatednerd, line 43: worth having this throw instead of returning a dummy string?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: (although split mode in GH does auto wrap)

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: I prefer @AnthonyMDev's suggestion, personally

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+      /// Generate a podspec file.
+      case cocoaPods
+      /// Generate a cartfile.
+      case carthage
+      /// Generate a package.swift file.
+      case swiftPackageManager
+    }
+
+    /// The namespace to use for generated operation objects. When combined with `.includedInTarget` this should be the main target name. When used with any other dependency manager this should be the name of the new shared module that will be created.
+    public let name: String
```
designatednerd, line 38: Why is this required? If stuff is in a module, are we still namespacing it?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// Automation to ease the integration of the generated schem types file with compatible dependency managers.
+    public let dependencyAutomation: DependencyManagerAutomation
+  }
+
+  /// Defines the location structure for the operation object files.
+  public enum OperationsFileOutput {
+    /// All operation object files will be located in the module with the schema types.
+    case inSchemaModule
+    /// Operation object files will be co-located relative to the defining `.graphql` file.
+    case relative(subpath: String?)
```
designatednerd, line 50: Would be helpful to define the `subpath` option here

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: I agree with @AnthonyMDev here too - I think the question for this enum is more "If there's a string literal, how should we format it" rather than "should we include the query string"

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: can you add more docs on this particular one, because I'm not 100% sure what's happening here

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+  public enum CompositionOption {
+    case include
+    case exclude
+  }
+
+  /// Enum to select how to handle properties using a custom scalar from the schema.
+  public enum CustomScalarFormat: Equatable {
+    /// Uses the default type of String.
+    case defaultAsString
+    /// Use your own types for custom scalars.
```
designatednerd, line 74: Probably worth specifying that this is being passed through from your schema

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -3,213 +3,5 @@ import ApolloCodegenTestSupport
...
-    _ = try ApolloCodegen.run(from: starWarsFolderURL,
-                              with: scriptFolderURL,
-                              options: options)
-    
-    XCTAssertTrue(FileManager.default.apollo.folderExists(at: outputFolder))
-    
-    let contents = try FileManager.default.contentsOfDirectory(atPath: outputFolder.path)
-    XCTAssertEqual(contents.count, 18)
-  }
+  #warning("TODO: Write extensive tests for codegen folder/module structure.")
```
designatednerd, line 6: is it worth leaving the test names with `XCTSkip` to make sure we have equivalent coverage?

docs/source/fetching-queries.md
```diff
@@ -194,7 +194,7 @@ Each query or mutation is identified by the SHA256 hash of its contents. If the
 
 To use APQs with the iOS SDK: 
 
-- When generating your code, pass a local path for output for the `--operationIdsPath` (or pass a file URL to the `operationIDsURL` on `ApolloCodegenOptions` if using Swift Scripting).  
+- When generating your code, pass a file URL to the `operationIDs` parameter on `ApolloCodegenConfiguration.FileInput`).
```
designatednerd, line 197: also should be updated to indicate this is output, not input

designatednerd (commented):
zomg excitement! really happy to see this moving forward

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
AnthonyMDev: It would become the name of the generated module

Apollo.xcodeproj/project.pbxproj
calvincestari: It's not the whole Quick framework and I do think it's an easier to read and understand matcher language. So yes, I'm good with us bringing it in and beginning to use it.

Is there any reason you particularly don't like it?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: I have my page guide set to 100 so yes I should have caught this; I'll break this and others.

On the whole I think the [Google Swift style guide](https://google.github.io/swift/) is pretty good and take direction from that generally. I believe it would form a good foundation for a [swift-format](https://github.com/apple/swift-format) and [SwiftLint](https://github.com/realm/SwiftLint) rule set.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: 👍🏻 - I agree `manuallyLinked` is more descriptive. Thanks for the extended documentation too.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: Great suggestion, thanks!

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: I intentionally didn't document the enum because I felt it was self-explanatory when looking at it in isolation. It takes on meaning in its application. I'll give it another pass though.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: Oops, my bad.

Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: I'm kinda meh on forcing the error here since we have the warning but I'm also game for anything that forcefully prevents us from forgetting. So yeah, I'll switch this out for `fatalError` since we don't actually want to be able to trap the error with a `throw`.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: I've instead opted to leave the test names as suggested but make them all fail with `XCTFail`. Using `XCTSkip` on the tests that don't throw is effectively the same as just leaving them empty and not serving as a hard-stop. Thanks for the great suggestion!

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: I added it as a way to give more expression to true/false type properties. I generalized it to `CompositionOption` because it is used in multiple properties.

An example is `let omitDeprecatedEnumCases: Bool = true` for which you had to read/grok into the property name to fully understand what `true` and `false` meant. It now becomes `let deprecatedEnumCases: CompositionOption = .include` which IMO is a more expressive way to make the statement. The `Bool` probably makes evaluation of it more terse when you actually have to take action based on the value but I favour the readability of the enum.

I still can't think of a good description for the enum itself but I've added clarifying documentation to the properties of that type; see e426028f.

AnthonyMDev (approved):
This looks great! Thanks Calvin1

Apollo.xcodeproj/project.pbxproj
designatednerd: No i was wondering because it's been linked due to transient dependencies for a while

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -34,31 +34,13 @@ public class ApolloCodegen {
...
-        }
-      } // else we're fine with JSON at this point for intermediate purposes.
-      
-      try FileManager.default.apollo.createContainingFolderIfNeeded(for: fileURL)
-    }
 
-    let cli = try ApolloCLI.createCLI(cliFolderURL: cliFolderURL, timeout: options.downloadTimeout)
-    return try cli.runApollo(with: options.arguments, from: folder)
+    #warning("TODO: Build folder structure from configuration")
+    fatalError("Not implemented!")
```
designatednerd, line 43: Doesn't need to happen now since this is already merged but I do like having an error that throws for this instead of a `fatalError` because a) If you try to test it, your whole test suite crashes and b) You can check that you're actually getting the correct `unimplemented` error, and then that will auto-fail when you actually implement it to remind you to write a test for the real thing 😃

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: Got it, this makes more sense now. I think it's throwing in a doc-comment on the type that this is to make what used to be boolean properties for whether to include or exclude certain things easier to reason about, because `CompositionOption` just as a name didn't really jump out to me.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
designatednerd: Yeah we've been meaning to add swiftlint for...some time

PR #1978, opened 2021-10-01
Disable CI on any `1.0` named branch

calvincestari:
**This is a temporary change** to stop skewing CI metrics until the `release/1.0-alpha-incubating` is in a state where tests are expected to pass; right now they are not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,18 +200,46 @@ workflows:
     jobs:
       - Swift_Build:
           name: Build with SPM
+          filters:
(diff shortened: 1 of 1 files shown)


AnthonyMDev:
Don't we need to ignore both `release/1.0**` and `1.0/**`?

calvincestari:
OK, finally got it working. Below is a screenshot from CI showing "No workflow" for branches with `1.0` in the name but still building for a non-1.0 branch name.

Screen Shot 2021-10-01 at 2 17 33 PM

PR #1985, opened 2021-10-11
Update tutorial-obtain-schema.md

martinbonnin:
Nitpick: I feel like the information that the user doesn't need to log in isn't really important and adds to the cognitive load so I'd just skip it.


designatednerd:
I think I'm gonna close this one out, I do think we'll get questions about the login button otherwise.

PR #2009, opened 2021-10-28
Create AST for code generation [1/x]

AnthonyMDev:
This PR creates `ASTSelectionSet`, which acts at the primary model for generated code.

This AST object and its related helpers handle the logic for transforming the JavaScript `CompilationResult` into the correct shape for code generation.
This includes:
- Deduplicating selection sets of the same type
- Computing merged selections for each type scope
- Sorting selections appropriately

The algorithmic logic in this PR is very complex. I'm happy to go over it if needed.


Sources/ApolloCodegenLib/AST/ScopedSelectionSetHashable.swift
```diff
@@ -0,0 +1,42 @@
...
+
+extension CompilationResult.Field: ScopedSelectionSetHashable {
+  var hashForSelectionSetScope: String {
+    return responseKey
+  }
+}
+
+extension CompilationResult.SelectionSet: ScopedSelectionSetHashable {
+  var hashForSelectionSetScope: String {
+#warning("What if there is a field with the same name as a type? Do we get a conflict? Write a test for this.")
```
designatednerd, line 33: it might have to be something where the field name and the type name are capitalization-equal - for instance `profile` and `Profile` shouldn't cause identical hashes

Sources/ApolloCodegenLib/AST/SortedSelections.swift
```diff
@@ -0,0 +1,131 @@
...
+  typealias Selection = CompilationResult.Selection
+  typealias Field = CompilationResult.Field
+  typealias TypeCase = CompilationResult.SelectionSet
+  typealias Fragment = CompilationResult.FragmentDefinition
+
+  fileprivate(set) var fields: OrderedDictionary<String, Field> = [:]
+  fileprivate(set) var typeCases: OrderedDictionary<String, TypeCase> = [:]
+  fileprivate(set) var fragments: OrderedDictionary<String, Fragment> = [:]
+
+  init() {}
```
designatednerd, line 14: is there a reason to have an empty init instead of just using the one with default values?

Sources/ApolloCodegenLib/AST/SortedSelections.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
+import OrderedCollections
+
+struct SortedSelections: Equatable, CustomDebugStringConvertible {
```
designatednerd, line 4: worth adding headerdoc here as well?

Sources/ApolloCodegenLib/AST/SortedSelections.swift
```diff
@@ -0,0 +1,131 @@
...
+    mergeIn(selections.values.elements)
+  }
+
+  var isEmpty: Bool {
+    fields.isEmpty && typeCases.isEmpty && fragments.isEmpty
+  }
+
+  // MARK: Selection Merging
+
+  @inlinable mutating func mergeIn(_ selections: SortedSelections) {
```
designatednerd, line 50: remind me: What is the benefit of making all these calls `@inlinable`?

Sources/ApolloCodegenLib/AST/TypeScopeDescriptor.swift
```diff
@@ -0,0 +1,51 @@
...
+typealias TypeScope = Set<GraphQLCompositeType>
+
+/// Defines the scope for an `ASTSelectionSet`. The scope is derived from the scope and all of its
+/// parent scopes.
+struct TypeScopeDescriptor {
+  let scope: TypeScope!
+
+  private init() {
+    self.scope = nil
+  }
```
designatednerd, line 12: Why does there need to ever be a case where the scope is `nil` here?

Sources/ApolloCodegenLib/Frontend/CompilationResult+SelectionSetMerging.swift
```diff
@@ -0,0 +1,97 @@
...
+extension CompilationResult.Selection: SelectionMergable {
+
+  var _selectionSet: CompilationResult.SelectionSet? {
+    switch self {
+    case let .field(selection as SelectionMergable),
+      let .inlineFragment(selection as SelectionMergable):
+      return selection._selectionSet
+
+    case let .fragmentSpread(fragment):
+      return fragment.selectionSet
```
designatednerd, line 80: why is this using the non-underscore `selectionSet` instead of the underscore one?

Sources/ApolloCodegenLib/Frontend/CompilationResult+SelectionSetMerging.swift
```diff
@@ -0,0 +1,97 @@
...
+  func merging(_ newSelectionSet: CompilationResult.SelectionSet) -> Self {
+    switch self {
+    case let .field(field):
+      return .field(field.merging(newSelectionSet))
+
+    case let .inlineFragment(selectionSet):
+      return .inlineFragment(selectionSet.merging(newSelectionSet))
+
+    case .fragmentSpread:
+      fatalError("Selections sets should never be merged into named fragments.")
```
designatednerd, line 93: This is a fatal rather than a throwing error since this basically represents an error on our part in the development of the code generator rather than anything the user can do anything about, correct?

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
```diff
@@ -0,0 +1,589 @@
...
+
+  static let compilationResult = try! frontend.compile(schema: schema, document: operations)
+
+  override func setUp() {
+    super.setUp()
+  }
+
+  override func tearDown() {
+    super.tearDown()
+  }
```
designatednerd, line 29: do these need to be here if they're not actually doing anything

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
```diff
@@ -0,0 +1,589 @@
...
+
+  func test__mergedSelections_AllAnimalsQuery_RootQuery__isCorrect() {
+    // given
+    let operation = Self.compilationResult.operations.first { $0.name == "AllAnimalsQuery" }
+    let rootSelectionSet = ASTSelectionSet(selectionSet: operation!.selectionSet,
```
designatednerd, line 34: i would recommend a `try XCTUnwrap` above rather than `!` since it'll let the rest of your tests keep going if it fails instead of crashing the test suite

(I'll stop myself from doing this for every single one of these but there's a fair few that could be switched to using this)

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
```diff
@@ -0,0 +1,589 @@
...
+
+    // then
+    expect(scope.type).to(equal(Interface_Animal))
+    expect(actual).to(shallowlyMatch(expected))
+  }
+
+  func test__mergedSelections_AllAnimalsQuery_AllAnimal_Height__isCorrect() {
+    // given
+    let operation = Self.compilationResult.operations.first { $0.name == "AllAnimalsQuery" }
+    guard case let .field(allAnimals) = operation!.selectionSet.selections[0],
```
designatednerd, line 93: worth doing these as `.first` instead of `[0]` for anti-crash reasons?

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
```diff
@@ -0,0 +1,589 @@
...
+    // then
+    expect(scope.type).to(equal(GraphQLObjectType.mock("Height")))
+    expect(actual).to(shallowlyMatch(expected))
+  }
+
+  func test__mergedSelections_AllAnimalsQuery_AllAnimal_Predator__isCorrect() {
+    // given
+    let operation = Self.compilationResult.operations.first { $0.name == "AllAnimalsQuery" }
+    guard case let .field(allAnimals) = operation!.selectionSet.selections[0],
+          case let .field(predator) = allAnimals.selectionSet?.selections[8] else { fail(); return }
```
designatednerd, line 123: how do we know this is at index 8?

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
```diff
@@ -0,0 +1,589 @@
...
+  func test__mergedSelections_AllAnimalsQuery_AllAnimal_AsCat_Height__isCorrect() {
+    // given
+    let operation = Self.compilationResult.operations.first { $0.name == "AllAnimalsQuery" }
+    guard case let .field(allAnimals) = operation!.selectionSet.selections[0] else { fail(); return }
+    let allAnimalsScope = ASTSelectionSet(selectionSet: allAnimals.selectionSet!,
+                                          compilationResult: Self.compilationResult)
+    let height = allAnimalsScope
+      .children.values[2]
+      .mergedSelections
+      .fields.values[1]
```
designatednerd, line 434: would be helpful to have some comments here about how we know these indexes

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
```diff
@@ -0,0 +1,2415 @@
...
+  /// query {
+  ///  allAnimals {
+  ///    ...AnimalDetails
+  /// }
+  ///
+  /// fragment AnimalDetails on Animal {
+  ///   species
+  /// }
+  /// Expected:
+  /// Birds.children should not include a type case for asAnimal
```
designatednerd, line 35: this is *extremely* helpful in figuring out what the test is doing, however I'm not seeing where the the subject type is being set as `Birds` - where is that?

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
```diff
@@ -0,0 +1,2415 @@
...
+  ///   ... on A {
+  ///     B
+  ///   }
+  /// }
+  ///
+  /// Expected:
+  /// Query.Children: { }
+  func test__children__givenInlineFragment_onSameType_mergesTypeCaseIn_doesNotHaveTypeCaseChild() {
+    // given
+    let Object_A = GraphQLObjectType.mock("A")
```
designatednerd, line 235: Is there a reason you're using abstract types here instead of examples from the schema? Honestly I find the examples from the schema a hell of a lot easier to follow.

Alternately, is it worth breaking these ones that are more abstract out into a separate file so it's clearer where we're testing actual use cases rather than purely conceptual ones?

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
```diff
@@ -0,0 +1,2415 @@
...
+    let expected: [CompilationResult.Selection] = [
+      .field(.mock(
+        "A",
+        type: .named(Object_A),
+        selectionSet: .mock(
+          parentType: Object_A,
+          selections: [
+            .field(.mock("B", type: .named(GraphQLScalarType.integer()))),
+            .field(.mock("C", type: .named(GraphQLScalarType.integer()))),
+            .field(.mock("D", type: .named(GraphQLScalarType.integer()))),
```
designatednerd, line 718: Is there a way to set up `.named` as a default value (since that's not really what you're testing here)?

Tests/ApolloCodegenTests/Frontend/DocumentParsingAndValidationTests.swift
```diff
@@ -124,7 +124,7 @@ class DocumentParsingAndValidationTests: XCTestCase {
     XCTAssertEqual(document.definitions.count, 3)
     
     let validationErrors = try codegenFrontend.validateDocument(schema: schema, document: document)
-    XCTAssertEqual(validationErrors, [])
+    XCTAssertEqual(validationErrors.count, 0)
```
designatednerd, line 127: also doable as `XCTAssertTrue(validationErrors.isEmpty)` but whatever you feel is clearer.

designatednerd (commented):
a lot of these questions are mostly to get more detail on how some of this works under the hood, but overall this looks amazing!

Sources/ApolloCodegenLib/Frontend/CompilationResult+SelectionSetMerging.swift
AnthonyMDev: Correct. This would only throw if we had an invalid schema. There is a consistent pattern throughout both the codegen and the Apollo lib (in the executor and caching layer) of assuming that the schema is valid. It means we can make much cleaner code with less useless try catch block that will NEVER run.

Everything has to run through the validator in `graphql-js` before getting to this point, so we don't have to handle this case.

Sources/ApolloCodegenLib/Frontend/CompilationResult+SelectionSetMerging.swift
AnthonyMDev: `FragmentDefinition` doesn't conform to `SelectionMergable`, so it doesn't have the underscored version.

Theoretically, line 80 should currently never be called. Since we can't actually merge fragments. But if for some reason i the future I wanted to use this method to retrieve the selectionSet from a `SelectionMergable` entity in another context where it would be valid, then it will still jut work.

This is not the same as the fatalError in `merging(_ newSelectionSet)`. As doing that is not just reading a property, but is actually an operation that is non-sensical and shouldn't be allowed to occur on a fragment. If we call that on a fragment, we WANT to fail fast and loud so we can fix it.

AnthonyMDev (commented):
Thanks for all the comments Ellen! This was super helpful!

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
AnthonyMDev: Oh I've never even thought to do that, because I've always used nimble over XCT whenever possible, but you're right that would be great. I'll start using that! Thanks

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomASTCreationTests.swift
AnthonyMDev: These indexes are specifically the order that the fields/selections should be generated in the generated models. It's really very precise and the point is to ensure the models are generated exactly right with a deterministic ordering. I can add a comment.

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
AnthonyMDev: Whoops, this should be `Animal` and it's set on line 43. My mistake!

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
AnthonyMDev: All of these are conceptual use cases really. None of this is using the actual schema. Each unit test creates the schema types it needs. (For example, In some of the tests, WarmBlooded implements Pet, in others it doesn't).

Most of these that just use "A, B, etc" are contrived examples that were difficult to create a real-world example of that made sense with the AnimalKingdom schema. I didn't want to waste a ton of time trying to come up with concepts that made sense for each example. For instance, Pet, WarmBlooded, and ClassroomPet doesn't always make sense for use cases with different Interfaces on the same type that implemented interfaces on other different sub types and selected the same fields.

Tests/ApolloCodegenTests/CodeGenAST/ASTSelectionSetTests.swift
AnthonyMDev: Done! Thanks for the suggestion!

Sources/ApolloCodegenLib/AST/ASTSelectionSet.swift
```diff
@@ -0,0 +1,147 @@
+import Foundation
+import ApolloUtils
+import OrderedCollections
+
+class ASTSelectionSet: CustomDebugStringConvertible, Equatable {
+  typealias Selection = CompilationResult.Selection
+  typealias SelectionSet = CompilationResult.SelectionSet
+  typealias ChildTypeCaseDictionary = OrderedDictionary<String, ASTSelectionSet>
```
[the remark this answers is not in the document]
designatednerd: Ooh right! Forgot about ~Dre~ that

Sources/ApolloCodegenLib/AST/SortedSelections.swift
designatednerd: Aha, that makes sense.

Issue #1981, opened 2021-10-05
GET option for ApolloSchemaDownloader


mikepitre:
## Feature request

It would be great if we had an option to use a GET http method for schema downloading/introspection as part of the ApolloCodegenLib.

## Motivation

A 3rd party CMS we're using only supports GET requests against their GraphQL API. They expect a GET request and the IntrospectionQuery passed as a query param.

designatednerd:
Out of sheer curiosity, what CMS is it?

mikepitre:
haha np, it's Prismic

calvincestari:
@mikepitre - this is now merged. Thanks for contribution! We'll get it out in the next release which should happen in the next week or so.

mikepitre:
@calvincestari great! Thanks for your help and sorry I couldn't dedicate more time to this at the moment. I really like the modifications you made!

PR #2010, opened 2021-10-28
GET method for `ApolloSchemaDownloader`

mikepitre:
https://github.com/apollographql/apollo-ios/issues/1981

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
@@ -46,6 +46,11 @@ public struct ApolloSchemaDownloadConfiguration {
     }
 
   }
+
+  public enum ApolloSchemaDownloadHTTPMethod {
+    case POST
+    case GET(introspectionQueryParameterName: String)
+  }
   
   public struct HTTPHeader: Equatable, CustomDebugStringConvertible {
     let key: String
@@ -63,6 +68,8 @@ public struct ApolloSchemaDownloadConfiguration {
 
   /// How to download your schema. Supports the Apollo Registry and GraphQL Introspection methods.
   public let downloadMethod: DownloadMethod
+  /// HTTP Method for Introspection. Setting this to GET requires passing the name of the query parameter for the IntrospectionQuery. Defaults to POST.
+  public let httpMethod: ApolloSchemaDownloadHTTPMethod
   /// The maximum time to wait before indicating that the download timed out, in seconds. Defaults to 30 seconds.
   public let downloadTimeout: Double
   /// Any additional headers to include when retrieving your schema. Defaults to nil.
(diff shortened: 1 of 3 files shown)


calvincestari:
I've made some minor changes to the enum and added an error case for trying to use GET with the Apollo Registry. Tomorrow I'll add tests.

calvincestari:
> Worth making this an option on the downloadMethod instead of on the initializer so you can't try to pass this in if you're using the registry?

Yes I'll take a pass at how that would look. There is an [error](https://github.com/apollographql/apollo-ios/pull/2010/commits/32dd79cb41aea5b4c0fceb0b796f06523478659e) for specifically this, but being able to completely rule out that possibility would be better.

designatednerd:
Yeah error good, making it actually impossible better 😇

calvincestari:
@designatednerd I think we're good to go with this PR now. There is more that I'd like to refactor with `ApolloSchemaDownloadConfiguration` but I'll do that in the 1.0 branch (#2012) because it will be breaking changes and right now supporting GET in this PR is all additive.

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
```diff
@@ -33,11 +33,29 @@ public struct ApolloSchemaDownloadConfiguration {
         self.variant = variant
       }
     }
+
+    /// The HTTP request method.
```
designatednerd, line 37: probably worth clarifying that this is for introspection

Sources/ApolloCodegenLib/ApolloSchemaDownloadConfiguration.swift
```diff
@@ -75,6 +93,7 @@ public struct ApolloSchemaDownloadConfiguration {
   /// - Parameters:
   ///   - downloadMethod: How to download your schema.
   ///   - downloadTimeout: The maximum time to wait before indicating that the download timed out, in seconds. Defaults to 30 seconds.
+  ///   - downloadHTTPMethod: HTTP Method for Introspection. Setting this to GET requires passing the name of the query parameter for the IntrospectionQuery. Defaults to POST.
```
designatednerd, line 96: I think this comment can also be removed

Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
```diff
@@ -29,7 +31,11 @@ public struct ApolloSchemaDownloader {
       case .couldNotCreateSDLDataToWrite(let schema):
         return "Could not convert SDL schema into data to write to the filesystem. Schema: \(schema)"
       case .couldNotConvertIntrospectionJSONToSDL(let underlying):
-          return "Could not convert downloaded introspection JSON into SDL format. Underlying error: \(underlying)"
+        return "Could not convert downloaded introspection JSON into SDL format. Underlying error: \(underlying)"
+      case .couldNotCreateURLComponentsFromEndpointURL:
+        return "Could not create URLComponents from EndpointURL for Introspection."
+      case .couldNotGetURLFromURLComponents:
+        return "Could not get URL from URLComponents."
```
designatednerd, line 38: Might be worth passing the components here so people can examine them. Also fine not to.

designatednerd (approved):
Couple suggestions, otherwise good to go!

PR #2014, opened 2021-11-02
Expose `cacheKey` function as `public`

### LINKED ISSUE #1972: Making GraphQLField's cacheKey method public instead of internal
## Feature request

Making GraphQLField's cacheKey method public instead of internal

## Motivation

We're using GraphQL and Apollo. I have several ideas which led me to this request. For one I like the idea of being able to log a nice description of a Query - basically matching what the Apollo client itself uses as the Query's key in QUERY_ROOT:
```
func some(query: GraphQLQuery, field: GraphQLField) {
        let key = try? field.cacheKey(with: query.variables)
        print(key)
}
```
There are other things I'm exploring related to tracking of queries and then being able to purge queries that haven't been executed within some time window. 

## Proposed solution
In Sources/Apollo/GraphQLSelectionSet.swift
```
  public func cacheKey(with variables: [String: JSONEncodable]?) throws -> String { ... }
```

## Outstanding Questions
I can certainly create a PR with this change if it seems appropriate.

calvincestari:
This changes the access modifier of the `GraphQLField.cacheKey` function from `internal` to `public`. 

The [caveat](https://github.com/apollographql/apollo-ios/issues/1972#issuecomment-941540807) to this change is:
> For the 1.0 release, I'd like to consider if there is a better way to expose cache keys that is safer and more stable, but since we are rewriting a large piece of the execution layer there, exposing this in the 0.x version for now seems fine.

Closes #1972

### DIFF
--- a/Tests/ApolloTests/CacheKeyForFieldTests.swift
+++ b/Tests/ApolloTests/CacheKeyForFieldTests.swift
@@ -1,5 +1,5 @@
 import XCTest
-@testable import Apollo
+import Apollo
 import ApolloTestSupport
 import StarWarsAPI
 
--- a/Sources/Apollo/GraphQLSelectionSet.swift
+++ b/Sources/Apollo/GraphQLSelectionSet.swift
@@ -61,7 +61,7 @@ public struct GraphQLField: GraphQLSelection {
     self.type = type
   }
 
-  func cacheKey(with variables: [String: JSONEncodable]?) throws -> String {
+  public func cacheKey(with variables: [String: JSONEncodable]?) throws -> String {
     if
       let argumentValues = try arguments?.evaluate(with: variables),
       argumentValues.apollo.isNotEmpty {


Tests/ApolloTests/CacheKeyForFieldTests.swift
```diff
@@ -1,5 +1,5 @@
 import XCTest
-@testable import Apollo
```
calvincestari, line 2: I'm in two minds about this change. I left it in as a forcing function of-sorts for anyone who one day thinks `public` might be the wrong choice. I could easily be convinced to revert this line if someone feels strongly that it should remain `@testable`.

designatednerd: I'm good with this, though we should probably do some audits of where we're using `@testable` import separate from this.

designatednerd (approved):
Approved given the caveat.

Issue #1943, opened 2021-09-18
Decision Record - Defer support of SQLite.swift 0.13.0+


calvincestari:
### Status
* Proposed - Sep 17, 2021
* Accepted - Sep 20, 2021

### Summary
SQLite.swift [0.13.0](https://github.com/stephencelis/SQLite.swift/releases/tag/0.13.0) was recently released and one of the changes included is an increase of the macOS minimum deployment target to 10.15 - [diff](https://github.com/stephencelis/SQLite.swift/commit/449edebfd6c91c2dd376ee61359b4c4b65032032#diff-76ba6f5c9f153314f6c7b47b72497c0526e0923dcd9eeb4c534d53560ab7cb2dR24).

PR #1925 began the work of bumping our own macOS minimum deployment target to 10.15, enabling us to support SQLite.swift 0.13.0. Dropping macOS 10.14 results in a number of warnings about deprecated APIs within the Apple Security.framework API. Note that the code in question is in the Starscream source code that we absorbed in PR #1906.
```
/apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/SSLSecurity.swift:161:13: warning: 'SecTrustEvaluate' was deprecated in macOS 10.15: renamed to 'SecTrustEvaluateWithError(_:_:)'
/apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/WebSocketStream.swift:101:23: warning: 'SSLSetEnabledCiphers' was deprecated in macOS 10.15: No longer supported. Use Network.framework.
/apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/WebSocketStream.swift:197:7: warning: 'SSLGetPeerDomainNameLength' was deprecated in macOS 10.15: No longer supported. Use Network.framework.
/apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/WebSocketStream.swift:201:9: warning: 'SSLGetPeerDomainName' was deprecated in macOS 10.15: No longer supported. Use Network.framework.
```

Replacing `SecTrustEvaluate` with `SecTrustEvaluateWithError(_:_:)` is easy enough - c0523d0569cfa368038572c2b831d111910b3517.

The other "No longer supported." warnings are a bit more complicated:
[...]

PR #2015, opened 2021-11-02
Update SQLite.swift to version 13.0

Iron-Ham:
## Overview

This PR: 
- Updates the version of SQLite being used in Apollo-iOS to version 13.0 

### What do we get by upgrading to version 13.0?

The release notes for version 13.0 aren't super descriptive (`Support for Xcode 12 and iOS 14 + fixes`) but version 13.0 includes new table functionality for SQLite.swift, such as `upsert` and `insertMany`.

I know that y'all are planning on [switching out SQLite.swift](https://github.com/apollographql/apollo-ios/issues/1658) for GRDB or FMDB, but this a small quality of life improvement for those of us who are also using `SQLite.swift` in our application which depend upon Apollo. 

### What should we test?

Everything... should be the same? The diff between versions is [fairly small](https://github.com/stephencelis/SQLite.swift/compare/0.12.2...0.13.0) and concerns around `next()` and `failableNext()` [don't seem to impact Apollo](https://github.com/stephencelis/SQLite.swift/blob/master/Sources/SQLite/Typed/Query.swift#L984-L990) since we're using [`map`](https://github.com/apollographql/apollo-ios/blob/0.47.1/Sources/ApolloSQLite/SQLiteDotSwiftDatabase.swift#L39) and not `next()` or `failableNext()` directly.

### DIFF
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -33,7 +33,7 @@ Pod::Spec.new do |s|
   s.subspec 'SQLite' do |ss|
     ss.source_files = 'Sources/ApolloSQLite/*.swift'
     ss.dependency 'Apollo/Core'
-    ss.dependency 'SQLite.swift', '~>0.12.2'
+    ss.dependency 'SQLite.swift', '~>0.13.0'
   end
 
   # Websocket and subscription support based on Starscream
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -33,7 +33,7 @@ Pod::Spec.new do |s|
   s.subspec 'SQLite' do |ss|
     ss.source_files = 'Sources/ApolloSQLite/*.swift'
     ss.dependency 'Apollo/Core'
-    ss.dependency 'SQLite.swift', '~>0.13.0'
+    ss.dependency 'SQLite.swift', '~>0.13.1'
   end
 
   # Websocket and subscription support based on Starscream
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -3429,7 +3429,7 @@
 			repositoryURL = "https://github.com/stephencelis/SQLite.swift.git";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 0.13.0;
+				minimumVersion = 0.13.1;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -3429,7 +3429,7 @@
 			repositoryURL = "https://github.com/stephencelis/SQLite.swift.git";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 0.12.2;
+				minimumVersion = 0.13.0;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/stephencelis/SQLite.swift.git",
         "state": {
           "branch": null,
-          "revision": "0a9893ec030501a3956bee572d6b4fdd3ae158a1",
-          "version": "0.12.2"
+          "revision": "9af51e2edf491c0ea632e369a6566e09b65aa333",
+          "version": "0.13.0"
         }
       }
     ]
--- a/Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/stephencelis/SQLite.swift.git",
         "state": {
           "branch": null,
-          "revision": "9af51e2edf491c0ea632e369a6566e09b65aa333",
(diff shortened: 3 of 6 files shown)


designatednerd:
FWIW the switch has been back-burnered for a bit until we get codegen done. I seem to remember @calvincestari looking into this and there was some reason it wasn't going to work...Calvin, care to elaborate?

calvincestari:
Hi @Iron-Ham, thanks for the PR - 👋🏻 

Your PR doesn't appear to update the Xcode package dependency for SQLite.swift. Opening the project you'll notice it's still only fetching `0.12.2` - that'll need to be updated too.

> Everything... should be the same? The diff between versions is fairly small and concerns around next() and failableNext() don't seem to impact Apollo since we're using map and not next() or failableNext() directly.

We switched to `prepareRowIterator` in https://github.com/apollographql/apollo-ios/pull/1917 because internally it uses `failableNext` and [the change in `0.12.2`](https://github.com/stephencelis/SQLite.swift/pull/1030/files#diff-512b841313be999788c91e9ee18e5fc4c62c4581ccfe6d950dc48290841bdfb8R211) would have hidden errors. After we'd merged our change I see they reversed the change in SQLite.swift but I still believe `failableNext` is the correct choice because the [old behaviour (now current again)](https://github.com/stephencelis/SQLite.swift/blob/master/Sources/SQLite/Core/Statement.swift#L211) will crash an app on error with `try!`.

@designatednerd is correct, I did look at updating to SQLite.swift `0.13.0` a while ago and we made a [decision](https://github.com/apollographql/apollo-ios/issues/1943) to defer it. We made that decision because of the bump to the minimum macOS deployment target from `10.10` to `10.15`.

While bumping to macOS `10.15` isn't impactful to the ApolloSQLite target it is for the ApolloWebSocket target because a bunch of the Security.framework API was deprecated. If you only ever built apollo-ios for the iOS platform you would be fine and never know about this because the minimum iOS deployment target for apollo-ios and SQLite.swift is `12.0` and `9.0` respectively.

Some thoughts:
* I'm not 100% sure of the impact with supporting divergent macOS deployment targets for ApolloSQLite and ApolloWebSocket.
* I'm confused why I don't get a warning when building ApolloSQLite for macOS where SQLite.swift requires `10.15` but apollo-ios still has a minimum of `10.14`?
* We are planning to rewrite the WebSocket implementation in apollo-ios 2.0, in amongst a larger network stack overhaul. That change will likely include us updating to minimum deployment targets of `13.0` for iOS and `10.15` for macOS. _That's not to say we can't find a resolution to your issue now, I'm sharing where we had deferred the SQLite.swift update to._

designatednerd:
> I'm confused why I don't get a warning when building ApolloSQLite for macOS where SQLite.swift requires 10.15 but apollo-ios still has a minimum of 10.14? 

Is that after you've updated the `xcodeproj` to try to use `0.13.0`? Otherwise I don't think it'll try to fetch.

designatednerd:
My least favorite category of thing to solve: It works, I have no idea why.

calvincestari:
> > I'm confused why I don't get a warning when building ApolloSQLite for macOS where SQLite.swift requires 10.15 but apollo-ios still has a minimum of 10.14?

> Is that after you've updated the xcodeproj to try to use 0.13.0? Otherwise I don't think it'll try to fetch.

@designatednerd yes. I verified it was showing `0.13.0` as the dependency version in Xcode and the updated has in the diff is `9af51e2edf491c0ea632e369a6566e09b65aa333`.

Iron-Ham:
Sticking this in draft. 
Context: Over Slack, we realized that this is working when built using SPM, but not so much when using CocoaPods. That's because in [SQLite.Swift's update to 0.13.0](https://github.com/stephencelis/SQLite.swift/commit/449edebfd6c91c2dd376ee61359b4c4b65032032), they updated the minimum version requirement in the `podspec`. There are no declared minimums in the `Package.swift` file.

calvincestari:
> Over Slack, we realized that this is working when built using SPM, but not so much when using CocoaPods.

Correct. When setting up a project with Cocoapods you get the expected warning due to the version mismatch.
```
Analyzing dependencies
Pre-downloading: `Apollo` from `https://github.com/Iron-Ham/apollo-ios.git`, branch `hs/update-sqliteswift`
[!] CocoaPods could not find compatible versions for pod "SQLite.swift":
  In Podfile:
    Apollo/SQLite (from `https://github.com/Iron-Ham/apollo-ios.git`, branch `hs/update-sqliteswift`) was resolved to 0.49.1, which depends on
      SQLite.swift (~> 0.13.0)

Specs satisfying the `SQLite.swift (~> 0.13.0)` dependency were found, but they required a higher minimum deployment target.
```

I've submitted a [PR](https://github.com/stephencelis/SQLite.swift/pull/1094) to SQLite.swift to get platform versions added to the SPM manifest. Hopefully this should give SPM the details it needs to give similar warnings.

calvincestari:
OK, we've gotten that [PR on SQLite.swift](https://github.com/stephencelis/SQLite.swift/pull/1094) merged and I'm following up with the originator of the macOS version change to see if it's actually needed; maybe we can get it rolled back.

calvincestari:
Good news - the change to require macOS 10.15 was [reverted](https://github.com/stephencelis/SQLite.swift/pull/1095). As soon as they publish another release we can get this PR merged after resolving to that new version.

Iron-Ham:
Thanks @calvincestari ! I really appreciate all of the follow-up you've been doing related to this PR, and I especially appreciate the follow-through with the `SQLite.swift` folks. 

❤️

calvincestari:
@Iron-Ham I've just bumped the minimum required version to `0.13.1` with the SQLite.swift [release](https://github.com/stephencelis/SQLite.swift/releases/tag/0.13.1) today. Once the tests on this updated PR pass I'll get it merged and we will release `0.50.0` of apollo-ios today with this included. 🎉

PR #2023, opened 2021-11-09
AST Field Model

AnthonyMDev:
The model structure for an AST Field. Next PR will implement construction of ASTField from within `MergedSelectionBuilder` to create merged selections for each field from it's enclosing entity.

### DIFF
--- a/Sources/ApolloCodegenLib/AST/SortedSelections.swift
+++ b/Sources/ApolloCodegenLib/AST/SortedSelections.swift
@@ -3,7 +3,7 @@ import OrderedCollections
 
 struct SortedSelections: Equatable, CustomDebugStringConvertible {
   typealias Selection = CompilationResult.Selection
-  typealias Field = ASTFieldType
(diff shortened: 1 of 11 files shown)


Sources/ApolloCodegenLib/AST/SortedSelections.swift
```diff
@@ -47,59 +47,63 @@ struct SortedSelections: Equatable, CustomDebugStringConvertible {
 
   // MARK: Selection Merging
 
-  @inlinable mutating func mergeIn(_ selections: SortedSelections) {
```
calvincestari, line 50: I agree this is not necessary - there's a good read on `@inlinable` over at https://swiftrocks.com/understanding-inlinable-in-swift.

Do you think there would be benefit in using `@inline` for `mergeIn(fragment:)` and `appendOrMerge<T: SelectionMergable>(selection:into:)`?

calvincestari (approved):

PR #2024, opened 2021-11-09
Attempted fix for integration test failing on CI

AnthonyMDev:
The integration test for `testConcurrentSubscribing()` has been failing intermittently on CI. I can't tell for certain what's causing it, because it seems to be a race condition and I can't reproduce it locally.

My current hypothesis is that the mutation is sometimes being performed before the subscribers are actually subscribed. The barrier on the `concurrentQueue` guarantees that we've called `subscribe`, but not that the subscription has been added, which happens async on the `processingQueue`.  The test comment even said "dispatched on processing queue", but it wasn't. It was dispatched on the websocket's callback queue.


calvincestari (approved):
Good hypothesis - I'm happy to try anything that makes this test more reliable.

PR #2025, opened 2021-11-09
Improve readability of code examples

CemYil03:
While reading the documentation the way the code examples where indented kind of triggered me. So I tried to improve their readability by optimizing their indentation.

### DIFF
--- a/docs/source/initialization.md
+++ b/docs/source/initialization.md
@@ -12,7 +12,7 @@ import Apollo
 
 class Network {
   static let shared = Network() 
-    
+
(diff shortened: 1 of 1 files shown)


designatednerd:
Hey @CemYil03!

You’ve kickstarted some discussions internally about creating a linting standard that we can use across our code and docs, but our standard may not line up with the decisions here (and would need to have many, many, many more changes once we agree on something). 

We definitely appreciate the work that went into the PR, but I'm going to go ahead and close this out, and we'll make a larger update once we've come to an agreement on a standard. Thank you!

CemYil03:
Personally I think little improvement is better than no improvement for now. But I am also happy if it kickstarted something for the longer term.

Issue #1611, opened 2021-01-06
Codegen not working on Apple Silicon / M1 without Rosetta


LinusU:
## Bug report

The `run-bundled-codegen.sh` script uses a bundled copy of `node` which is an Intel binary. This means that it cannot run on the new M1 Macs (without Rosetta), and fails with the following message:

> env: node: Bad CPU type in executable

I believe that an easy fix is to upgrade the bundled Node.js version to a universal binary.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.39.0
- Xcode version: 12.3
- Swift version: 5.3.2
- Package manager: 5.3.0

## Steps to reproduce

Run the `run-bundled-codegen.sh` script on an M1 Mac.

LinusU:
If anyone needs a quick workaround, it can be solved by simply replacing the bundled Node with your system binary (e.g. the one installed via Brew):

```sh
cp $(which node) ~/Library/Developer/Xcode/DerivedData/PROJECT-xxxxxxxxxxxxxxxxxxxxxxxxxxxx/SourcePackages/checkouts/apollo-ios/scripts/apollo/bin/node
```

designatednerd:
I'd respectfully disagree with "not working" as it's working when using rosetta. 

I'll take a look at what we can do about getting a silicon binary to run natively though.

LinusU:
> I'd respectfully disagree with "not working" as it's working when using rosetta.

Sure thing 👍 I've updated the title

designatednerd:
Yeah I think the overwhelming majority of folks are going to have rosetta installed, at least to start with. 

Unfortunately most of what I'd need to take a look at is on our end around the packaging, but thank you for the offer!

LinusU:
@designatednerd any update on this? Ran into this today again 😅 

Node.js 14 is getting Apple Silicon support soon (https://github.com/nodejs/node/pull/38051) and the newly released 16 version already have it 🚀

designatednerd:
Talked this through with some colleagues - we'll take a look at updating once the Node 14 universal binary gets released. Right now the version bundled for iOS use is 12 since that was the LTS version when we first started putting this workaround together.

sigmadeltasoftware:
> @designatednerd any update on this? Ran into this today again 😅
> 
[...]

PR #2037, opened 2021-11-17
Workaround: Add option to use Homebrew's ARM64 node binary for Apple Silicon users

sigmadeltasoftware:
In context of the discussions held about [issue #1611](https://github.com/apollographql/apollo-ios/issues/1611) , this workaround will allow Apple Silicon users to make use of Homebrew's ARM64 `node` binary without changing anything about the current workings of the system.

### DIFF
--- a/scripts/run-bundled-codegen.sh
+++ b/scripts/run-bundled-codegen.sh
@@ -99,12 +99,12 @@ download_apollo_cli_if_needed
 # Make sure we're using an up-to-date and valid version of the Apollo CLI
 validate_codegen_and_extract_if_needed
 
+# Add the binary directory to the beginning of PATH so included binary verson of node is used.
+ PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"
+
 # Give Apple Silicon users the option to install & use an ARM-based binary of node through Homebrew
 if [[ `uname -m` == 'arm64' ]]; then
-  PATH="/opt/homebrew/bin:${SCRIPT_DIR}/apollo/bin:${PATH}"
-else
-  # Add the binary directory to the beginning of PATH so included binary verson of node is used.
-  PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"
+  PATH="/opt/homebrew/bin:${PATH}
 fi
 
 # Use the bundled executable of the Apollo CLI to generate code
--- a/scripts/run-bundled-codegen.sh
+++ b/scripts/run-bundled-codegen.sh
@@ -99,8 +99,13 @@ download_apollo_cli_if_needed
 # Make sure we're using an up-to-date and valid version of the Apollo CLI
 validate_codegen_and_extract_if_needed
 
-# Add the binary directory to the beginning of PATH so included binary verson of node is used.
-PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"
+# Give Apple Silicon users the option to install & use an ARM-based binary of node through Homebrew
+if [[ `uname -m` == 'arm64' ]]; then
+  PATH="/opt/homebrew/bin:${SCRIPT_DIR}/apollo/bin:${PATH}"
+else
+  # Add the binary directory to the beginning of PATH so included binary verson of node is used.
+  PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"
+fi
 
 # Use the bundled executable of the Apollo CLI to generate code
 APOLLO_CLI="${SCRIPT_DIR}/apollo/bin/run"


AnthonyMDev:
@designatednerd Will this work with the SwiftScripts method, or does that not use this script at all?

scripts/run-bundled-codegen.sh
```diff
@@ -99,8 +99,13 @@ download_apollo_cli_if_needed
 # Make sure we're using an up-to-date and valid version of the Apollo CLI
 validate_codegen_and_extract_if_needed
 
-# Add the binary directory to the beginning of PATH so included binary verson of node is used.
-PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"
+# Give Apple Silicon users the option to install & use an ARM-based binary of node through Homebrew
```
AnthonyMDev, line 102: Would this be better as this?

```
 # Add the binary directory to the beginning of PATH so included binary verson of node is used.
 PATH="${SCRIPT_DIR}/apollo/bin:${PATH}"

# Give Apple Silicon users the option to install & use an ARM-based binary of node through Homebrew
if [[ `uname -m` == 'arm64' ]]; then
  PATH="/opt/homebrew/bin:${PATH}
fi
```
Seems cleaner and doesn't repeat parts of the path.

aravasio: @AnthonyMDev 

I think a better approach would be to extend the current bundling strategy. Apollo already bundles a binary for apollo-cli in it's Pod (`${SCRIPT_DIR}/apollo/bin`), so I think it makes sense to just bundle another binary (lets call it `${SCRIPT_DIR}/apollo/bin-arm64` for clarity's sake) and check architecture and define `PATH` based on your proposed solution.

This would keep in place the solution to avoid dependency hell with NPM and allow for Apple Silicon to work natively and out of the box, without having to figure out if users have homebrew, npm,  yarn, or whatever else they might (or might not) have.

```
PATH="${SCRIPT_DIR}/apollo/bin:${PATH}
if [[ `uname -m` == 'arm64' ]]; then
  PATH=${SCRIPT_DIR}/apollo/bin-arm64:${PATH}
fi
```

sigmadeltasoftware: @aravasio Apollo currently doesn't prepackage any binaries in the release, @designatednerd explained in #1611 they download the necessary source files to build the `node` (x86_64) binary for usage. 

So while I agree that from a pragmatic & long term POV building the ARM64 binary of the library as well would be better, given the fact that the upcoming `1.0` release is going to move away from `node` completely, the effort of incorporating & maintaining the ARM64 build of `node` could outweigh the benefits.

Taking into account that the largest developer userbase on macOS is likely already using & acquainted with Homebrew, the currently proposed solution is merely a minimally intrusive workaround until `1.0` is released. 

@AnthonyMDev I like your solution better indeed, I'll adapt the current proposal.

designatednerd:
So there are a number of scenarios where this would a) Not work and b) Lead to even further confusion. Let me outline a few: 

- **The user does not have the version of node which works with our CLI**. My recollection is that the version of node we have bundled is Node 14 (I'd need to double check this), but different versions of node will interact differently with our CLI. Some of them will be compatible, some won't. This is the entire reason we started downloading the whole CLI including node in the first place: Dealing with node and `npm` versioning sucks, and is not a thing iOS developers have to do on a regular basis, and therefore they don't have experience troubleshooting it. 

- **The user does not have Homebrew set up yet**. It's not clear what happens here if the user hasn't set up Homebrew - if that folder doesn't exist when you try to add it to `PATH`, what happens then? 

- **The user has Homebrew set up, but they don't have an M1 compatible version of node installed**: This leaves us exactly where we are right now, since the system would be looking for a compatible version of node that doesn't exist, and then bomb out when it can't find it. 

- **The user has installed an M1 compatible version of node from a non-Homebrew source**: Since that may or may not already be in the `PATH`, it could either work perfectly or it could bomb out identically to how it's bombing out now.

Dealing with problems like this was seriously 50% of the issues when I first started at Apollo, and I am _extremely_ reluctant to re-open that rabbit hole. Particulary because a) The vast majority of people are using Rosetta, where this is not an issue and b) This entire system is getting replaced within the next few months with version 1.0, which will be using `JavaScriptCore` which can run on M1 natively. 

Would be interested in hearing your thoughts on these scenarios, but my instinct is that handling all of these (and any others we run across in the issues once we officially say "Yes, this is code Apollo is going to support") is going to create an amount of work that will materially delay the new codegen, which is a result nobody wants.

PR #2040, opened 2021-11-24
Implements a glob path builder

### LINKED ISSUE #2030: Use `ApolloCodegenConfiguration.FileInput.glob` to build a list of operation paths
We have the schema and we are given a glob of file inputs. The task here is to be able to generate a list of paths for all operation files.

_in:_ path, glob pattern
_out:_ operation file paths

calvincestari:
This is a path pattern matcher that will be used to ingest the glob pattern from `ApolloCodegenConfiguration.FileInput.glob` and output a set of matched file paths.

* There is no standard implementation for globstar (`**`) support, it is custom logic.
* It uses the underlying system function [glob](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/glob.3.html) to perform the actual file path matching. glob in turn uses [fnmatch](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/fnmatch.3.html) pattern definitions.

| Pattern | Meaning |
| -- | -- |
| * | matches everything |
| ? | matches any single character |
| [seq] | matches any character in seq |
| [!seq] | matches any character not in seq |
| ** | searches all subdirectories for the path pattern |

Closes #2030.


martinbonnin:
Out of curiosity, why not Regexes? 

As a user, half of my brain is already filled with Regexes details and I always dread having to learn another globing scheme.

calvincestari:
This uses the underlying glob library function of the operating system, which isn't regex based. It's the same behaviour you'd get using it on the command line.

Going down the reflex path we'd have to implement the matching logic ourselves, granted regex would do much of the heavy lifting for us.

calvincestari:
I just realized this is the old PR that I closed because of the branch name firing off CI.

There is a new PR with correct branch naming for CI - https://github.com/apollographql/apollo-ios/pull/2041

martinbonnin:
> It's the same behaviour you'd get using it on the command line.

Gotcha 👍 . Gradle has the same thing too.

PR #2041, opened 2021-11-24
Implements a glob path builder

### LINKED ISSUE #2030: Use `ApolloCodegenConfiguration.FileInput.glob` to build a list of operation paths

calvincestari:
This is a path pattern matcher that will be used to ingest the glob pattern from `ApolloCodegenConfiguration.FileInput.glob` and output a set of matched file paths.

* There is no standard implementation for globstar (`**`) support, it is custom logic.
* It uses the underlying system function [glob](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/glob.3.html) to perform the actual file path matching. glob in turn uses [fnmatch](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/fnmatch.3.html) pattern definitions. I intentionally only mention a few of the characters/patterns supported by glob/fnmatch because it's difficult to communicate the full ability of those libraries. I think the table below should be our only supported pattern matching options.

Each path matching pattern can include the following characters:
| Character | Meaning | Example |
| -- | -- | -- |
| `*` | matches everything but the directory separator (shallow) |`*.graphql` |
| `?` | matches any single character | `file-?.graphql` |
| `**` | matches all subdirectories (deep) | `**/*.graphql` |
| `!` | excludes any match | `*.graphql,!file.graphql` |

Closes #2030.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -13,7 +13,7 @@ public struct ApolloCodegenConfiguration {
     /// - `*` matches everything but the directory separator (shallow), eg: `*.graphql`
     /// - `?` matches any single character, eg: `file-?.graphql`
     /// - `**` matches all subdirectories (deep), eg: `**/*.graphql`
-    /// - `!` excludes any match, eg: `*.graphql,!file.graphql`
+    /// - `!` excludes any match only if the pattern starts with a `!` character, eg: `!file.graphql`
     public let includes: [String]
 
     /// Designated initializer.
@@ -25,7 +25,7 @@ public struct ApolloCodegenConfiguration {
     ///     - `*` matches everything but the directory separator (shallow), eg: `*.graphql`
     ///     - `?` matches any single character, eg: `file-?.graphql`
     ///     - `**` matches all subdirectories (deep), eg: `**/*.graphql`
-    ///     - `!` excludes any match, eg: `*.graphql,!file.graphql`
+    ///     - `!` excludes any match only if the pattern starts with a `!` character, eg: `!file.graphql`
     public init(schemaPath: String, includes: [String]) {
       self.schemaPath = schemaPath
       self.includes = includes
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -13,9 +13,12 @@ public struct ApolloCodegenConfiguration {
     ///
     /// - Parameters:
     ///  - schemaPath: Local path to the GraphQL schema file. Can be in JSON or SDL format.
-    ///  - glob: Glob of files to search for GraphQL operations. This should be used to find queries and any client
-    ///  schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of
-    ///  the folder where the script is run.
+    ///  - glob: Glob pattern used to search for GraphQL operations, such as queries and mutations. The pattern accepts a
+    ///  number of special characters:
+    ///      - `*` matches everything but the directory separator, eg: `*.graphql`
+    ///      - `?` matches any single character, eg: `file-?.graphql`
+    ///      - `{,}` matches all patterns separated by a comma, eg: `{query.*,operation.*}`
+    ///      - `**` matches all subdirectories, eg: `**/*.graphql`
     public init(schemaPath: String, glob: String = "./**/*.graphql") {
       self.schemaPath = schemaPath
       self.glob = glob
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -9,9 +9,10 @@ public struct ApolloCodegenConfiguration {
     /// A comma-delimited string of path matching patterns to search for GraphQL operations, such as queries and mutations.
     /// 
     /// Each path matching pattern can include the following characters:
-    /// - `*` matches 0 or more characters in a single path portion, eg: `*.graphql`
-    /// - `?` matches 1 character, eg: `file-?.graphql`
-    /// - `**` matches zero or more directories and subdirectories searching for matches, eg: `**/*.graphql`
+    /// - `*` matches zero or more characters in a single path portion, eg: `*.graphql`
+    /// - `?` matches one character in a single path portion, eg: `file-?.graphql`
+    /// - `**` includes zero or more directories and subdirectories searching for matches, eg: `**/*.graphql`
+    /// - `!` excludes any match, eg: `a/*.graphql,!a/file.graphql`
     public let includes: String
 
     /// Designated initializer.
@@ -20,9 +21,10 @@ public struct ApolloCodegenConfiguration {
     ///  - schemaPath: Local path to the GraphQL schema file. Can be in JSON or SDL format.
     ///  - includes: A comma-delimited string of path matching patterns to search for GraphQL operations, such as queries and
     ///  mutations. Each path matching pattern can include the following characters:
-    ///     - `*` matches 0 or more characters in a single path portion, eg: `*.graphql`
-    ///     - `?` matches 1 character, eg: `file-?.graphql`
-    ///     - `**` matches zero or more directories and subdirectories searching for matches, eg: `**/*.graphql`
+    ///     - `*` matches zero or more characters in a single path portion, eg: `*.graphql`
+    ///     - `?` matches one character in a single path portion, eg: `file-?.graphql`
+    ///     - `**` includes zero or more directories and subdirectories searching for matches, eg: `**/*.graphql`
+    ///     - `!` excludes any match, eg: `a/*.graphql,!a/file.graphql`
     public init(schemaPath: String, includes: String) {
       self.schemaPath = schemaPath
       self.includes = includes
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -15,13 +15,13 @@ public struct ApolloCodegenConfiguration {
     /// - `?` matches any single character, eg: `file-?.graphql`
     /// - `**` matches all subdirectories (deep), eg: `**/*.graphql`
     /// - `!` excludes any match only if the pattern starts with a `!` character, eg: `!file.graphql`
-    public let includes: [String]
+    public let searchPaths: [String]
 
     /// Designated initializer.
     ///
     /// - Parameters:
     ///  - schemaPath: Local path to the GraphQL schema file. Can be in JSON or SDL format.
-    ///  - includes: An array of path matching pattern strings used to find files, such as GraphQL operations (queries, mutations,
+    ///  - searchPaths: An array of path matching pattern strings used to find files, such as GraphQL operations (queries, mutations,
     ///  etc.), included for code generation. You can use absolute or relative paths for the path portion of the pattern. Relative paths
     ///  will be based off the current working directory from `FileManager`. Each path matching pattern can include the following
     ///  characters:
@@ -29,9 +29,9 @@ public struct ApolloCodegenConfiguration {
     ///     - `?` matches any single character, eg: `file-?.graphql`
     ///     - `**` matches all subdirectories (deep), eg: `**/*.graphql`
     ///     - `!` excludes any match only if the pattern starts with a `!` character, eg: `!file.graphql`
-    public init(schemaPath: String, includes: [String]) {
+    public init(schemaPath: String, searchPaths: [String]) {
       self.schemaPath = schemaPath
-      self.includes = includes
+      self.searchPaths = searchPaths
     }
   }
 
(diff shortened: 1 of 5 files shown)


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -6,19 +6,26 @@ public struct ApolloCodegenConfiguration {
...
     public let schemaPath: String
-    /// Glob of files to search for GraphQL operations. This should be used to find queries and any client schema extensions.
-    public let glob: String
+    /// A comma-delimited string of path matching patterns to search for GraphQL operations, such as queries and mutations.
+    /// 
+    /// Each path matching pattern can include the following characters:
+    /// - `*` matches 0 or more characters in a single path portion, eg: `*.graphql`
+    /// - `?` matches 1 character, eg: `file-?.graphql`
+    /// - `**` matches zero or more directories and subdirectories searching for matches, eg: `**/*.graphql`
+    public let includes: String
```
designatednerd, line 15: Is this worth setting up as an enum with associated values instead of a globstar string? Then you can have an array of them and have things defined more clearly

designatednerd (commented):
Particularly since we're not actually using much of glob's power, should we consider using something more swift-based?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: Your comment is along the lines of what I'm still working on; namely dropping the term 'glob' (because it's difficult to express its full ability and that ability is different on each system) and turning it into comma-delimited list of path matching patterns which allow for some special characters.

I'm interested to hear more about your enum idea though..

designatednerd: I was thinking something like:

```swift
enum SearchOption {
   case shallow(value: String)
   case matchingAnyCharacter(in: String)
   case anyMatchInSubdirectories(using: String)
   case excluding(using: String)
}
```

Although it'd need to be an option set to allow combinations, and I can't remember if option  sets work with enums with associated values right now. Will poke at this some on monday.

calvincestari: With https://github.com/apollographql/apollo-ios/pull/2041/commits/3723bbc4b8760543e14ddbef37e53d696f81cc07 it's an array of strings. Which gets rid of some logic and is a _slightly_ more Swifty way to pass these values in.

_I'm not yet sold on the enum idea._

designatednerd: Yeah on further investigation it looks like turning it to an option set won't work, at least not without a bunch of heavy lifting: 

Screen Shot 2021-11-29 at 11 53 46 AM

And I don't know if it'd work well if you just passed in an array of `SearchOption`. 

In general I'm not a *giant* fan of things like globstar and regex because I think they make it a lot harder for people who aren't familiar with those structures to understand what's happening. But they do exist for a reason...

calvincestari: I agree they do get complicated to explain when you start trying to express the full ability of them. That's why I've reduced it to a few special characters that should give users all they need to tell us where to find the operation files.

```
  /// - `*` matches everything but the directory separator (shallow), eg: `*.graphql`
  /// - `?` matches any single character, eg: `file-?.graphql`
  /// - `**` matches all subdirectories (deep), eg: `**/*.graphql`
  /// - `!` excludes any match, eg: `*.graphql,!file.graphql`
```

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -0,0 +1,196 @@
...
+        let patternMatches = try matches(for: String(pattern.dropFirst()))
+        excludeMatches.append(contentsOf: patternMatches)
+
+      } else {
+        let patternMatches = try matches(for: pattern)
+        includeMatches.append(contentsOf: patternMatches)
+      }
+    }
+
+    return Set<String>(includeMatches).subtracting(excludeMatches)
```
calvincestari, line 85: Excludes are processed at the end and not in order - I think this is the better logic.

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    CodegenLogger.log("Expanding globstar \(pattern)", logLevel: .debug)
+
+    let isExclude = pattern.isExclude
+    var parts = pattern.components(separatedBy: String.Globstar)
+    var firstPart = parts.removeFirst()
+    let lastPart = parts.joined(separator: String.Globstar)
+
+    if isExclude {
+      firstPart = String(firstPart.dropFirst())
```
calvincestari, line 114: Leaving the `!` in the pattern would cause `glob` to return an empty result. It is added again below so we can continue to track excludes.

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    return Set<String>(includeMatches).subtracting(excludeMatches)
+  }
+
+  /// Separates a comma-delimited string into paths, expanding any globstars and removes duplicates.
+  private func expand(_ patterns: [String]) throws -> Set<String> {
+    var paths: Set<String> = []
+    for pattern in patterns {
+      if pattern.containsExclude && !pattern.isExclude {
+        throw MatchError.invalidExclude(path: pattern)
```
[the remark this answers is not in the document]
AnthonyMDev: You might want to document this inline with the code.

Sources/ApolloCodegenLib/Glob.swift
AnthonyMDev: Document this as well probably.

AnthonyMDev (approved):
This is awesome! Thank you!

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -0,0 +1,196 @@
+import Foundation
+
+private extension String {
```
designatednerd, line 3: worth making this an extension on `ApolloExtensions` even though it's private so that it's clear this is functionality we added?

Tests/ApolloCodegenTests/ApolloCodegenConfigurationTests.swift
```diff
@@ -41,7 +44,8 @@ class ApolloCodegenConfigurationTests: XCTestCase {
   func test_validation_givenSchemaPath_doesNotExist_shouldThrow() throws {
     // given
     let schemaPath = directoryURL.appendingPathComponent(UUID().uuidString)
-    let config = ApolloCodegenConfiguration(input: .init(schemaPath: schemaPath.path),
+    let config = ApolloCodegenConfiguration(input: .init(schemaPath: schemaPath.path,
+                                                         includes: ["**/*.graphql"]),
```
designatednerd, line 48: I think this param may need to be renamed from `includes` since it also can give information about what paths need to be excluded as well. 

`searchPaths` is what comes to mind off the top of my head but there's probably a better one

Tests/ApolloCodegenTests/GlobTests.swift
```diff
@@ -0,0 +1,359 @@
...
+      self.baseURL.appendingPathComponent("other/file.oye").path // <outputFolder>/Glob/other/file.oye
+    ]
+
+    for file in files {
+      expect(try fileManager.createFile(atPath: file)).to(beTrue())
+    }
+  }
+
+  override func tearDownWithError() throws {
+    try FileManager.default.apollo.deleteDirectory(atPath: baseURL.path)
```
designatednerd, line 35: does this also need to call `super` after doing this?

Tests/ApolloCodegenTests/GlobTests.swift
```diff
@@ -0,0 +1,359 @@
...
+      baseURL.appendingPathComponent("file.one").path,
+      baseURL.appendingPathComponent("a/file.one").path,
+      baseURL.appendingPathComponent("a/b/file.one").path,
+      baseURL.appendingPathComponent("a/b/c/file.one").path,
+      baseURL.appendingPathComponent("a/b/c/d/e/f/file.one").path,
+      baseURL.appendingPathComponent("other/file.one").path
+    ]))
+  }
+
+  func test_match_givenRelativePattern_usingSingleDotPrefix_shouldUseCurrentDirectory() throws {
```
designatednerd, line 332: Probably worth also testing that if the current directory is the `a` directory, you're not getting `file.one` or `other/file.one` for this same search

Tests/ApolloCodegenTests/GlobTests.swift
```diff
@@ -0,0 +1,359 @@
...
+
+  // MARK: Tests
+
+  func test_match_givenSinglePattern_whenNoMatch_shouldReturnEmpty() throws {
+    // given
+    let pattern = baseURL.appendingPathComponent("*.xyz").path
+
+    // when
+    // <baseURL>/file.one
+    // <baseURL>/file.two
```
designatednerd, line 46: For clarity: These are all just comments because `files` are created in the setup function, correct? 

The one thing I'd be a little worried about with this method is that someone could change what's in `files` but not update these comments and then you'd get some nonsensical results

Sources/ApolloCodegenLib/Glob.swift
AnthonyMDev: If it's private, I don't think we need to do that. I'd definitely agree if it was shared outside of this file.

Tests/ApolloCodegenTests/GlobTests.swift
AnthonyMDev: Probably better to do so. Doubt it will actually be a problem, but not calling super might cause memory leaks or something?

designatednerd: I don't know about memory leaks but generally if you're overriding something and you're not explicitly trying to avoid the effects of what you're overridden, it's a good idea to call `super` in case something in the `super` implementation changes.

calvincestari: Yup, I'll add the call to `super`. An oversight on my part.

Tests/ApolloCodegenTests/GlobTests.swift
calvincestari: Yes - these are comments to easier see what files are present at the test paths. I agree they could get easily become stale. I'm happy to remove them if they've got little value; maybe the value was more for me as I was building the tests? 🤔

calvincestari: With 5db89203 the file creation is now part of each test. Cleans up a couple things; no more comments and only create what you need to satisfy the test. Overall a good change. 🙇🏻

designatednerd: I think it actually might be worth it to have all the files created in one place and have them be the same set of files for each test, especially if what you're trying to test is that certain things are *not* being found - but maybe just make the `createFiles` call one in each test in the `when` rather than the setup?

Tests/ApolloCodegenTests/GlobTests.swift
calvincestari: Good call - a test for this was added in https://github.com/apollographql/apollo-ios/pull/2041/commits/3d0d633363ef6104b7ea28a0f1925cfa80e6cb7d.

Tests/ApolloCodegenTests/GlobTests.swift
calvincestari: I prefer having duplication and each test case able to be read independently. If you're not strongly opposed I'm going to leave the files array in each test case to prevent having to refer outside the test to know what files were created. I did expand on which paths are created to cover the edges though - https://github.com/apollographql/apollo-ios/pull/2041/commits/33b961e9cbe7377b9257db6f523758babd441e1f. I'm confident in the tests.

designatednerd: Nope, not strongly opposed, fine by me.

Tests/ApolloCodegenTests/GlobTests.swift
```diff
@@ -54,7 +60,8 @@ class GlobTests: XCTestCase {
     // when
     try create(files: [
       baseURL.appendingPathComponent("file.one").path,
-      baseURL.appendingPathComponent("file.two").path
+      baseURL.appendingPathComponent("file.two").path,
+      baseURL.appendingPathComponent("other/file.one").path
```
designatednerd, line 64: complete wonkery: I like adding trailing commas to lists like this because then when you add or delete an item you don't have the comma change in the diff 🤓

Tests/ApolloCodegenTests/GlobTests.swift
```diff
@@ -29,6 +29,11 @@ class GlobTests: XCTestCase {
     }
   }
 
+  private func changeCurrentDirectory(to directory: String) throws {
```
designatednerd, line 32: 👍 makes a bunch of stuff more readable

Sources/ApolloCodegenLib/Glob.swift
designatednerd: I'm fine with that

designatednerd (approved):
Awesome work!

Tests/ApolloCodegenTests/ApolloCodegenConfigurationTests.swift
calvincestari: This last change was done in https://github.com/apollographql/apollo-ios/pull/2041/commits/afe85ad3850ee76a1def5fa9655a4b6a3d107754.

PR #2044, opened 2021-11-30
Fix `FileManagerExtensionTests` matching function

calvincestari:
I was incorrectly using the [`match`](https://github.com/Quick/Nimble#strings) function instead of [`equal`](https://github.com/Quick/Nimble#equivalence) to test String equality.
This also helped fix an incorrect test. (https://github.com/apollographql/apollo-ios/commit/a074bd5710db8229f180dfbad238c1fbcb277504) 🥳


designatednerd (approved):
Makes sense - for the record, what's the difference in the comparison mechanisms?

calvincestari:
> Makes sense - for the record, what's the difference in the comparison mechanisms?

`match` is for regex matching. So in these cases it was treating `expected` as a regex expression. Pure luck it was working. 😓 

> // Passes if 'actual' matches the regular expression defined in 'expected':
> expect(actual).to(match(expected))

PR #2046, opened 2021-12-01
Update FileManager extension function return

calvincestari:
This was the only function that had a Boolean return when the rest are throwing functions. This change brings aligns it with the rest of the extension API. Tests were updated as needed.

_Another option considered was using `@discardableResult` but I believe this was the better choice._


calvincestari:
@AnthonyMDev here's that change we spoke about earlier - thanks!

designatednerd (approved):

PR #2045, opened 2021-12-01
Compute MergedSelections as a memoized tree

AnthonyMDev:
Memoizes computation of merged selections using `MergedSelectionTree`. This requires keeping track of type scopes for all parent types in the path to the selection set.

This PR enables merged selection computation across nested child selection sets across an entire operation.

PR #2051, opened 2021-12-06
Remove `ApolloAPI` import

calvincestari:
~IR.Swift introduced the `ApolloAPI` framework to `ApolloCodegenLib` - [diff](https://github.com/apollographql/apollo-ios/pull/2045/files#diff-32108a7446a136617be773da0ed4e7ec65a7aeeda8fd594a4d62dfe2914772e7R3). The `ApolloAPI` framework was not linked to the `ApolloCodegenLib` framework though - this fixes that.~

IR.Swift incorrectly introduced the `ApolloAPI` framework to `ApolloCodegenLib` - this fixes that. See https://github.com/apollographql/apollo-ios/pull/2051#issuecomment-988212303.

#### How to test
1. Clean and build the `release/1.0-alpha-incubating` branch (latest head is d20d795a)
2. The build will fail with the error `"No such module 'ApolloAPI'"` in [IR.Swift](https://github.com/apollographql/apollo-ios/blob/release/1.0-alpha-incubating/Sources/ApolloCodegenLib/IR/IR.swift#L3).
3. Clean and build the branch for this PR (`1.0/fix-framework-linking`)
4. No more errors 🎉

### DIFF
--- a/Sources/ApolloCodegenLib/IR/IR.swift
+++ b/Sources/ApolloCodegenLib/IR/IR.swift
@@ -1,6 +1,5 @@
 import OrderedCollections
 import ApolloUtils
-import ApolloAPI
 
 class IR {
 


designatednerd:
I'm almost positive this dependency is intended, but I'd check with @AnthonyMDev when he gets back later this week to make sure.

calvincestari:
> I'm almost positive this dependency is intended, but I'd check with @AnthonyMDev when he gets back later this week to make sure.

Your comment made me question my original assumption - "that it must be necessary". Turns out it's not and I've amended this PR to simply remove the `import` statement.

I think this is a demonstration of the danger of Xcode automatically importing modules for you. From the [Xcode 13 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13-release-notes):
> Code completion helps you finish expressions that aren’t correct yet. It finds and identifies types in modules you haven’t imported, and automatically adds the necessary import. When the surrounding source has errors, code completion still infers what you mean. It offers the completion you’re looking for with a message describing how to make it valid.

I guess ["Xcode 13 - turn off auto import feature"](https://stackoverflow.com/questions/69839535/xcode-13-turn-off-auto-import-feature) on stackoverflow is an indication that other people are also being burned by it.

designatednerd (approved):

AnthonyMDev:
@designatednerd Thanks for looking out there! When I saw that e-mail, i agreed with you from memory, but after refreshing my memory, we are okay to not link `ApolloAPI` to `ApolloCodegenLib`.

`ApolloAPI` contains the code that needs to be accessible by both the `Apollo` library and the _generated models_ at run time. So the generated models import `ApolloAPI`, not `Apollo`. That way, parts of the user's codebase that use generated models, but aren't directly doing networking with `ApolloClient` don't need to import the entire `Apollo` library.

`ApolloUtils` is the module for code that is shared between the `Apollo`, `ApolloAPI`, and `ApolloCodegenLib`.

So this PR was all good. Thanks @calvincestari 

Because this isn't really clear, I'm still considering renaming `ApolloAPI` prior to `1.0` release. Would like opinions on that naming from you @calvincestari & @designatednerd. I know we went back and forth a bit on the name before and settled on `ApolloAPI`, but as I'm seeing confusion still, I'm not sure if that is right.

I wanted to go with `ApolloInternal`, but it's not actually completely internal. the generated models are probably going to `@_exported import ApolloAPI`, because files that consume the models are going to need to also be able to consume things like `GraphQLEnum` and `GraphQLNullable`. So it's not totally internal.

PR #2052, opened 2021-12-06
Change framework embed policy for UploadAPI and AnimalKingdomAPI

calvincestari:
This PR:
* changes the `ApolloUtils.framework` link from "Embed & Sign" to "Do Not Embed" in the `AnimalKingdomAPI` target.
* changes the `Apollo.framework` link from "Embed & Sign" to "Do Not Embed" in the `UploadAPI` target.

Neither of these targets are published and I can't find any reason why we would need them to be embedded and signed. The default in Xcode appears to be "Embed & Sign" which makes me think it's an oversight.


calvincestari:
https://stackoverflow.com/questions/57687170/do-not-embed-embed-sign-embed-without-signing-what-are-they-what-th

designatednerd (approved):

Issue #2033, opened 2021-11-17
Create a 'builder coordinator'


calvincestari:
The 'builder coordinator' is the orchestration layer between the code (file) builders and the codegen engine. This is likely to have some internal configuration (concurrency maximums, etc.) that should not be exposed just yet. Once it is stable then we can let users configure the builder environment.

_in:_ compilation result, configuration
_out:_ spawn file builders

calvincestari:
Closing this issue as it's no longer needed. The coordinator was only going to be necessary when concurrently generating files and we've decided to simply serially generate the files for now. It's possible that at some time in the future concurrent file generation might be needed but we'll revisit that at another time.

PR #2055, opened 2021-12-08
Generate CompilationResult

### LINKED ISSUE #2035: Generate `CompilationResult`
Dependent on completion of #2030.

Once we have all operations feed it into the graphql-js frontend to produce a compilation result. This will in turn be used with the AST builders to produce the AST used in code generation.

_in:_ schema, operation paths
_out:_ compilation result

calvincestari:
~_Created as a draft until needed changes to the graphql-js frontend are available._~

This calls into the graphql-js frontend to load, parse, merge, validate and compile all the GraphQL inputs (schema + operations) needed for code generation. The results from here will go on to the coordinator and get fed into the Swift AST generators which is coming in #2033.

Closes #2035

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -42,7 +42,9 @@ public class ApolloCodegen {
 
     let graphqlErrors = try frontend.validateDocument(schema: graphqlSchema, document: mergedDocument)
     guard graphqlErrors.isEmpty else {
-      throw Error.validationFailed(atLines: graphqlErrors.flatMap({ $0.logLines }))
+      let errorlines = graphqlErrors.flatMap({ $0.logLines })
+      CodegenLogger.log(String(describing: errorlines), logLevel: .error)
+      throw Error.validationFailed(atLines: errorlines)
     }
 
     return try frontend.compile(schema: graphqlSchema, document: mergedDocument)
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -5,12 +5,47 @@ import Foundation
 
 /// A class to facilitate running code generation
 public class ApolloCodegen {
+  public enum Error: Swift.Error, LocalizedError {
+    case validationFailed(atLines: [String])
+
+    public var errorDescription: String? {
+      switch self {
+      case let .validationFailed(lines):
+        return "Schema and Operation validation failed! Check \(lines)"
+      }
+    }
+  }
   /// Executes the code generation engine with a specified configuration.
   ///
   /// - Parameters:
   ///   - configuration: The configuration to use to build the code generation.
   public static func build(with configuration: ApolloCodegenConfiguration) throws {
     try configuration.validate()
+
+    let compilationResult = try compileResults(using: configuration.input)
+    #warning("TODO - compilationResult will be passed into the next step")
+  }
+
+  static func compileResults(
+    using input: ApolloCodegenConfiguration.FileInput
+  ) throws -> CompilationResult {
+    let frontend = try ApolloCodegenFrontend()
+
+    let schemaURL = URL(fileURLWithPath: input.schemaPath)
+    let graphqlSchema = try frontend.loadSchema(from: schemaURL)
+
+    let matches = try Glob(input.searchPaths).match()
+    let documents = try matches.map({ path in
+      return try frontend.parseDocument(from: URL(fileURLWithPath: path))
+    })
+    let mergedDocument = try frontend.mergeDocuments(documents)
+
+    let graphqlErrors = try frontend.validateDocument(schema: graphqlSchema, document: mergedDocument)
+    guard graphqlErrors.isEmpty else {
+      throw Error.validationFailed(atLines: graphqlErrors.flatMap({ $0.logLines }))
+    }
+
+    return try frontend.compile(schema: graphqlSchema, document: mergedDocument)
   }
 }
 
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -29,7 +29,7 @@ public class ApolloCodegen {
   static func compileResults(
     using input: ApolloCodegenConfiguration.FileInput
   ) throws -> CompilationResult {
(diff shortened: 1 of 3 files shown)


calvincestari:
_Created as a draft until needed changes to the graphql-js frontend are available._

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -5,12 +5,47 @@ import Foundation
...
+
+    let matches = try Glob(input.searchPaths).match()
+    let documents = try matches.map({ path in
+      return try frontend.parseDocument(from: URL(fileURLWithPath: path))
+    })
+    let mergedDocument = try frontend.mergeDocuments(documents)
+
+    let graphqlErrors = try frontend.validateDocument(schema: graphqlSchema, document: mergedDocument)
+    guard graphqlErrors.isEmpty else {
+      throw Error.validationFailed(atLines: graphqlErrors.flatMap({ $0.logLines }))
```
AnthonyMDev, line 45: This can be a project for another day but I'd like us to log the actual graphQL errors here instead of just "Failed, check lines 134 & 178". Can we create an issue for 1.0 to improve this please?

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -0,0 +1,193 @@
...
+          title
+          name
+        }
+      }
+      """.data(using: .utf8)!
+    createOperationFile(for: data, inDirectory: directoryURL)
+
+    let config = ApolloCodegenConfiguration.FileInput(
+      schemaPath: schemaPath,
+      searchPaths: [searchPath]
```
AnthonyMDev, line 114: Looking at this now, I think it would be clearer if we named this `operationSearchPaths`.

One thing I love about tests is it forces you to consume the API you are building so you can see how clear the usage of it is. haha :)

AnthonyMDev (approved):
LGTM! Let's keep moving forward! :)

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -0,0 +1,193 @@
...
+  /// }
+  ///
+  /// OPERATION:
+  /// query getBooks {
+  ///   books {
+  ///     title
+  ///     name (this will cause the validation error)
+  ///   }
+  /// }
+  /// ```
```
designatednerd, line 94: I like this comment from a theoretical perspective in terms of showing what's happening in this test, but I again worry about the comment and the actual test diverging at some point. I think given that the schema being used is right in this file throwing an explanation of the error into a `// when` comment might be a little more obvious.

AnthonyMDev: Yes, 100% agree. Not every test needs to (or should) be commented like this. I only did that because the actual setup for each test and the expected values were not easy to read and understand. 

In most cases, the tests should be self explanatory and clear just from reading them. Only when we feel like we can’t make that happen should we by commenting every test to this extent. 

I’m always fine with a quick little comment explaining why we are testing something or giving a little detail. But having the entire given and expected laid out in a comment like this is not a best practice I want to create. It’s fine when needed, but not a best practice. 

Like @designatednerd foretold, I’m already having issues with the comments and tests deviating in the tests I wrote like this, and I didn’t even know it. So this is not great. 

I’m thinking I could make those tests a little clearer to read without the comments if I created a result builder for it. But that’s a lot of work for a test utility.

Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: Are you meaning log via `CodegenLogger`? That's easy enough to add into this PR.

Something up for debate is whether we should throw `GraphQLError`s as I've got here. graphql-js doesn't throw but IMO if there are validation errors _we_ should halt code generation until they're resolved. Happy to be convinced otherwise if there is a strong opposing argument.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: It's not only operation files being consumed here though. I need to add a test where `searchPaths` includes both a schema extension and operation to ensure that they're both being merged into the compilation result. I'm not 100% certain on the schema extension syntax though, will figure that out tomorrow.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: I went back and forth on leaving in this comment for exactly the reason of being brittle and divergence - it's not a best practice I agree.

> throwing an explanation of the error into a // when comment might be a little more obvious.

I like this suggestion, thanks.

Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: I've added logging of these errors in https://github.com/apollographql/apollo-ios/pull/2055/commits/8e624b6d5633bc3e78889a2418448014b5c81597.

There are two parts to your comment:
> log the actual graphQL errors here instead of just "Failed, check lines 134 & 178"

1. The `logLines` property on GraphQLError isn't simply a line number indicator. It is instead a combination of all relevant parts that constitute the error, both for validation and syntax errors - see [here](https://github.com/apollographql/apollo-ios/blob/release/1.0-alpha-incubating/Sources/ApolloCodegenLib/Frontend/GraphQLError.swift#L37).
2. Console logging for useful debugging output.

So, this change will result in console messages like below with the same being thrown (the relevant error message I think you're looking for is at the end):
```
[ERROR - ApolloCodegenLib:ApolloCodegen.swift:46] - ["/Users/calvincestari/Projects/apollo-ios/Tests/ApolloCodegenTests/Output/Codegen/6B8FF0BF-AB8A-4B15-B697-04CDBDA23B86.graphql:4:error:Cannot query field \"name\" on type \"Book\"."]
```

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: I'm struggling to get this test written, so I've created https://github.com/apollographql/apollo-ios/issues/2058 to keep track of it and I'll sync up with @IvanGoncharov to figure out where I'm going wrong in using graphql-js.

designatednerd (approved):

PR #2056, opened 2021-12-08
Invert Dependency on JavaScriptBridge

AnthonyMDev:
JavaScriptObjects now retain JavaScriptBridge
JavaScriptBridge `wrapperMap` now has weak references to `JavaScriptObject`s

Rename `ApolloCodgenFrontend` to `GraphQLJSFrontend`

I kept the JavaScript frontend as `ApolloCodgenFrontend` because it's not just `graphql-js`. It's a wrapper for graphql-js that specifically exposes the necessary pieces for the Apollo Codegen and has a few functions that handle parsing the graphls-js return results into the compilation result.


calvincestari:
> I kept the JavaScript frontend as ApolloCodgenFrontend because it's not just graphql-js.

Do you mean the bundle? Looks like the class was [renamed](https://github.com/apollographql/apollo-ios/pull/2056/files#diff-d3c9d6c2762c2fda075e54e92fee8ff9f3a44422c6fbefbcdc56de00e5ee5522R104).

Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
```diff
@@ -266,9 +266,11 @@ public struct ApolloSchemaDownloader {
     """
   
   
-  static func downloadViaIntrospection(from endpointURL: URL,
-                                       httpMethod: ApolloSchemaDownloadConfiguration.DownloadMethod.HTTPMethod,
-                                       configuration: ApolloSchemaDownloadConfiguration) throws {
+  static func downloadViaIntrospection(
```
calvincestari, line 269: These are mostly formatting changes right - am I missing anything else?

Sources/ApolloCodegenLib/Frontend/JavaScript/rollup.config.js
```diff
@@ -10,7 +10,7 @@ const options = {
     file: "../dist/ApolloCodegenFrontend.bundle.js",
     format: "iife",
     name: "ApolloCodegenFrontend",
-    sourcemap: true,
+    sourcemap: false,
```
calvincestari, line 13: Does this remove the `ApolloCodegenFrontend.bundle.js.map` file? It looks like the file is still in the `dist` directory but empty. 😕

AnthonyMDev:
I mean the `bundle.js` file. :)

Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
```diff
@@ -126,15 +153,19 @@ class JavaScriptBridge {
...
-  // all objects that are passed over the bridge both on the Swift side and in JavaScript
-  // ('JSValue` is an Objective-C object that uses `JSValueProtect` to mark the underlying JavaScript
-  // object as inelligable for garbage collection.)
-  private var wrapperMap: [JSValue: JavaScriptObject] = [:]
+  /// We keep a map between `JSValue` objects and wrapper objects, to avoid repeatedly creating new
+  /// wrappers and to guarantee referential equality.
+  /// TODO: We may want to consider making the keys weak here, because this does mean we'll be
+  /// keeping alive all objects that are passed over the bridge in JavaScript.
+  /// ('JSValue` is an Objective-C object that uses `JSValueProtect` to mark the underlying
+  /// JavaScript object as inelligible for garbage collection.)
```
designatednerd, line 161:
```suggestion
  /// JavaScript object as ineligible for garbage collection.)
```

Sources/ApolloCodegenLib/Frontend/dist/ApolloCodegenFrontend.bundle.js
```diff
@@ -1,2 +1 @@
...
-//# sourceMappingURL=ApolloCodegenFrontend.bundle.js.map
```
designatednerd, line 2: Why is this no longer necessary? And why did the bundle map get deleted?

AnthonyMDev: It was never necessary, this is just cleanup of legacy. I changed the rollup config that creates the bundled js to remove the source mapping. As far as I can tell. that is only used for TypeScript libraries, and ours is pure JS.

Sources/ApolloCodegenLib/Frontend/JavaScript/rollup.config.js
designatednerd: ^that explains the answer to one of my questions, but now I have a new one: Why are we turning this off?

Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
AnthonyMDev: Yupp, just cleaning up the formatting here.

Sources/ApolloCodegenLib/Frontend/JavaScript/rollup.config.js
calvincestari: Do we need a source map? It's possible `true` is the default for that value in Rollup.

AnthonyMDev: Yeah that's what it does. The file was always empty, which is why I'm removing it. You're saying it's not being removed? My commit shows it deleted.
image

calvincestari: Yes you're correct, it is being deleted. I didn't look at the change properly. 👍🏻

calvincestari: > The file was always empty

It had content in before this deletion. I don't believe it was ever used though.

calvincestari (approved):

PR #2057, opened 2021-12-09
Allow periods in arguments to be ignored when parsing cacheKeys

### LINKED ISSUE #1071: Cannot return data from the SQLite cache if the query has "."
For example here is the query: 

> QUERY_ROOT.repository(name:apollo-ios,owner:apollographql).object(expression:master:Package.swift)

Apollo-ios version: 0.23.1

Iron-Ham:
This PR closes #1071 by introducing a regular expression when targeting the components of a `CacheKey`. Instead of simply splitting on `.`, we use a regular expression that matches `.` only if it is not contained within parentheses. 

Performance:
As far as performance goes... maybe this isn't the best way to go about quantifying that, but it seems that any performance hit here is... trackable, but perhaps pretty negligible?

Screen Shot 2021-12-09 at 10 21 54 AM


calvincestari:
@Iron-Ham was that performance screenshot from a test run on your desktop? Judging by [this](https://support.apple.com/en-ca/guide/iphone/iphe3fa5df43/12.0/ios/12.0) it looks like the oldest iPhone apollo-ios would support (iOS deployment target is 12.0) is an iPhone 5s. It would be great to run that same quick-dirty test on a physical older device if possible.

Iron-Ham:
@calvincestari can do!

AnthonyMDev:
@Iron-Ham Thanks for this. I'm still a bit concerned about the performance implications, just because these cache key functions are called many times during the execution of GraphQL operations. 

We have some rudimentary performance tests in the repo. I'm going to run those and see if I can get a better idea of the impact of this.

I'd also like to see more test cases to ensure this regex is correct. One test for the happy path is great, but are we certain that this is correctly ignoring `.` in all places outside of the arguments?

AnthonyMDev:
After running our performance tests, it doesn't seem to create a noticeable regression, so I'm for merging this in. @Iron-Ham  would you be willing to provide a few more test cases, to ensure that this only ignoring `.` within the arguments and that everything else still works correctly. I'm pretty certain it does, but would like to have the test coverage for posterity.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -117,3 +117,14 @@ extension SQLiteNormalizedCache: NormalizedCache {
     try self.database.clearDatabase(shouldVacuumOnClear: self.shouldVacuumOnClear)
   }
 }
+
+private extension String {
+  func split(usingRegex pattern: String) -> [String] {
+    guard let regex = try? NSRegularExpression(pattern: pattern) else { return [] }
```
AnthonyMDev, line 123: Since `NSRegularExpression is an `NSObject`, in order to reduce the overhead incurred by initializing `NSObject`, we might be able to improve performance even just slightly more if we just make this:
```
/// A regex for splitting the cache key for a graphQLField into its path components.
///
/// Matches `.` so long as it's not contained within a parentheses (ie. being used in a field argument value).
static let fieldCacheKeyComponentRegex = NSRegularExpression(pattern: "\\.(?![^()]*\\))")

func splitIntoCacheKeyComponents() -> CacheKey {
...
}
```

AnthonyMDev (changes requested):

Iron-Ham:
Writing this down for future track-ability, as this was originally posted on Slack:

I spent time reading up on the documentation around `NSRegularExpression`. Using regular expressions will not work for all cases given the flavor that `NSRegularExpression` uses: ICU. ICU doesn't support: recursion, non-fixed-width lookbehinds, or control-verbs. That means it cannot handle certain cases of nesting. 

The original solution produced the following outputs:
```
Input: my.chemical.romance 
Output: ["my", "chemical", "romance"]

Input: my.chemical.romance(the.hidden.albums)
Output: ["my", "chemical", "romance(the.hidden.albums)"]

Input: my.chemical.romance(the(very).hidden.albums)
Output: ["my", "chemical", "romance(the(very).hidden.albums)"]

Input: my.chemical.romance(the.(very)hidden.albums)
Output: ["my", "chemical", "romance(the", "(very)hidden.albums)"]
```

Note that the final example is wrong: The last two items in the array should be one string. The ICU regex spec only supports known-count nesting, so supporting one degree would look like this: `\.(?![^)(]*(?:\([^)(]*\)[^)(]*)*\))`

Instead, we should create these tokens by just scanning the string manually.

Sources/ApolloSQLite/SQLiteNormalizedCache.swift
```diff
@@ -117,3 +117,44 @@ extension SQLiteNormalizedCache: NormalizedCache {
...
+    for character in self where ["(", ")"].contains(character) {
+      if character == "(" {
+        stack.append(character)
+      } else if !stack.isEmpty && character == ")" {
+        _ = stack.popLast()
+      }
+    }
+
+    return stack.isEmpty
+  }
```
Iron-Ham, line 135: Maybe this isn't the best name for this since this isn't _strictly_ checking for balanced parentheses pairs.

AnthonyMDev (approved):
This LGTM! Anything else that needs to be done here before merging it in?

Iron-Ham:
@AnthonyMDev not afaict, so if you're good with it: merge away!

Tests/ApolloTests/Cache/CacheKeyConstructionTests.swift
```diff
@@ -0,0 +1,46 @@
+import XCTest
+@testable import ApolloSQLite
+
+final class CacheKeyConstructionTests: XCTestCase {
```
calvincestari, line 4: Would be worth adding another test that has no periods in the cache key, i.e.:
```
func testCacheKeyNoSplits() {
    let input = "mychemicalromance"
    let expected = ["mychemicalromance"]

    XCTAssertEqual(input.splitIntoCacheKeyComponents(), expected)
  }
```

AnthonyMDev (approved):

AnthonyMDev:
Thanks so much for this contribution @Iron-Ham!

Issue #573, opened 2019-06-07
Codegen should handle Int32 JS limit better


CraigSiemens:
On iOS `Int` defaults to `Int64` but Apollo Server only accepts `Int32`. If you try passing a int larger that what can fit in a `Int32`, Apollo Server will fail with an internal server error.

```
Int cannot represent non 32-bit signed integer value: 123412341234
```

js:
Sadly the GraphQL spec specified that [integers should be 32 bits](https://graphql.github.io/graphql-spec/June2018/#sec-Int), because Javascript I suppose…

If the codegen was change to use Int32 it would be correct, but UX/ergonomics wise it would a regression as most interoperability with existing code would suffer, since most swift code would use `Int` and would have to attempt to convert it to `Int32`. Not sure what is worst, a `EXC_BAD_INSTRUCTION` crash or an http error if it doesn't fit into 32 bits?

designatednerd:
I agree with @js - It'd be a pretty significant hit to usability to switch from `Int` to `Int32`. I might see if there's a way to add some kind of assertion failure or something though.

designatednerd:
Any kind of validation on this is going to have to wait until we're doing codegen in Swift

kevinmbeaulieu:
> Any kind of validation on this is going to have to wait until we're doing codegen in Swift

@designatednerd Do you mean for technical reasons or just prioritization? It'd be possible to add something like `assert(someInt.leadingZeroBitCount > 32)` to the getters/setters of `GraphQLInt`-type fields, right?

designatednerd:
Mostly prioritization. It'd likely also add some significant overhead to add that validation on every single setter for `Int` properties, so I'd like to come up with a strategy that lets us do this a bit more cleanly. I think this'll be easier once the codegen is moved to swift, tbh.

AnthonyMDev:
We've made the decision not to make the change to `Int32` for input values in the `1.0` release. The ergonomics/usability concerns came into consideration, but we also recognize that the 32 bit integer limitation is pretty artificial. Though this is a defined part of the specification, any server implementation in a language that handles 64 bit integers would have no problem accepting them from a client. By making this change, we would be imposing that limitation on everyone in a way that's not very flexible.

Furthermore, as @js questioned, we feel that the field error on a network response is a more desirable failure path than a runtime crash here.

[...]

PR #2065, opened 2021-12-14
Allow Doubles as JSONDecodables too

niklassaers:
I'm not sure why only Int would be parsed, so giving Doubles the same treatment as Int

### DIFF
--- a/Sources/Apollo/JSONStandardTypeConversions.swift
+++ b/Sources/Apollo/JSONStandardTypeConversions.swift
@@ -7,6 +7,8 @@ extension String: JSONDecodable, JSONEncodable {
         self = string
     case let int as Int:
         self = String(int)
+    case let double as Double:
+        self = String(double)
     default:
         throw JSONDecodingError.couldNotConvert(value: value, to: String.self)
     }


AnthonyMDev:
I don't really see any good reason why we shouldn't merge this in. We should probably as `Float` as well then?

@designatednerd @calvincestari Do you think this should be default behavior? For `Int`, it makes a bit more sense to me, because it's common convention to store identifiers as strings, but in `JSON`, they are going to come across as an `Int`. For a `Double` or `Float`, they can have a decimal point in them, and should be converted to `NSNumber` during JSON decoding.

designatednerd:
I think `Double` would be misleading because [GQL only natively supports `Float`](https://graphql.org/learn/schema/#scalar-types) and that could lead to some potential errors. 

We [already have baked-in support for `Float](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/JSONStandardTypeConversions.swift#L33)

niklassaers:
Hi, I've signed the license agreement.

I figure with treating it as double, we catch float and similar decimal types too, given that they can be represented as doubles? But yes, it'd probably be good to support commonly used data types that other services vend, even if it's not strictly in the spec

designatednerd:
> it'd probably be good to support commonly used data types that other services vend, even if it's not strictly in the spec

I get this on a philosophical level - the problem is that because they're not in the spec, custom scalar setup is subject to the individual whims of the server. That means that different servers can and do handle setting up how a `Double` is sent down differently, and that means any handling we try to do out of the box has a fairly good chance of being wrong. I'd much rather make it clear to people that this is something they have to know how to handle than try to handle something we _think_ is how most people do it, but have zero standards around.

I definitely get that this is super-frustrating, but I think it would be considerably more frustrating to have the baked-in handling be totally wrong.

niklassaers:
@designatednerd I get not wanting to support the whims of the server, and thus sticking to spec. However, I don't expect most users of the library to be able to easily do anything about the server. That's probably developed by someone else, and the details of that can be buried in the framework they use. At the moment, then, the only recourse is to do as I did, fork the project and add in the support. I think adding more support even out of spec is preferred, and even if we then get to a different string than they expected, they can manipulate the string afterwards. An alternative could be adding an extension point here where they can put in their logic to decode this out-of-spec result themselves.

designatednerd:
@niklassaers Is there a reason why you weren't able to use our [instructions on using custom scalars](https://www.apollographql.com/docs/ios/fetching-queries/#notes-on-working-with-custom-scalars)? That should allow you to handle custom scalars in your app of any type, including `Double`, without needing to fork.

designatednerd:
FYI, this kicked up a larger discussion internally that boils down to "Turns out [what GraphQL calls a `Float`](https://spec.graphql.org/October2021/#sec-Float) is what Swift (and Java and ObjC and Kotlin) calls a `Double` and we've been inappropriately truncating things to single-point precision all along 🤦‍♀️ 

If you scroll up in the spec from where I've linked, you'll see a note about 32 bit things needing to be sent as `Strings`. On further examination, it appears that's only about `Int` representing a 32 bit _integer_ rather than `Float` only representing a 32 bit floating-point number, because the `Float` section linked explicitly specifies that it's a double-precision floating point number. 

SO. What that means is that things presently represented as `Float` in GraphQL should be getting represented as `Double` in Swift, rather than requiring a specific `Double` custom scalar, and [we should address this long-standing issue about `Int32` validation](https://github.com/apollographql/apollo-ios/issues/573) since Swift `Int` is generally `Int64` under the hood. 

We're gonna have to back up and figure out how to deal with this, but the ultimate end of it is: I was wrong, and I based my arguments on the misunderstanding I had. We'll get it sorted, though.

niklassaers:
Happy New Year, Ellen, and thank you very much for the updates. I look forward to the updates, and will happily test them out and report back. In the meanwhile, I'll leave this patch here in case it helps others as a stop-gap

AnthonyMDev:
We had some more conversations about this today. We're going to go ahead and make this change for the new 1.0 release. This will make Custom Scalars that are sent as Doubles to convert to a String and function normally by default. We are also changing our handling of the GraphQL Float to be treated as a Swift Double.

Thanks for the input @niklassaers

PR #2066, opened 2021-12-14
Generate Operation Definition DocumentType template

AnthonyMDev:


designatednerd (approved):
Not seeing anything obvious here - I do find it a little weird that we have to wrap everything in a `DocumentType` based on whether the server is supporting APQs or not, but I think if we document _why_ that's necessary properly and mostly leave it under the hood we should be OK

AnthonyMDev:
The reason for that is because the `APQ` config determines what information we actually include here.

It's documented in `DocumentType` and `APQConfig`

For `disabled` we need to send the entire source of the query with the network request, so it has to be generated here.

For `automaticallyPersist`, we need the `operationIdentifier`, but if the id is not recognized, we still need to send the source of the query, so we generate both.

For `persistedOperationsOnly` (allowlisting), we just need the `operationIdentifier`, if the id is not recognized, we just fail the request, so we don't need to generate the entire source of the query into the operations.

Customers using `persistedOperationsOnly` have requested in the past that we provide a way for them to omit the source of their operations from the compiled code (both for code size and security purposes). So this allows for that!

calvincestari (approved):

PR #2069, opened 2021-12-15
Refactor OperationBuilder to enable building of Fragments

AnthonyMDev:
The portion of the `IR.OperationBuilder` that builds the root entity field and it's tree of selection sets needs to also be used to create the same data structures when creating a fragment.

This PR:
- Creates a new `IR.NamedFragment` class
- Refactors the root field building into a new `IR.RootEntityFieldBuilder`.
  - `IR.OperationBuilder` now configures the root entity from the operation definition and calls the `RootEntityFieldBuilder` to build the field tree.
  - Creates `IR.FragmentBuilder`, which configures the root entity for a `NamedFragment` based on the fragment name and type condition and uses `RootEntityFieldBuilder` to build the field tree.
- Refactors all of the unit tests for this.
- Stubs out the `FragmentGenerator` and it's test file, to be completed in the next PR.


calvincestari (approved):

Sources/ApolloTestSupport/XCTAssertHelpers.swift
```diff
@@ -153,9 +153,7 @@ public struct XCTFailure: Error, CustomNSError {
 }
 
 public extension Optional {
-  var xctUnwrapped: Wrapped {
-    get throws {
-      try XCTUnwrap(self)
-    }
+  func xctUnwrapped(file: String = #file, line: Int = #line) throws -> Wrapped {
+    try XCTUnwrap(self)
```
designatednerd, line 157: you actually have to pass the `file` and `line` parameters here for this to work, FYI

AnthonyMDev: Hahahah right... I knew that. Lol thanks for noticing. I'll fix it in the next PR.

designatednerd: real easy to miss when it doesn't complain about the unused params!

PR #2003, opened 2021-10-26
Say Goodbye to Apollo CLI

calvincestari:
Closes #2001 

_This PR is branched off of PR #1999 but I do not intend to merge it into that branch. Once the parent PR (#1999) is merged GitHub should automagically change the base branch to `release/1.0-alpha-incubating` which is the intended target._

On to the business of this PR:
* Removes the Apollo CLI from codegen. I believe I've caught all related code too; extensions, tests, helpers, etc. however if you know of something I've missed please let me know.
* Does not include any changes to the Swift scripts; those haven't caught up with the recent changes to `ApolloCodegenConfiguration` anyways so I'll put that into a different PR, together with a couple other related bits.

Goodbye CLI, it's been grand, you can see yourself out now - _Totsiens, Au Revoir, Auf Wiedersehen, Yasou, Aloha, Namaste, Arrivederci, Sayōnara, Adios, Hamba kahle, Do pobachennia!_

adieu

PR #2071, opened 2021-12-16
Refactor FragmentProtocols to remove HasFragments

AnthonyMDev:
And other cleanup


designatednerd:
Why is `HasFragments` no longer needed? Is it worth updating the RFC, which IIRC mentions this?

calvincestari (approved):
Nicely done!

calvincestari:
Thanks for cleaning up the CLI files. _For anyone interested they were not deleted from the file system in my PR #2003 and mistakenly only removed from the project._

AnthonyMDev:
> Why is `HasFragments` no longer needed? Is it worth updating the RFC, which IIRC mentions this?

@designatednerd, It was always an ugly solution and i figured out a better way of doing this. Rather than a `HasFragments` protocol, the `SelectionSet` now has an associated type for `Fragments` itself with a default type of a stubbed type called `NoFragments`, now you don't need to make them conform to `HasFragments`, if we generate a `Fragments` type into a `SelectionSet`, then you can access the fragments, if not, then the `fragments` property won't exist (because `fragments` is a property on a protocol extension on `SelectionSet where Fragments: FragmentContainer`).

AnthonyMDev:
As for the RFC, I plan on updating it at some point soon, I've been making a few changes and things are still in flux a little bit, so I'll make a batch update to it around the time we release Alpha 1

PR #2072, opened 2021-12-16
Add tutorial on subscriptions to match Android

designatednerd:
Pulls over the section on Subscriptions from the Android tutorial and makes it work™ for the iOS tutorial. 

Updated code is in the `main` branch of the [iOS Tutorial Repo](https://github.com/apollographql/iOSTutorial) since it doesn't interfere with anything already in there.


docs/source/tutorial/tutorial-subscriptions.md
```diff
@@ -0,0 +1,198 @@
...
+
+<img alt="A new trip was booked (rocket)" class="screenshot" src="images/screenshot_trip_booked.png" width="300"/>
+
+And you've done it! You've completed the tutorial. 
+
+## More resources
+
+There are way more things you can do with the Apollo iOS SDK, and the rest of this documentation includes info on more advanced topics like:
+
+- Using [fragments](/fragments/)
```
calvincestari, line 189: These links 404 but I assume that's because you need to view this doc in the actual tutorial environment for the paths to route correctly?

calvincestari (approved):
Looks good with minor comments on some of the copy. Nice work!

docs/source/tutorial/tutorial-subscriptions.md
```diff
@@ -0,0 +1,198 @@
+---
+title: "9. Write your first subscription"
+---
+
+In this section, you will use subscriptions to get notified whenever someone books a flight 🚀! [Subscriptions](https://graphql.org/blog/subscriptions-in-graphql-and-relay/) allow to be notified in real time whenever an event happens on your server. The [fullstack backend](https://apollo-fullstack-tutorial.herokuapp.com) supports subscriptions based on [WebSockets](https://en.wikipedia.org/wiki/WebSocket).
```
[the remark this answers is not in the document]
designatednerd: I think "you" is probably better here but yeah defintiely a word there

docs/source/tutorial/tutorial-subscriptions.md
designatednerd: I think so? My recollection is that if it were to be something where it'd actually 404 that would fail the build

PR #2076, opened 2021-12-22
`GraphQLObjectType` File Generator

calvincestari:
This is the first of the Swift file generators based on output from the IR. For this first PR I chose to implement the Schema Object (Type) file generator. There are a number of other Schema files to be generated too (Interfaces, Enums, Unions) so once we're happy with the pattern for this PR the others will be quick to implement.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -27,8 +27,16 @@ public class ApolloCodegen {
   public static func build(with configuration: ApolloCodegenConfiguration) throws {
     try configuration.validate()
 
-    #warning("TODO - compilationResult will be passed into the next step")
     let compilationResult = try compileGraphQLResult(using: configuration.input)
+
+    let ir = IR(
+      schemaName: configuration.output.schemaTypes.moduleName,
+      compilationResult: compilationResult
+    )
+
+    try generateSchemaTypeFiles(for: ir.schema.referencedTypes, configuration: configuration)
+    #warning("TODO - generate operation files")
+    #warning("TODO - generate package manager manifest")
   }
 
   static func compileGraphQLResult(
@@ -53,6 +61,21 @@ public class ApolloCodegen {
     }
 
     return try frontend.compile(schema: graphqlSchema, document: mergedDocument)
+  }
+
+  static func generateSchemaTypeFiles(
+    for referencedTypes: IR.Schema.ReferencedTypes,
+    configuration: ApolloCodegenConfiguration
+  ) throws {
+    let schemaTypesDirectory = configuration.output.schemaTypes.modulePath
+
+    try referencedTypes.objects.forEach { graphqlObject in
+      try TypeFileGenerator.generate(for: graphqlObject, in: schemaTypesDirectory)
+    }
+
+    #warning("TODO - generate enum files")
+    #warning("TODO - generate interface files")
+    #warning("TODO - generate union files")
   }
 }
 
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -1,4 +1,5 @@
 import Foundation
+import OrderedCollections
 
 // Only available on macOS
 #if os(macOS)
@@ -34,10 +35,15 @@ public class ApolloCodegen {
       compilationResult: compilationResult
     )
 
-    try generateSchemaFiles(
-      for: ir.schema.referencedTypes,
-      configuration: configuration.output.schemaTypes
-    )
+    try fileGenerators(
+      for: ir.schema.referencedTypes.objects,
+      directoryPath: configuration.output.schemaTypes.modulePath
+    ).forEach({ try $0.generateFile() })
+
+    #warning("TODO - generate schema enum files")
+    #warning("TODO - generate schema interface files")
+    #warning("TODO - generate schema union files")
+    #warning("TODO - generate schema file")
     #warning("TODO - generate operation/fragment files")
     #warning("TODO - generate package manager manifest")
   }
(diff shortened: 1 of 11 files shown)


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -44,11 +60,27 @@ public class ApolloCodegen {
...
       let errorlines = graphqlErrors.flatMap({ $0.logLines })
       CodegenLogger.log(String(describing: errorlines), logLevel: .error)
-      throw Error.validationFailed(atLines: errorlines)
+      throw Error.graphQLSourceValidationFailed(atLines: errorlines)
     }
 
     return try frontend.compile(schema: graphqlSchema, document: mergedDocument)
   }
+
+  static func generateSchemaFiles(
```
calvincestari, line 69: This function right now is better suited to an integration test rather than a unit test; arguably the fact that this function doesn't return anything also goes against it being a static function. 🤔 

A better way might be to have separate functions that return the generators for each referenced type and `build(with:)` simply gathers and calls the generate function on them sequentially. At least that would be testable to ensure `ApolloCodegen` is creating the correct generators from a given schema. That also means the file generators would need to hold on to values (referenced type and path) so their generate function doesn't need parameters.

designatednerd: I have no objections to making this not static. I think your way sounds reasonable.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -1,32 +1,54 @@
...
   public enum Error: Swift.Error, LocalizedError {
-    case validationFailed(atLines: [String])
+    /// An error occured during validation of the GraphQL schema or operations.
+    case graphQLSourceValidationFailed(atLines: [String])
 
     public var errorDescription: String? {
       switch self {
-      case let .validationFailed(lines):
-        return "Schema and Operation validation failed! Check \(lines)"
+      case let .graphQLSourceValidationFailed(lines):
```
calvincestari, line 18: Getting more specific about what part of validation failed.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
```diff
@@ -28,10 +28,31 @@ public class MockFileManager: FileManager {
   private var closures: [String: Closure] = [:]
   private var closuresToBeCalled: Set<String> = []
 
+  /// If `true` then all called closures must be mocked otherwise the call will fail. When `false` any called closure
+  /// that is not mocked will fall through to `super`. As a byproduct of `false`, all mocked closures must be called otherwise
+  /// the test will fail.
+  let strict: Bool
```
calvincestari, line 34: The previous behaviour was always strict, it's now strict be default. Turning this into a property allows for less mocking code and cleaner tests. There is the added 'safety net' in `deinit` when strict is false.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
```diff
@@ -28,10 +28,31 @@ public class MockFileManager: FileManager {
...
+  /// - Parameters:
+  ///  - strict: If `true` then all called closures must be mocked otherwise the call will fail. When `false` any called closure
+  ///  that is not mocked will fall through to `super`. As a byproduct of `false`, all mocked closures must be called otherwise
+  ///  the test will fail.
+  public init(strict: Bool = true) {
+    self.strict = strict
+  }
+
+  deinit {
+    if strict == false && allClosuresCalled == false {
```
calvincestari, line 47: This ensures that when `strict` is `false` and calls are allowed to fall through to `super`, that intentionally mocked closures aren't skipped, i.e.: they must be called.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -1,32 +1,54 @@
...
 
 /// A class to facilitate running code generation
 public class ApolloCodegen {
+  // MARK: Public
+
+  /// Errors that can occur during code generation.
   public enum Error: Swift.Error, LocalizedError {
-    case validationFailed(atLines: [String])
+    /// An error occured during validation of the GraphQL schema or operations.
+    case graphQLSourceValidationFailed(atLines: [String])
```
AnthonyMDev, line 14: I usually name my errors as nouns, not verbs. So, `graphQLSourceValidationFailure` is what I would call this.
But we don't have a style guide (and we really should), so this isn't wrong. Any opinion on which way you like this more? 

We really need to start putting together a style guide.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -348,3 +361,20 @@ extension ApolloCodegenConfiguration {
     }
   }
 }
+
+// MARK: Helper Extensions (Internal Only)
+
+extension ApolloCodegenConfiguration.SchemaTypesFileOutput {
+  var moduleName: String {
+    switch self.dependencyAutomation {
+    case let .manuallyLinked(namespace):
```
AnthonyMDev, line 370: You could actually combine these all on to one line. The compiler doesn't care that your parameter is named differently, just that they are all strings. (You'd need to bind the value to the same `moduleName` variable` though.)

Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import ApolloUtils
+
+/// The methods to conform to when building a code generation Swift file generator.
+protocol FileGenerator: Equatable {
+  associatedtype graphQLType
```
AnthonyMDev, line 6: Associated types should be capitalized. `GraphQLType`.

Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
```diff
@@ -0,0 +1,12 @@
...
+
+  var objectType: GraphQLObjectType { get }
```
AnthonyMDev, line 8: What is the reason we need to even have this variable on the protocol? Making this a generic protocol feels like it overcomplicates things. Where are we using this `objectType` property?

Sources/ApolloCodegenTestSupport/MockFileManager.swift
AnthonyMDev: At this point, we really should just consider using Cuckoo, you're building some complicated mocks that do the same thing that framework does. Not urgent to change now, but the next time you are building a complicated mock, lets just look at using a library for it.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
```diff
@@ -28,10 +28,31 @@ public class MockFileManager: FileManager {
...
+    }
+  }
+
   /// Provide a mock closure for the `FileManager` function.
   ///
   /// - Parameter closure: The mocked function closure.
-  public func set(closure: Closure) {
+  public func mock(closure: Closure) {
     closures[closure.description] = closure
     closuresToBeCalled.insert(closure.description)
```
AnthonyMDev, line 57: This is super ingenious. I don't think I would have ever thought of this!

AnthonyMDev (approved):
Overall looks good! I'm wondering if we can simply the protocol a bit, but I might be missing something.

Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: I don't have a strong opinion either way. In many other parts of code I use the noun-naming convention so I'm happy to do the same here.

Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
calvincestari: My initial response was that maybe it would make more sense when the other file generators are introduced but looking at that more I think I'm still subconsciously still planning for concurrency where all file generators have a common interface so the concurrency coordinator doesn't need to care what file generator it actually is. So when you remove concurrency then the entire need for this protocol falls away.

PR #2077, opened 2021-12-22
Refactor `MockFileManager`

calvincestari:
#### What
This is a refactor of my ill-fated attempt (PR #1999) at building a generic file manager provider that could easily be mocked. While the original `MockFileManager` worked for isolated testing of the `apollo` namespace extension it does not work in practice as a type that can be passed around with conformance to `ApolloCompatible`.

#### Why
A property of `FileManagerProvider` works fine with the previous `MockFileManager` but you then lose the `.apollo` namespace extension functions because extensions cannot inherit, so this fails.
```
extension FileManagerProvider: ApolloCompatible {}

Results in compiler error "Extension of protocol 'FileManagerProvider' cannot have an inheritance clause"
```

#### Changes
* `MockFileManager` is now just a subclass of `FileManager` with stub-able functions.
* It is 'strict' meaning that any stub-able function called without being stubbed will result in a failed test.
* A downside is that `FileManager` functions not in `MockFileManager.Closure` will pass through undetected - this was a problem with `FileManagerProvider` too. At the moment there are function stubs for all `base` functions called in the `apollo` extension. I acknowledge this is brittle though.

### DIFF
--- a/Sources/ApolloCodegenLib/FileManager+Apollo.swift
+++ b/Sources/ApolloCodegenLib/FileManager+Apollo.swift
@@ -6,23 +6,10 @@ import ApolloUtils
 
 public typealias FileAttributes = [FileAttributeKey : Any]
 
-/// A protocol to decouple `ApolloExtension` from `FileManager`. Use it to build objects that can support
-/// `ApolloExtension` behavior.
-public protocol FileManagerProvider {
-  func fileExists(atPath path: String, isDirectory: UnsafeMutablePointer<ObjCBool>?) -> Bool
-  func removeItem(atPath path: String) throws
-  @discardableResult func createFile(atPath path: String, contents data: Data?, attributes attr: FileAttributes?) -> Bool
-  func createDirectory(atPath path: String, withIntermediateDirectories createIntermediates: Bool, attributes: FileAttributes?) throws
-}
-
 /// Enables the `.apollo` etension namespace.
 extension FileManager: ApolloCompatible {}
 
-/// `FileManager` conforms to the `FileManagerProvider` protocol. If it's method signatures change both the protocol and
-/// extension will need to be updated.
-extension FileManager: FileManagerProvider {}
-
-extension ApolloExtension where Base: FileManagerProvider {
+extension ApolloExtension where Base: FileManager {
 
   public enum PathError: Swift.Error, LocalizedError, Equatable {
     case notAFile(path: String)
--- a/Sources/ApolloCodegenTestSupport/MockFileManager.swift
+++ b/Sources/ApolloCodegenTestSupport/MockFileManager.swift
@@ -45,13 +45,17 @@ public class MockFileManager: FileManager {
 
   // MARK: FileManager overrides
 
+  private func missingClosureMessage(_ function: String) -> String {
+    return "\(function) closure must be set before calling it! Check your MockFileManager instance."
+  }
+
   public override func fileExists(atPath path: String, isDirectory: UnsafeMutablePointer<ObjCBool>?) -> Bool {
     let key = #function
 
     guard
       let closure = closures[key],
       case let .fileExists(handler) = closure else {
-        XCTFail("\(key) closure must be set before calling it!")
+        XCTFail(missingClosureMessage(key))
         return false
       }
 
@@ -68,7 +72,7 @@ public class MockFileManager: FileManager {
     guard
       let closure = closures[key],
       case let .removeItem(handler) = closure else {
-        XCTFail("\(key) closure must be set before calling it!")
+        XCTFail(missingClosureMessage(key))
         return
       }
 
@@ -85,7 +89,7 @@ public class MockFileManager: FileManager {
     guard
       let closure = closures[key],
       case let .createFile(handler) = closure else {
-        XCTFail("\(key) closure must be set before calling it!")
+        XCTFail(missingClosureMessage(key))
         return false
       }
 
@@ -102,7 +106,7 @@ public class MockFileManager: FileManager {
     guard
       let closure = closures[key],
       case let .createDirectory(handler) = closure else {
-        XCTFail("\(key) closure must be set before calling it!")
+        XCTFail(missingClosureMessage(key))
         return
       }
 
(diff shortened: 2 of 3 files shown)


Sources/ApolloCodegenTestSupport/MockFileManager.swift
```diff
@@ -3,77 +3,113 @@ import ApolloCodegenLib
...
+/// Used to mock a `FileManager` instance that is compatible with the `.apollo` namespace extension.
+public class MockFileManager: FileManager {
+  /// Translates to the `FileManager` functions that can be mocked.
+  public enum Closure: CustomStringConvertible {
+    case fileExists(_ handler: (String, UnsafeMutablePointer<ObjCBool>?) -> Bool)
+    case removeItem(_ handler: (String) throws -> Void)
+    case createFile(_ handler: (String, Data?, FileAttributes?) -> Bool)
+    case createDirectory(_ handler: (String, Bool, FileAttributes?) throws -> Void)
+
+    public var description: String {
```
calvincestari, line 15: These descriptions match the output from the `#function` macro which is used in each overridden function to lookup the closure.

Tests/ApolloCodegenTests/FileManagerExtensionsTests.swift
```diff
@@ -24,643 +24,714 @@ class FileManagerExtensionTests: XCTestCase {
...
+    let mocked = MockFileManager()
+
+    mocked.set(closure: .fileExists({ path, isDirectory in
       expect(path).to(equal(self.uniquePath))
       expect(isDirectory).notTo(beNil())
 
       isDirectory?.pointee = false
       return true
 
-    }, removeItem: { (path: String) in
```
calvincestari, line 276: The `removeItem` closure was never actually being called here.

Tests/ApolloCodegenTests/FileManagerExtensionsTests.swift
```diff
@@ -24,643 +24,714 @@ class FileManagerExtensionTests: XCTestCase {
...
-    }, createFile: { (path: String, data: Data?, attr: FileAttributes?) in
+    }))
+    mocked.set(closure: .createFile({ path, data, attr in
       expect(path).to(equal(self.uniquePath))
       expect(data).to(equal(self.uniqueData))
       expect(attr).to(beNil())
 
       return false
 
-    }, createDirectory: { (path: String, createIntermediates: Bool, attr: FileAttributes?) in
```
calvincestari, line 361: The 'strict' nature of `MockFileManager` will fail tests calling functions that aren't stubbed, i.e.: don't need fail blocks like this.

Sources/ApolloCodegenLib/FileManager+Apollo.swift
```diff
@@ -6,23 +6,10 @@ import ApolloUtils
...
-
 /// Enables the `.apollo` etension namespace.
 extension FileManager: ApolloCompatible {}
 
-/// `FileManager` conforms to the `FileManagerProvider` protocol. If it's method signatures change both the protocol and
-/// extension will need to be updated.
-extension FileManager: FileManagerProvider {}
-
-extension ApolloExtension where Base: FileManagerProvider {
+extension ApolloExtension where Base: FileManager {
```
designatednerd, line 12: This should be able to be imported from either `ApolloCore` or ApolloCodegenLib` (I forget which it's declared in) without having to re-declare it.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
```diff
@@ -3,77 +3,113 @@ import ApolloCodegenLib
...
-    guard let block = createFile else {
-      XCTFail("createFile block must be set before calling this method!")
-      return false
+  public override func createFile(atPath path: String, contents data: Data?, attributes attr: FileAttributes?) -> Bool {
+    let key = #function
+
+    guard
+      let closure = closures[key],
+      case let .createFile(handler) = closure else {
+        XCTFail("\(key) closure must be set before calling it!")
```
designatednerd, line 88: You probably want to pass in `#file` and `#line` that this is being called from in the test so it's clear where in a particular test this is failing (ditto for everything else in here)

For what it's worth I've found it to be a good rule of thumb that if you're using an assertion not directly in a `testBlahBlah` method, you should pass in the file and line of the caller so that the error shows up in the test code rather than in the helper code.

Sources/ApolloCodegenLib/FileManager+Apollo.swift
calvincestari: I don't quite understand what you mean. This is the only place where the extension for `FileManager` is declared; it's the same [currently in `main`](https://github.com/apollographql/apollo-ios/blob/main/Sources/ApolloCodegenLib/FileManager%2BApollo.swift#L9). This change is essentially reverting what I'd done with `FileManagerProvider`.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
calvincestari: These are overrides of `FileManager` functions though. Changing the signature they won't match and anything in the `.apollo` namespace extension will no longer route through `MockFileManager`. I'll add another sentence into the error message to indicate the problem lies with `MockFileManager` but I think this is one place we'll have to live without `#file` and `#line`.

calvincestari: I've improved the error message to hint at where the problem will be located. Let me know if you're happy with that compromise.

Sources/ApolloCodegenLib/FileManager+Apollo.swift
designatednerd: Oh derp, I got confused, this IS in `ApolloCodegenLib`. I read good.

Sources/ApolloCodegenTestSupport/MockFileManager.swift
designatednerd: Might be worth throwing a note in here about that in case for some stupid reason the `#function` macro changes in the future

designatednerd (approved):
One suggestion, otherwise :shipit:

Sources/ApolloCodegenTestSupport/MockFileManager.swift
designatednerd: ahhhhh right, dammit. compromise LGTM

PR #2079, opened 2021-12-29
ApolloClient supports async/await

twof:
This change is non-breaking

### DIFF
--- a/Sources/Apollo/ApolloClient.swift
+++ b/Sources/Apollo/ApolloClient.swift
@@ -68,7 +68,6 @@ public class ApolloClient {
 // MARK: - ApolloClientProtocol conformance
 
 extension ApolloClient: ApolloClientProtocol {
-
   public var cacheKeyForObject: CacheKeyForObject? {
     get {
       return self.store.cacheKeyForObject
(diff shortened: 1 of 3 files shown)


designatednerd:
Hey @twof - We'll take a longer look at this after the holidays but while it's not changing anything in existing requirements, this PR does add protocol requirements to `ApolloClientProtocol` that would cause compilation failures for targets that don't support anything below the required versions for the concurrency backport. 

There's a legit argument that that's not technically breaking, but I would say anything that causes compilation failures is generally _treated_ as breaking.

Sources/Apollo/ApolloClient.swift
```diff
@@ -154,6 +153,124 @@ extension ApolloClient: ApolloClientProtocol {
                                       callbackQueue: queue,
                                       completionHandler: resultHandler)
   }
+
+#if swift(>=5.5.2) && canImport(_Concurrency)
```
twof, line 157: I'm not super clear on if `canImport(_Concurrency)` is necessary. I saw Vapor doing this for their async/await rollout.

Sources/Apollo/ApolloClient.swift
```diff
@@ -154,6 +153,124 @@ extension ApolloClient: ApolloClientProtocol {
...
+  @available(macOS 10.15, iOS 13, watchOS 8, tvOS 15, *)
```
twof, line 158: TODO: Check what the minimum available versions are for watchOS and tvOS. Does this library care about linux support?

Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -82,4 +82,89 @@ public protocol ApolloClientProtocol: AnyObject {
...
+  ///   - files: An array of `GraphQLFile` objects to send.
+  ///   - queue: A dispatch queue on which the result handler will be called. Should default to the main queue.
+  ///   - completionHandler: The completion handler to execute when the request completes or errors. Note that an error will be returned If your `networkTransport` does not also conform to `UploadingNetworkTransport`.
+  /// - Returns: An object that can be used to cancel an in progress request.
+  @available(macOS 10.15, iOS 13, watchOS 8, tvOS 15, *)
+  func upload<Operation: GraphQLOperation>(operation: Operation,
+                                           files: [GraphQLFile],
+                                           queue: DispatchQueue) async throws -> GraphQLResult<Operation.Data>
+
+#elseif swift(>=5.5.0) && swift(<5.5.2) && canImport(_Concurrency)
```
twof, line 128: @designatednerd I made the conditional compilation flags more specific, and it looks like `testAsync` is passing on both Xcode 13.0.0, and Xcode 13.2.1. There's a ton of code repetition though. I can see what I can do about that if you decide we should move forward with this PR.

twof:
> I would say anything that causes compilation failures is generally treated as breaking.

I finally understood what you meant by this. Yeah you're totally right that this is a breaking change. I'll take another crack at this later when I find the time.

Sources/Apollo/ApolloClient.swift
```diff
@@ -154,6 +153,124 @@ extension ApolloClient: ApolloClientProtocol {
                                       callbackQueue: queue,
                                       completionHandler: resultHandler)
   }
+
+#if swift(>=5.5.2)
+  @available(macOS 10.15, iOS 13, watchOS 8, tvOS 15, *)
+  public func fetch<Query>(query: Query, cachePolicy: CachePolicy, contextIdentifier: UUID?, queue: DispatchQueue) async throws -> GraphQLResult<Query.Data> where Query : GraphQLQuery {
+    return try await withCheckedThrowingContinuation { continuation in
```
TizianoCoroneo, line 160: I think this implementation will crash if we set a cachePolicy of `.returnCacheDataAndFetch`: the client will first return the cached data (if present) and then proceed to return data from the network call, invalidating the invariant from `withCheckedThrowingContinuation`:

> [You must call a resume method exactly once on every execution path throughout the program.](https://developer.apple.com/documentation/swift/checkedcontinuation)

I didn't have time to verify this, sorry

twof: I think you're right about that unfortunately 😕. Since that's the case, async/await may not be a good fit for the library. I'll poke around a little more and update later.

AnthonyMDev: Yeah, we are planning on rethinking these APIs after the 1.0 release. It may be more suitable to using some sort of `Combine` stream as the return from these methods. The other alternative would be to break the `fetch` for `.returnCacheDataAndFetch` out into a separate method.

This is certainly going to be an issue for watchers, which will have to use `Combine` or some other method there.

None of those changes are backwards compatible though, which is why they are currently slated for 2.0.

AnthonyMDev (commented):
While I'm eager to implement async/await, I'm not sure if this is something we should be doing until we come up with a better way to handle the issue of multiple fetches for the `.returnCacheDataAndFetch` policy. It's likely going to be a breaking change in that case.

@designatednerd Any other thoughts on how/if we should proceed?

Sources/Apollo/ApolloClientProtocol.swift
calvincestari: [Here's](https://github.com/swift-server/swift-service-lifecycle/issues/110) a discussion on `canImport(_Concurrency)` that was helpful.

Sources/Apollo/ApolloClient.swift
calvincestari: That's a really great catch @TizianoCoroneo! Thanks for making it known.

calvincestari (commented):
Thanks for the PR @twof. I'm not completely sold on the value of introducing async/await into the SDK right now though. The [multiple-returns-cache-policy issue](https://github.com/apollographql/apollo-ios/pull/2079#discussion_r777369161) is a great example of why we shouldn't rush into adding support for this.

IMO until we get to 2.0 and modernize the networking stack it would be better for developers to add async/await support that suits their specific needs rather than adding it for everyone.

twof:
Makes sense to me! I'll close this out for now.

PR #2080, opened 2021-12-29
Always provide a non-empty WebSocket connection close reason

s3cur3:
This is a tiny fix for something I encountered. When the connection close `data` was empty, it was being interpreted as a valid, zero-length string,  and therefore replacing the default `closeReason` of "connection closed by server". The default in this case is more informative, so it probably makes sense to guard against the empty case before replacing it.

### DIFF
--- a/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
+++ b/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
@@ -814,7 +814,7 @@ public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSock
       if receivedOpcode == .connectionClose {
         var closeReason = "connection closed by server"
         if let customCloseReason = String(data: data, encoding: .utf8),
-           customCloseReason.isEmpty {
+           customCloseReason.apollo.isNotEmpty {
           closeReason = customCloseReason
         } else {
           closeCode = CloseCode.protocolError.rawValue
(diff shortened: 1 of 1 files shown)


Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
```diff
@@ -813,7 +813,8 @@ public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSock
 
       if receivedOpcode == .connectionClose {
         var closeReason = "connection closed by server"
-        if let customCloseReason = String(data: data, encoding: .utf8) {
+        if let customCloseReason = String(data: data, encoding: .utf8),
+           customCloseReason.isEmpty {
```
designatednerd, line 817: I think this is inverted - this would mean that if the custom close reason is empty, it would fall into the `if`, while if it was not empty, it'd fall into the default case. 

There's a `.apollo.isNotEmpty` extension floating around somewhere in `ApolloUtils` that would make this clearer than just a `!`, fwiw.

s3cur3: 🤦‍♂️ 

This is what I get for trying to write one line of code without a test. I'll fix it up.

designatednerd (approved):

Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
AnthonyMDev: Thanks for this. Could we get a simple test added for it as well? I'd be happy to merge this once we have a test!

AnthonyMDev:
We are going to update this code to be compatible with recent changes and add a unit test prior to merging this for the next patch release.

PR #2093, opened 2022-01-13
File generators for other schema types

calvincestari:
This PR adds file generators for the schema types Enum, Interface, Union, InputObject and the schema metadata file.

_This is replication of the file generation pattern from #2076. The next set of PRs will implement each template to begin generating real Swift content. There will likely be changes to some of the file generators as the template requires more or less data upon initialization but I expect the `FileGenerator` protocol shouldn't need to change._

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
@@ -1,0 +1,22 @@
+import Foundation
+import OrderedCollections
+
+/// Generates a file containing schema metadata used when converting data to `Object` types at runtime.
+struct SchemaFileGenerator: FileGenerator, Equatable {
+  let name: String
+  let objectTypes: OrderedSet<GraphQLObjectType>
+  let path: String
+
+  var data: Data {
+    #warning("TODO: need correct data template")
+    return "public enum Schema {}".data(using: .utf8)!
+  }
+
+  init(name: String, objectTypes: OrderedSet<GraphQLObjectType>, directoryPath: String) {
+    self.name = name
+    self.objectTypes = objectTypes
+
+    self.path = URL(fileURLWithPath: directoryPath)
+      .appendingPathComponent("Schema.swift").path
+  }
+}
--- a/Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
@@ -3,7 +3,9 @@ import OrderedCollections
 
 /// Generates a file containing schema metadata used when converting data to `Object` types at runtime.
 struct SchemaFileGenerator: FileGenerator, Equatable {
+  /// The schema module name.
   let name: String
+  /// The `OrderedSet` of `GraphQLObjectType` objects used to build the file content.
   let objectTypes: OrderedSet<GraphQLObjectType>
   let path: String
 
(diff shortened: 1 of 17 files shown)


Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/EnumFileGeneratorTests.swift
```diff
@@ -0,0 +1,64 @@
...
+    let ratingType = try ir.schema[enum: "Rating"].xctUnwrapped()
+
+    let rootURL = URL(fileURLWithPath: CodegenTestHelper.outputFolderURL().path)
+    let fileURL = rootURL.appendingPathComponent("Rating.swift")
+
+    let mockFileManager = MockFileManager(strict: false)
+
+    mockFileManager.mock(closure: .createFile({ path, data, attributes in
+      expect(path).to(equal(fileURL.path))
+      expect(String(data: try! data.xctUnwrapped(), encoding: .utf8))
```
AnthonyMDev, line 50: We don't need to be testing the actual data in these tests. If the test is to ensure that we are generating output to the right path, then just test that.

Every time we change we the rendered output shape looks like, this test shouldn't fail and have to be changed.

Tests for the actual generated content are in the unit tests for the templates. The `EnumFileGenerator` and all other file generators should really be calling into templates that are located in the "Templates" folder, like I have for `SchemaTemplate` and the others.

Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
```diff
@@ -0,0 +1,30 @@
...
+struct SchemaFileGenerator: FileGenerator, Equatable {
+  /// The schema module name.
+  let name: String
+  /// The `OrderedSet` of `GraphQLObjectType` objects used to build the file content.
+  let objectTypes: OrderedSet<GraphQLObjectType>
+  let path: String
+
+  var data: Data? {
+    #warning("TODO: need correct data template")
+    return "public enum Schema {}".data(using: .utf8)
```
AnthonyMDev, line 14: `SchemaTemplate` already exists. This should just be `SchemaTemplate(schema: schema).render()`.

Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  /// Designated initializer.
+  ///
+  /// - Parameters:
+  ///  - name: The schema module name.
+  ///  - interfaceType: The `OrderedSet` of `GraphQLObjectType` objects used to build the file content.
+  ///  - directoryPath: The **directory** path that the file should be written to, used to build the `path` property value.
+  init(name: String, objectTypes: OrderedSet<GraphQLObjectType>, directoryPath: String) {
```
AnthonyMDev, line 23: SchemaFileGenerator should be initialized with the actual `IR.Schema` object.

AnthonyMDev (approved):
Very minor comments. This looks awesome!!

Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
AnthonyMDev: This is addressed in #2097

Sources/ApolloCodegenLib/FileGenerators/SchemaFileGenerator.swift
AnthonyMDev: This is addressed in #2097

Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/EnumFileGeneratorTests.swift
calvincestari: I've reduced the test area for all these tests in https://github.com/apollographql/apollo-ios/pull/2093/commits/894646b4ea6b550be952c7618c261bf78a3971ef.

PR #2098, opened 2022-01-15
EnumFileGenerator Template Content

calvincestari:
This implements `TemplateString` for Swift Enum types, uses it in `EnumFileGenerator`, completing the schema enum file generator.


designatednerd:
Is functionality to exclude deprecated cases going to be added later or should that be part of this?

calvincestari:
> Is functionality to exclude deprecated cases going to be added later or should that be part of this?

Good question! I don't believe we'll need deprecated enum cases for the alpha release but shortly afterwards definitely. So to track to your question I've created https://github.com/apollographql/apollo-ios/issues/2101.

designatednerd:
Cool - My recollection is that there were some other things we wanted in enums that I can't remember off the top of my head  besides respecting case from the server (something in the back of my mind is saying alphabetization, but that may just been of parameter names). 

It's probably worth combing the issues marked `codegen` to make sure all that stuff is implemented.

calvincestari:
I've been working from the [Codegen Proposal](https://github.com/apollographql/apollo-ios/blob/1.0/codegen-proposal-updates/CodegenProposal.md#object-types) as a guide for what-should-go-into-what-template but that's not to say the proposal is complete either. I've got a draft PR up for changes to the proposal as I'm working through this; if you've got changes we should get in there please feel free to add commits to it.

AnthonyMDev (approved):
Looking good.

Deprecated cases and any thing else we want can be added post alpha. Right now lets just focus on generating the bare minimum to get generated objects that actually compile and work correctly for the 90% use case.

For example, I know we will want to also add documentation to the generated code from the schemas type descriptions. That stuff can come later.

PR #2102, opened 2022-01-19
InterfaceFileGenerator Template Content

calvincestari:
This implements `TemplateString` for GraphQL Schema Interface types generating a Swift Class, and uses it in InterfaceFileGenerator.

_This is not a complete implementation of the Swift output because the `@Field` property wrapper is not complete yet._


AnthonyMDev (approved):
LGTM!

PR #2112, opened 2022-01-25
Implements `InputObjectTemplate` for GraphQL schema `input` types

calvincestari:
This is the template, and its use in the relevant file generator, that will generate Swift code for a GraphQL schema `input` object.

Supporting changes:
* Loosens the access control for `GraphQLInputField` properties to support mocking them
* Changes `GraphQLInputField.fields` to be an ordered collection for consistent output

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/InputObjectFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/InputObjectFileGenerator.swift
@@ -7,8 +7,9 @@ struct InputObjectFileGenerator: FileGenerator, Equatable {
   let path: String
 
   var data: Data? {
-    #warning("TODO: need correct data template")
-    return "public struct \(inputObjectType.name) {}".data(using: .utf8)
+    InputObjectTemplate(graphqlInputObject: inputObjectType)
+      .render()
+      .data(using: .utf8)
   }
 
   /// Designated initializer.
--- a/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
+++ b/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
@@ -68,13 +68,13 @@ public class GraphQLInputObjectType: GraphQLNamedType {
 }
 
 public class GraphQLInputField: JavaScriptObject {
-  private(set) lazy var name: String = self["name"]
+  lazy var name: String = self["name"]
   
-  private(set) lazy var type: GraphQLType = self["type"]
+  lazy var type: GraphQLType = self["type"]
   
   private(set) lazy var description: String? = self["description"]
   
-  private(set) lazy var defaultValue: Any? = self["defaultValue"]
+  lazy var defaultValue: Any? = self["defaultValue"]
     
   private(set) lazy var deprecationReason: String? = self["deprecationReason"]
 }
--- a/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
+++ b/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
@@ -1,4 +1,5 @@
 import JavaScriptCore
+import OrderedCollections
 
 // These classes correspond directly to the ones in
 // https://github.com/graphql/graphql-js/tree/master/src/type
@@ -63,8 +64,8 @@ public class GraphQLEnumValue: JavaScriptObject {
 
 public class GraphQLInputObjectType: GraphQLNamedType {
   private(set) lazy var description: String? = self["description"]
-  
-  private(set) lazy var fields: [String: GraphQLInputField] = try! invokeMethod("getFields")
+
+  lazy var fields: OrderedDictionary<String, GraphQLInputField> = try! invokeMethod("getFields")
 }
 
 public class GraphQLInputField: JavaScriptObject {
--- a/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
@@ -83,22 +83,33 @@ fileprivate extension GraphQLInputField {
 }
 
 fileprivate extension GraphQLType {
-  func render(containedInNonNull: Bool = false) -> String {
+  enum NullabilityContainer {
+    case none
+    case graphqlNullable
+    case swiftOptional
+  }
+
+  func render(nullability: NullabilityContainer = .graphqlNullable) -> String {
     switch self {
     case let .entity(type as GraphQLNamedType),
       let .enum(type as GraphQLNamedType),
       let .scalar(type as GraphQLNamedType),
       let .inputObject(type as GraphQLNamedType):
 
-      return containedInNonNull ? type.swiftName : "GraphQLNullable<\(type.swiftName)>"
+      switch nullability {
+      case .none: return type.swiftName
+      case .graphqlNullable: return "GraphQLNullable<\(type.swiftName)>"
+      case .swiftOptional: return "\(type.swiftName)?"
+      }
 
     case let .nonNull(ofType):
-      return ofType.render(containedInNonNull: true)
+      return ofType.render(nullability: .none)
 
     case let .list(ofType):
-      let inner = "[\(ofType.render(containedInNonNull: false))]"
+      let inner = "[\(ofType.render(nullability: .swiftOptional))]"
 
-      return containedInNonNull ? inner : "GraphQLNullable<\(inner)>"
+      if nullability == .graphqlNullable { return "GraphQLNullable<\(inner)>" }
+      else { return inner }
     }
   }
 }
--- a/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
@@ -1,0 +1,108 @@
+import Foundation
+import JavaScriptCore
+
+struct InputObjectTemplate {
+  let graphqlInputObject: GraphQLInputObjectType
+
+  func render() -> String {
+    TemplateString(
+    """
+    struct \(graphqlInputObject.name.firstUppercased): InputObject {
+      private(set) public var dict: InputDict
+
+      init(
+        \(InitializerParametersTemplate())
+      ) {
+        dict = InputDict([
+          \(InputDictInitializerTemplate())
+        ])
+      }
+
+      \(graphqlInputObject.fields.map({ "\(FieldPropertyTemplate($1))" }), separator: "\n\n")
+    }
+    """
+    ).value
+  }
+
+  private func InitializerParametersTemplate() -> TemplateString {
+    TemplateString("""
+    \(graphqlInputObject.fields.map({
+      "\($1.name): \($1.renderType(includeDefault: true))"
+    }), separator: ",\n")
+    """)
+  }
+
+  private func InputDictInitializerTemplate() -> TemplateString {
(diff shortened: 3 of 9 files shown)


Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
```diff
@@ -0,0 +1,108 @@
...
+  func render(containedInNonNull: Bool = false) -> String {
+    switch self {
+    case let .entity(type as GraphQLNamedType),
+      let .enum(type as GraphQLNamedType),
+      let .inputObject(type as GraphQLNamedType):
+
+      return containedInNonNull ? type.name : "GraphQLNullable<\(type.name)>"
+
+    case let .scalar(type as GraphQLNamedType):
+      let typeName = (type.name == "Boolean" ? "Bool" : type.name)
```
calvincestari, line 95: I wanted to have this come from JS but it looks like the frontend gets [this type from graphql-js](https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts#L176) and not our bridging/frontend code so here seems like the 'right' place, _could possible be moved somewhere common for the rest of the code to use if needed though._

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
```diff
@@ -0,0 +1,312 @@
...
+
+    let expected = """
+      init(
+        stringField: GraphQLNullable<String> = nil,
+        intField: GraphQLNullable<Int> = nil,
+        boolField: GraphQLNullable<Bool> = nil,
+        floatField: GraphQLNullable<Float> = nil,
+        enumField: GraphQLNullable<EnumValue> = nil,
+        inputField: GraphQLNullable<InnerInputObject> = nil,
+        listField: GraphQLNullable<[GraphQLNullable<String>]> = nil
```
calvincestari, line 155: Please double-check me on this Swift output type.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
```diff
@@ -0,0 +1,312 @@
...
+      ),
+      GraphQLInputField.mock(
+        "listField",
+        type: .list(.scalar(.string())),
+        defaultValue: nil
+      )
+    ])
+
+    let expected = """
+      init(
```
calvincestari, line 148: Let me know your thoughts on whether we should care about styling the output to obey column widths, etc. OR simply care about the most concise output possible.

It could generate warnings for some projects if they lint code and don't tell the linter to ignore these generated files.

designatednerd: > It could generate warnings for some projects if they lint code and don't tell the linter to ignore these generated files.

I don't think linters should _ever_ be run on generated files, because the code that they generate is generally set to use the standards of the project that generates them rather than the project that consumes them. 

I do think having the `init` on a separate line is worthwhile here because then you can see more clearly if there are any changes in the params of the initializer.

Sources/ApolloCodegenLib/FileGenerators/InputObjectFileGenerator.swift
```diff
@@ -7,8 +7,9 @@ struct InputObjectFileGenerator: FileGenerator, Equatable {
   let path: String
 
   var data: Data? {
-    #warning("TODO: need correct data template")
-    return "public struct \(inputObjectType.name) {}".data(using: .utf8)
+    InputObjectTemplate(graphqlInputObject: inputObjectType)
+      .render()
+      .data(using: .utf8)
```
designatednerd, line 12: We should probably be wary of how this affects things for character-based languages like Japanese and Chinese which take heavy advantages of UTF-16. Not sure exactly how to test it but definitely worth at least making an issue on it

Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
```diff
@@ -63,18 +64,18 @@ public class GraphQLEnumValue: JavaScriptObject {
...
-  private(set) lazy var name: String = self["name"]
+  lazy var name: String = self["name"]
   
-  private(set) lazy var type: GraphQLType = self["type"]
+  lazy var type: GraphQLType = self["type"]
   
   private(set) lazy var description: String? = self["description"]
   
-  private(set) lazy var defaultValue: Any? = self["defaultValue"]
+  lazy var defaultValue: Any? = self["defaultValue"]
```
designatednerd, line 78: Why are these all no longer `private (set)`?

Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
designatednerd: I'm not a giant fan of doing this as a one-liner instead of an if/else - this isn't going into client code so we don't need to be wary of saving lines. 

Might be worth doing it as a `swiftName` extension and hiding the logic in there?

calvincestari: See f9eb935ac7eee92446ce76da25b1e5450801310c for the implementation.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
calvincestari: > I don't think linters should ever be run on generated files

Agreed.

Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
calvincestari: If the setter is private then the mocking functions can't work.

Sources/ApolloCodegenLib/FileGenerators/InputObjectFileGenerator.swift
calvincestari: That's an interesting point. I created https://github.com/apollographql/apollo-ios/issues/2115 so we follow up before 1.0 GA release.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
designatednerd: Seems legit to me for an optional list of optional strings - @AnthonyMDev?

designatednerd (approved):

Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
calvincestari: The alternative to that extension is a simple `typealias` but I'm not sure how I feel about all our generated code being `Boolean` vs. `Bool` - should we care about it?

calvincestari: @AnthonyMDev do you have a preference?

designatednerd: I think a typealias is way more confusing, personally

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
AnthonyMDev: I think this is wrong? Even if the strings are optional, there is no concept of omitting a value from inside the list. It's either a value or a null. The `listField` itself can be omitted from the request data, but an element inside the list can't. 

I think this should just be `GraphQLNullable<[String?]>`. @designatednerd @martinbonnin Please tell me if I'm wrong about that.

Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
calvincestari: I generally agree because `typealias` names are usually dissimilar from the origin but in this case I could get comfortable with it given that `Boolean` as a type name is pretty clear.

AnthonyMDev: I like the `swiftName` property in the extension. Not a typealias for sure.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
AnthonyMDev: Thinking about it more, I'm pretty sure I'm correct here. This minor fix and then I think we are good to go with this one!

AnthonyMDev (changes requested):

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
designatednerd: ^yeah this makes sense. gonna be a fun one to explain in the docs 🙃

calvincestari: Is there a good case for even allowing optionals in a list? I'm questioning whether we should just always treat elements as being annotated with `!`. Realistically what action would be taken by a server when it encounters `null` in the list? Is it going to persist the list with nulls in-place? What is the semantic difference between multiple nulls in a list?

calvincestari: Bear in mind that these are for inputs so we don't need to support _receiving_ null elements in a list.

AnthonyMDev: I mean, it's supported in the GraphQL spec, so we probably need to support it. People could use it to delete elements from a list perhaps? If your server schema says it expects a list of optionals, then we have to assume that your implementation has a use case for sending something like `["a", "b", null, "d"]`.

calvincestari: The GraphQL spec examples for "[Combining List and Non-Null](https://spec.graphql.org/draft/#sec-Combining-List-and-Non-Null)" has one example that I don't understand. Could have bearing here.

| Expected Type | Internal Value | Coerced Result |
| - | - | - |
| [Int!] | [1, 2, null] | null (With logged coercion error) |

calvincestari: > People could use it to delete elements from a list perhaps?

This seems the list likely to me, although just sending a new complete (but reduced) list is essentially the same on assignment.

calvincestari: Decision - we should remain spec-compliant so if the user wants nullable list items then a Swift optional is the correct code to generate. _Feelings aside of whether input list items should ever be nullable._

calvincestari: Swift code generation fixed in https://github.com/apollographql/apollo-ios/pull/2112/commits/84453e00ff10bf04d3ebb99b5892b633dd984953 followed by more tests for list nullability in https://github.com/apollographql/apollo-ios/pull/2112/commits/259af49837c7117dbc42af967d1459b18458301a.

PR #2116, opened 2022-01-26
Implements `TypeTemplate` for GraphQL schema object types

calvincestari:
This change is for the `TypeTemplate` Swift code template and using it in `TypeFileGenerator`.

I'm open to better names than `TypeTemplate` and `TypeFileGenerator` if you have any suggestions. I considered `ObjectTemplate` and `ObjectFileGenerator` but haven't committed yet. `Type` is very broad though so the `Object` name might be more representative of what is being generated. 🤷🏻

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/TypeTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/TypeTemplate.swift
@@ -1,0 +1,20 @@
+import Foundation
+
+struct TypeTemplate {
+  let graphqlObject: GraphQLObjectType
+
+  func render() -> String {
+    TemplateString("""
+    public final class \(graphqlObject.name.firstUppercased): Object {
+      override public class var __typename: String { \"\(graphqlObject.name.firstUppercased)\" }
+
+      override public class var __metadata: Metadata { _metadata }
+      private static let _metadata: Metadata = Metadata(implements: [
+        \(graphqlObject.interfaces.map({ interface in
+      "\(interface.name.firstUppercased).self"
+        }), separator: ",\n")
+      ])
+    }
+    """).value
+  }
+}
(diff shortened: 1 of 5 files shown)


Sources/ApolloCodegenLib/Templates/TypeTemplate.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+struct TypeTemplate {
+  let graphqlObject: GraphQLObjectType
+
+  func render() -> String {
+    TemplateString("""
+    public final class \(graphqlObject.name.firstUppercased): Object {
```
designatednerd, line 8: you may want to make this `swiftName` instead of `name`

designatednerd (commented):
I think type template works fine, particularly in the context of the other templates. 

Are you planning to do some integration testing of including things like fields within a template separately?

Sources/ApolloCodegenLib/Templates/TypeTemplate.swift
calvincestari: This is to create the Swift representation of the GraphQL schema concrete types though, I believe `name` directly from the schema is what we want?

Schema type:
```
type Dog {
  name: String
}
```

Swift output:
```
public final class Dog: Object {
  ..
}
```

designatednerd: ah fair. I think maybe a bunch of what I'm thinking of is dealing with pluralization of names for arrays of things, which is more of a field template issue

calvincestari: Yes - all that 'fun' stuff will come after the first alpha release.

designatednerd: yes fun is definitely the correct word to describe what that will be 🙃

PR #2120, opened 2022-01-28
Dependency Automation - Swift Package Manager

calvincestari:
Adds support for Swift Package Manager dependency automation, generating an SPM manifest file.

_The other three automation types (`.cocoapods`, `.carthage` and `.manuallyLinked`) intentionally throw an exception._

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
@@ -30,7 +30,12 @@ struct SchemaModuleFileGenerator: FileGenerator, Equatable {
   }
 
   static func == (lhs: SchemaModuleFileGenerator, rhs: SchemaModuleFileGenerator) -> Bool {
-    return lhs.path == rhs.path
+    switch (lhs.fileGenerator, rhs.fileGenerator) {
+    case let (lhs as SwiftPackageManagerFileGenerator, rhs as SwiftPackageManagerFileGenerator):
+      return lhs == rhs
+    default:
+      return false
+    }
   }
 }
 
--- a/Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
@@ -1,0 +1,58 @@
+import Foundation
+
+/// Generates a package manifest file for the releveant depdency manager.
+struct SchemaModuleFileGenerator: FileGenerator, Equatable {
+  var path: String { fileGenerator.path }
+  var data: Data? { fileGenerator.data }
+
+  private let fileGenerator: FileGenerator
+
+  /// Designated initializer.
+  ///
+  /// - Parameters:
+  ///  - configuration: A configuration object that specifies properties such as which dependency manager and where the file
+  ///  should be output to.
+  init(_ configuration: ApolloCodegenConfiguration.SchemaTypesFileOutput) {
+    switch configuration.dependencyAutomation {
+    case let .swiftPackageManager(moduleName):
+      self.fileGenerator = SwiftPackageManagerFileGenerator(
+        moduleName: moduleName,
+        directoryPath: configuration.modulePath
+      )
+
+    default:
+      fatalError("Only Swift Package Manager is supported at the moment!")
+    }
+  }
+
+  func generateFile(fileManager: FileManager = FileManager.default) throws {
+    try fileGenerator.generateFile(fileManager: fileManager)
+  }
+
+  static func == (lhs: SchemaModuleFileGenerator, rhs: SchemaModuleFileGenerator) -> Bool {
+    return lhs.path == rhs.path
+  }
+}
+
+/// Generates a Package.swift file for Swift Package Manager support of the generated schema module files.
+fileprivate struct SwiftPackageManagerFileGenerator: FileGenerator, Equatable {
+  let path: String
+  /// The name of the library and target of the module.
+  let moduleName: String
+
+  var data: Data? {
+    SwiftPackageManagerModuleTemplate(moduleName: moduleName)
+      .render()
+      .data(using: .utf8)
+  }
+
+  /// Designated initializer.
+  ///
+  /// - Parameters:
+  ///   - moduleName: The name of the library and target of the module.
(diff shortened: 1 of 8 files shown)


Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
```diff
@@ -0,0 +1,29 @@
...
+        .iOS(.v12),
+        .macOS(.v10_14),
+        .tvOS(.v12),
+        .watchOS(.v5),
+      ],
+      products: [
+        .library(name: "\(moduleName)", targets: ["\(moduleName)"]),
+      ],
+      targets: [
+        .target(name: "\(moduleName)"),
```
calvincestari, line 24: Questionable whether we should have a `test` target? We _could_ provide some basic sanity tests for the generated code but compilation _should_ be enough.

calvincestari: On second thought maybe sanity tests of the generated queries to ensure we're providing all expected type casts isn't a bad idea?

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
```diff
@@ -0,0 +1,58 @@
...
+      throw NSError(domain: "ApolloCodegen", code: -1, userInfo: [NSLocalizedDescriptionKey: "Only Swift Package Manager is supported at the moment!"])
+    }
+  }
+
+  func generateFile(fileManager: FileManager = FileManager.default) throws {
+    try fileGenerator.generateFile(fileManager: fileManager)
+  }
+
+  static func == (lhs: SchemaModuleFileGenerator, rhs: SchemaModuleFileGenerator) -> Bool {
+    return lhs.path == rhs.path
```
AnthonyMDev, line 33: This is not right. Two generators for a Cocoapods config and a SPM config could have the same path, but not be equal.

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
AnthonyMDev: We shouldn't be adding generated tests to our user's projects. We need to be testing that everything works properly and generates all valid code in the Apollo tests, but generated code should just be assumed to work.

AnthonyMDev: We do need to add dependencies on ApolloAPI here though!

Sources/ApolloCodegenTestSupport/MockApolloCodegenConfiguration.swift
```diff
@@ -22,4 +22,16 @@ extension ApolloCodegenConfiguration {
       apqs: apqs
     )
   }
+
+  public static func mock(
```
AnthonyMDev, line 26: I added something similar in my most recent PR, so we should look and combine these after merging everything in.

AnthonyMDev (changes requested):

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
calvincestari: They should never have the same path because `SwiftPackageManagerFileGenerator` and (the future) `CocoaPodsFileGenerator` manipulate their own paths for the manifest filename. `path` on `SchemaModuleFileGenerator` routes to the private `fileGenerator` created on initialization.

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
calvincestari: > We do need to add dependencies on ApolloAPI here though!

Dang, I completely forgot about that!

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
AnthonyMDev: Okay... but that seems like an implementation detail to me. It should be checking lhs.fileGenerator == rhs.fileGenerator in my opinion. But this is really only relevant for unit tests right?

calvincestari: Should only be needed in unit tests, correct. I can change to `fileGenerator` equality, not an issue.

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
calvincestari: @AnthonyMDev, the manifest now pulls in `ApolloAPI` - https://github.com/apollographql/apollo-ios/pull/2120/commits/4e2cb16a1e646e28bc766f710a7f5b5094b3886a.

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
```diff
@@ -0,0 +1,34 @@
...
+        .iOS(.v12),
+        .macOS(.v10_14),
+        .tvOS(.v12),
+        .watchOS(.v5),
+      ],
+      products: [
+        .library(name: "\(moduleName)", targets: ["\(moduleName)"]),
+      ],
+      dependencies: [
+        .package(url: "https://github.com/apollographql/apollo-ios.git", from: "1.0.0"),
```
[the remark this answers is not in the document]
AnthonyMDev: Oh maybe not? Because it's "from: 1.0.0" it will resolve with whatever version you are pulling in from your projects package file right?

AnthonyMDev (approved):
Looking good!

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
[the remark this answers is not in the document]
calvincestari: I believe so. It will be tricker in the alpha release as we discussed though because `1.0.0` won't be a tag yet.

PR #2124, opened 2022-01-31
SelectionSet Template Rendering [4/X]

AnthonyMDev:
- Renders TypeCase and Fragment Accessors
- Renders nested entities for fields and type cases
- Fixed some bugs in the string interpolation on `TemplateString`
  - Added new "section" interpolation to `TemplateString`
- Boolean is now generated as "Bool"


calvincestari (approved):
LGTM! 🚀

PR #2126, opened 2022-02-01
Add Union and root types to referenced types in JS FrontEnd

AnthonyMDev:


Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -265,7 +275,7 @@ export function compileToIR(
           ? (typeFromAST(schema, typeNode) as GraphQLCompositeType)
           : parentType;
 
-        referencedTypes.add(typeCondition);
+        addReferencedType(typeCondition);
```
calvincestari, line 278: Is it intentional that this use of `typeCondition` doesn't need to be wrapped in `getNamedType`? _I'm not 100% sure whether TS would catch this if it's not a `GraphQLNamedType`._

Sources/AnimalKingdomAPI/Generated/Query.swift
```diff
@@ -0,0 +1,9 @@
+import ApolloAPI
+
+public final class Query: Object {
```
calvincestari, line 3: Is `Query` ever going to be useful/needed? Seems like an unintended consequence of pulling in all referenced types.

AnthonyMDev: I intentionally added it in this PR. It is 100% needed for the generated code to build and execute correctly. :)

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
AnthonyMDev: `typeFromAST` returns a named type. It replaces calling `getNamedType`. I'm not 100% sure that it's always right either. But it hasn't broken yet. If this breaks for someone we will know soon in the alpha. I don't think we could even wrap this in `getNamedType` because the `typeCondition` is already a composite (named) type. We get do `getNamedType(typeNode)` possibly. But it's redundant when we've already called `typeFromAST`.

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -127,10 +138,9 @@ export function compileToIR(
     );
 
     const source = print(operationDefinition);
-    const rootType = getOperationRootType(
-      schema,
-      operationDefinition
-    ) as GraphQLObjectType;
+    const rootType = schema.getRootType(operationType) as GraphQLObjectType;
+
+    referencedTypes.add(getNamedType(rootType));
```
calvincestari, line 143: Do I understand this change correctly that this is where `Query` is being added?

calvincestari (approved):
Looks good!

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
AnthonyMDev: That's correct!

PR #2128, opened 2022-02-01
Render Selection Set - Minor fixes

AnthonyMDev:


Tests/ApolloCodegenTests/CodeGeneration/Templates/OperationDefinitionTemplateTests.swift
```diff
@@ -147,6 +147,31 @@ class OperationDefinitionTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected, atLine: 3, ignoringExtraLines: true))
   }
 
+  func test__generate__givenQueryWithNameEndingInQuery_generatesQueryOperationWithoutDoubledTypeSuffix() throws {
```
calvincestari, line 150: 👍🏻

calvincestari (approved):

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -82,12 +83,12 @@ public struct Glob {
       }
     }
 
-    return Set<String>(includeMatches).subtracting(excludeMatches)
+    return OrderedSet<String>(includeMatches).subtracting(excludeMatches)
```
designatednerd, line 86: remind me, why are we making this an ordered set? I would think ordered set has worse perf than a vanilla set.

Tests/ApolloCodegenTests/CodeGeneration/Templates/OperationDefinitionTemplateTests.swift
designatednerd: So if you have a query that someone's incorrectly named the `BlahDeBlahMutation`, would its correctly generated name be `BlahDeBlahMutationQuery`?

Sources/ApolloCodegenLib/Glob.swift
calvincestari: It does affect performance but I expect it aids predictability and testability where items are always in the same order. I encountered that recently when dealing with input object fields.

Tests/ApolloCodegenTests/CodeGeneration/Templates/OperationDefinitionTemplateTests.swift
calvincestari: I don't think that's a bad thing; adhere to the type rather than more-likely-than-not incorrect naming.

Sources/ApolloCodegenLib/Glob.swift
designatednerd: Ah, yeah, we've also had requests for consistent ordering of fields so that makes sense

Tests/ApolloCodegenTests/CodeGeneration/Templates/OperationDefinitionTemplateTests.swift
designatednerd: I wasn't implying that's a bad thing, just validating my assumption. Might be worth adding a test that validates this behavior if it's what we want.

AnthonyMDev: Yes, that's correct. I can add another test.

Sources/ApolloCodegenLib/Glob.swift
AnthonyMDev: Yeah, this has to be ordered because it determines the order that we pass the graphQL documents into the JS compiler, which affects the order of collecting the referenced types. The schema file was rendering with objects in different orders before this.

PR #2130, opened 2022-02-02
Add documentation with selection set name to rendered selection sets

AnthonyMDev:


Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -29,20 +30,30 @@ struct SelectionSetTemplate {
...
     public struct As\(typeCase.renderedTypeName): \(schema.name).TypeCase {
       \(BodyTemplate(typeCase))
     }
     """
     ).description
   }
 
+  func SelectionSetNameDocumentation(_ selectionSet: IR.SelectionSet) -> TemplateString {
+    """
+    /// \(generatedSelectionSetName(
```
designatednerd, line 43: Do we want to add `MARK: ` here so the names show up in the jump bar in xcode?

AnthonyMDev: Sure. Will do

calvincestari (approved):
No comment other than echoing @designatednerd, happy for this to get merged once that change is in. 👍🏻

PR #2132, opened 2022-02-02
Refactor file generators to be more functional

calvincestari:
This change is to have the file generators hold onto less data as the swift output files are being generated, which is particularly relevant for operation and fragment file generators which build large parts of the IR in their generation process.


Apollo.xcodeproj/project.pbxproj
```diff
@@ -347,20 +348,21 @@
...
 		E623FD2A2797A6F4008B4CED /* InterfaceTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E623FD292797A6F4008B4CED /* InterfaceTemplate.swift */; };
 		E64F7EB827A0854E0059C021 /* UnionTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EB727A0854E0059C021 /* UnionTemplate.swift */; };
 		E64F7EBA27A085D90059C021 /* UnionTemplateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EB927A085D90059C021 /* UnionTemplateTests.swift */; };
 		E64F7EBC27A11A510059C021 /* GraphQLNamedType+Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EBB27A11A510059C021 /* GraphQLNamedType+Swift.swift */; };
 		E64F7EBF27A11B110059C021 /* GraphQLNamedType+SwiftTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EBE27A11B110059C021 /* GraphQLNamedType+SwiftTests.swift */; };
-		E64F7EC127A122300059C021 /* TypeTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EC027A122300059C021 /* TypeTemplate.swift */; };
+		E64F7EC127A122300059C021 /* ObjectTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = E64F7EC027A122300059C021 /* ObjectTemplate.swift */; };
```
calvincestari, line 357: `TypeTemplate` and `TypeFileGenerator` have been renamed to `ObjectTemplate` and `ObjectFileGenerator` (and their relevant test files) to be consistent with naming concepts used elsewhere.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
```diff
@@ -31,6 +31,21 @@ public class CompilationResult: JavaScriptObject {
     override public var debugDescription: String {
       "\(name) on \(rootType.debugDescription)"
     }
+
+    lazy var nameWithSuffix: String = {
```
calvincestari, line 35: Moved out of `OperationDefinitionTemplate` so the file generator can use the logic too. Also lazy because it'll be called more than once.

Sources/ApolloCodegenLib/Templates/FragmentTemplate.swift
```diff
@@ -2,7 +2,7 @@ struct FragmentTemplate {
 
   let fragment: IR.NamedFragment
   let schema: IR.Schema
-  let config: ApolloCodegenConfiguration
+  let config: ApolloCodegenConfiguration.FileOutput
```
calvincestari, line 5: These changes are narrowing the surface area of the configuration struct that gets passed around.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -148,144 +148,77 @@ class ApolloCodegenTests: XCTestCase {
...
 
     // then
-    expect(try ApolloCodegen.compileGraphQLResult(using: config).operations).to(beEmpty())
+    expect(try ApolloCodegen.compileGraphQLResult(config).operations).to(beEmpty())
   }
 
   // MARK: File Generator Tests
 
-  func test_fileGenerators_givenSchema_shouldCreateFileGeneratorsForUsedSchemaTypes() throws {
+  func test_fileGenerators_givenSchemaAndMultipleOperationDocuments_shouldGenerateSchemaAndOperationsFiles() throws {
```
calvincestari, line 156: This test is about verifying the complete file list output and not about the file paths specifically, therefore I don't intend to write additional tests with different configurations. My next PR that address the correct subfolder structure for file output will test the varying combinations and ensure that the correct file paths are generated.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -104,52 +70,101 @@ public class ApolloCodegen {
...
-    return objectTypes.map({ graphqlObjectType in
-      TypeFileGenerator(objectType: graphqlObjectType, directoryPath: path)
-    })
-  }
+  /// Generates Swift files for the compiled schema, ir and configured output structure.
+  static func generateFiles(
+    compilationResult: CompilationResult,
+    ir: IR,
+    config: ApolloCodegenConfiguration,
+    fileManager: FileManager = FileManager.default
```
calvincestari, line 78: I'm not a huge fan of having to expose `fileManager` in many more function definitions but I don't see another way around it for mocked testing; the alternative is actually writing to the file system.

At least this function is _internal_ so users calling ApolloCodegen won't ever know this exists.

AnthonyMDev: Yeah, it's annoying but necessary. We should not be writing to the file system in unit tests.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -104,52 +70,101 @@ public class ApolloCodegen {
...
+          schema: ir.schema,
+          config: config.output,
+          directoryPath: config.output.schemaTypes.path,
+          fileManager: fileManager
+        )
+      }
+    }
+
+    for operation in compilationResult.operations {
+      try autoreleasepool {
```
AnthonyMDev, line 151: Yay for `autoreleasepool`!

Sources/ApolloCodegenLib/FileGenerators/EnumFileGenerator.swift
```diff
@@ -1,26 +1,23 @@
...
-  let path: String
-
-  var data: Data? {
-    EnumTemplate(graphqlEnum: self.enumType)
-      .render()
-      .data(using: .utf8)
-  }
-
-  /// Designated initializer.
+struct EnumFileGenerator {
```
AnthonyMDev, line 4: Nit: We don't need to change this right now, but I usually use caseless enums for objects that are never initialized (only have static functions). That way it's literally impossible to even make an instance of them.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -148,144 +148,77 @@ class ApolloCodegenTests: XCTestCase {
...
-    let businessObject = try ir.schema[object: "Business"].xctUnwrapped()
-    let contactObject = try ir.schema[object: "Contact"].xctUnwrapped()
-    let businessTypeEnum = try ir.schema[enum: "BUSINESS_TYPE"].xctUnwrapped()
-    let searchResultUnion = try ir.schema[union: "SearchResult"].xctUnwrapped()
-    let contactInput = try ir.schema[inputObject: "ContactInput"].xctUnwrapped()
-    let mutationObject = try ir.schema[object: "Mutation"].xctUnwrapped()
+    let fileManager = MockFileManager(strict: false)
+
+    var filePaths: Set<String> = []
+    fileManager.mock(closure: .createFile({ path, data, attributes in
```
AnthonyMDev, line 174: Love this MockFileManager implementation!

AnthonyMDev (approved):
Great work! Thanks for all the clean up in this PR!

PR #2134, opened 2022-02-03
Update swift scripting documentation for alpha & refactor codegen con…

AnthonyMDev:
…figs to make template easier to use


docs/source/swift-scripting.md
```diff
@@ -82,26 +82,12 @@ Note that these instructions are updated along with newer versions of the librar
 
 This section will walk you through the already-set up code in the template project. There are two files: `main.swift` and `FileStructure.swift`. 
 
-### FileStructure
```
calvincestari, line 85: Any particular reason to remove this bit?

calvincestari (approved):

docs/source/swift-scripting.md
AnthonyMDev: I removed that entire object. Config has been moved inline to the `main.swift` file. And that config is described in those sections.

PR #2140, opened 2022-02-04
Prepare alpha release

calvincestari:
Final things in preparation of the 1.0.0 Alpha 1 release.


calvincestari:
@AnthonyMDev we can add a commit to here with the release notes, unless you've already got that prepped in another PR?

PR #2143, opened 2022-02-07
Split advanced "Creating a client" content into separate article

StephenBarlow:


docs/source/initialization.mdx
```diff
@@ -0,0 +1,44 @@
...
+public init(networkTransport: NetworkTransport,
+            store: ApolloStore)
+```
+
+Apollo iOS provides the following classes that conform to the [`NetworkTransport` protocol](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/NetworkTransport.swift):
+
+| Class | Description |
+|-------|-------------|
+| `RequestChainNetworkTransport` | Passes a request through a chain of interceptors that can interact with the request both before and after it's transmitted. Uses standard HTTP requests to communicate with the server. |
+| `WebSocketTransport` | Transmits _all_ GraphQL operations via WebSocket. Requires the `Apollo/WebSocket` sub-spec. |
```
calvincestari, line 40: > Requires the `Apollo/WebSocket` sub-spec.

The old text mentioned this only in the context of CocoaPods but I believe even for Swift Package Manager the user needs to specifically select the `WebSocket` package, so this new language is fine.

docs/source/request-pipeline.mdx
```diff
@@ -0,0 +1,636 @@
...
+
+</ExpansionPanel>
+
+> These built-in interceptors are described [below](#built-in-interceptors).
+
+### Custom interceptor providers
+
+> [See an example interceptor provider.](#example-interceptor-provider)
+
+If your use case requires it, you can create a custom class that conforms to the [`InterceptorProvider` protocol](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/InterceptorProvider.swift).
```
calvincestari, line 76:
```suggestion
If your use case requires it, you can create a custom struct or class that conforms to the [`InterceptorProvider` protocol](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/InterceptorProvider.swift).
```

calvincestari (approved):
This looks great Stephen! I really like how you've split them up and restructured the content - nice work.

PR #2147, opened 2022-02-09
[v1.0.0 Alpha 1] JavascriptError @unchecked Sendable

### LINKED ISSUE #2146: [v1.0.0 Alpha] 'Sendable' class 'JavaScriptError' cannot inherit from another class other than 'NSObject'
## Bug report

I'm trying out the new alpha version for v1.0.0 using Xcode 13.3 beta. I encountered this error while trying to run the codegen using Swift Scripting.

## Versions

- `apollo-ios` SDK version: v1.0.0-alpha-1
- Xcode version: 13.3 beta 1
- Swift version: 5.6
- Package manager: SPM

## Steps to reproduce

Try to run a Swift Scripting setup of the Apollo codegen.

Screen Shot 2022-02-09 at 12 20 22

TizianoCoroneo:
Added unchecked sendable to JavascriptError to be able to compile with Swift 5.6.

Since this is just an `Error` object I don't foresee big trouble out of this change. Sure, I would prefer to not have this issue in the first place (using value types instead of classes, and composition instead of inheritance), but that's a bigger refactor that the core team needs to evaluate separately.

Fixes #2146

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
+++ b/Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
@@ -6,7 +6,7 @@ import JavaScriptCore
 
 /// An errror thrown during JavaScript execution.
 /// See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
-public class JavaScriptError: JavaScriptObject, Error {
+public class JavaScriptError: JavaScriptObject, Error, @unchecked Sendable {
   lazy var name: String? = self["name"]
   
   lazy var message: String? = self["message"]
--- a/Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
+++ b/Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
@@ -4,7 +4,7 @@ import JavaScriptCore
 // and can be safely force unwrapped. (Even when an exception is thrown they would still return
 // a `JSValue` representing a JavaScript `undefined` value.)
 
-/// An errror thrown during JavaScript execution.
+/// An error thrown during JavaScript execution.
 /// See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
 public class JavaScriptError: JavaScriptObject, Error, @unchecked Sendable {
   // These properties were changed to read-only when `@unchecked Sendable` was added for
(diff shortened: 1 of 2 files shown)


TizianoCoroneo:
In hindsight, of course this fails CI 😁 
Maybe we should have a separate branch for Xcode 13.3 beta?

calvincestari:
OK, I think we have a workable solution here. Thanks for the base of this PR @TizianoCoroneo, I made a couple other commits that should wrap this up.

1. Updated CircleCI to use the latest images; I'm not sure why `Sendable` was failing with Xcode 13.0.0, since the [documentation](https://developer.apple.com/documentation/swift/sendable) states 13.0+, but we already had the 1.0 branch updated to use the latest so I pulled some of [that](https://github.com/apollographql/apollo-ios/commit/f6d14d3c46a109e9a5eac4c8018f7f6eba1300be) across here. No need for beta branches or conditional compiles, etc.

2. I think the root cause of this error is that [`Error` types must conform to `Sendable`](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#thrown-errors) but because `JavaScriptError` also inherits from `JavaScriptObject`, and is not a `final` class, that conformance is broken. Adding `@unchecked` overrides the type checking behavior but requires that we understand the implications and make a guarantee that the object is thread-safe.    
  Making the properties of `JavaScriptError` computed properties, i.e.: read-only, seems the safest way to offer thread-safety to callers. We do lose some performance by dropping `lazy` but it's negligible given that this is an error type and not typically accessed repeatedly in large volume.  
 An alternative was to declare them with `private (set)` but I think that leaves the door open for someone who didn't read the comment to simply create a public function that sets the values which, without locking or synchronizing, would not be thread-safe. This way you cannot set the values without explicitly changing their declaration and by that stage I hope they've read the comment.  

@TizianoCoroneo let me know if you agree and @AnthonyMDev would love to get your thoughts on this solution too.

_This PR can get merged into `main` because I need to merge `main` into the two `release/1.0` branches anyways. Once I've done that the fix will be available everywhere._

Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
```diff
@@ -6,12 +6,13 @@ import JavaScriptCore
 
 /// An errror thrown during JavaScript execution.
```
TizianoCoroneo, line 7: Somehow we both missed this typo 😆 
```suggestion
/// An error thrown during JavaScript execution.
```

TizianoCoroneo (commented):
I agree with the changes 👍

PR #2163, opened 2022-02-15
Generate Operation Variables & Field Arguments

AnthonyMDev:


calvincestari (commented):
Glad we talked about the default value stuff, feels like we got it right in that conversation.

AnthonyMDev:
@trevorblades Not sure why Netlify deploy preview is failing. Can you look at this? Netlify logs are showing 

```
Error: ENOSPC: System limit for number of file watchers reached, watch '/opt/b  uild/repo/docs/gatsby-config.js'
```

trevorblades:
> @trevorblades Not sure why Netlify deploy preview is failing. Can you look at this? Netlify logs are showing
> 
> ```
> Error: ENOSPC: System limit for number of file watchers reached, watch '/opt/b  uild/repo/docs/gatsby-config.js'
> ```

Not sure why this happened, but I just re-ran the deploy and it seems to be working properly now. In the future, this kind of thing will be avoided since we'll only build docs deploy previews for PRs that change docs content.

Sources/ApolloAPI/GraphQLEnum.swift
```diff
@@ -34,6 +34,10 @@ public enum GraphQLEnum<T: EnumType>: CaseIterable, Equatable, RawRepresentable
     self = .case(caseValue)
   }
 
+  public init(_ rawValue: String) {
```
calvincestari, line 37: Should we just make this the designated initializer signature instead of having two public methods that do the same thing?

Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
```diff
@@ -75,7 +75,7 @@ public class GraphQLInputField: JavaScriptObject {
   
   private(set) lazy var description: String? = self["description"]
   
-  lazy var defaultValue: Any? = self["defaultValue"]
+  lazy var defaultValue: GraphQLValue? = (self["astNode"] as JavaScriptObject)["defaultValue"]
```
calvincestari, line 78: 🥇

calvincestari (approved):
Nice work here, going to be great to get this out in the next alpha!

Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
AnthonyMDev: Haha thank you. This was painful to figure out... The other solution was going to require a lot of additional code in the TypeScript library that was going to further slow down the compilation of the AST. This is a little hacky, but definitely a better solution.

Long term, I'd really like us to re-write the `JavascriptFrontend` in Swift and just access the cold hard JavaScript objects in Swift directly from the start. It will be ugly and unsafe code, casting the JS objects directly without the TypeScript compiler checking and converting types for us, but it should improve performance further. Not urgent at all, but possibly a cool intern project or something haha. :)

Sources/ApolloAPI/GraphQLEnum.swift
AnthonyMDev: I might be misunderstanding you, but `RawRepresentable` requires you to have a `public init(rawValue: String)` initializer, so we can't do that.

PR #2168, opened 2022-02-19
Add `graphql-ws` protocol support

### LINKED ISSUE #1622: graphql-ws Protocol Support
## Feature request

Enable support for the [graphql-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) protocol so servers can leverage the security and maintenance improvements.

## Motivation

subscriptions-transport-ws has fallen into disrepair in recent years, and its successor, graphql-ws, introduces a completely new [protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) which would need to be implemented for compatibility. I've been seeing a lot of issues with the server library and would like not to fork subscriptions-transport-ws 😅

calvincestari:
This introduces support for the `graphql-transport-ws` WebSocket sub-protocol implemented in the [graphql-ws](https://github.com/enisdenjo/graphql-ws) library. This is a modern, actively maintained successor to the `graphql-ws` sub-protocol implemented in the [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) library.

### Notable changes:
1. Provides the user with a number of ways to select which WebSocket protocol is used. ~but defaults to using the `graphql_ws` protocol.~ This is a breaking change that requires the user to select a protocol when initializing a `WebSocket` instance.
2. Adds a new target (`SubscriptionAPI`) which is built with a single test to validate the implementation of the `graphql-transport-ws` protocol in an end-to-end test using an [Apollo Server documentation example](https://github.com/apollographql/docs-examples/tree/main/apollo-server/v3/subscriptions-graphql-ws).
3. Adds the required Apollo Server documentation example service to the chain of servers used in CI integration tests.
4. Switches to a default of node v16 to be used for integration test services, while node v12 is still used for `SimpleUploadServer`.

### How to configure which WebSocket protocol is used:
1. ~Providing a `URLRequest` with the `Sec-WebSocket-Protocol` header preconfigured. WebSocket will respect this value if preset otherwise will default to using the `graphql_ws` protocol.~
2. Use one of the initializers `WebSocket(request:webSocketProtocol:)`, `WebSocket(url:webSocketProtocol:)` or `WebSocket(url:writeQueueQOS:webSocketProtocol:)`

### Protocol differences and how we support them
All protocol messages are defined in [OperationMessage](https://github.com/apollographql/apollo-ios/blob/c18863756dfd2c1ae3ee3db48750cd28d2f2222e/Sources/ApolloWebSocket/OperationMessage.swift#L6) and the behavior logic is in [WebSocketTransport](https://github.com/apollographql/apollo-ios/blob/c18863756dfd2c1ae3ee3db48750cd28d2f2222e/Sources/ApolloWebSocket/WebSocketTransport.swift#L135). The way this is implemented now effectively means there is no segmentation of protocol behaviour and an active web socket could _receive_ a message from _any_ protocol and it would process it; not ideal.

### Future changes
This is an aged implementation that has served us well so far, and I've added tests where we had none before but is not scalable nor ideal. Major version 2.0 will focus on the networking stack and at that time we can completely overhaul how we support GraphQL over WebSocket protocols.

Closes #1622


.circleci/config.yml
```diff
@@ -43,18 +45,29 @@ commands:
           name: Start StarWars Server
           background: true
       - run:
-          command: cd SimpleUploadServer && npm install && npm start
+          command: cd SimpleUploadServer && nvm use && npm install && npm start
```
calvincestari, line 48: Required to tell nvm to actually use the `.nvmrc` file created to lock SimpleUploadServer to node v12.

Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,6 +1,33 @@
...
+        "package": "CwlPreconditionTesting",
+        "repositoryURL": "https://github.com/mattgallagher/CwlPreconditionTesting.git",
+        "state": {
+          "branch": null,
+          "revision": "c21f7bab5ca8eee0a9998bbd17ca1d0eb45d4688",
+          "version": "2.1.0"
+        }
+      },
+      {
+        "package": "Nimble",
```
calvincestari, line 23: Adding Nimble to `main`, currently it only exists in the 1.0 branches.

SimpleUploadServer/index.js
```diff
@@ -64,6 +64,8 @@ const server = new ApolloServer({
   }
 });
 
-server.listen().then(({ url }) => {
+server.listen({
+  port: 4001
```
calvincestari, line 68: The Apollo Server documentation example defaults to port `4000`. Since it's maintained by other teams and serves as a documentation example it's easier for us to move the SimpleUploadServer service to port `4001`.

Sources/ApolloWebSocket/OperationMessage.swift
```diff
@@ -7,6 +7,7 @@ final class OperationMessage {
   enum Types : String {
     case connectionInit = "connection_init"            // Client -> Server
     case connectionTerminate = "connection_terminate"  // Client -> Server
+    case subscribe = "subscribe"                       // Client -> Server
```
calvincestari, line 10: Messages that are additive so that we can continue to support both protocols.

calvincestari:
That's a good question @alexdavid. We do have ping/pong as WebSocket infrastructure, i.e.: transport level, but nothing at the protocol level. I'm going to change this to a draft PR until I'm 100% certain about the requirement. Thanks for the heads-up!

Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
```diff
@@ -68,8 +68,11 @@ public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSock
     public let code: Int
   }
 
+  /// The WebSocket sub-protocols supported.
   public enum WSProtocol: CustomStringConvertible {
```
phanikavi, line 72: I think these comments are interchanged for both protocols.

-   The transport protocol 'graphql-transport-ws' is implemented as per [graphql-ws](https://github.com/enisdenjo/graphql-ws) package.
-     The transport protocol 'graphql_ws' is implemented as per [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) package.  This is considered a legacy implementation.

calvincestari: Thanks @phanikavi, you're correct. Honestly the naming of these libraries and sub-protocols is a mess and an endless source of confusion. I'm going to rename the cases in `WSProtocol` to match [apollo-kotlin](https://github.com/apollographql/apollo-kotlin/tree/main/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws) and be closer to the library names instead. It's what most people will be familiar with rather than the sub-protocol names.

phanikavi (approved):

calvincestari:
> Thank you for implementing this.
> 
> Quick question though: the new `graphql-transport-ws` protocol specifies that [clients must send a `Pong` event to `Ping` messages](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping). Is this handled in this implementation?

@alexdavid I've added message support for ping/pong in https://github.com/apollographql/apollo-ios/pull/2168/commits/348dcf0078185ebfd94a2e6b73718379ae5d455d and it's tested in https://github.com/apollographql/apollo-ios/pull/2168/commits/6edea9bc12f346ae84daea5e3da95e8eac585991. Thanks for calling that out.

SimpleUploadServer/.nvmrc
```diff
@@ -0,0 +1 @@
+v12.22.10
```
calvincestari, line 1: SimpleUploadServer needs to be upgraded to the latest Apollo Server at some point but for now we have to do this. ☹️

Tests/ApolloTests/WebSocket/GraphqlTransportWsProtocolTests.swift
```diff
@@ -0,0 +1,276 @@
+import XCTest
+@testable import ApolloWebSocket
+import ApolloTestSupport
+import Nimble
+import Apollo
+import SubscriptionAPI
+
+class GraphqlTransportWsProtocolTests: XCTestCase {
+  private let asyncTimeout: DispatchTimeInterval = .seconds(3)
```
calvincestari, line 9: A longer async wait interval for slower environments such as CI.

Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
```diff
@@ -183,8 +199,12 @@ public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSock
...
   }
 
-  /// Used for setting protocols.
-  public init(request: URLRequest) {
+  /// Designated initializer.
+  ///
+  /// - Parameters:
+  ///   - request: A URL request object that provides request-specific information such as the URL.
+  ///   - webSocketProtocol: Protocol to use for communication over the web socket.
+  public init(request: URLRequest, webSocketProtocol: WSProtocol) {
```
AnthonyMDev, line 207: I'm fine with leaving this as is if you want, but I personally would have probably just named these parameters `protocol: WSProtocol`. We are already in the context that this is a parameter to the initialization of a `WebSocket` and the type name is `WSProtocol`, so having the parameter named `webSocketProtocol` seems redundant to me.

But this is a minor nit.

Tests/ApolloTests/WebSocket/GraphqlTransportWsProtocolTests.swift
AnthonyMDev: These tests look great, but I'm really confused why CI would need a 3 second time out. If we are mocking the Socket and the Transport, it shouldn't need all that time. These are unit tests, not integration tests, right? 

If it's taking that long, I'm wondering if there is something going wrong with a locking thread or something?

AnthonyMDev (commented):
> All protocol messages are defined in [OperationMessage](https://github.com/apollographql/apollo-ios/blob/c18863756dfd2c1ae3ee3db48750cd28d2f2222e/Sources/ApolloWebSocket/OperationMessage.swift#L6) and the behavior logic is in [WebSocketTransport](https://github.com/apollographql/apollo-ios/blob/c18863756dfd2c1ae3ee3db48750cd28d2f2222e/Sources/ApolloWebSocket/WebSocketTransport.swift#L135). The way this is implemented now effectively means there is no segmentation of protocol behaviour and an active web socket could receive a message from any protocol and it would process it; not ideal.

Yeah, we should separate our concerns better when we refactor all of this in 2.0. But this behavior shouldn't be likely to cause any tangible problems for the time being, right? Are you just calling this out to be thorough, or are you concerned about this?

Tests/ApolloTests/WebSocket/GraphqlTransportWsProtocolTests.swift
calvincestari: The longer timeout is my 'easy' fix attempt at resolving failed CI test. It's not that though, still working on resolving the failed tests.

Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
calvincestari: I've no objection to renaming them, good point about the enclosing context for a shorter parameter name.

calvincestari:
> Yeah, we should separate our concerns better when we refactor all of this in 2.0. But this behavior shouldn't be likely to cause any tangible problems for the time being, right? Are you just calling this out to be thorough, or are you concerned about this?

Not concerned since servers implement a single protocol. Calling it out as behaviour we should not carry forward into 2.0.

AnthonyMDev (approved):
Awesome job!

PR #2169, opened 2022-02-21
Ensure SelectionSet field accessors are generated with swift camelCasing

### LINKED ISSUE #2167: [v1.0.0 alpha] PascalCase GraphQL operation names generate code that doesn't compile
## Bug report

When generating code from a GraphQL schema that has `PascalCase` query / mutation names, the generated code does not compile because of "ambiguous use of..." errors.

## Versions

- `apollo-ios` SDK version: v1.0.0-alpha.1
- Xcode version: Version 13.2.1 (13C100)
- Swift version: 5.5.3
- Package manager: SPM

## Steps to reproduce

1. Use a GraphQL server that has `PascalCase` operation names. In my case, https://graphql.anilist.co/ (GraphiQL here: https://anilist.co/graphiql)
    - For this example, I'll use the `Staff` query
2. Write some GraphQL 👨‍💻 
    ```graphql
    query GetStaffDetail($id: Int!) {
        Staff(id: $id) {
            name {
                first
                last
                native
            }
        }
    }
    ```
3. Generate code
4. Build

The generated code looks like this:

```swift
public struct Data: AniListAPI.SelectionSet {
    // ...snip
    public static var selections: [Selection] { [
      .field("Staff", Staff?.self), // Compiler error: ⛔️ Ambiguous use of 'Staff'
    ] }

    public var Staff: Staff? { data["Staff"] }

    /// Staff
    public struct Staff: AniListAPI.SelectionSet {
        // ...snip
    }
}
```

This code doesn't compile because of the naming collision between `public var Staff` and `public struct Staff`.

The same query resulted in code that compiled successfully using apollo-ios v0.x.

## Further details

I'm able to work around this by using aliases on _all_ operations. If I tweak the example above:

```graphql
query GetStaffDetail($id: Int!) {
    staff: Staff(id: $id) {
        name {
            first
            last
            native
        }
    }
}
```

This generates code that compiles:

```swift
public struct Data: AniListAPI.SelectionSet {
    // ...snip
    public static var selections: [Selection] { [
      .field("Staff", alias: "staff", Staff?.self),
    ] }
    
    public var staff: Staff? { data["staff"] }
    
    /// Staff
    public struct Staff: AniListAPI.SelectionSet {
        // ...snip
    }
}
```

This is less than ideal, since I'm likely to forget to alias new queries I write 😅

AnthonyMDev:
Fixes #2167.


calvincestari (approved):
LGTM!

PR #2173, opened 2022-02-24
[v1.0.0 alpha] Make fragments property public

klanchman:
I'm getting a lot of compiler errors in my project using the alpha since the `fragments` property isn't accessible. I'm hoping that it was meant to be `public` 😄


AnthonyMDev:
Yes, this was just a simple oversight. Whoops 😅 . Thank you!

Issue #2171, opened 2022-02-22
Experimental Client Controlled Nullability support


twof:
## Feature request

Support for [Client Controlled Nullability](https://github.com/graphql/graphql-js/pull/3418).
Spec: https://github.com/graphql/graphql-spec/pull/895
Client Controlled Nullability (CCN) is still in Stage 1 of the RFC process, but has gotten the go-ahead to be released on an experimental basis. I would be happy to begin doing the integration for Apollo iOS.

## Motivation

CCN is a highly requested feature. Apollo Android introduced @required which is basically an equivalent. The intent is to get early feedback from real world use cases. 

## Proposed solution

CCN is going to be introduced in parts as they go through review, starting with code for parsing and validation, which is enough to do code generation. The feature is available behind [a parameter](https://github.com/graphql/graphql-js/pull/3418/files#diff-297a41e54c25bbd5c846bb55d7acaaca9f2c9f5fa654b19004da4efbf9a88154R123) so it can be turned off and on as needed. It's being released in the form of unstable canary releases, so there is no guarantee with regards to breaking changes. CCN is likely to evolve as aspects of it fall in and out of favor with the GraphQL working group. 

With the current implementation, code generation should more or less work out of the box. Decisions will need to be made around caching behaviour.

## Outstanding Questions

Should we branch off of the WIP 1.0 branch? Could CCN support be a part of 1.0?
Tagging @AnthonyMDev since he said [we should talk when we were ready to start integrating.](https://github.com/apollographql/apollo-ios/pull/1876#issuecomment-989053888)

AnthonyMDev:
Really excited about this! I think it would only make sense to add it in 1.0. You can branch off the `1.0-alpha-incubating` branch to being work on this. We'll be merging things into that branch pretty regularly as we work on 1.0, so rebasing will need to be frequent. It's not a stable branch yet (hence alpha), so I don't want to create too much churn for you on the implementation as things change on our side.

I'm happy to help explain anything about what we are doing or how the code for 1.0 works. I suspect the bulk of the work for this is going to be in Code Generation engine, which is quite a complex beast.

calvincestari:
Excited to see this progress, thanks for championing it!

twof:
I think I've got the parsing step working: https://github.com/apollographql/apollo-ios/pull/2172

[...]

PR #2178, opened 2022-02-25
Experimental CCN

twof:
Related: https://github.com/apollographql/apollo-ios/issues/2171

Notes for reviewers:
- This PR enables CCN for the `.animalKingdom` target in SwiftScripts. Incantation: `swift run Codegen --target=AnimalKingdom --package-type=SwiftPackageManager`
- In the AnimalKingdomAPI, operations that utilize CCN are stored under `graphql/ccnGraphql`. Tests and SwiftScripts that don't have CCN enabled will only search for operations with `*.graphql` and those that have CCN enabled will search with `**/*.graphql`.


Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -28,6 +28,113 @@ import {
 import * as ir from "./ir";
 import { valueFromValueNode } from "./values";
 
+import type { GraphQLOutputType } from 'graphql';
```
twof, line 31: This was pulled over from the CCN implementation. I'll upstream it once I've got everything else working

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -28,6 +28,113 @@ import {
...
+
+        // We're working with the inner-most type
+        const nextType = typeStack.pop();
+
+        // There's no way for nextType to be null if both type and nullabilityNode are valid
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return getNullableType(nextType!);
+      },
+    },
+    // ListNullabilityDesignator: {
```
twof, line 96: This bit wasn't compiling. Gotta come back to it

AnthonyMDev:
I had a conversation yesterday with @calvincestari and @martinbonnin about this. We have some ideas about validation. At some point, once CCN is a stable part of the spec and all the major GraphQL servers have fully implemented support on the back end, this will work great. But if this feature is used currently with a server that doesn't support CCN, you could get run time crashes when data we expect to exist just doesn't. 

However in the mean time, we believe a client implementation of this is viable without server support. A client could use the CCN modifiers for code generation and strip them from the operation before sending it to the server. In that case, the client would need to validate that all required data does exist during execution on the response data. This means we would need to implement the entire behavior of the spec in our executor client-side, but it gives us the ability to provide CCN support without waiting for servers to implement it.

@twof What are your thoughts on this? Does it seem viable? You are much more intimately familiar with this than we are. Are we missing any pieces here that require us to depend on a spec-compliant server implementation of CCN?

twof:
@AnthonyMDev 
> But if this feature is used currently with a server that doesn't support CCN, you could get run time crashes when data we expect to exist just doesn't.

I *think* what will happen is that the server will respond with an operation parsing error and no data if you send something with CCN syntax to a server that doesn't support it, but I'll write some tests to make sure that's the case. 

This is a good point though. I see that [Apollo's server supports stream/defer](https://www.apollographql.com/blog/community/backend/introducing-defer-in-apollo-server/). Do any of the clients? Did yall run into problems with feature mismatch between clients and servers for stream/defer?

Sources/AnimalKingdomAPI/Generated/Package.resolved
```diff
@@ -1,43 +0,0 @@
-{
```
twof, line 1: gotta bring this back

Sources/AnimalKingdomAPICCN/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
```diff
@@ -0,0 +1,63 @@
+// @generated
+// This file was automatically generated and should not be edited.
+
+import ApolloAPI
+
+public class AllAnimalsQueryCCNQuery: GraphQLQuery {
+  public let operationName: String = "AllAnimalsQueryCCN"
+  public let document: DocumentType = .notPersisted(
```
twof, line 8: code generation looks like it's working wonderfully for `!`/`?` 😊

Next up, list nullability syntax

Sources/AnimalKingdomAPICCN/GeneratedCCN/Package.swift
```diff
@@ -3,22 +3,22 @@
 import PackageDescription
```
twof, line 3: I'm not sure why it says this file moved. I'll have to look at it more.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -268,6 +268,8 @@ public struct ApolloCodegenConfiguration {
   /// See `APQConfig` for more information on Automatic Persisted Queries.
   public let apqs: APQConfig
 
+  public let experimentalClientControlledNullability: Bool
```
twof, line 271: We might want to move this into its own struct to set us up for other experimental features that might crop up in the future.

Sources/ApolloCodegenLib/Frontend/GraphQLJSFrontend.swift
```diff
@@ -87,14 +87,26 @@ public final class GraphQLJSFrontend {
...
   /// Syntax errors will result in throwing a `GraphQLError`.
-  public func parseDocument(from fileURL: URL) throws -> GraphQLDocument {
-    return try library.call("parseDocument", with: makeSource(from: fileURL))
+  public func parseDocument(
+    from fileURL: URL,
+    experimentalClientControlledNullability: Bool = false
+  ) throws -> GraphQLDocument {
+    return try library.call(
+      "parseDocument",
+      with: makeSource(from: fileURL), experimentalClientControlledNullability
```
twof, line 108: fix syntax

Sources/ApolloCodegenTestSupport/Resources.swift
```diff
@@ -8,11 +8,23 @@ public struct Resources {
...
+
+  public static let CCNGraphQLOperations: [URL] = Bundle.urls(
+    forResourcesWithExtension: "graphql",
+    subdirectory: "ccnGraphql"
+  )!
+
+  public static let CCNAnimalKingdomSchema = Bundle.url(
+    forResource: "AnimalSchema",
+    withExtension: "graphqls",
+    subdirectory: "ccnGraphql"
```
twof, line 28: I've made a copy of `AnimalKingdomAPI` that has CCN enabled, and then the `.graphql` and the `.graphqls` files from both targets are copied to bundle resources by `ApolloCodegenTestSupport` into separate directories so everything can be tested in `ApolloCodegenTests`.

Sources/AnimalKingdomAPICCN/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
twof: Done!

Apollo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,6 +1,15 @@
...
     "pins": [
+      {
+        "package": "Apollo",
+        "repositoryURL": "https://github.com/apollographql/apollo-ios.git",
+        "state": {
+          "branch": null,
+          "revision": "9729f7431c6003b2b61e694e1bf94b8038069cb6",
+          "version": "1.0.0-alpha.2"
+        }
+      },
```
twof, line 12: This seems like it shouldn't be here

twof:
@AnthonyMDev 

> But if this feature is used currently with a server that doesn't support CCN, you could get run time crashes when data we expect to exist just doesn't.

I checked on this, and it will qualify as a malformed request and throw a parsing error, so Apollo iOS will handle this situation however it currently handles parsing errors. Is that acceptable? 

> However in the mean time, we believe a client implementation of this is viable without server support. A client could use the CCN modifiers for code generation and strip them from the operation before sending it to the server. In that case, the client would need to validate that all required data does exist during execution on the response data. This means we would need to implement the entire behavior of the spec in our executor client-side, but it gives us the ability to provide CCN support without waiting for servers to implement it.

This is possible. One way this might be done without reimplementing the server spec is that I could expose GraphQL.js methods that do null propagation, so Apollo-iOS would take an operation, strip out the designators, send it, get the response, pass it off to GraphQL.js for null propagation, and then handle the rest like normal. Doing that might also make it easier for other clients to do the same thing if that's something they want. 

My one concern about all this is the tradeoff. It's a stop-gap solution until CCN officially enters the spec. If we think that's a ways off, it may be worth the work, but it will be quite a bit of work. I'll let yall decide on that.

Another thing we could do is release without the behavior you're describing, see if this is a common problem for people to run into, and then address it. Yall know your users better than I do, so you might already have enough information to know that people are definitely going to run into issues.

I'm going to start writing up a rollout plan for Yelp on Friday. Let me know if you want to meet over Zoom at some point to coordinate, or if there's anything else I can do to make things smooth. I think this PR is nearly ready for review, probably next week or the week after.

AnthonyMDev:
@twof I'll set aside a few hours to go over this PR thoroughly early next week! Looking forward to having this included.

I don't think it's necessary to support CCN strictly client side right now in order to get it merged in for 1.0. If we are seeing slow adoption server-side and want to make that happen in the future, I'll consider it.

twof:
Great! There are a few things wrong with SPM that I've noted, but everything else is looking alright. I know this is a pretty long PR, so let me know if it would be helpful for you if I broke it into smaller chunks. I think I've covered the relevant tests, but if I've missed any, please point me in the right direction and I'll be happy to fill in any gaps. Thanks!

AnthonyMDev (changes requested):
This is looking awesome! Excited to get this in. :) Only a few minor notes:

- Instead of creating a whole separate target for `AnimalKingdomAPICCN`, we should just include the CCN operations in the same `AnimalKingdomAPI` target we already have. I'd like to confirm that having the experimental feature flag turned on doesn't break operations that __don't__ use CCN. It will also just make this easier to maintain.
- Changing `ParseOptions` to `ExperimentalFeatures` as noted in inline PR comments.

Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
```diff
@@ -13,7 +13,7 @@
     "npm": ">=7"
   },
   "dependencies": {
-    "graphql": "^16.0.1",
+    "graphql": "16.3.0-canary.pr.3510.5099f4491dc2a35a3e4a0270a55e2a228c15f13b",
```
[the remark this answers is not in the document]
twof: I would love to merge to master and have CCN be part of a release, and I agree that relying on me to put out updated canary releases is not ideal. The current blocker is that Ivan is the sole active person with maintainer permissions for the GraphQL.js. No one else readily involved in the project has permissions to merge or do releases. There are some folks from The Guild who would be interested in helping out, so I'm going to float that at the next working group meeting.

I'll leave it up to yall if you want to hold off on merging until we get an official release or not.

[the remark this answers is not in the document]
hwillson: @AnthonyMDev @twof I just talked to Ivan about this; each canary PR release has an npm tag assigned that follows the format of `canary-pr-XXXX`. This means we can use `canary-pr-3510` here instead.

Sources/AnimalKingdomAPI/Generated/Package.resolved
twof: We don't actually. It looks like it's been .gitignore'd for a while, but was never untracked

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/AllAnimalsCCNQuery.swift
```diff
@@ -0,0 +1,63 @@
...
+    public struct AllAnimal: AnimalKingdomAPI.SelectionSet {
+      public let data: DataDict
+      public init(data: DataDict) { self.data = data }
+
+      public static var __parentType: ParentType { .Interface(AnimalKingdomAPI.Animal.self) }
+      public static var selections: [Selection] { [
+        .field("height", Height?.self),
+      ] }
+
+      public var height: Height? { data["height"] }
```
twof, line 45: Here's an example of a designator on a selection set

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
```diff
@@ -0,0 +1,63 @@
+// @generated
```
twof, line 1: @AnthonyMDev I changed the name of the query, and it looks like the old query was kept around. idk if you're already aware of this bug.

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/ClassroomPetDetailsCCN.swift
```diff
@@ -0,0 +1,52 @@
...
+// This file was automatically generated and should not be edited.
+
+import ApolloAPI
+
+public struct ClassroomPetDetailsCCN: AnimalKingdomAPI.SelectionSet, Fragment {
+  public static var fragmentDefinition: StaticString { """
+    fragment ClassroomPetDetailsCCN on ClassroomPet {
+      ... on Animal {
+        height {
+          inches!
```
twof, line 11: Example of a designator in an inline fragment within a standard fragment.

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/ClassroomPetsCCNQuery.swift
```diff
@@ -0,0 +1,70 @@
...
+
+import ApolloAPI
+
+public class ClassroomPetsCCNQuery: GraphQLQuery {
+  public let operationName: String = "ClassroomPetsCCN"
+  public let document: DocumentType = .notPersisted(
+    definition: .init(
+      """
+      query ClassroomPetsCCN {
+        classroomPets[!]? {
```
twof, line 12: List syntax example that uses both an element designator and a designator on the field as a whole.

Sources/AnimalKingdomAPI/graphql/AnimalSchema.graphqls
```diff
@@ -1,6 +1,6 @@
 type Query {
   allAnimals: [Animal!]!
-  classroomPets: [ClassroomPet!]!
+  classroomPets: [ClassroomPet]
```
twof, line 3: I also removed the Non-Nulls from classroomPets to give me something modify in tests. Hope that's okay.

Sources/AnimalKingdomAPICCN/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
twof: I think I've gone and covered each of these, but let me know if you want more. 

> CCN can only be done on fields, right? You can't designate named or inline fragments with CCN?

Fields within fragments can have designators, but fragments themselves cannot, no.

Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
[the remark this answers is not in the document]
twof: I can push breaking changes to the branch at any time. Any idea if that will cause problems if you depend on a release based on a branch rather than a specific commit? I'm not the most familiar with NPM.

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
calvincestari: Do you mean a file with the old query name was still present? If so then it's not a 'bug' as we haven't coded for it to clean up, not sure how we'd do that yet.

twof:
@AnthonyMDev Let me know when you're ready to take another pass and I'll rebase.

Sources/AnimalKingdomAPI/GeneratedCCN/Operations/AllAnimalsQueryCCNQuery.swift
AnthonyMDev: Yeah, you need to handle that manually currently. We may do something about it in the future, but it's not a priority at the moment.

AnthonyMDev:
@twof Now that I've got `include/skip` implemented I have time to look at this again! Let's rebase and get it back on track! Thanks for the patience.

twof:
@AnthonyMDev Alrighty I've removed a bunch of duplicate files and rebased, so the diff is quite a bit less unwieldy. Have at it :)

Sources/ApolloCodegenLib/Frontend/JavaScript/package.json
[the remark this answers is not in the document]
AnthonyMDev: I imagine it would. I'm not super familiar with NPM either. Can we create a stable branch for CCN and point to that instead?

Then you can continue to push breaking changes to your branch for development, but only merge that dev branch into the stable CCN branch when you are mostly confident it won't break things.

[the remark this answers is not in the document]
twof: Yeah I can give that a shot.

SwiftScripts/Sources/Codegen/Target.swift
```diff
@@ -6,6 +6,7 @@ enum Target {
   case gitHub
   case upload
   case animalKingdom
+  case ccnAnimalKingdom
```
AnthonyMDev, line 9: This file and `Module.swift` were moved in my latest PR. But I don't think we need a separate `ccnAnimalKingdom` target. I don't really see the value in that. If CCN works and is supported and everything is going into the same generated module, then lets just do it all the time.

In that case, all you need is the `target.ccnEnabled` and we are good to go!

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -142,6 +142,75 @@ class ApolloCodegenTests: XCTestCase {
...
+
+    let config = ReferenceWrapped(value: ApolloCodegenConfiguration.mock(input: .init(
+      schemaPath: schemaPath,
+      searchPaths: [directoryURL.appendingPathComponent("*.graphql").path]
+    )))
+
+    // then
+    expect(try ApolloCodegen.compileGraphQLResult(config).operations).to(throwError { error in
+      guard let error = error as? GraphQLError else {
+        fail("Expected .graphQLSourceValidationFailure, got .\(error)")
```
AnthonyMDev, line 207: What about this causes the validation error?

AnthonyMDev (changes requested):
Sorry to make you rebase AGAIN! I can handle the rebase if you'd like, but since it will be a force push, I don't wanna mess up your local history.

I want to release another alpha version by end of day tomorrow (Friday Apr. 15th). I'd like to get this merged in before then so we can publish it in the next Alpha.

The only other things let to resolve is pointing to a more stable branch of `graphql-js`. Let me know what you think about that.

AnthonyMDev:
This is looking SO SO Awesome! I'm super excited to get this published in the next Alpha!! Thanks for all your work on this @twof

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
twof: In this test we're attempting to compile a document that contains CCN operators without CCN enabled. I've added some more info to the failure message to make that more obvious.

AnthonyMDev: Ahhhh thank you!

SwiftScripts/Sources/Codegen/Target.swift
twof: I've deleted `ccnAnimalKingdom` and made CCN enabled all the time for the `animalKingdom` target, but left it disabled for the other targets. I think that's what you wanted, but let me know if I should do something else!

twof:
@AnthonyMDev Here's the stable branch: https://github.com/graphql/graphql-js/pull/3535
It looks like the bot that does automated canary releases [may be busted](https://github.com/graphql/graphql-js/pull/3535#issuecomment-1099702198). I've tagged some maintainers and hopefully they're able to fix it.

AnthonyMDev:
Just realizing this isn't a serious problem, because we don't actually update the npm packages very often anyways. So as long as it's working now, we just won't run `npm update` while it's not working.

I'm going to merge this in and push an alpha release this afternoon that includes CCN!

twof:
Awesome! Let me know if y'all run into any trouble surrounding CCN. Thanks for all your help!

PR #2184, opened 2022-03-03
Enable lazy access to the request body creation for leverage in custo…

rickfast:
Enable lazy access to the request body creation for leverage in custom built interceptors, since `JSONRequest.toURLRequest()` encapsulates the creation. This enables the `GraphQLMap` to be accessed without re-creating the body


AnthonyMDev:
Thank you for this @rickfast!

kylebrowning:
This PR seems to have broken auto persisted retries, which work in 0.51.0 but do not in 0.51.1

Issue #2034, opened 2021-11-17
Dependency automation for supported module types


calvincestari:
This task is one of the final steps in code generation; to take the generated schema files and package them together in the required [dependency manager format](https://github.com/apollographql/apollo-ios/blob/release/1.0-alpha-incubating/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift#L100).

- [x] Swift Package Manager
- [x] None (previously "Manually Linked")
- [x] Other (previously "Carthage" and "CocoaPods")
- [ ] Swift Package Manager (snippet - TBD)

calvincestari:
Sequencing could be an issue re. having this package built before any other package manager build begins.

calvincestari:
I'm marking this issue as "Todo" again as SPM support is all we're going to do for the first alpha release. Supporting the other dependency managers will come afterwards.

calvincestari:
A bit more detail on how we're changing support for dependency managers and the generated schema module.

Initially we decided to support named implementations for Swift Package Manager, CocoaPods and Carthage. SPM is easy enough since the Package.swift file is fairly concise and we can generate all required metadata with what we know about the schema module.

CocoaPods makes this more complicated with the [podspec format](https://guides.cocoapods.org/syntax/podspec.html#group_root_specification) which would require us to gather a lot more detail from the user for the metadata. Carthage takes this a step further and [requires an Xcode project file](https://github.com/Carthage/Carthage#differences-between-carthage-and-cocoapods). 😢 

Instead of growing the already large codegen configuration struct we've opted to support 'other' dependency managers by requiring the user to create the infrastructure needed, such as the podspec for CocoaPods or the Xcode project for Carthage. Combine that with the generated schema module files output to a location of the user's choosing and we can support a much wider range of dependency managers than we can write specific implementations for.

calvincestari:
I'm closing this issue as done and tracking the last item (Swift Package Manager - snippet) in https://github.com/apollographql/apollo-ios/issues/2236, if we want to slot it into any of the 1.0 milestones.

PR #2185, opened 2022-03-04
Dependency manager support - CocoaPods

calvincestari:
Relates to #2034 

This adds support for the schema module, and operation files if co-located, to be exported as a CocoaPods Pod.

_Right now the PR is a draft until I finalize how we let users specify details for the required podspec fields that we can't autogenerate._

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/CocoaPodsModuleTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/CocoaPodsModuleTemplate.swift
@@ -1,0 +1,24 @@
+import Foundation
+
+struct CocoaPodsModuleTemplate {
+  let name: String
+  let version: String
+  let license: String
+  let homepage: URL
+  let source: URL
+
+  func render() -> String {
+    TemplateString("""
+    Pod::Spec.new do |spec|
+      spec.name = '\(name)'
+      spec.version = '\(version)'
+      spec.authors = 'Apollo Codegen'
+      spec.license = '\(license)'
+      spec.homepage = '\(homepage.absoluteString)'
+      spec.source = { :git => '\(source.absoluteString)', :tag => '\(version)' }
+      spec.summary = 'Automatically generated API code that helps you execute all forms of GraphQL operations, as well as parse and cache operation responses.'
+      spec.source_files = './**/*.swift'
+    end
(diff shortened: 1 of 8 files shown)


Sources/ApolloCodegenLib/Templates/CocoaPodsModuleTemplate.swift
```diff
@@ -0,0 +1,24 @@
...
+
+struct CocoaPodsModuleTemplate {
+  let name: String
+  let version: String
+  let license: String
+  let homepage: URL
+  let source: URL
+
+  func render() -> String {
+    TemplateString("""
```
calvincestari, line 11: These are the [minimum required fields for a podspec](https://guides.cocoapods.org/syntax/podspec.html#group_root_specification), some of them we can autogenerate the content but most are really quite specific to the developer/organization/project and not something we can glean from the codegen configuration.

Right now these are stored as associated values with [the `.cocoapods` enum case of ModuleType](https://github.com/apollographql/apollo-ios/pull/2185/files#diff-c75e72335519c23675c4ac6e74cd5913ed6265a4450bdc32915044f6b65680c0R162). That means a [larger configuration](https://github.com/apollographql/apollo-ios/pull/2185/files#diff-b3d18e19585b1dcca04806cc30099bebe006b1f067427fb9eebda32ed083af29R6) but does provide a lot of flexibility to the user. The podspec specification seems to be quite stable so I'm confident this isn't something we'll have to change often, if at all.

An alternative is to simply have a single freeform `String` that the user must specify that is the 'guts' of the podspec but that opens a whole bunch of potential error scenarios where users have to consult the podspec documentation and then format it correctly - it's just not a good idea.

@AnthonyMDev would love your thoughts on this and whether you have alternative suggestions.

calvincestari: This also isn't complete because the `source_files` field has a bunch of variability to it which will need more associated values.

calvincestari:
Closing this PR as we've settled on a different set of options.

calvincestari:
Some more detail on why this was closed - https://github.com/apollographql/apollo-ios/issues/2034#issuecomment-1064527884.

PR #2194, opened 2022-03-09
Refactor `SchemaTypesFileOutput.ModuleType`

calvincestari:
PR 1 of 3 for #2034 

* Refactors `ModuleType` to be more general in the third-party dependency managers that can be supported.
* Refactored `ApolloCodegenConfigurationTests` with more reusable code, dropped a few tests that are no longer relevant with the module helpers being removed.

PR #2195, opened 2022-03-10
Finish support for `.other` module type

calvincestari:
PR 2 of 3 for #2034 

This wraps up support for dependency managers that are _not_ Swift Package Manager. The behaviour was already correct, the tests just needed a bit of refactoring and wider test coverage.

* Removed the import statement tests from both fragment and operation template tests and instead created ImportStatementTemplateTests that bring wider test coverage.
* Fixed a bug in the line-by-line comparison function that would invalidly pass tests when the expected value had more lines than the actual.


AnthonyMDev (approved):

calvincestari:
@AnthonyMDev https://github.com/apollographql/apollo-ios/pull/2195/commits/d0a677f651b71d540055a7ae7903b18b797770af relaxes the error condition from https://github.com/apollographql/apollo-ios/pull/2194#discussion_r824186724.

PR #2197, opened 2022-03-14
Updates for new docs infra

trevorblades:


calvincestari:
@trevorblades how does this work with multiple versions? The `release/1.0` branch will still have the old-style documentation configuration but when we merge `main` into that it'll overwrite what we've got there.

PR #2200, opened 2022-03-15
Set up docs preview and publish workflows

trevorblades:
This branch adds actions to deploy docs changes to prod and preview changes in PRs.


.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,16 @@
+name: Preview docs on Netlify
+
+on:
+  pull_request:
+    branches:
```
calvincestari, line 5: 👍🏻

PR #2205, opened 2022-03-18
Add codegen option for excludes

bannzai:
## Abstract
Add property of`excludes` option to `ApolloCodegenOptions`. 
ref: https://github.com/apollographql/apollo-tooling#apollo-clientcodegen-output

I need this option.

## Question
How to update [ApolloCodegenOptions.md](https://github.com/apollographql/apollo-ios/blob/main/docs/source/api/ApolloCodegenLib/structs/ApolloCodegenOptions.md)? Manually?

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenOptions.swift
@@ -59,6 +59,7 @@ public struct ApolloCodegenOptions {
   
   let codegenEngine: CodeGenerationEngine
   let includes: String
+  let excludes: String?
   let mergeInFieldsFromFragmentSpreads: Bool
   let namespace: String?
   let modifier: AccessModifier
@@ -77,6 +78,7 @@ public struct ApolloCodegenOptions {
   /// - Parameters:
   ///  - codegenEngine: The code generation engine to use. Defaults to `CodeGenerationEngine.default`
   ///  - includes: Glob of files to search for GraphQL operations. This should be used to find queries *and* any client schema extensions. Defaults to `./**/*.graphql`, which will search for `.graphql` files throughout all subfolders of the folder where the script is run.
+  ///  - excludes: Glob of files to exclude for GraphQL operations. Caveat: this doesn't currently work in watch mode
   ///  - mergeInFieldsFromFragmentSpreads: Set true to merge fragment fields onto its enclosing type. Defaults to true.
   ///  - modifier: [EXPERIMENTAL SWIFT CODEGEN ONLY] - The access modifier to use on everything created by this tool. Defaults to `.public`.
   ///  - namespace: [optional] The namespace to emit generated code into. Defaults to nil.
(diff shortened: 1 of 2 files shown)


calvincestari:
Hi @bannzai - thanks for PR.

> I need this option.

@bannzai could you describe your use case a bit more please.

AnthonyMDev:
> How to update [ApolloCodegenOptions.md](https://github.com/apollographql/apollo-ios/blob/main/docs/source/api/ApolloCodegenLib/structs/ApolloCodegenOptions.md)? Manually?

You just need to run the DocumentationGenerator in the `SwiftScripts` project. We can do that when we make an actual release including this. 

But definitely a good addition to add this. @bannzai have you tested that this works with your code?

bannzai:
> @bannzai could you describe your use case a bit more please.

I added more details :)

bannzai:
> But definitely a good addition to add this. @bannzai have you tested that this works with your code?

Yes, I already test and it's looks good to me 👍

bannzai:
> You just need to run the DocumentationGenerator in the SwiftScripts project. We can do that when we make an actual release including this.

Thank you for your response . I will try it!

calvincestari:
It's probably better not to include the documentation change in this PR since doc updates are published immediately and the codegen parameter change will only be 'published' in the next release. That's why we usually update docs at release.

calvincestari:
> My approach is to make a difference in naming the schema file. Like Shopify Storefront API schema file name to *.storefront.graphql, our server API schema file name to *.graphql.
This approach and executing codegen two times with correct codegen option can be enable to separate generation code for each schema correspond to endpoint.
> 
> But, our server's API case has problem. Because specifying*.grpahql include *.storefront.graphql. So, I got error when codegen.
I need exclude option.

Thanks for the detail. I'm not against exposing `exclude` since the underlying tool supports it but you could achieve what you want already by using subdirectories for Shopify vs. Server files instead of one common directory. Then the `include` parameter would change between codegen executions.

`Shopify/*.graphql`
`Server/*.graphql`

bannzai:
> It's probably better not to include the documentation change in this PR since doc updates are published immediately and the codegen parameter change will only be 'published' in the next release. That's why we usually update docs at release.

I see. OK, I will not execute `DocumentationGenerator`

bannzai:
> but you could achieve what you want already by using subdirectories for Shopify vs. Server files instead of one common directory. Then the include parameter would change between codegen executions.

Thank you for the useful advice :+1: But in my case is not resolved...
In my project architecture, the rule is to put the .graphql file and the .swift file that uses it in the same directory.
It becomes clearly which View is using which Query, Mutation, Fragment ...etc. I'm sure the project structure will continue to change in the future, so I hope that include this PR to apollo-ios!

bannzai:
> You might be pleased to know then that in the upcoming 1.0 release ([Alpha 2](https://github.com/apollographql/apollo-ios/releases/tag/1.0.0-alpha.2) is the current latest) the code generator will have an option to co-locate the .swift file with the .graphql operation file. So you could put them anywhere in your project. It also only supports multiple files so you won't end up with one single massive API.swift file anymore. 🎉

That's cool! I really can't wait!

> @bannzai withholding the doc updates it looks like this PR is complete and ready to merge; code change + updated test. Is there anything else you wanted to add here?

No further updates anymore. This PR is complete and ready to merge :+1: Thank you!

calvincestari (approved):

PR #2219, opened 2022-03-24
Add note to tutorial about the `graphql-ws` subscription library

rkoron007:


docs/source/tutorial/tutorial-subscriptions.md
```diff
@@ -61,6 +61,8 @@ Build your project, and the subscription will be picked up and added to your `AP
 
 ## Configure your ApolloClient to use subscriptions
 
+> This tutorial uses the `graphql-ws` protocol, implemented by the `subscriptions-transport-ws` library. **This library is no longer actively maintained.** We recommend using the [`graphql-ws`](https://www.npmjs.com/package/graphql-ws) library instead, which implements its own WebSocket subprotocol, `graphql-transport-ws`. Note that the two libraries do not use the same WebSocket subprotocol, so you need to ensure that your servers, clients, and tools use the **same library and subprotocol**. For more information and examples, see [GraphQL over WebSocket protocols](../subscriptions#graphql-over-websocket-protocols).
```
calvincestari, line 64:
```suggestion
> This tutorial uses the [`graphql-ws`](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) protocol, implemented by the [`subscriptions-transport-ws`](https://github.com/apollographql/subscriptions-transport-ws) library. **That library is no longer actively maintained.** We recommend using the [`graphql-ws`](https://www.npmjs.com/package/graphql-ws) library instead, which implements its own WebSocket subprotocol, `graphql-transport-ws`. Note that the two libraries do not use the same WebSocket subprotocol and you need to ensure that your servers, clients, and tools use the **same library and subprotocol**. For more information and examples, see [GraphQL over WebSocket protocols](../subscriptions#graphql-over-websocket-protocols).
```

calvincestari: Thanks for clarifying this in the tutorial @rkoron007. I made some minor edits to include links to the old protocol and library, plus change wording to be clear that it's the `subscriptions-transport-ws` library and not `apollo-ios` that is no longer maintained.

calvincestari (approved):

PR #2229, opened 2022-03-29
feat: Add support for `.none` module type - Part 1: File Templates

calvincestari:
_PR 1 for #2034 - This refactor grew larger than I anticipated so I'm trying to limit the changesets and keep the PRs focused; **this PR refactors the file templates** to support `.none` module type._

* Supporting the `.none` module type by wrapping the template in a namespace seemed simple enough until I got to having to split the header and import statements from the rest of the template body; that was complicated by each template handling the header/import statement additions.
* Individual templates have now been pared back to focus on their single responsibility and conform to `TemplateRenderer`.
* `TemplateRenderer` is a protocol that is responsible for combining individual templates based on the target of the template and the codegen configuration. Each individual template conforming to the protocol provides the base properties and the extension handles rendering and combining templates as needed.
* Also introduced is a reference wrapper for value types; `ReferenceWrapped`. This allows us to pass around the codegen configuration struct internally without copying it each time it's passed to a new function. 
* The individual template tests don't validate header/import statements anymore since they don't pull them in.
* There is a bit of duplication between the `TemplateRenderer` and `ImportStatementTemplate` tests for checking the import scenarios.

**Note: This PR will not pass CI checks until Part 2 is merged into it.**


Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
```diff
@@ -10,43 +10,24 @@ struct SchemaModuleFileGenerator {
     _ config: ApolloCodegenConfiguration.SchemaTypesFileOutput,
     fileManager: FileManager = FileManager.default
   ) throws {
+
```
calvincestari, line 13: Commits to this file should not have been included in this PR - ignore it. The next PR, based off this one, addresses file generators.

Sources/ApolloCodegenLib/Templates/ImportStatementTemplate.swift
```diff
@@ -1,20 +1,21 @@
+import ApolloUtils
+
+/// Provides the format to import Swift modules required by the template type.
```
calvincestari, line 3: Note that both `ImportStatementTemplate` (this) and `HeaderCommentTemplate` don't conform to `TemplateRenderer`. They're not 'traditional' templates since they're never output to a file alone, they get pulled into other templates.

I considered making them private in `TemplateRenderer.swift` to avoid confusion about why they don't conform to the protocol but that complicates testing. I'm open to suggestions on how to better structure that.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -0,0 +1,104 @@
...
+  var target: TemplateTarget { get }
+  /// The swift code format.
+  var template: TemplateString { get }
+
+  /// Renders the template converting all input values and generating a final String representation
+  /// of the template.
+  ///
+  /// - Parameter config: Shared codegen configuration.
+  /// - Returns: Swift code derived from the template format.
+  func render(forConfig config: ReferenceWrapped<ApolloCodegenConfiguration>) -> String
```
calvincestari, line 31: I don't think this should be a function in the protocol and instead just have the extension implement it. Having it as part of the protocol might lead someone to creating a new template to implement it, which would potentially cause problems for the new template.

Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/SchemaModuleFileGeneratorTests.swift
```diff
@@ -4,23 +4,23 @@ import ApolloCodegenTestSupport
 import Nimble
 
 class SchemaModuleFileGeneratorTests: XCTestCase {
+  let rootURL = URL(fileURLWithPath: CodegenTestHelper.outputFolderURL().path)
```
calvincestari, line 7: Ignore this too, shouldn't have been in this PR. Subsequent PR, based on this one, focuses on the file generators.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SwiftPackageManagerModuleTemplateTests.swift
```diff
@@ -14,20 +20,20 @@ class SwiftPackageManagerModuleTemplateTests: XCTestCase {
...
     // when
-    let actual = subject.render()
+    let actual = renderSubject()
 
     // then
     expect(actual).to(equalLineByLine(expected, ignoringExtraLines: true))
   }
 
-  func test_boilerplate__generatesRequiredImports() {
+  func test__boilerplate__generatesRequiredImports() {
```
calvincestari, line 29: These tests don't need to be validating import statement behaviour - will change.

Sources/ApolloCodegenLib/Templates/FragmentTemplate.swift
```diff
@@ -1,24 +1,25 @@
...
+/// into Swift code.
+struct FragmentTemplate: TemplateRenderer {
+  /// IR representation of source [GraphQL Fragment](https://spec.graphql.org/draft/#sec-Language.Fragments).
   let fragment: IR.NamedFragment
+  /// IR representation of source GraphQL schema.
   let schema: IR.Schema
-  let config: ApolloCodegenConfiguration.FileOutput
 
-  func render() -> String {
+  var target: TemplateTarget { .operationFile }
```
AnthonyMDev, line 11: If a fragment is being shared across operations and those operations are in separate modules, the Fragment would need to be include in the schema module. We probably need to put some more thought into this.

Sources/ApolloCodegenLib/Templates/ImportStatementTemplate.swift
AnthonyMDev: That makes sense to me!

AnthonyMDev (approved):
This is awesome! We should talk about what to do about shared Fragments, but otherwise this is good to go!

Sources/ApolloCodegenLib/Templates/ImportStatementTemplate.swift
calvincestari: Done in 77fc9028dc72a00db3dcf3a8dafa23faa1dd45da.

PR #2234, opened 2022-04-04
feat: Add support for `.none` module type - Part 2: File Generators

calvincestari:
PR 2 for https://github.com/apollographql/apollo-ios/issues/2034 - This refactor grew larger than I anticipated so I'm trying to limit the changesets and keep the PRs focused; **this PR refactors the file generators** to support .none module type. _This PR will merge into #2229, that PR will then pass CI, and they both then get merged into `release/1.0`._

* All file generators now conform to the `FileGenerator` protocol making them simpler by providing just properties and sharing the rendering, file system code in the protocol extension.
* `FileGenerator` handles resolving the path for the file, rendering the file into a data object and then writing that data to the resolved path in the file system.
* `PathResolver` moved to `FileGenerator.swift` and renamed to `FileTarget`; the logic is only used in the shared file generator code.
* The codegen configuration is now passed around as a reference wrapped struct to prevent copying it everywhere that needs configuration data. The downside is that they all get access to the full struct and not only their relevant parts of the configuration. _This struct may change further when the codegen CLI is designed/built._

**Note: This PR may fail the documentation checks of CI. That should be OK because `main` has not been merged into `release/1.0` yet which means the latest changes to the documentation build workflow are not in this branch yet.**


Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
```diff
@@ -1,19 +1,128 @@
...
+    forConfig config: ReferenceWrapped<ApolloCodegenConfiguration>,
+    fileManager: FileManager = FileManager.default
+  ) throws {
+    let directoryPath = target.resolvePath(forConfig: config)
+    let filePath = URL(fileURLWithPath: directoryPath).appendingPathComponent(fileName).path
+
+    let rendered: String = template.render(forConfig: config)
+
+    try fileManager.apollo.createFile(atPath: filePath, data: rendered.data(using: .utf8))
+  }
```
AnthonyMDev, line 29: 😍

Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
```diff
@@ -1,19 +1,128 @@
...
+    case .enum: return "Enums"
+    case .interface: return "Interfaces"
+    case .union: return "Unions"
+    case .inputObject: return "InputObjects"
+    case .fragment, .operation: return "Operations"
+    case .schema: return ""
+    }
+  }
+
+  static func ==(lhs: FileTarget, rhs: FileTarget) -> Bool {
```
AnthonyMDev, line 56: I'm actually surprised if you need this. If `FragmentDefinition` and `OperationDefinition` are `Equatable` this should be synthesized for you I thought. Could be wrong though!

AnthonyMDev (approved):
This looks AMAZING!

Not important now, but wondering if `ReferenceWrapped` would be better as a property wrapper. Not sure if that is the right way to go, but now that we can use property wrappers in function arguments, could be cool.
https://www.swiftbysundell.com/tips/attaching-property-wrappers-to-function-arguments/

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
```diff
@@ -7,27 +8,28 @@ struct SchemaModuleFileGenerator {
   ///   - config: A configuration object specifying output behavior.
   ///   - fileManager: `FileManager` object used to create the file. Defaults to `FileManager.default`.
   static func generate(
```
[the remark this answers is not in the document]
calvincestari: Thanks for the feedback, I'll leave it as-is but great suggestion on the two protocols.

Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift
calvincestari: Correct - removed.

PR #2237, opened 2022-04-05
Refactor Execution Models to accept Compound conditional selections

AnthonyMDev:


Sources/ApolloAPI/Selection.swift
```diff
@@ -53,18 +54,83 @@ public enum Selection {
...
     }
+
+    @inlinable
+    public init(stringLiteral value: StringLiteralType) {
+      self.variableName = value
+      self.inverted = false
+    }
+
+    @inlinable
+    public static prefix func !(value: Condition) -> Condition {
```
calvincestari, line 108: 👍🏻

Tests/ApolloTests/Cache/SQLite/CachePersistenceTests.swift
```diff
@@ -79,67 +79,68 @@ class CachePersistenceTests: XCTestCase {
...
-            }
-          }
-        }
-      }
-
-      self.waitForExpectations(timeout: 2, handler: nil)
-    }
-  }
+  
+  #warning("TODO")
```
calvincestari, line 83: Was this meant to make it into the PR and fixed in subsequent PRs? If so we'll need to ensure it's removed before the next alpha release.

Sources/Apollo/SelectionCondition+Evaluation.swift
```diff
@@ -0,0 +1,39 @@
+#if !COCOAPODS
+import ApolloAPI
+#endif
+import Foundation
+
+// MARK: Conditions - Or Group
+extension Selection.Conditions {
+  func evaluate(with variables: GraphQLOperation.Variables?) throws -> Bool {
+    for andGroup in value {
```
calvincestari, line 9: `andGroup` is the correct name for the 'inner' arrays but this function is difficult to grok as it reads now - _we're evaluating the outer 'or' array with a quick exit should any of the inner arrays evaluates to true._ I wonder if there is a better way to express that?

Tests/ApolloTests/GraphQLExecutor_SelectionSetMapper_FromResponse_Tests.swift
```diff
@@ -1414,4 +1466,53 @@ class GraphQLExecutor_SelectionSetMapper_FromResponse_Tests: XCTestCase {
...
+
+    // then
+    expect(data.name).to(beNil())
+  }
+
+  func test__booleanCondition_bothSkipAndInclude_mergedAsComplexLogicalCondition_correctlyEvaluatesConditionalSelections() throws {
+    // given
+    class GivenSelectionSet: MockSelectionSet {
+      override class var selections: [Selection] {[
+        .include(if: ("a" && !"b" && "c") || "d" || !"e", .field("name", String.self))
```
calvincestari, line 1491: The parenthesis confused me here but that's not how this condition is being evaluated. Did you intend to have grouping of the conditions in the parenthesis; variables below would have to be grouped as arrays for it to function like that?

Depending on your intent for this test I think we should either remove the parenthesis here or create another test that evaluates both && and || groups.

calvincestari (commented):
Looks good! Had a question on the last test that could lead to another needed test.

Tests/ApolloTests/Cache/SQLite/CachePersistenceTests.swift
AnthonyMDev: It's the only test that isn't BUILDING in Apollo tests right now, because it references a query generated from StarWarsAPI. Don't think it's crucial for the next alpha, but before Beta for sure! 

It's not commented out because it's failing and there is broken behavior, it just can't compile without the StarWarsAPI, which will be generated once we finish include/skip and generate the `__typename`s.

Issue #2238, opened 2022-04-06
Support custom scalars


calvincestari:
* Namespacing custom scalars in selection set template
* Pulling out list of all custom scalars into IR (ReferencedTypes, add new customScalars
* Iterate through and create file for each scalar
* Do not overwrite scalar type files, allowing for customizability

calvincestari:
This is done! 🎉

PR #2240, opened 2022-04-08
fix: collect all referenced interfaces in js-frontend

calvincestari:
1 of 2 for #2238

tl;dr - not all referenced interfaces were being collected into the compiled result from the JS frontend.

This PR:
* loops through all interfaces referenced by an object and adds them to the compiled result
* added a quick-exit check if the type to be referenced is already in the set


AnthonyMDev (approved):
Perfect! <3

PR #2241, opened 2022-04-08
Code generation of include/skip directives

AnthonyMDev:
Also fixes some bugs in the calculation of merged fields with inclusion conditions in entity selection tree


AnthonyMDev:
@calvincestari Please review the generated code from the `AllAnimalsIncludeSkipQuery` to make sure it makes sense. I looked at it pretty thoroughly, but it's a pretty silly contrived example with lots of weird cases, and it would be valuable to get a second set of eyes on it to make sure there are no bugs that slipped through the cracks.

AnthonyMDev:
We may want to rename `TypeCase` to something different since it now can be type or inclusion conditions, but haven't decided on a new name yet.

calvincestari (approved):
Generated code looks good. Let's sync later on the `TypeCase` name and see if there is anything better we can think of.

PR #2242, opened 2022-04-08
chore: Update Xcode build tool versions on CircleCI

calvincestari:
This brings CI up to the same versions we use daily.


calvincestari:
I was motived to put this up given the recent questions/issues about `Sendable` on Xcode 13.3. I use 13.3 daily and have no issue but this should ensure that it works in environments other than my development one.

PR #2244, opened 2022-04-11
feat: Add support for custom scalars

### LINKED ISSUE #2238: Support custom scalars

calvincestari:
Closes #2238


Sources/ApolloCodegenLib/FileGenerators/EnumFileGenerator.swift
```diff
@@ -7,5 +7,5 @@ struct EnumFileGenerator: FileGenerator {
 
   var template: TemplateRenderer { EnumTemplate(graphqlEnum: graphqlEnum) }
   var target: FileTarget { .enum }
-  var fileName: String { "\(graphqlEnum.name).swift" }
+  var fileName: String { "\(graphqlEnum.name.firstUppercased).swift" }
```
calvincestari, line 10: This is a mistake I noticed, just threw the fix into this PR.

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
```diff
@@ -30,6 +30,10 @@ struct SchemaModuleFileGenerator {
       return
     }
 
-    try fileManager.apollo.createFile(atPath: filePath, data: rendered.data(using: .utf8))
+    try fileManager.apollo.createFile(
+      atPath: filePath,
+      data: rendered.data(using: .utf8),
+      overwrite: true
```
calvincestari, line 36: This file generator doesn't conform to `FileGenerator` so it has to pass in the `overwrite` value.

Sources/ApolloCodegenLib/IR/IR+Schema.swift
```diff
@@ -60,3 +65,17 @@ extension IR {
...
+    if let _ = self.specifiedByURL { return true }
+
+    return !isString && !isInt && !isBool && !isFloat && !isID
+  }
+
+  var isString: Bool { self.name == String(describing: String.self) }
+  var isInt: Bool { self.name == String(describing: Int.self) }
+  var isBool: Bool { self.name == "Boolean" }
+  var isFloat: Bool { self.name == String(describing: Float.self) }
+  var isID: Bool { self.name == "ID" }
```
calvincestari, line 80: `ID` comes through as a scalar type, had to be checking for it too.

Sources/AnimalKingdomAPI/AnimalKingdomAPI/Schema/CustomScalars/CustomDate.swift
```diff
@@ -0,0 +1,6 @@
+// @generated
+// This file was automatically generated and should not be edited.
```
AnthonyMDev, line 2: We need to give this a different header comment, but that can be done in a follow up PR.

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
AnthonyMDev: We should make overwrite default to true at the function declaration.

Sources/ApolloCodegenLib/IR/IR+Schema.swift
```diff
@@ -60,3 +65,17 @@ extension IR {
     }
   }
 }
+
+extension GraphQLScalarType {
+  var isCustom: Bool {
+    if let _ = self.specifiedByURL { return true }
```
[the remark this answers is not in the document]
AnthonyMDev: Nice!

Sources/ApolloCodegenLib/IR/IR+Schema.swift
AnthonyMDev: This seems like a lot of extra work. This could just be

```swift
switch name {
case "String", "Int", "Float", "Boolean", "ID":
  return false
default: 
  return true
}
```

Sources/AnimalKingdomAPI/AnimalKingdomAPI/Schema/CustomScalars/CustomDate.swift
calvincestari: Doh!

Sources/ApolloCodegenLib/FileGenerators/SchemaModuleFileGenerator.swift
calvincestari: I debated that and then decided that function being generic to multiple file uses should require you to specify the value. I could be convinced otherwise though, `createFile` does imply action by default. Plus overwrite is the default behaviour of `FileManager`.

calvincestari: OK..I just talked myself into it. It'll cut down on this diff size too.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -125,7 +125,7 @@ struct SelectionSetTemplate {
     """
     .field("\(field.name)"\
     \(ifLet: field.alias, {", alias: \"\($0)\""})\
-    , \(typeName(for: field)).self\
+    , \(if: field.isCustomScalar, "\(schema.name).")\(typeName(for: field)).self\
```
AnthonyMDev, line 128: This doesn't handle lists of custom scalars correctly.

AnthonyMDev:
Refactored and merged in by #2245

PR #2246, opened 2022-04-12
fix: Minor changes

calvincestari:
Some minor things that I think missed in the branch shenanigans yesterday afternoon:
- Committed `CustomDate.swift` with the updated header comment
- `createFile` extension function should overwrite by default
- Updated `AnimalKingdomAPI` dependency versions


AnthonyMDev (approved):
Thanks for fixing this stuff!

PR #2247, opened 2022-04-13
Ensure generated fragment Swift types use TitleCasing if GraphQL frag ment is lowercased

eliperkins:
This causes compile-time failure as other parts of codegen assume that types are TitleCased.

Here's an example of a compile-time failure from the GitHub codebase:

| Fragment Defintion | Fragment Usage |
| --- | --- |
| Screen Shot 2022-04-13 at 6 54 31 PM | Screen Shot 2022-04-13 at 6 55 23 PM |

cc @calvincestari since we were DM'ing about this!

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/FragmentTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/FragmentTemplateTests.swift
@@ -88,6 +88,42 @@ class FragmentTemplateTests: XCTestCase {
     expect(String(actual.reversed())).to(equalLineByLine("}", ignoringExtraLines: true))
   }
 
+  func test__render__givenLowercaseFragment_generatesTitleCaseTypeName() throws {
+    // given
+    document = """
+    fragment testFragment on Query {
+      allAnimals {
+        species
+      }
+    }
+    """
+
+    let expected =
+    """
+    public struct TestFragment: TestSchema.SelectionSet, Fragment {
+      public static var fragmentDefinition: StaticString { ""\"
+        fragment testFragment on Query {
+          __typename
+          allAnimals {
+            __typename
+            species
+          }
+        }
+        ""\" }
+
+      public let data: DataDict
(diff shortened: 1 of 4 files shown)


calvincestari:
@eliperkins,. those tests are failing because we don't have the whole project passing on the `release/1.0` branch yet. It's only important that the `ApolloCodegenLib` tests pass right now; which we have a hack for in that if you create a PR on a `1.0/` prefixed branch then CI will only run the codegen tests.

tl;dr - don't worry about the tests for now, I'll make sure the relevant ones pass. 👍🏻

Tests/ApolloCodegenTests/CodeGeneration/Templates/FragmentTemplateTests.swift
```diff
@@ -88,6 +88,42 @@ class FragmentTemplateTests: XCTestCase {
...
+      }
+    }
+    """
+
+    let expected =
+    """
+    public struct TestFragment: TestSchema.SelectionSet, Fragment {
+      public static var fragmentDefinition: StaticString { ""\"
+        fragment testFragment on Query {
+          __typename
```
calvincestari, line 106: This test probably doesn't need anything beyond the first three lines of the expected output; matching TitleCase struct name and ensuring that the definition is preserved with the correct casing. The rest isn't important to the test passing and would get swallowed by the `ignoringExtraLines` param on `equalLineByLine`.

Doesn't need to change for this test, just a handy tip for it you ever need to write tests using that matcher again.

calvincestari (approved):
@eliperkins I made the same change to the file generator, which would have suffered the same bug.

Thanks for this contribution, LGTM!

calvincestari:
All codegen related tests are passing, this PR is good to go.
> Test Suite 'ApolloCodegenTests.xctest' passed at 2022-04-13 21:50:27.051.
>    Executed 581 tests, with 0 failures (0 unexpected) in 4.946 (5.261) seconds

PR #2251, opened 2022-04-15
refactor: All Swift type names must have first letter uppercased

calvincestari:
This is a follow-on from #2247 to ensure we correctly case all generated Swift types.

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/CustomScalarFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/CustomScalarFileGenerator.swift
@@ -7,6 +7,6 @@ struct CustomScalarFileGenerator: FileGenerator {
 
   var template: TemplateRenderer { CustomScalarTemplate(graphqlScalar: graphqlScalar) }
   var target: FileTarget { .customScalar }
-  var fileName: String { "\(graphqlScalar.name.firstUppercased).swift" }
+  var fileName: String { "\(graphqlScalar.name).swift" }
   var overwrite: Bool { false }
 }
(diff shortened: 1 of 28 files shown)


Sources/ApolloCodegenLib/FileGenerators/CustomScalarFileGenerator.swift
```diff
@@ -7,6 +7,6 @@ struct CustomScalarFileGenerator: FileGenerator {
 
   var template: TemplateRenderer { CustomScalarTemplate(graphqlScalar: graphqlScalar) }
   var target: FileTarget { .customScalar }
-  var fileName: String { "\(graphqlScalar.name.firstUppercased).swift" }
+  var fileName: String { "\(graphqlScalar.name).swift" }
```
AnthonyMDev, line 10: Why are we removing `.firstUppercased` here?

AnthonyMDev (approved):
Confused about the places you _removed_ the uppercasing. But tests look good, so if it works lets do it!

Sources/ApolloCodegenLib/FileGenerators/CustomScalarFileGenerator.swift
calvincestari: Just because they were in each file generator implementation and now the protocol extension handles it. It does no harm here.

calvincestari: I thought it would be clearer and more consistent if we had it the same in each of the file generators rather than asking ourselves questions later, when we forget this change, why some do it and others don't.

AnthonyMDev: Got it. Thanks!

PR #2254, opened 2022-04-21
Fix a few sandbox links

martinbonnin:
The fullstack tutorial now requires the `/graphql` path


calvincestari:
Thanks for the this @martinbonnin! _Ignore the failing test, it's a Node issue._

Issue #2166, opened 2022-02-17
Codegen CLI


hwillson:
- Swift command line tool
  - JSON file input
  - JSON file string
  - arguments (from convenience initializers
- Parameters
  - JSON input path
    - Default is path of execution
  - High level convenience parameters
    - Too many parameters to provide for all config options. _Full config is supported via the JSON input file._
- Command to dump standard configuration to json file
  - Let user specify output path 
- Available via Homebrew and GitHub Releases
  - _See if CI can handle this workflow_
- Refactor ApolloCodegenConfiguration options into the following groups
  - Input / output / rules / experimental
    - _Need a better name than 'rules'_
  - Verify that the passing around of config is efficient in ApolloCodegen with the new groups
- Compile the Apollo codegen lib with optimization (`-O unchecked`)
- Can we use Package Manager Extensive Build Tools ([SE-0303](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md)).

calvincestari:
#2166 is the bulk of this work. Still to do is refining the build configuration for release builds and creating a Formula for getting the executable added to Homebrew.

hwillson:
Done (excluding homebrew work broken out in https://github.com/apollographql/apollo-ios/issues/2305). Thanks!

PR #2255, opened 2022-04-21
refactor: Wrap codegen output options in a struct + cleanup options

calvincestari:
PR #1 of #2166:
* Moves the properties that affect the generated code output into another struct; `OutputOptions`. There are now groups for codegen config of:
  * `input`: All input file locations
  * `output`: All output file locations and behaviour
  * `options`: All options that affect the generated code
  * `experimentalFeatures`: All codegen output related to experimental feature sets
* Removed the `customScalarFormat` property because the behaviour of custom scalars is to always be output as a `typealias`'d `String`. The user is then free to modify the generated file and those changes will not be overwritten.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -128,7 +128,50 @@ public struct ApolloCodegenConfiguration {
     case absolute(path: String)
   }
 
-  // MARK: General Types
+  public struct OutputOptions {
+    /// Any non-default rules for pluralization or singularization you wish to include.
+    public let additionalInflectionRules: [InflectionRule]
+    /// Formatting of the GraphQL query string literal that is included in each
+    /// generated operation object.
+    public let queryStringLiteralFormat: QueryStringLiteralFormat
+    /// How to handle properties using a custom scalar from the schema.
+    public let customScalarFormat: CustomScalarFormat
+    /// How deprecated enum cases from the schema should be handled.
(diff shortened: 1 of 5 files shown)


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -210,22 +251,12 @@ public struct ApolloCodegenConfiguration {
   public let input: FileInput
   /// The paths and files output by code generation.
   public let output: FileOutput
-  /// Any non-default rules for pluralization or singularization you wish to include.
-  public let additionalInflectionRules: [InflectionRule]
-  /// Formatting of the GraphQL query string literal that is included in each
-  /// generated operation object.
-  public let queryStringLiteralFormat: QueryStringLiteralFormat
-  /// How to handle properties using a custom scalar from the schema.
-  public let customScalarFormat: CustomScalarFormat
```
calvincestari, line 219: You might notice that this property is not present in `OutputOptions`. That is because the workflow for custom scalars in code generation 1.0 is different; by default custom scalars are always a `typealias` to `String` and the user is free to edit that generated file without being overwritten. So we no longer need an options for default or passthrough behaviour.

AnthonyMDev: I love this new behavior :) I hope users like it too haha.

AnthonyMDev (approved):
Looks good! Does the `SwiftScripts` project need to be updated to use this new format?

Issue #2256, opened 2022-04-23
Swift Scriptings Pauses Automatic Previews


asbhat:
## Bug report

When using [swift scripting](https://www.apollographql.com/docs/ios/swift-scripting) to generate code, after adding it to the [build phase](https://www.apollographql.com/docs/ios/swift-scripting#running-code-generation-from-your-main-project) of the target, SwiftUI previews will continuously be paused.

<img width="726" alt="auto-preview-paused" src="https://user-images.githubusercontent.com/3885798/164875305-38191bfd-e474-4998-8970-42d802f2a3ba.png">

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: `v0.51.2`
- Xcode version: `v13.3.1`
- Swift version: `v5.6`
- Package manager: SPM

## Steps to reproduce

1. Install and setup [Swift scripting](https://www.apollographql.com/docs/ios/swift-scripting)
2. Make sure `$ swift run ApolloCodegen generate` runs successfully
3. Add a run script to the build phase, as detailed [here](https://www.apollographql.com/docs/ios/swift-scripting#running-code-generation-from-your-main-project)
4. Notice SwiftUI previews will continue to pause themselves

calvincestari:
Thanks for raising the issue @asbhat, let's keep the conversation in the PR for now - #2257

PR #2257, opened 2022-04-23
fix: Swift build phase documentation (#2256)

asbhat:
Adds an environment variable check in the build phase run script so it doesn't run during SwiftUI Previews. Otherwise Xcode thinks files have been updated and the preview needs to pause.

### DIFF
--- a/docs/source/swift-scripting.md
+++ b/docs/source/swift-scripting.md
@@ -189,6 +189,8 @@ This is best achieved with a Run Script Build Phase.
     ```
     # Don't run this during index builds
     if [ $ACTION = "indexbuild" ]; then exit 0; fi
+    # Don't run during SwiftUI previews
+    if [ $ENABLE_PREVIEWS = "YES" ]; then exit 0; fi
 
     cd "${SRCROOT}"/ApolloCodegen
     xcrun -sdk macosx swift run ApolloCodegen generate


docs/source/swift-scripting.md
```diff
@@ -189,6 +189,8 @@ This is best achieved with a Run Script Build Phase.
     ```
     # Don't run this during index builds
     if [ $ACTION = "indexbuild" ]; then exit 0; fi
+    # Don't run during SwiftUI previews
+    if [ $ENABLE_PREVIEWS = "YES" ]; then exit 0; fi
```
calvincestari, line 193: I'm not convinced `ENABLE_PREVIEWS` is the correct environment variable to be watching for. What is the value when not building for preview vs. building for preview? The name of it reads as a variable that will be set simply if your build has previews enabled.

I went looking for documentation on Xcode build variables and found [this from Apple](https://developer.apple.com/documentation/xcode/build-settings-reference). Unfortunately it doesn't list the values for these variables. It does have `ENABLE_PREVIEWS` though; "If enabled, the product will be built with options appropriate for supporting previews". IMO that supports my earlier suspicion of it not being the value you need.

[This question on stackoverflow](https://stackoverflow.com/questions/58759987/how-do-you-check-if-swiftui-is-in-preview-mode/) suggests there is a different variable to watch (`XCODE_RUNNING_FOR_PREVIEWS `) which isn't documented in by Apple.

Getting this wrong could lead to a very frustrating time trying to debug why the codegen script isn't executing unless you're aware of this change. I don't feel we can merge this change in until we better understand the potential values of these variables.

calvincestari (changes requested):

docs/source/swift-scripting.md
calvincestari: For what it's worth v1.0 should not have this problem; we intend to build codegen to only output generated files when there is a change in the input schema or operations.

calvincestari: I would prefer not to merge this change in. If Apple does get clearer on their documentation or anyone can definitively provide more details about the `ENABLE_PREVIEWS` environment variable then we can revisit.

Thanks for raising the issue and being proactive with the PR - we do appreciate the contributions even if they don't end up getting merged.

calvincestari:
Closing - see https://github.com/apollographql/apollo-ios/pull/2257#discussion_r859937345.

docs/source/swift-scripting.md
charlesanim499: So my team actually has encountered this issue when we added Apollo iOS package to our app. All our previews failed to run. Mind you this is around December 2021. Fast forward to today, i just added the below script in the Apollo build phase and magically 😄 all our previews came back to life and working. I think the larger issue here is building in previews with Apollo fails when a view refrences the shared instance of the apollo client. We're happy to be using previews again, at least this solution can be adopted temporarily until a more `concrete` fix by Team Apollo or Apple is released. For now, it just works. Also, doesnt affect builds to device or builds to App store. 🤞🏼 

```
# Don't run this during index builds
if [ $ACTION = "indexbuild" ]; then exit 0; fi

# skip this build phase if in previews
if [ "${ENABLE_PREVIEWS}" = "YES" ]; then
  echo "Previews enabled, quitting to prevent 'preview paused'."
  exit 0;
fi
```

PR #2270, opened 2022-05-12
Refactor ApolloCodegenConfguration to add TestMockFileOutput

AnthonyMDev:


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -91,43 +122,46 @@ public struct ApolloCodegenConfiguration {
...
+    /// file. If `subpath` is specified a subfolder will be created relative to the `.graphql` file
+    /// and the operation object files will be generated there. If no `subpath` is defined then all
+    /// operation object files will be generated alongside the `.graphql` file.
     case relative(subpath: String?)
     /// All operation object files will be located in the specified path.
     case absolute(path: String)
   }
 
+  /// The local path structure for the generated test mock object files.
+  public enum TestMockFileOutput: Equatable {
```
AnthonyMDev, line 141: @calvincestari Let me know what you think about this approach. After a lot more consideration, this is what I felt was the simplest and cleanest way to provide all of the configuration we wanted with the least complexity added to the configuration structure.

SwiftScripts/Sources/Codegen/main.swift
```diff
@@ -8,27 +8,13 @@ import TargetConfig
...
-
   @Option(
+    wrappedValue: [],
     name: [.customLong("target"), .customShort("t")],
-    help: "The target to generate code for - Required."
+    parsing: .upToNextOption,
+    help: "The target to generate code for."
   )
-  var targetName: String
+  var targetNames: [String]
```
AnthonyMDev, line 17: This allows us to pass multiple targets to run codegen on at once. And if you pass nothing, we will generate all targets.

SwiftScripts/Sources/TargetConfig/Target.swift
```diff
@@ -3,48 +3,49 @@ import ApolloCodegenLib
 
 public enum Target: CaseIterable {
   case starWars
-  case gitHub
+//  case gitHub
```
AnthonyMDev, line 6: In order to get generation of all targets to work, I commented this one out for now, since we haven't set it up yet and it would cause the whole script to fail.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -91,43 +122,46 @@ public struct ApolloCodegenConfiguration {
...
+    ///
+    ///- Note: Generated files must be manually added to your test target. Test mocks generated
+    /// this way may also be manually embedded in a test utility module that is imported by your
+    /// test target.
+    case absolute(path: String)
+    /// Generated test mock files will be included in a target defined in the generated
+    /// `Package.swift` file that is suitable for linking the generated test mock files to your
+    /// test target using Swift Package Manager.
+    ///
+    /// The name of the test mock target can be specified with the `name` value.
```
calvincestari, line 155:
```suggestion
    /// The name of the test mock target can be specified with the `targetName` value.
```
Missed parameter name update, I think.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: It should satisfy what we're wanting to allow for. As more complicated use cases come up we can reconsider.

SwiftScripts/Sources/TargetConfig/Module.swift
```diff
@@ -6,7 +6,7 @@ public struct Module {
 
   public init?(module: String) {
     switch module.lowercased() {
-    case "none": self.moduleType = .none
+    case "none": self.moduleType = .embeddedInTarget(name: "")
```
calvincestari, line 9: Is a change away from `none` required here too?

calvincestari (approved):
👍🏻 - let's get this in so we can unblock ourselves.

SwiftScripts/Sources/TargetConfig/Module.swift
AnthonyMDev: Not at the moment. We aren't using it, so really this could just be `fatalError()`. If we ever use it, we should fix this. But this is just our custom SwiftScripts that we use for generating OUR example schemas, so it's okay that it's not comprehensive.

PR #2272, opened 2022-05-13
Add `Codable` support to `ApolloCodegenConfiguration` and `InflectionRule`

calvincestari:
This work modifies `ApolloCodegenConfiguration` and `InflectionRule` to conform to `Codable` and changed a couple of the enums to use a raw type of `String`, so they are serialized as strings instead of JSON objects themselves; see `QueryStringLiteralFormat` as an example.

The added tests are limited to only the enums that were specifically changed to be serialized differently, e.g.: `QueryStringLiteralFormat` was given a raw type of `String` so that it is serialized as a string and not an embedded JSON object.


AnthonyMDev (approved):
This rocks. Thanks!

PR #2276, opened 2022-05-19
Added exported import of GraphQLNullable to generated operations

AnthonyMDev:


Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -145,6 +145,7 @@ private struct ImportStatementTemplate {
     static func template(forConfig config: ReferenceWrapped<ApolloCodegenConfiguration>) -> TemplateString {
       """
       \(ImportStatementTemplate.template)
+      @_exported import enum ApolloAPI.GraphQLNullable
```
calvincestari, line 148: I'm skeptical this needs to be in each generated file. [This article](https://davedelong.com/blog/2018/01/19/simplifying-swift-framework-development/) makes me think we could get away with a top-level `Exports.swift` file that has this in it and it would apply module-wide.

calvincestari (commented):
I'm not against including this but very weary given the ["strongly discouraged" notice from the Swift team on inclusion of underscored attributes](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#underscored-attributes-reference). What would our plan be if this were deprecated?

calvincestari (approved):

PR #2281, opened 2022-05-24
Make SelectionSet Hashable with JSON conversion to AnyHashable

### LINKED ISSUE #1956: As an iOS developer, I would like to have an optional way of creating inputs and objects equatables for my tests.
## Feature request

Give the tool the ability to create equatable extensions of my objects and inputs, maybe even on a different folder if you don't want to ship this to the app store.

## Motivation

The swift way of doing unit testing a lot of times requires using protocols to swap implementations on runtime. There are actually a lot of different libraries that do code generation of mocks using the protocols in your projects (ex: [mockingbird](https://github.com/birdrides/mockingbird), [cuckoo](https://github.com/Brightify/Cuckoo), [swiftymocky](https://github.com/MakeAWishFoundation/SwiftyMocky)). When doing that I would like to assert my network interface is receiving a correctly built input, which currently requires me to write manual conformance to equatable on my test just to check if an input was correctly built.

## Proposed solution

Give the tool the ability to generate equatables of your inputs.

## Outstanding Questions

I would need feedback to check what the solution looks like.

AnthonyMDev:
Fixes #1956 

By making `JSONValue == AnyHashable` instead of `Any`, we can add `Hashable` and `Equatable` conformance to all `SelectionSet` objects, having them just match their `DataDict`s.

### DIFF
--- a/Sources/Apollo/GraphQLSelectionSetMapper.swift
+++ b/Sources/Apollo/GraphQLSelectionSetMapper.swift
@@ -10,7 +10,7 @@ final class GraphQLSelectionSetMapper<SelectionSet: AnySelectionSet>: GraphQLRes
          let .customScalar(decodable as JSONDecodable.Type):
       // This will convert a JSON value to the expected value type,
       // which could be a custom scalar or an enum.
-      return try decodable.init(jsonValue: scalar)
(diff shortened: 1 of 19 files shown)


calvincestari (approved):
This is going to be really helpful to users - kudos on the great work!

PR #2282, opened 2022-05-24
1.0/anyhashable inputvalue

### LINKED ISSUE #1645: Make operationName a Type property
## Feature request

We would like to be able to access the `operationName` of a GraphQL operation given just the type of the operation. We're happy to spend time creating a PR for this, but want to make sure it makes sense as a feature request. 😄 

## Motivation

- Our understanding is that `operationName` will be the same across all instances of a `GraphQLOperation`, so we think it would be a good candidate for a type property.
- We currently (indirectly) use the `operationName` to map e.g. some metadata to GraphQL operations to be sent with requests, or map mock responses to GraphQL operations. Right now we just hardcode the operation names in these places (e.g. `operationName: "MyQuery"`), but this is brittle:
  - If `MyQuery`'s operation name changes, but the type name does not change, and we neglect to update the string for these mappings, we will fail at runtime. With this change, we can reference `MyQuery.operationName` instead of hardcoding the string, so it will stay up to date.
  - If `MyQuery`'s operation _and_ type name changes, we still fail at runtime. With this change, our reference to `MyQuery.operationName` will fail the build (since `MyQuery` will no longer be a valid type), which is preferable to failing at runtime.

## Outstanding Questions

- Are there cases where `operationName` is different between instances of the same `GraphQLOperation`?

AnthonyMDev:
Fixed #1645 and #1956 

Make `InputObject` Hashable as well.
Make the metadata fields on `GraphQLOperation` static


calvincestari (approved):
LGTM!

PR #2286, opened 2022-05-26
Add local mutation operation directive

AnthonyMDev:
- Adds compilation of directives to operations in frontend
- Adds default schema extension to be added to all schemas during codegen for custom codegen directives
  - Add `@apollo_client_ios_localCacheMutation` directive to default schema extension


Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
```diff
@@ -69,16 +71,30 @@ export function mergeDocuments(documents: DocumentNode[]): DocumentNode {
...
   schema: GraphQLSchema,
   document: DocumentNode
 ): readonly GraphQLError[] {
-  return validate(schema, document, defaultValidationRules);
+  const codegenSchema = mergeSchemaExtensions(buildSchemaExtensions(), schema)
+  return validate(codegenSchema, document, defaultValidationRules);
 }
 
 export function compileDocument(
   schema: GraphQLSchema,
```
AnthonyMDev, line 95: We are now able to easily add a parameter `schemaExtensions: readonly [Source]` here in the future and then merge them into the schema before running codegen!

Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
```diff
@@ -46,8 +48,8 @@ export function loadSchemaFromIntrospectionResult(
 }
 
 export function loadSchemaFromSDL(source: Source): GraphQLSchema {
-  const document = parse(source);
-
+  let document = parse(source);
```
calvincestari, line 51: I think this is a remnant from the earlier change before we discovered `extendSchema`. It doesn't look like `document` is modified anywhere so it can remain `const` defined?

Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
```diff
@@ -69,16 +71,31 @@ export function mergeDocuments(documents: DocumentNode[]): DocumentNode {
...
+  const codegenSchema = mergeSchemaExtensions(buildSchemaExtensions(), schema)
+  return validate(codegenSchema, document, defaultValidationRules);
 }
 
 export function compileDocument(
   schema: GraphQLSchema,
   document: DocumentNode
 ): CompilationResult {
-  return compileToIR(schema, document);
+  console.log(schema.getDirectives());
```
calvincestari, line 98: Was this `log` command meant to stick around?

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -413,9 +413,9 @@ export function compileToIR(
     directiveNode: DirectiveNode,
     directiveDef: GraphQLDirective
   ): ir.InclusionCondition | undefined {
-    if (directiveDef == GraphQLIncludeDirective || directiveDef == GraphQLSkipDirective) {      
+    if (directiveDef.name == "include" || directiveDef.name == "skip") {
```
calvincestari, line 416: Is there a benefit to this change?

Sources/ApolloCodegenLib/Frontend/JavaScript/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.base",
   "compilerOptions": {
-    "types": [],
+    "types": ["node"],
```
calvincestari, line 4: If I recall correctly this is for certain Node types to be recognized by the compiler correct?

calvincestari (approved):
This looks good, great implementation for the user not to have to manually extend their schema for our apollo-ios directive!

Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
AnthonyMDev: Good catch

Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
AnthonyMDev: Nope, thought I caught that. Thanks

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
AnthonyMDev: It's to get around a bug in graphql-js. `extendSchema` copies the directives, so it's no longer got the same identity as the original `GraphQLIncludeDirective`.

Sources/ApolloCodegenLib/Frontend/JavaScript/tsconfig.json
AnthonyMDev: Yeah, I needed it in order for `console.log` to work. But that stuff shouldn't get compiled into the rolled up stuff, so we should be good there.

PR #2291, opened 2022-06-01
Fixes nested fragment type case merging for selection set generation

### LINKED ISSUE #2220: Unable to generate code with non-nullable value
Hi! When trying out the new codegen on our project I get a crash when trying to generate code for an `INPUT_OBJECT`, the object looks like this:
```
{
  "kind": "INPUT_OBJECT",
  "name": "MyInput",
  "description": null,
  "fields": null,
  "inputFields": [
    {
      "name": "thisIsTheId",
      "description": null,
      "type": {
        "kind": "NON_NULL",
        "name": null,
        "ofType": {
          "kind": "SCALAR",
          "name": "UUID",
          "ofType": null
        }
      },
      "defaultValue": null
    }
  ],
  "interfaces": null,
  "enumValues": null,
  "possibleTypes": null
}
``` 

What happens is that I get a crash on [`JavaScriptBridge.swift:200`](https://github.com/apollographql/apollo-ios/blob/9729f7431c6003b2b61e694e1bf94b8038069cb6/Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift#L200), saying that `jsValue` is `undefined`. Tracing the crash back a bit it seems to be because we're trying to resolve the default value (although using a field called [`astNode`](https://github.com/apollographql/apollo-ios/blob/9729f7431c6003b2b61e694e1bf94b8038069cb6/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift#L78)), but this is undefined.
Going even further up this seems to be called because [`isSwiftOptional`](https://github.com/apollographql/apollo-ios/blob/9729f7431c6003b2b61e694e1bf94b8038069cb6/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLInputField%2BRendered.swift#L8) returns true for this `INPUT_OBJECT`.

With my limited understanding it seems like maybe `isSwiftOptional` is the culprit here, and maybe the conditional has gotten mixed up.

## Versions

- `apollo-ios` SDK version: 1.0-alpha.2
- Xcode version: XCode 13.3
- Swift version: 5.6
- Package manager: SPM

## Steps to reproduce

Run new codegen on a schema defining optional values, but I assume this just happens in some cases.

AnthonyMDev:
Fixes #2220


Sources/AnimalKingdomAPI/graphql/WarmBloodedDetails.graphql
```diff
@@ -1,7 +1,4 @@
 fragment WarmBloodedDetails on WarmBlooded {
   bodyTemperature
-  height {
-    meters # TODO: Use HeightInMeters fragment?
-    yards
-  }
+  ...HeightInMeters
```
AnthonyMDev, line 3: All of the changes to the generated code are because I did this. It's just a sanity check that merging works for fragments referenced from inside other fragments.

Sources/ApolloCodegenLib/IR/IR+EntitySelectionTree.swift
```diff
@@ -443,7 +443,29 @@ extension IR.EntitySelectionTree.EnclosingEntityNode {
     withNodeRootType nodeRootType: GraphQLCompositeType,
     using entityStorage: IR.RootFieldEntityStorage
   ) {
-    switch otherTree.rootNode.child {
+    let rootNode = otherTree.rootNode
```
AnthonyMDev, line 446: This fixes a bug where, when merging the entity selection tree from a fragment, we forgot to merge type conditions that were at the root of the fragment.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -396,43 +396,72 @@ fileprivate extension IR.MergedSelections.MergedSource {
   }
 
   private func generatedSelectionSetNameForMergedEntity(in fragment: IR.NamedFragment) -> String {
-    let selectionSetName = ApolloCodegenLib.generatedSelectionSetName(
-      from: typeInfo.scopePath.head.next!,
-      withFieldPath: Array(typeInfo.entity.fieldPath.toArray().dropFirst())
-    )
+    var selectionSetNameComponents: [String] = [fragment.definition.name]
```
AnthonyMDev, line 399: There is some clean up here as well, but the primary thing happening here is that we now look for type conditions at the root of the fragment and add those to the generated name of the selection set to point to as well. This is what fixes #2220

calvincestari (approved):

PR #2296, opened 2022-06-06
Fix compile errors in test mock initializers

AnthonyMDev:
Thanks for catching this @calvincestari


calvincestari (approved):
Looks good, glad we caught these. Definitely need to put a CI in check to keep us honest in the very near future.

PR #2298, opened 2022-06-07
fix: Union test mock generator

calvincestari:
This PR:
1. Adds a target dependency to the `ApolloTestSupport` package.
2. Adds the generation of test mock extensions for `Union` types that conform to `MockFieldValue`.

Alternatives solutions for `Union` conforming to `MockFieldValue`
- Change the `Union` protocol declaration to add conformance to `MockFieldValue`; the extension then to implement the typealias can be in that file or elsewhere. The problem with this approach is that `MockFieldValue` references `AnyMock` and attempting to `import ApolloTestSupport` creates a circular dependency. Bringing the `AnyMock` declaration into `ApolloAPI` resolves build errors in `Union.swift` but makes `Union` unusable in `ParentType` because of the associated type requirement - `Protocol 'Union' can only be used as a generic constraint because it has Self or associated type requirements`.
-


calvincestari:
@AnthonyMDev let's discuss this again in the morning but I think generating the union test mock extensions may be the best solution.

AnthonyMDev:
Looks like you've got a test that needs updating, but otherwise I'm good to ship this!

PR #2300, opened 2022-06-07
SYM-41: Expose cache.

jasperblues:


jasperblues:
Sorry for the PR, I meant to raise it on our fork, not the master repo. 

However, for note: For us it is useful to have the store change events publicly exposed. Perhaps others will find this useful.

AnthonyMDev:
Thanks again! Will definitely consider this for the re-write of this part of the library. I agree that it's not ideal having these strongly referenced here!

Issue #2101, opened 2022-01-19
Implement the output of `ApolloCodegenConfiguration`


calvincestari:
The work here is to ensure that all configuration properties of `ApolloCodegenConfiguration` affect the output in the desired way.

- [x] `queryStringLiteralFormat`
- [x] `deprecatedEnumCases` (done in #2303)
- [x] `schemaDocumentation`
- [x] `apqs`
  - [x]  `FileOutput.operationIdentifiersPath`
- [x] `additionalInflectionRules`
- [x] `warningsOnDeprecatedUsage`

_This was prompted by https://github.com/apollographql/apollo-ios/pull/2098#issuecomment-1016073269._

AnthonyMDev:
- `queryStringLiteralFormat`
  - `singleLine`:
```swift
public static let document: DocumentType = .notPersisted(
    definition: .init(
      "query AllAnimalsQuery { allAnimals {__typename height { __typename feet inches } ...HeightInMeters ...WarmBloodedDetails species skinCovering ...",
      fragments: [HeightInMeters.self, WarmBloodedDetails.self, PetDetails.self]
    ))
```

AnthonyMDev:
- `apq`
  - `persistedOperationsOnly`
```swift
public static let document: DocumentType = .persistedOperationsOnly(
    operationIdentifier: "a121309ofsah80j")))
```

AnthonyMDev:
- `schemaDocumentation`
  - SelectionSet Headers
```swift
/// AllAnimal.Predator
///
/// Parent Type: `Animal`
  public struct Predator: AnimalKingdomAPI.SelectionSet {
  public let data: DataDict
  public init(data: DataDict) { self.data = data }

  public static var __parentType: ParentType { .Interface(AnimalKingdomAPI.Animal.self) }
```

calvincestari:
`warningsOnDeprecatedUsage` - from https://github.com/apollographql/apollo-ios/pull/2303#discussion_r893846978

> For deprecated cases, we probably should have the deprecated annotation to provide a deprecation warning as well. https://stackoverflow.com/questions/71354470/deprecate-enum-value-in-swift
> 
> That warning should also apply to fields marked as deprecated in the schema.
> 
[...]

PR #2303, opened 2022-06-08
feat: Evaluate `deprecatedEnumCases` configuration option in generated code

calvincestari:
This PR
* Implements the logic for correct evaluation and output of the `ApolloCodegenConfiguration.OutputOptions.deprecatedEnumCases` codegen configuration option.
* Includes a JS frontend test to ensure that `deprecationReason` is correctly parsed and included with the compilation results.
* Adds the JS frontend test suite to the 'Swift Code Generation' CI test step

PR 1 of #2101


Sources/ApolloCodegenLib/Frontend/JavaScript/src/__tests__/deprecatedEnumCasesCompilationTests.ts
```diff
@@ -0,0 +1,64 @@
+import {
```
AnthonyMDev, line 1: Awesome job on this! We don't have many tests for the JS, nice to be adding more.

Worth calling out: I don't think we should spend too much time on writing tests here b/c ideally we move over to use `apollo-rs` at some point in the future! Easy additive tests like this are great, but I wouldn't take the time to go back and write comprehensive tests for the whole thing until we move over to `apollo-rs` (if that ever happens).

Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
```diff
@@ -1,22 +1,35 @@
...
     )
   }
+
+  private func evaluateDeprecation(
+    graphqlEnumValue: GraphQLEnumValue,
+    config: ReferenceWrapped<ApolloCodegenConfiguration>
+  ) -> String? {
+    switch (config.options.deprecatedEnumCases, graphqlEnumValue.deprecationReason) {
+    case (.exclude, .some): return nil
+    default: return "case \(graphqlEnumValue.name)"
```
AnthonyMDev, line 32: This made me think of something.

For deprecated cases, we probably should have the deprecated annotation to provide a deprecation warning as well. https://stackoverflow.com/questions/71354470/deprecate-enum-value-in-swift

That warning should also apply to fields marked as deprecated in the schema.

Though some people may not want those warning clogging up their project, even when things are deprecated... So  we should add another configuration option for this. Maybe add `public let warningsOnDeprecatedUsage: Composition`. This would add the `@available(*, deprecated, message: "Deprecated: \(deprecationReason)") to enum cases and deprecated fields if set to `.include`.

Can we add this to the list of config options you're working on?

Tests/ApolloCodegenInternalTestHelpers/MockGraphQLType.swift
```diff
@@ -98,18 +99,29 @@ public extension GraphQLEnumType {
   class func mock(
     name: String,
     values: [String] = []
+  ) -> Self {
```
AnthonyMDev, line 102: I like the addition of this. Nice

Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
```diff
@@ -14,9 +15,23 @@ class EnumTemplateTests: XCTestCase {
...
+    values: [(String, String?)] = [("ONE", nil), ("TWO", nil)],
+    config: ApolloCodegenConfiguration = ApolloCodegenConfiguration.mock()
+  ) {
     subject = EnumTemplate(
-      graphqlEnum: GraphQLEnumType.mock(name: name, values: values)
+      graphqlEnum: GraphQLEnumType.mock(
+        name: name,
+        values: values.map({ GraphQLEnumValue.mock(name: $0.0, deprecationReason: $0.1) })
+      ),
+      config: ReferenceWrapped(value: config)
```
AnthonyMDev, line 34: We should really fix `ReferenceWrapped` to not have the `value` name on the initializer. I'm going to do that at some point.

AnthonyMDev (approved):
Perfect! No changes necessary for this. Comments are all about future additions we should make. This is ready to merge!

Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
calvincestari: At the time I thought it read nicely and fit with the Swift naming guidelines. Not opposed to renaming it though.

Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
calvincestari: This is a pretty nice suggestion, I think that could be really useful. Yes I'll add it to #2101.

PR #2306, opened 2022-06-09
API for Local Cache Mutations

AnthonyMDev:
This defines `LocalCacheMutation` and refactors the `ApolloStore` to handle read/write transactions using them.
It also makes `DataDict` mutable to be used in `LocalCacheMutation`.

Next PRs:
[] Supporting local cache mutations for `mutation`, `subscription`, and `fragment`. (Currently only supports queries).
[] Generating the local cache mutations


calvincestari (approved):
LGTM!

PR #2307, opened 2022-06-09
Support local mutation on fragment and operations

AnthonyMDev:


Sources/ApolloCodegenLib/Frontend/JavaScript/src/utilities/apolloCodegenSchemaExtension.ts
```diff
@@ -2,10 +2,10 @@ import { DirectiveDefinitionNode, DocumentNode, Kind, NameNode, StringValueNode
 
 const directive_apollo_client_ios_localCacheMutation: DirectiveDefinitionNode = {
   kind: Kind.DIRECTIVE_DEFINITION,
-  description: stringNode("A directive used by the Apollo iOS client to annotate queries that should be used for local cache mutations instead of standard query operations."),
+  description: stringNode("A directive used by the Apollo iOS client to annotate operations or fragments that should be used exclusively for generating local cache mutations instead of as standard operations."),
```
calvincestari, line 5: Do you think we might need to clarify further that these will be separate data structures to regular operations? "used exclusively for" is quite telling already though. 🤔

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
```diff
@@ -414,7 +414,70 @@ class ReadWriteFromStoreTests: XCTestCase, CacheDependentTesting, StoreLoading {
...
+        XCTAssertEqual(graphQLResult.source, .cache)
+        XCTAssertNil(graphQLResult.errors)
+
+        let data = try XCTUnwrap(graphQLResult.data)
+        XCTAssertEqual(data.hero.name, "Artoo")
+      }
+    }
+  }
+
+  func test_updateCacheMutation_givenMutationOperation_updateNestedField_updatesObjectAtMutationRoot() throws {
```
calvincestari, line 428: Are we missing a test using `MockLocalCacheMutationFromSubscription`? _Or did I miss it in this PR?_

calvincestari (approved):

Tests/ApolloTests/Cache/ReadWriteFromStoreTests.swift
AnthonyMDev: I chose not to make one because it doesn't have any new behavior that isn't tested in the one for mutation. We need a test that we are calling `CacheReference.rootCacheReference(for: CacheMutation.operationType).key` at the call site in `ApolloStore`. Once we have one test verifying that, I didn't feel the need to add a redundant test for subscriptions. It's overkill that provides next to no value in so far as actually verifying real behavior. If you disagree I can add one though.

Sources/ApolloCodegenLib/Frontend/JavaScript/src/utilities/apolloCodegenSchemaExtension.ts
AnthonyMDev: Yeah, that's why I added "exclusively". I think we should make that clear in documentation. Very few people are ever going to be looking at the directive description here.

PR #2311, opened 2022-06-13
Generate local cache mutations 

AnthonyMDev:
Generate Local Cache Mutation Templates!

Also needed to strip the `@apollo_client_ios_localCacheMutation` directive from the definition we send in the network request. While doing that, I did some clean up to the code that adds the `__typename` field. This way we just make one pass to convert the client definition into one suitable for sending to the server. Also added tests for that!


Tests/ApolloCodegenTests/CodeGeneration/Templates/OperationDefinitionTemplateTests.swift
```diff
@@ -399,8 +399,6 @@ class OperationDefinitionTemplateTests: XCTestCase {
 
     let actual = renderSubject()
 
-//    print(actual)
```
calvincestari, line 402: Hehe, this was probably me - 😬

Tests/ApolloCodegenTests/TestHelpers/LineByLineComparison.swift
```diff
@@ -66,6 +68,20 @@ public func equalLineByLine(
   }
 }
 
+fileprivate func PrettyPrintedFailureResult(
+  actual: String?,
+  message: ExpectationMessage
+) -> PredicateResult {
+  if let actual = actual {
+    print ("Actual Document:")
+    print(actual)
```
calvincestari, line 77: Nice quality of life change for us - thanks!

Tests/TestPlans/Apollo-CodegenTestPlan.xctestplan
```diff
@@ -13,6 +13,7 @@
   },
   "testTargets" : [
     {
+      "parallelizable" : true,
```
calvincestari, line 16: I'm not certain what level of service we pay for in CircleCI. This could work against us if the shared machines we run CI on become saturated with other jobs too. Appears to give a nice improvement but worth keeping an eye on and remembering if we see odd test behaviour. Nice find!

calvincestari (commented):
I had a comment re. this in one of the TS test files but we need to consider the implication of stripping things from the query string on Operation Safelisting. I feel like it may force us to implement the safelisting logic in apollo-ios because our client is the only one that would understand the local-only directives.

Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Schema/Schema.swift
```diff
@@ -11,6 +11,12 @@ where Schema == AnimalKingdomAPI.Schema {}
 public protocol InlineFragment: ApolloAPI.SelectionSet & ApolloAPI.InlineFragment
 where Schema == AnimalKingdomAPI.Schema {}
 
+public protocol MutableSelectionSet: ApolloAPI.MutableRootSelectionSet
```
calvincestari, line 14: If I'm not mistaken I think this PR is missing the changes to `SchemaTemplate.swift` to generate this code. I just rebased my PR (#2313) and this file isn't having these protocols generated. Let's sync on this in the morning.

PR #2323, opened 2022-06-21
refactor: Various uses of `ApolloCodegenConfiguration`

calvincestari:
This PR cleans up the usage of the shared configuration property in the templates and tests.
* Moves the `config` property into the protocol so it's available in the protocol extension
* Removes the need to pass `config` into various functions since it's now available on the protocol
* Refactors some test repetition
* Enforces consistency of template properties


AnthonyMDev (approved):
Thanks for this!

PR #2328, opened 2022-06-24
Support Schema Extension Compilation

### LINKED ISSUE #2058: Extend GraphQLJSFrontend to work with schema extension files
Backstory to this is https://github.com/apollographql/apollo-ios/pull/2055#discussion_r765447128.

tl;dr - the schema + extensions needs to be parsed and merged, like operation documents, then the 'complete' AST can be created from the combined document.

AnthonyMDev:
closes #2058


Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
```diff
@@ -31,7 +33,37 @@ export {
...
+export function loadSchemaFromSources(sources: Source[]): GraphQLSchema {
+  var introspectionJSONResult: Source | undefined
+
+  var documents = new Array<DocumentNode>()
+  for (const source of sources) {
+    if (source.name.endsWith(".json")) {
+      if (!introspectionJSONResult) {
+        introspectionJSONResult = source        
+      } else {
+        throw new Error(`Schema search paths can only include one JSON schema definition.
```
calvincestari, line 45: Where does this constraint come from? Is graphql.js unable to merge multiple JSON sources? IMO extensions make sense in SDL but I wonder if there is an edge case where the schema is broken up into multiple JSON files; that may not even be possible though.

Either way I think this constraint should make it's way up into the header doc for `schemaSearchPaths`.

calvincestari (approved):
This looks great - thank you! The new test coverage is spot on.

Sources/ApolloCodegenLib/Frontend/JavaScript/src/index.ts
calvincestari: The precondition just states the required extension but doesn't mention there can only be one. "schema files" could be interpreted to mean more than one.

> Precondition: JSON format schema files must have the file extension ".json".

AnthonyMDev: Oh youre right! I'll add that.

PR #2335, opened 2022-06-27
fix: Test mock extension, fields and initializers should use module namespace

### LINKED ISSUE #2332: Test Mocks don't compile for `.embeddedInTarget` modules
Using a schema module type of `.embeddedInTarget` results in test mock objects not being generated with the correct namespace.

calvincestari:
Fixed #2332 

Test mock field types and convenience initializers were not being generated with the schema module namespace when required.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
@@ -1,7 +1,10 @@
 extension GraphQLType {
 
   enum RenderContext {
+    /// Renders the type for use in an operation selection set.
     case selectionSetField(forceNonNull: Bool = false)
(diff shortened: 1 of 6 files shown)


Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
```diff
@@ -1,7 +1,10 @@
 extension GraphQLType {
 
   enum RenderContext {
+    /// Renders the type for use in an operation selection set.
     case selectionSetField(forceNonNull: Bool = false)
+    /// Renders the type for use in a test mock object.
+    case testMockField(forceNonNull: Bool = false)
```
AnthonyMDev, line 7: Yay for `RenderContext`!

AnthonyMDev (approved):
Thanks so much for solving this! Tests look on point!

PR #2336, opened 2022-06-27
1.0/introspection json output

### LINKED ISSUE #2159: Add configuration option to schema downloading for introspection output as json
This was raised in https://github.com/apollographql/apollo-ios/issues/2155 - need a decision on whether to add the config option.

AnthonyMDev:
Closes #2159

This doesn't have any tests, because the Schema Downloader tests seem to be all in the integration tests which are not currently running. Those tests are pretty bare bones and need to be improved in the future as well, but this should work as intended.


calvincestari (approved):
We'll convert these tests to unit tests at some point, they'll be more valuable like that.

PR #2338, opened 2022-06-27
chore: SwiftScripts update for 1.0.0-alpha.8 release

calvincestari:
* Updates Package.swift to link to the `1.0.0-alpha.8` tag
* Removes Package.resolved from AnimalKingdomAPI. This should ensure that any other package dependencies would pull the latest version allowed instead of being tied to the resolved version. This can give us some predictability that all versions in the stated dependency continue to build.


calvincestari:
The apollo-ios-cli failing tests are fixed in https://github.com/apollographql/apollo-ios/pull/2340.

PR #2339, opened 2022-06-27
Generate APQ operation ids

AnthonyMDev:
Next PR will generate the `operationIDs.json` correctly.
Still need to do additional normalization of the operation source for generating operation IDs for legacy safelisting of operations.

PR #2341, opened 2022-06-28
feature: Support deprecation warnings with the Swift `@available` attribute

calvincestari:
_This PR is layered on top of #2339 to avoid conflicts with the additional configuration options added in that PR._

Adds a configuration option that supports annotating deprecated enum cases with the Swift `@available` attribute. The default is to include the attribute.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -289,6 +289,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///
     /// See `APQConfig` for more information on Automatic Persisted Queries.
     public let apqs: APQConfig
+    /// Annotate generated Swift code with the Swift `available` attribute and `deprecated`
+    /// argument for parts of the GraphQL schema annotated with the built-in `@deprecated`
+    /// directive.
+    ///
+    /// This is currently limited to enum values.
+    public let warningsOnDeprecatedUsage: Composition
 
     /// Designated initializer.
     ///
@@ -305,13 +311,15 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       queryStringLiteralFormat: QueryStringLiteralFormat = .multiline,
       deprecatedEnumCases: Composition = .include,
       schemaDocumentation: Composition = .include,
-      apqs: APQConfig = .disabled
+      apqs: APQConfig = .disabled,
+      warningsOnDeprecatedUsage: Composition = .include
     ) {
       self.additionalInflectionRules = additionalInflectionRules
       self.queryStringLiteralFormat = queryStringLiteralFormat
       self.deprecatedEnumCases = deprecatedEnumCases
       self.schemaDocumentation = schemaDocumentation
       self.apqs = apqs
+      self.warningsOnDeprecatedUsage = warningsOnDeprecatedUsage
     }
   }
 
--- a/Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
@@ -25,9 +25,22 @@ struct EnumTemplate: TemplateRenderer {
   }
 
   private func evaluateDeprecation(graphqlEnumValue: GraphQLEnumValue) -> String? {
-    switch (config.options.deprecatedEnumCases, graphqlEnumValue.deprecationReason) {
-    case (.exclude, .some): return nil
-    default: return "case \(graphqlEnumValue.name)"
+    switch (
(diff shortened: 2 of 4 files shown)


CodegenCLI/Tests/Commands/InitializeTests.swift
```diff
@@ -214,7 +215,7 @@ class InitializeTests: XCTestCase {
       let expectedPath = URL(fileURLWithPath: outputPath).standardizedFileURL.path
 
       expect(actualPath).to(equal(expectedPath))
-      expect(data).to(equal(self.expectedJSON.data(using: .utf8)!))
+      expect(data?.asString).to(equal(self.expectedJSON))
```
calvincestari, line 218: This change is simply to get more readable test failure messages. Comparing strings will give us the human-readable string instead of this obscure failure message.
> failed - expected to equal <Data<hash=205197838,length=820>>, got <Data<hash=205197838,length=821>>

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -289,6 +289,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///
     /// See `APQConfig` for more information on Automatic Persisted Queries.
     public let apqs: APQConfig
+    /// Annotate generated Swift code with the Swift `available` attribute and `deprecated`
+    /// argument for parts of the GraphQL schema annotated with the built-in `@deprecated`
+    /// directive.
+    ///
+    /// This is currently limited to enum values.
```
calvincestari, line 296: This implementation is currently limited to only deprecated enum values but if it can be extended to the other parts of a GraphQL schema that can be deprecated we should consider it.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -289,6 +289,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+    public let warningsOnDeprecatedUsage: Composition
```
martinbonnin, line 297: I initially understood `warningsOnDeprecatedUsage` as an option to display warnings during the codegen itself (not as a way to add annotations in the generated code). Maybe it's worth doing the distinction? Warning during codegen could be used together with something like "warningAsErrors" to fail the CI in case a field becomes deprecated?

calvincestari: I couldn't settle on a clean name for the new option but yes it could be better stating it'll result in Swift compiler warnings; `swiftDeprecationWarnings` or similar? I personally don't think there is much value in warning during code generation, those logs are easy to ignore.

martinbonnin: > swiftDeprecationWarnings or similar?

I can suggest `addAvailableAnnotations`, `annotateDeprecatedUsages` or if a noun is needed `annotationOnDeprecatedUsages`. I wouldn't trust me on this though as I'm not familiar enough with the ecosystem to know what would feel "natural"

 > those logs are easy to ignore

True. It's only useful with something like `allWarningsAsErrors` for catching these early in the CI

AnthonyMDev: This is specifically within the `OutputOptions` struct, so it seems clear to me, but fair point that others may not make that assumption. I don't think any of the other name suggestions here are more clear than what we have honestly. The added documentation should be enough here IMO.

Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
```diff
@@ -25,9 +25,22 @@ struct EnumTemplate: TemplateRenderer {
   }
 
   private func evaluateDeprecation(graphqlEnumValue: GraphQLEnumValue) -> String? {
```
AnthonyMDev, line 27: Can we change the name of this function to `EnumCaseTemplate(_ graphQLEnumValue: GraphQLEnumValue)`.

This is returning the template string to render, not a value indicating an evaluation of deprecation reason.

AnthonyMDev (changes requested):
I'd like us to also get the Field deprecations in here if we can. If you wanna do them in a follow up PR, that's fine, but they should be in before considering this feature complete.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
calvincestari: > We should 100% de doing deprecations warnings for deprecated fields as well if this is set to true.

To be done in a separate PR.

AnthonyMDev (approved):
Beautiful. Let's go!

PR #2350, opened 2022-06-29
feature: Implement generation of operation identifiers file

calvincestari:
Enables generation of the operation identifiers file which is used to support APQs.


Apollo.xcodeproj/project.pbxproj
```diff
@@ -2291,11 +2299,12 @@
 				DE5FD60A276970FC0033EE23 /* FragmentTemplate.swift */,
 				E69BEDA42798B86D00000D10 /* InputObjectTemplate.swift */,
 				E623FD292797A6F4008B4CED /* InterfaceTemplate.swift */,
+				DE71FDBB2853B67C005FA9CC /* LocalCacheMutationDefinitionTemplate.swift */,
 				DE9CEAEE282C62B700959AF9 /* MockObjectTemplate.swift */,
 				E6203343284F1D1100A291D1 /* MockUnionTemplate.swift */,
 				E64F7EC027A122300059C021 /* ObjectTemplate.swift */,
 				DE5FD5FC2769222D0033EE23 /* OperationDefinitionTemplate.swift */,
-				DE71FDBB2853B67C005FA9CC /* LocalCacheMutationDefinitionTemplate.swift */,
```
calvincestari, line 2298: These changes are from sorting the files by name.

Sources/ApolloCodegenLib/FileGenerators/OperationIdentifiersFileGenerator.swift
```diff
@@ -0,0 +1,56 @@
...
+  let filePath: String
+
+  /// `OperationIdentifier` objects to be written to `path`.
+  private var operationIdentifiers: [OperationIdentifier] = []
+
+  /// Designated initializer.
+  ///
+  /// Parameters:
+  ///  - config: A configuration object specifying output behavior.
+  init?(config: ApolloCodegen.ConfigurationContext) {
```
calvincestari, line 24: This uses a failable initializer which I prefer since it places the logic of path evaluation in this struct. The alternative is to always initialize an instance but then `collectOperationIdentifier` would need to evaluate `path` for a quick exit; I think that's inefficient.

Sources/ApolloCodegenLib/Templates/OperationIdentifiersTemplate.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import ApolloUtils
+
+/// Provides the format to output a file used for APQ registration.
+struct OperationIdentifiersTemplate {
```
calvincestari, line 5: I debated simply using `Codable` on the `OperationIdentifier` struct but it resulted in needing to write a custom encoder/decoder which at that point our templating system is quicker.

Sources/ApolloCodegenLib/Templates/OperationIdentifiersTemplate.swift
```diff
@@ -0,0 +1,35 @@
...
+    )
+  }
+
+  func render() -> String {
+    template.description
+  }
+}
+
+fileprivate extension String {
+  func jsonEscaped() -> String {
```
calvincestari, line 31: I'm not sure of another 'standard' way to do this.

calvincestari: I'm second-guessing myself now - maybe it would be better to just implement custom `Codable` conformance and know that JSON escaping/formatting will be 100% correct?

Sources/ApolloCodegenLib/FileGenerators/OperationIdentifiersFileGenerator.swift
AnthonyMDev: Yeah, I'm down with the approach!

AnthonyMDev (changes requested):
This looks great, but we need to resolve the issue with the referenced fragments not being included in the source.

Sources/ApolloCodegenLib/Templates/OperationIdentifiersTemplate.swift
calvincestari: I tried `Codable` yesterday but a plain `Dictionary` didn't preserve order which made testing fail randomly and `OrderedDictionary` serialized with array opening/closing braces and the object format wasn't right.

I'll try this approach but I suspect the lack of ordering could still cause an issue with tests.

AnthonyMDev:
Interesting idea: Could you add JSON encoding as a custom interpolation method to `TemplateString`?
https://www.avanderlee.com/swift/string-interpolation/#case-2-printing-json

calvincestari:
That custom interpolation looks interesting..I'll explore that a bit more. Thanks for the link!

Sources/ApolloCodegenLib/Templates/OperationIdentifiersTemplate.swift
```diff
@@ -0,0 +1,34 @@
...
+    )
+  }
+
+  func render() throws -> String {
+    try template().description
+  }
+}
+
+fileprivate extension String.StringInterpolation {
+  mutating func appendInterpolation(json jsonData: Data) {
```
AnthonyMDev, line 31: Awesome!

AnthonyMDev (approved):

calvincestari:
OK, the fragment collection is now fixed. I also moved the collection logic from the file generator into the template so that the template is now wholly responsible for collection and rendering which makes it easier to keep test logic contained in a single test file. The file generator now routes collection to the template and is only responsible for writing the rendered data to the path.

AnthonyMDev (approved):
Awesome!

PR #2351, opened 2022-06-29
Generate documentation for generated objects

AnthonyMDev:
Also added newline at end of each generated file. They should be there, and Xcode automatically adds them when you open the file. So it causes weird issues where you get source control modifications for generated files when you regenerate them or view them.


calvincestari:
> Also added newline at end of each generated file.

Nice! Thanks for doing this.

calvincestari (approved):
This is a great addition to the generated output, nice work!

PR #2356, opened 2022-07-01
feature: Generate deprecation warnings on InputObject when fields are deprecated

calvincestari:
This PR is the final piece to adding deprecation warnings to the Swift generated code.

* Generates a second InputField initializer that doesn't contain any of the deprecated fields
* Annotates initializer with deprecated fields as deprecated using Swift's `@available` attribute
* Annotates fields as deprecated using Swift's `@available` attribute


AnthonyMDev (approved):
Nice work!

PR #2361, opened 2022-07-07
Expose GraphQLResultError Path String

joshuashroyer-toast:
# Summary
Adds a new publicly available computed property to `GraphQLResultError` which just exposes the `path` description.

# Motivation
My team uses apollo-ios SDK for most of our networking needs, however, we also include crash and error reporting within our project. While using Apollo, if there are errors thrown and my team has identified it's something we care about reporting, we log those to our reporting tools.

I go into more detail in the [community post I made](https://community.apollographql.com/t/apollo-ios-remove-potential-pii-from-graphqlresulterror-jsondecodingerror-couldnotconvert/3714), but the problem we have is when the thrown errors contain PII (Personally Identifiable Information). In this specific instance, we saw this with the `JSONDecodingError.couldNotConvert` error when combined with the `GraphQLResultError`. You could have a situation like the following:
> Error at path “accountHolder.birthDate)”: couldNotConvert(value: 2000-01-01T00:00:00.000Z, to: Foundation.Date)

For this, we decided we needed to map the errors we cared about reporting into new errors - removing the parts that could be considered PII. We could regenerate the `JSONDecodingError.couldNotConvert` error without a problem, but the `GraphQLResultError` type couldn't be re-created as the underlying `ResponsePath` type is not public and therefore neither is the `path` property.

# Solution
Considering the above, we had a couple of options we could take:
- Parse the `errorDescription` String to take what we needed?
  - Fragile - what if the string format changed between versions?
- Only recreate the `JSONDecodingError.couldNotConvert` error?
  - This would work, but we'd lose the information for which json key had the issue.
- Fork the repo and make an adjustment?
  - Likely the best outcome to maintain error fidelity in our logs while ensuring no PII is logged.

We opted to fork the repo and only add in a `pathString` computed property to the `GraphQLResultError` type so we could map to new errors while still including the json key which had the issue.

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -249,6 +249,7 @@
 		DED46035261CEA660086EF63 /* ApolloTestSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9F8A95781EC0FC1200304A2D /* ApolloTestSupport.framework */; };
 		DED46042261CEA8A0086EF63 /* TestServerURLs.swift in Sources */ = {isa = PBXBuildFile; fileRef = DED45C172615308E0086EF63 /* TestServerURLs.swift */; };
 		DED46051261CEAD20086EF63 /* StarWarsAPI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9FCE2CFA1E6C213D00E34457 /* StarWarsAPI.framework */; };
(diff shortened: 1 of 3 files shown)


calvincestari:
Thanks for the contribution @joshuashroyer-toast. I don't have any issue with the change, but one thing that I think is required in order to merge it in is a test for `ResponsePath.description` to ensure it's always returning what we're expecting. Especially since that will be the public output. @AnthonyMDev any additional comment?

_ResponsePath is before my time on the team but I don't see any tests for it currently._

calvincestari:
@AnthonyMDev I've added unit tests for `ResponsePath`, let me know if you have any objection to merging in this change.

AnthonyMDev (approved):

PR #2366, opened 2022-07-11
Add a Netlify config to the docc directory

trevorblades:
This PR adds a `netlify.toml` that will enable deployments of docc archives on Netlify.


AnthonyMDev:
You are the best @trevorblades !

PR #2367, opened 2022-07-12
Re-enable all CI test suites for 1.0

### LINKED ISSUE #2364: Re-enable all CI tests for `release/1.0` branch
While building out the 1.0 feature set we disabled a bunch of CI test suites for any branch that has 1.0 in the branch name. With Beta out now it's time to bring those tests back and ensure we've not indirectly broken other parts of the codebase.

calvincestari:
Closes #2364 

This PR updates and enabled all CI test suites on the `release/1.0` branch.
* Removes the 1.0 branch filter for all CI jobs
* Fixed: ApolloTestSupport was not using the right xcconfig for some configurations.
* Fixed: The 'Header' build phase was moved above all instances of 'Compile Sources'. I'm not sure when/why they were moved down in the order but Xcode threw a warning that building would be problematic unless it was moved up the build order.
* Library search path was not needed in the xcconfig for test helpers.
* TestMockTests changed to use NSObject equality function.
* Replaced legacy generated swift code with 1.0 codegen generated models.
* Enables parallelization of CI, Integration and Unit test plans


Apollo.xcodeproj/project.pbxproj
```diff
@@ -437,10 +437,9 @@
...
 		E610D8E1278F8F3D0023E495 /* UnionFileGeneratorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E610D8E0278F8F3D0023E495 /* UnionFileGeneratorTests.swift */; };
 		E616B6D126C3335600DB049E /* ExecutionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E616B6D026C3335600DB049E /* ExecutionTests.swift */; };
-		E6180AA528093EB7004EC2A3 /* API.swift in Sources */ = {isa = PBXBuildFile; fileRef = E6180A9F28093EB7004EC2A3 /* API.swift */; };
 		E6180AA628093EB7004EC2A3 /* schema.graphqls in Resources */ = {isa = PBXBuildFile; fileRef = E6180AA128093EB7004EC2A3 /* schema.graphqls */; };
-		E6180AA728093EB7004EC2A3 /* operation_ids.json in Resources */ = {isa = PBXBuildFile; fileRef = E6180AA228093EB7004EC2A3 /* operation_ids.json */; };
 		E6180AA828093EB7004EC2A3 /* subscription.graphql in Resources */ = {isa = PBXBuildFile; fileRef = E6180AA328093EB7004EC2A3 /* subscription.graphql */; };
+		E61D426A28803F3600C00F97 /* SubscriptionAPI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E66B836628093D7C001B3F2D /* SubscriptionAPI.framework */; };
```
calvincestari, line 442: The `SubscriptionAPI` framework was missing as a library for `ApolloServerIntegrationTests` target.

Apollo.xcodeproj/project.pbxproj
```diff
@@ -2888,8 +2900,9 @@
 		E66B836728093D7C001B3F2D /* SubscriptionAPI */ = {
 			isa = PBXGroup;
 			children = (
+				E6FDD746288041F300BD0ADA /* Operations */,
+				E6FDD749288041F300BD0ADA /* Schema */,
 				E6180AA028093EB7004EC2A3 /* graphql */,
-				E6180A9F28093EB7004EC2A3 /* API.swift */,
```
calvincestari, line 2892: This is deleting the old 0.x codegen generated files.

SwiftScripts/Sources/TargetConfig/Target.swift
```diff
@@ -6,6 +6,7 @@ public enum Target: CaseIterable {
 //  case gitHub
   case upload
   case animalKingdom
+  case subscription
```
calvincestari, line 9: The SubscriptionAPI target generated files added to our internal tooling.

Tests/ApolloServerIntegrationTests/StarWarsServerCachingRoundtripTests.swift
```diff
@@ -61,9 +62,7 @@ class StarWarsServerCachingRoundtripTests: XCTestCase, CacheDependentTesting {
   }
   
   func testHeroAndFriendsNamesWithIDsQuery() {
-    client.store.cacheKeyForObject = { $0["id"] }
```
calvincestari, line 64: This doesn't appear to have any affect on the test and the mechanism for cache key resolution has changed to be class-based.

Tests/TestPlans/Apollo-CITestPlan.xctestplan
```diff
@@ -13,6 +13,7 @@
   },
   "testTargets" : [
     {
+      "parallelizable" : true,
```
AnthonyMDev, line 16: I'm a little bit concerned about this causing some weird test failures. When doing tests in parallel, you have to be really careful about test isolation in a more specific way, which is a **good thing**, but I don't know if all of our tests are handling that correctly right now.

But if they are passing right now, then it's probably okay. Guess we'll find out!

AnthonyMDev (approved):
Thanks for getting this cleaned up! I know this was a tedious one to work through.

My only concern is the concurrency subscription tests that used to fail intermittently. If they start failing again, we are going to need to do something about that or disable those specific tests.

Tests/TestPlans/Apollo-CITestPlan.xctestplan
calvincestari: I can't say that I saw a massive improvement in time-to-run after parallelization was enabled, so I'm not opposed to disabling it if it begins to cause problems; I'd rather tackle the root cause of those anyways and enforce better test isolation.

PR #2368, opened 2022-07-12
Fix: Update Index Links 

lleadbet:
The current links aren't working on the docs, so this replaces them with standard MD links to become functional.


AnthonyMDev:
Would like to get @trevorblades to look at this, but I'm pretty sure that since we moved over to the central docs repo, we don't need that `ButtonLink` component anymore.

StephenBarlow:
Fixed with https://github.com/apollographql/apollo-ios/pull/2369

trevorblades:
@calvincestari @AnthonyMDev sorry about that, there was a regression in the docs framework caused by a couple of lines being accidentally removed in a recent refactor that we did. This comment has more info: https://github.com/apollographql/apollo-ios/pull/2369#issuecomment-1182468343

Anyway, I opened a new PR that reverts @StephenBarlow's recent change and the `ButtonLink` stuff should be working again as normal :)

PR #2369, opened 2022-07-12
Fix buttons on docs root

StephenBarlow:


calvincestari:
Thanks @StephenBarlow!

PR #2371, opened 2022-07-12
Revert "Fix buttons on docs root"

trevorblades:
Reverts apollographql/apollo-ios#2369


trevorblades:
@StephenBarlow @calvincestari not sure why the codgen CI step fails here, but the docs change does seem to work now 🤷

StephenBarlow (approved):

calvincestari:
@trevorblades I've seen that a bit recently. It's odd and random; has to do with validating the checksum of the downloaded CLI. Can ignore in this PR.

PR #2379, opened 2022-07-18
Fix typo in tutorial text

mischa-hildebrand:
Removed duplicate "the"


calvincestari:
I'm merging this in since we still have the 0.x documentation but the tutorial will be discontinued in 1.0.

PR #2380, opened 2022-07-18
Fix incorrect parameter name in tutorial

mischa-hildebrand:
`id` was set as a parameter, but the query expects `launchId`.


calvincestari (approved):
Thanks for the contribution @mischa-hildebrand!

PR #2382, opened 2022-07-19
ci: Remove unused CocoaPodsTrunk job

calvincestari:
This PR removes the `CocoaPodsTrunk` job which was never being used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,17 +196,6 @@ jobs:
       - common_test_setup
       - codegen_cli_run_tests
 
-  CocoaPodsTrunk:
-    macos:
-      xcode: << pipeline.parameters.xcode_version >>
-    steps:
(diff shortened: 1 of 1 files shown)


calvincestari:
@AnthonyMDev I don't believe we've ever used the CocoaPodsTrunk job; has it been used during your time on the project, do we need to keep it around for anything?

AnthonyMDev:
We haven't been using this. We haven't pushed the 1.0 alphas to Cocoapods, but even for the 0.x releases, I have not done a manual `pod trunk push` in a while... Glad you've been remembering to do so.

I'm fine with removing this from CI, since I'm not even clear on how we would get CircleCI to trigger this job to run? But we should probably be automating more of this when we can...

AnthonyMDev (approved):

PR #2389, opened 2022-07-22
tests: Add cache key resolution tests

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/2390.

Fixed `CacheKeyProvider` protocol conformance to work with extensions and adds test cases to ensure that cache key resolution behaves [as described in the 1.0 Codegen Proposal](https://github.com/apollographql/apollo-ios/blob/release/1.0/CodegenProposal.md#cache-key-resolution).
* Removed the `__cacheKeyProvider` property from the `Object` class because while schema objects are subclasses and can override class properties, extensions cannot. Users would be forced to subclass their schema objects to implement `CacheKeyProvider` = no bueno. The implementation now checks if the object class conforms to `CacheKeyProvider` which is possible within an extension.
* Added tests for both `SchemaUnknownTypeCacheKeyProvider` and `CacheKeyProvider`.


AnthonyMDev (changes requested):
These tests look great! Thanks so much for doing that!

Having cache key resolution check for the `Object` to conform to `CacheKeyProvider` is actually how I originally had it set up. But I realized that being able to share the same `CacheKeyProvider` on different `Object` types was easier when it was it's own type. I do see the limitation you're talking about, which I totally missed. 

It's awkward to me that we still have the `SchemaConfiguration` return a provider seperately, but not `Object`s. But it feels more ergonomic to do it this way. You can do what youre doing in the test and have the schema conform itself and return `self` as the provider, but if you want different providers for different types, it's cleaner to be able to return different objects.

I also think that, for flexibility, we should make the functions on `CacheKeyProvider` not static. The minor overhead in allocating a value is negligible and it does allow people to be able to make their providers more robust (ie. they can initialize them with different properties and configurations of their own creation). If they don't want to allocate them constantly, they can store them statically still.

```swift
private struct AlphaCacheKeyProvider: CacheKeyProvider {
  static let shared = Self.init()
  // ...
}

class Alpha: Object {
  static var __cacheKeyProvider {  AlphaCacheKeyProvider.shared }
}
```

What if we made an `ObjectProtocol` that had the `__cacheKeyProvider` property on it, would that allow it to be overridden in an extension?

```swift
public protocol ObjectProtocol {
  static var __cacheKeyProvider: CacheKeyProvider? { get }
}
extension ObjectProtocol {
  static var __cacheKeyProvider: CacheKeyProvider? { nil }
}

public class Object: ObjectProtocol {}
```

Let's experiment with that, and make `CacheKeyProvider` methods not `static`. I've continued to iterate on this piece, as I have never been totally happy about it. Let's lock in the best API we can come up with before the 1.0 GA where we can't change it anymore haha.

calvincestari:
@AnthonyMDev the latest changes reflect what we spoke about earlier today; still static methods but no longer requiring `SchemaConfiguration` to return a cache key provider.

I think the changes are an improvement to what we had before but tomorrow I'll iterate on the non-static implementation.

calvincestari:
We've put these changes on hold until we conclude discussions about changes to the schema object models.

calvincestari:
We're getting close to a potential refactor of the schema configuration. I'm closing this PR and will submit another once we've gotten the cache key resolution mechanism in place for the refactor.

PR #2396, opened 2022-07-27
Xcode 14 has deprecated bitcode and so we should disable it

stareque-atlassian:
This document explains how bitcode is not supported in Xcode 14:

> Starting with Xcode 14, bitcode is no longer required for watchOS and tvOS applications, and the App Store no longer accepts bitcode submissions from Xcode 14.
Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with > bitcode is deprecated. Please update your project and/or target settings to disable bitcode.” The capability to build with bitcode will be removed in a future Xcode release. IPAs that contain bitcode will have the bitcode stripped before being submitted to the App Store. Debug symbols for past bitcode submissions remain available for download. (86118779)

https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -2153,7 +2153,6 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 9FC750581D2A532D00458D91 /* Build configuration list for PBXNativeTarget "Apollo" */;
 			buildPhases = (
-				90690D322243442F00FC2E54 /* Ensure no build settings are in the Xcode project */,
 				9FC7503F1D2A532C00458D91 /* Sources */,
 				9FC750411D2A532C00458D91 /* Headers */,
 				9FC750421D2A532C00458D91 /* Resources */,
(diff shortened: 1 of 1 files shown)


calvincestari:
Thanks for the contribution @stareque-atlassian! This project uses xcconfig files so we can't remove that build phase from the project though; it's there to ensure we don't have a mixed configuration. I'm pretty sure there is a way we can do this with xcconfig files.

calvincestari:
@stareque-atlassian I've created #2397 to keep track of this issue. I'm going to close this PR as we can't merge a solution that removes the project settings check build phase. If you find another way to achieve this please feel free to reopen this PR or create a new one - thanks.

stareque-atlassian:
Thanks so much for the quick response. Raised a PR for this with changes only in xcconfig files https://github.com/apollographql/apollo-ios/pull/2398

Tested it manually by using `carthage` to build it successfully with Xcode 14.

Can I request to make a hotfix tag `0.51.3`. This PR branch is rebased from the tag `0.51.2`

PR #2399, opened 2022-07-27
Get rid of ApolloUtils

AnthonyMDev:
Removes the `ApolloUtils` target and all references to it.

**Nets a deletion of 3,616 lines of code and documentation!**

We decided we don't need this anymore. The things that used to be shared here are actually no longer shared. There is no code shared between the `Apollo` and `ApolloCodegenLib` targets. Rendering this unnecessary.

Removing additional targets is better for startup time. All the `ApolloExtensions` were being used only to expose internal extension functionality. None of it should be exposed publicly, so the namespacing is not necessary. This should reduce our binary size and memory footprint (though very minor).


calvincestari (approved):
Nice work here, thank you! Much needed 'modernization' and gets rid of a bunch of old unnecessary code = awesome!

kylebrowning:
🥳

PR #2400, opened 2022-07-28
Update API reference URLs in v1 sidebar

trevorblades:
Updates the API reference links with the correct URLs, and adds an "Overview" link.


AnthonyMDev (approved):
Thank you!

PR #2408, opened 2022-08-01
refactor: Removes ApolloCodegenConfiguration.validation

### LINKED ISSUE #2404: Apollo 1.0 Beta: Destination paths should not need to exist.
## Bug report

When you are missing certain directories in output paths, it fails codegen with missing file system path errors. Based on working with other similar tools, my expectation is that Codegen creates the output paths in their entirety.

calvincestari:
Closes #2404 

This PR removes the validation logic from `ApolloCodegenConfiguration`.
* It was not helpful in determining possible configuration errors and incorrectly required output parent paths to exist prior to execution.
* The validation logic that is _deterministic_ has been moved into `ApolloCodegen` and checked before any time is spent parsing input or generating models.
* The validation logic that is _indeterministic_ , such as missing input paths or inaccessible output paths, is now left to the file system and will be raised inline as encountered.


CodegenCLI/Sources/main.swift
```diff
@@ -5,10 +5,9 @@ struct CodegenCLI: ParsableCommand {
   static var configuration = CommandConfiguration(
     commandName: "apollo-ios-cli",
     abstract: "A command line utility for Apollo iOS code generation.",
-    version: "1.0.0-alpha.8",
+    version: "1.0.0-beta.1",
```
AnthonyMDev, line 8: Is there some way for us to have this automatically update when we bump the version of the main repo?

AnthonyMDev (approved):
Thanks for this! There were no changes to the actual Codegen code though. Will it already do the creation of the directories that don't exist? No changes needed there?

calvincestari:
> There were no changes to the actual Codegen code though. Will it already do the creation of the directories that don't exist? No changes needed there?

Yes:
* The [`generate`](https://github.com/apollographql/apollo-ios/blob/release/1.0/Sources/ApolloCodegenLib/FileGenerators/FileGenerator.swift#L31) function in the `FileGenerator` extension calls into our `ApolloFileManager` function,
* which creates the parent directory if it does not exist with [`createContainingDirectoryIfNeeded`](https://github.com/apollographql/apollo-ios/blob/release/1.0/Sources/ApolloCodegenLib/FileManager%2BApollo.swift#L81).

It was always behaving this way so the validation logic in `ApolloCodegenConfiguration` was only ever checking that the parent directory existed = completely useless since it would have been created anyways.

CodegenCLI/Sources/main.swift
calvincestari: If we could link it to apollo-ios xcconfig files maybe but they are outside the project folder so I'm not sure about whether that's possible.

PR #2416, opened 2022-08-03
Update version in Swift Scripting docs

redryerye:
Updated version in Swift Scripting docs


calvincestari (approved):
Thanks for the contribution @redryerye!

Issue #2381, opened 2022-07-18
Automate versioned CLI installation


calvincestari:
This can be automated by using:
- [x] CocoaPods (prepare_command) - see https://github.com/Brightify/Cuckoo/blob/master/Cuckoo.podspec
- [x] Swift Package Plugin - See Swift docc plugin as an example.

AnthonyMDev:
This is finished now and is shipping with 1.0.0-beta.3.

PR #2417, opened 2022-08-03
feat: Builds CLI locally on CocoaPods installs

calvincestari:
PR 1 of #2381

This change is to ensure that the CLI and ApolloCodegenLib are the same version. It does this by simply building the CLI locally during a CocoaPods installation.

To manually test this do the following:
1. Create an Xcode project
2. Use `pod init` to create a Podfile and use `pod 'Apollo', :git => 'https://github.com/apollographql/apollo-ios', :branch => '1.0/versioning-cli-cocoapods'` as the Apollo dependency.
3. After installation you should have a file `apollo-ios-cli` in the `Pods/Apollo` folder.

### DIFF
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -1,24 +1,27 @@
 Pod::Spec.new do |s|
-  s.name         = 'Apollo'
-  s.version      = `scripts/get-version.sh`
-  s.author       = 'Meteor Development Group'
-  s.homepage     = 'https://github.com/apollographql/apollo-ios'
-  s.license      = { :type => 'MIT', :file => 'LICENSE' }
-
-  s.summary      = "A GraphQL client for iOS, written in Swift."
-
-  s.source       = { :git => 'https://github.com/apollographql/apollo-ios.git', :tag => s.version }
-
+  s.name = 'Apollo'
+  s.version = `scripts/get-version.sh`
+  s.author = 'Meteor Development Group'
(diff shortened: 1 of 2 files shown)


Apollo.podspec
```diff
@@ -1,25 +1,28 @@
...
   s.osx.deployment_target = '10.14'
   s.tvos.deployment_target = '12.0'
   s.watchos.deployment_target = '5.0'
 
+  cli_directory = 'CodegenCLI'
+  cli_binary_name = 'apollo-ios-cli'
+  s.preserve_paths = ['CodegenCLI/**/*', cli_binary_name]
+  s.prepare_command = <<-CMD
+    make --directory=CodegenCLI
+    cp #{cli_directory}/.build/release/#{cli_binary_name} #{cli_binary_name}
```
calvincestari, line 22: At the moment the destination of the final CLI binary will be `<project root>/Pods/Apollo`. This is the same behaviour as Cuckoo and I'm not sure we can actually copy something outside of the `Pods` directory. The problem with this destination is that it's not very discoverable. Thoughts?

AnthonyMDev: Pods/Apollo is the pretty standard expectation for Cocoapods users. I think it's fine. We just document this in the usage guide and should be fine!

AnthonyMDev (approved):
Awesome! That was fast :)

PR #2420, opened 2022-08-04
fix: "No such module 'ApolloAPI'"

### LINKED ISSUE #2418: 1.0 Beta 1 -  "No such module 'ApolloAPI'"
## Bug report

I'm trying to use the 1.0 Beta 1 release with Cocoapods. I get a compilation error in `GraphQLResult` which says `No such module 'ApolloAPI'` . When looking at the pod contents, i see ApolloAPI is not included. It looks like it is included in the SPM file but not in the PodSpec. 

## Versions

- `apollo-ios` SDK version: 1.0 Beta 1
- Xcode version:  13.4
- Swift version:  5.6
- Package manager: Cocoapods

## Steps to reproduce

- Setup a project using cocoapods. 
- Add the following line to your podfile: `pod 'Apollo', :git => 'https://github.com/apollographql/apollo-ios.git', :branch => '1.0.0-beta.1'`
- Run `pod install`
- Build the project.

calvincestari:
Closes #2418 

Fixes a couple recent bugs:
1. Don't import ApolloAPI if the environment is CocoaPods
2. ApolloAPI subfolder source files should be included in the podspec
3. Remove duplicate typealias definitions and import ApolloAPI in the modules where those definitions are needed

### DIFF
--- a/Apollo.podspec
+++ b/Apollo.podspec
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
   CMD
 
   s.subspec 'Core' do |ss|
-    ss.source_files = 'Sources/Apollo/*.swift','Sources/ApolloAPI/*.swift'
+    ss.source_files = 'Sources/Apollo/*.swift','Sources/ApolloAPI/**/*.swift'
   end
 
   # Apollo provides exactly one persistent cache out-of-the-box, as a reasonable default choice for
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -259,7 +259,7 @@ public class ApolloStore {
       )
     }
 
-    public func updateObject<SelectionSet: ApolloAPI.MutableRootSelectionSet>(
+    public func updateObject<SelectionSet: MutableRootSelectionSet>(
       ofType type: SelectionSet.Type,
       withKey key: CacheKey,
       variables: GraphQLOperation.Variables? = nil,
--- a/Sources/Apollo/GraphQLResult.swift
+++ b/Sources/Apollo/GraphQLResult.swift
@@ -1,5 +1,7 @@
 import Foundation
+#if !COCOAPODS
 import ApolloAPI
+#endif
 
 /// Represents the result of a GraphQL operation.
 public struct GraphQLResult<Data: RootSelectionSet> {
--- a/Sources/ApolloAPI/Selection.swift
+++ b/Sources/ApolloAPI/Selection.swift
@@ -6,7 +6,7 @@ public enum Selection {
   /// A fragment spread of a named fragment definition.
   case fragment(Fragment.Type)
   /// An inline fragment with a child selection set nested in a parent selection set.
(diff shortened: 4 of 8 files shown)


Apollo.podspec
```diff
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
   CMD
 
   s.subspec 'Core' do |ss|
-    ss.source_files = 'Sources/Apollo/*.swift','Sources/ApolloAPI/*.swift'
+    ss.source_files = 'Sources/Apollo/*.swift','Sources/ApolloAPI/**/*.swift'
```
AnthonyMDev, line 27: Do we need to add the `**/*` to Apollo also? You've only added it to ApolloAPI

calvincestari:
Changed this PR to a draft while I work on the fix for https://github.com/apollographql/apollo-ios/issues/2418#issuecomment-1205438140.

AnthonyMDev (commented):
Aren't we also going to need to add this to the generated files? For those, I imagine it would be:
```swift
#if COCOAPODS
import Apollo
#else
import ApolloAPI
#endif
```

In that case, maybe we should make a module type specifically for Cocoapods in the codegen options? It would work just like `other` but would alter the import statement here. Then you wouldn't need the macro above, you would just do `import Apollo` instead of `import ApolloAPI` for cocoapods.

Am I thinking about this correctly, or is that change not needed?

Apollo.podspec
calvincestari: Not unless we start better organizing the Apollo folder source files into subdirectories. Right now they're all in one folder and the only subfolder in there is `Documentation.docc` which won't contain any source files. So we're fine for now but might need to later.

calvincestari:
> Aren't we also going to need to add this to the generated files?

Yes that change is needed and what you've got there is the fix.

> In that case, maybe we should make a module type specifically for Cocoapods in the codegen options? It would work just like other but would alter the import statement here. Then you wouldn't need the macro above, you would just do import Apollo instead of import ApolloAPI for cocoapods.

Hmm, I hadn't considered this approach but I'm not against it. How much value does `other` still have then?

calvincestari:
> How much value does other still have then?

At the moment it'll be for Carthage and any other dependency manager that does not behave like SPM nor CocoaPods.

calvincestari:
@AnthonyMDev I'm not completely sure what we spoke about is the correct fix. Let's [gather more details about his setup](https://github.com/apollographql/apollo-ios/issues/2418#issuecomment-1206149459) before we make unnecessary changes.

calvincestari:
@AnthonyMDev please don't forget to include these changes in Beta 2.

AnthonyMDev:
Thanks for the reminder. I'll look at these again as soon as I can!

calvincestari:
@AnthonyMDev let's separate the fixes for this into apollo-ios files vs. generate code files. This one tackles the apollo-ios files.

AnthonyMDev:
Closing in favor of #2442

PR #2428, opened 2022-08-05
Add test projects for different codegen module configurations

AnthonyMDev:
This PR:
- Adds the example projects for the different types of module configurations @calvincestari  had set up.
- Created a script that runs the codegen cli against each of the added example projects
- Added the ability for the Codegen engine to resolve the relative paths in the cli config json file against the root URL.


calvincestari (approved):
Thanks for leading me through these changes, and thanks for getting them into the project.

PR #2431, opened 2022-08-09
Fixes for Swift 5.7 and Xcode 14 Beta 

### LINKED ISSUE #2360: No type for 'Value' can satisfy both
## Bug report

Apollo 1.0.0-beta doesn't compile on Xcode 14 beta. 

## Versions

- `apollo-ios` SDK version: 1.0.0-beta.1
- Xcode version: 14 beta 2
- Swift version: 5.7
- Package manager:

## Steps to reproduce
Build project in Xcode 14.

## Further details
No type for 'Value' can satisfy both 'Value == any JSONEncodable' and 'Value : Equatable'
No type for 'Value' can satisfy both 'Value == any JSONEncodable' and 'Value : Hashable'

Screenshot 2022-07-03 at 10 29 04

AnthonyMDev:
This closes #2360 and also cleans up some warnings when using Apollo iOS with Swift 5.7.

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXBuildFile section */
@@ -4496,6 +4496,7 @@
 /* Begin PBXShellScriptBuildPhase section */
 		90690D322243442F00FC2E54 /* Ensure no build settings are in the Xcode project */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
 			buildActionMask = 2147483647;
 			files = (
 			);
--- a/Sources/Apollo/GraphQLDependencyTracker.swift
+++ b/Sources/Apollo/GraphQLDependencyTracker.swift
@@ -1,3 +1,7 @@
+#if !COCOAPODS
+import ApolloAPI
+#endif
+
 final class GraphQLDependencyTracker: GraphQLResultAccumulator {
   private var dependentKeys: Set<CacheKey> = Set()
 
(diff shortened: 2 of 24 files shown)


Apollo.xcodeproj/project.pbxproj
```diff
@@ -4496,6 +4496,7 @@
 /* Begin PBXShellScriptBuildPhase section */
 		90690D322243442F00FC2E54 /* Ensure no build settings are in the Xcode project */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
```
calvincestari, line 4499: Any further details on what this new attribute is for?

AnthonyMDev: It indicates that the build script running is NOT dependent on any input or output sources.

@calvincestari I know you know this part, but for the sake of a complete answer: If you provide any of these sources it will only run them when the input sources have changed or the output sources do not exist.

In past versions of Xcode, if you don't provide any input or output sources, it will just always run the script. In Xcode 14, not providing input sources will show a warning. I don't want to go find the actual warning message right now, but its something along the lines of "This script will run on every build of your target. Please provide input/output sources. If you want this script to run every time, check the checkbox for 'not dependent on any sources'"

Checking that checkbox on the script build phase will make `alwaysOutOfDate = 1;`.

AnthonyMDev: Maybe we want to setup an input source for that build phase to only run when the xcproject file changes actually though!

calvincestari: > In past versions of Xcode, if you don't provide any input or output sources, it will just always run the script. In Xcode 14, not providing input sources will show a warning.

This is the key bit I didn't know about Xcode 14 - thanks.

calvincestari: > Maybe we want to setup an input source for that build phase to only run when the xcproject file changes actually though!

It seems you need to specify both input _and_ output files for that setting to not generate any warnings. Apple's [documentation](https://developer.apple.com/documentation/Xcode/improving-the-speed-of-incremental-builds) says to generate a static file in the case of no output.

The execution time of that custom script is 0.013s? We _can_ generate a static file as 'output' but I think the implementation would take way longer than the eventual time you'd end up saving.

Screen Shot 2022-08-15 at 3 10 35 PM

PR #2443, opened 2022-08-16
Fix reference to fragment with lowercase name

### LINKED ISSUE #2441: Apollo 1.0 Beta: fragments that are not capitalized seems to cause issues
## Bug report

If you have a fragment like this:
```
fragment myThing on Thing {
   stuff
}
```

Apollo codegen seems to be confused as to what to call it and reference it.

It is generated with a type `MyThing`, but it seems to be referenced as `myThing.self`.

AnthonyMDev:
Closes #2441


calvincestari (approved):
👍🏻

Issue #2129, opened 2022-02-02
[WebSocket] updateConnectingPayload Websocket is not properly disconnected before reconnection 


CureleaAndrei:
**Context** 
I have a GraphQL server using NestJS that checks the subscriptions connectionPayload for JWT and returns the data if JWT is valid or "Forbidden" if the JWT is null or invalid.

I've been dealing with the following issue: after initialising a WebSocket connection with payload that contains a nil JWT I have the correct server response that is "Forbidden", after I update the connection payload with a valid JWT I get the same response "Forbidden" which is wrong, even if the server get's the valid JWT. I've tested this on a new project too and the problem persists.

After doing some testing with some delays like this : 
- pause web socket
- delay
- update connection payload
- resume connection

it started working every time, but I've been using a sleep on the thread which I don't like. This led me to think that the websocket doesn't disconnects from the server in time for the reconnection with the new JWT.
 
So I've done some digging and I've saw that in the reconnectWebSocket() from updateConnectionPayload() that connection happens without a callback validating the disconnection. Going deeper in the WebSocket implementation I've found that the disconnect method uses an other disconnect method that has a forceTimeout parameter that is not used and the websocket waits for a response from the server when updating payload. I've tested with a forceTimeout set to 0 and it started working every time, closing the websocket immediately and reconnecting it. Unfortunately the forceTimeout parameter cannot be set from outside. 

1. Is this the intended behavior or it is a bug? or did I do something wrong ? 
2. Is there a way I can set from outside the forceTimeout parameter ?
3. Is there an other recommended way of updating the connectionPayload while the websocket is connected (and without losing the existing subscriptions) other than using the "updateConnectingPayload" method ?
4. What is the recommended way of handling this case?

Notes: 

- I initialize the WebSocketTransport with the default WebSocket implementation with connection on init and a ["Authorization":nil] connection payload. 
- Sometimes it works, but most of time it doesn't.
- the JWT comes from Firebase Auth and the servers checks also in Firebase Auth

- `apollo-ios` SDK version: 0.50.0
- Xcode version: 13.2.1
- Swift version: 5.5.2

Issue #1918, opened 2021-08-24
Unable to resume websocket connection after pausing/closing


mitchellporter:
## Bug report

Unable to resume websocket connection and receive events after pausing/closing the connection.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 0.47.1
- Xcode version: 12.5.1
- Swift version: 5
- Package manager: Cocoapods

## Steps to reproduce

1. Create/setup apollo clientt
2. Resume websocket connection and begin receiving events
3. Pause/close websocket connection to stop receiving events
4. Resume websocket connection again to begin receiving events again
5. Unable to receive websocket events

## Further details

We receive websocket events on a single screen in our app. When we create/setup our apollo client, we pass in a value of `false` for the `connectOnInit` param for the `WebSocketTransport` because we want to have control over pausing/resuming websocket events and only receive them when the user is viewing this specific screen.

When the user visits this screen, we check if the websocket is connected and if it's not, then we call `resumeWebSocketConnection` to begin receiving events. This works fine the first time the user visits this screen.

When the user leaves the screen, the view controller's `deinit` function is called and we call `pauseWebSocketConnection` on the websocket transport to stop receiving events (have also tested `closeConnection` with similar results).

We receive the following error as expected, just as described in the apollo docs, when we pause the connection:

`"websocket is disconnected: WSError(type: Apollo.WebSocket.WSError.ErrorType.protocolError, message: \"\", code: 1000)"
`

When the user revisits the screen, we check the connection again and if it's not connected (it shouldn't be), then we call `resumeWebSocketConnection` on the websocket transport to begin receiving events again. However this time, we receive the following error:

`Error with subscription result: Websocket network error. Error: Optional(Apollo.WebSocket.WSError(type: Apollo.WebSocket.WSError.ErrorType.protocolError, message: "", code: 1000))`

From this point on we are unable to receive websocket events.

Here is our apollo client setup/creation code:

```
// Setup HTTP transport
        let client = URLSessionClient()
        let provider = NetworkInterceptorProvider(client: client, store: store)
        self.httpTransport = RequestChainNetworkTransport(interceptorProvider: provider, endpointURL: httpsUrl)
        
        // Setup websocket transport
        let request = URLRequest(url: wssUrl)
        let webSocket = WebSocket(request: request)
        self.webSocketTransport = WebSocketTransport(websocket: webSocket, connectOnInit: false)
[...]

PR #2449, opened 2022-08-18
Fix issues 1918 and 2129 by websocket disconnect forceTimeout

STomperi:
This fixes problem mentioned in issue, https://github.com/apollographql/apollo-ios/issues/1918
I also have been able to reproduce this by calling pauseWebSocketConnection() when app is going to background.
Like mentioned in comment https://github.com/apollographql/apollo-ios/issues/1918#issuecomment-1012505222

Issue https://github.com/apollographql/apollo-ios/issues/2129 is fixed by calling disconnect with timeout of 0:
    
`self.websocket.disconnect(forceTimeout: 0)`

Timeout of 0 is required since 
`self.websocket.connect()`
Is called on the immediately following code line and this is unreliable if connection has not closed.

I don't really know how to write unit test for these issues.

### DIFF
--- a/Sources/ApolloTestSupport/MockWebSocket.swift
+++ b/Sources/ApolloTestSupport/MockWebSocket.swift
@@ -29,7 +29,7 @@ public class MockWebSocket: WebSocketClient {
   open func write(ping: Data, completion: (() -> ())?) {
   }
 
-  public func disconnect() {
+  public func disconnect(forceTimeout: TimeInterval?) {
   }
   
   public func connect() {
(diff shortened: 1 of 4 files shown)


calvincestari:
Thanks for the contribution @STomperi, these changes all seem reasonable. I think there are some improvements we can make to the documentation for `WebSocket.disconnect(forceTimeout:closecode:)` and I'll take a look at tests for this too.

calvincestari (approved):
I've updated the method documentation to be correct for the behaviour.

I looked into tests and found that we have one at [`StarWarsSubscriptionTests. testPausingAndResumingWebSocketConnection`](https://github.com/apollographql/apollo-ios/blob/main/Tests/ApolloServerIntegrationTests/StarWarsSubscriptionTests.swift#L438) for this behaviour. Additional tests that could be worthwhile are in the default implementation to ensure that the timeout values are being adhered to but that would mean making private methods internal. v2 will be an extensive rewrite of the networking layer at which point this code will likely be rewritten.

PR #2453, opened 2022-08-22
Re-locate CodegenCLI module and tests

calvincestari:
This PR removes the separate package for CodegenCLI and adds it to the main apollo-ios package as well as Xcode workflows. The reason for the separate CodegenCLI package was because we were originally going to distribute it via Homebrew and build it locally at install time..we're not going to be using Homebrew anymore.

* The apollo-ios package defines an executable target and this is how we build the binary, via a make file.
* Xcode only has the CodegenCLI framework and test targets; I didn't see the need for the executable target in Xcode.
* I've bumped the swift-tools-version to 5.6 because [5.6 is included with Xcode 13.3](https://www.swift.org/blog/swift-5.6-released/) and [as of April 2022 Xcode 13 is the minimum version you can use to submit to the App Store](https://developer.apple.com/ios/submit/).
* I renamed the CI jobs for better clarity on what they're tasked to do.
* There are two tests being skipped in InitializeTests because they link to the generated binary and since Xcode doesn't have an executable target for apollo-ios-cli they can't execute. I have some ideas but haven't figured out to resolve this yet.

Part of #2381.


AnthonyMDev (approved):

AnthonyMDev:
Thanks! I'm going to take a look at those two tests and see if I can come up with anything

AnthonyMDev:
I think rather than actually creating a process and calling the CLI with arguments, we should just be initializing the command like in the other test and calling `run()`. The best thing to do is probably just give the command an `output` property that we feed the output to. I can make up a quick PR for this.

AnthonyMDev:
Actually probably won't get to this today. Will do tomorrow.

calvincestari:
The skipped tests are re-enabled in https://github.com/apollographql/apollo-ios/pull/2455.

PR #2454, opened 2022-08-22
Generate CacheKeyResolution extension in Schema

AnthonyMDev:
This generates the `Schema+CacheKeyResolution.swift` file that you can edit to implement your cache key logic. This ensures that this function is declared properly and is in the right location at all times. We need to make sure to document this entry point in the usage guide.


AnthonyMDev:
Still got some work to do on this tomorrow actually. Need to fix some tests and think through some minor edge cases of cache keys for unknown object types. Will have PR finished tomorrow.

AnthonyMDev:
Okay this is ready for review now!

SwiftScripts/.swiftpm/xcode/xcshareddata/xcschemes/SwiftScripts-Package.xcscheme
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
calvincestari, line 1: Does this file need to be committed?

Sources/ApolloAPI/SchemaConfiguration.swift
```diff
@@ -59,7 +52,8 @@ extension SchemaConfiguration {
     guard let typename = object["__typename"] as? String else {
       return nil
     }
-    return objectType(forTypename: typename)
+    return objectType(forTypename: typename) ??
+    Object(typename: typename, implementedInterfaces: [])
```
calvincestari, line 56: Is this to handle unknown types?

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -79,6 +80,10 @@ extension TemplateRenderer {
 
   private func renderSchemaFile(_ type: TemplateTarget.SchemaFileType) -> String {
     let namespace: String? = {
+      if case .cacheKeyResolutionExtension = type {
```
calvincestari, line 83: Do none of the combinations below work for this file? At first look it seems like `case (true, nil):` should return the right string.

calvincestari (commented):
Had some questions in review. It's good to see how extensible the file template/generator pattern is; easy to add new ones as we need to.

Sources/ApolloAPI/SchemaConfiguration.swift
AnthonyMDev: Yes! We aren't handling them as far as we'd like to in the future (where we actually request the implemented interfaces), but for now this just gives parity to 0.x, because you should be able to store them in the normalized cache if you have enough information to normalize them.

This doesn't handle casting to other type scopes because we don't know about interfaces and unions.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
AnthonyMDev: By not having a namespace name, we don't call `wrappedinNamespace`. This needs an extension on `MySchemaAPI.Schema`. You can't do an extension on `Schema` inside of an extension on `MySchemaAPI`.

Though we could actually do the `case let (true, .some(schemaTypeNamespace)):` if we provided `Schema` as the `namespaceComponent`.

But then the template is just the static function, and it's **always** wrapped in the namespace. That's not really the intended use of this `TemplateRenderer`. It's to determine the wrapping context based on the config options.

SwiftScripts/.swiftpm/xcode/xcshareddata/xcschemes/SwiftScripts-Package.xcscheme
AnthonyMDev: It's going to be added everytime we open the project in Xcode 14. Doesn't matter. Probably easier to keep it than to `.gitignore` it.

PR #2455, opened 2022-08-23
tests: Fix CodegenCLI skipped tests

calvincestari:
Provides an optional closure to receive output from the `Initialize` command; if not provided then the output is written to `Swift.print`.

Related of #2381.


AnthonyMDev (approved):
LGTM thanks!

PR #2457, opened 2022-08-23
SchemaMetadata and SchemaConfiguration refactor

AnthonyMDev:


Sources/ApolloAPI/SchemaConfiguration.swift
```diff
@@ -1,38 +1,15 @@
...
-  /// returns `nil`.
-  static func objectType(forTypename typename: String) -> Object?
-}
-
 public protocol SchemaConfiguration {
   /// The entry point for configuring the cache key resolution
   /// for the types in the schema, which is used by `NormalizedCache` mechanisms.
   ///
-  /// The default implementation always returns `nil`, disabling all cache normalization.
+  /// The default generated implementation always returns `nil`, disabling all cache normalization.
```
calvincestari, line 9: I wonder if it's worth mentioning here that disabling normalization does not mean no caching. Something simple like `...disabling all cache normalization. Responses will be cached by query path instead.`

Sources/ApolloCodegenLib/Templates/SchemaConfigurationTemplate.swift
```diff
@@ -1,27 +1,29 @@
...
-struct SchemaCacheKeyResolutionTemplate: TemplateRenderer {
+struct SchemaConfigurationTemplate: TemplateRenderer {
 
   /// Source IR schema.
   let schema: IR.Schema
   /// Shared codegen configuration
   let config: ApolloCodegen.ConfigurationContext
 
-  let target: TemplateTarget = .schemaFile(type: .cacheKeyResolutionExtension)
+  let target: TemplateTarget = .schemaFile(type: .schemaConfiguration)
```
calvincestari, line 11: This and the renaming of the `.schema` case too are much better names, I'm glad we did this change.

calvincestari:
Nice work on the rest of the PR! 👍🏻

PR #2462, opened 2022-08-24
Fix casing of type names for schema types with lowercase names

### LINKED ISSUE #2460: Incorrect handling of types that start lower-case
## Bug report

**This is v1.0.0 beta issue**
If there is type in schema that starts with lower-case letter, then type file will generate making it uppercase, but the operations will try to use it with lower case

Screenshot 2022-08-24 at 21 19 53
Screenshot 2022-08-24 at 21 23 40

## Versions

- `apollo-ios` SDK version: 1.0.0-beta.2
- Xcode version: 13.4.1
- Swift version: 5.6
- Package manager: SPM

## Steps to reproduce

Generate code with schema that contains lower-case starting types

AnthonyMDev:
Fixed #2460

Also handles the same issue for input objects, enums, and custom scalars with lowercase names


calvincestari (approved):
Thanks for fixing these - LGTM!

PR #2463, opened 2022-08-25
Fixed String jsonValue initialiser for large numbers

Almaz5200:
Large numbers is interpreted as Int64 instead of just Int (in my case is was timestamps). In this case parsing to string would fail.

### DIFF
--- a/Sources/ApolloAPI/JSONStandardTypeConversions.swift
+++ b/Sources/ApolloAPI/JSONStandardTypeConversions.swift
@@ -15,6 +15,8 @@ extension String: JSONDecodable, JSONEncodable {
         self = string
     case let int as Int:
       self = String(int)
+    case let int64 as Int64:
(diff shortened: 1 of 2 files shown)


AnthonyMDev:
Interesting... It's so odd that its doing that, considering `Int` defaults to 64 bit in Swift anyways. Thanks for catching this and the PR! We'll need to get a unit test case to ensure this doesn't regress in the future before we can pull it in. If you don't have time to write the test case, I'll start a new PR off your branch with the test added!

Almaz5200:
I've added one test case, but I feel like It's not enough because I've hardcoded Int64 type in there and it does not cover any real parsing (so It won't catch it if all of a sudden swift will start interpret something as Int32 or something like this). But I don't think I understand Apollo enough to figure out where do I put test for that, so you can add that if you think that is needed too.

PR #2465, opened 2022-08-25
Fixed casing for lowercase starting types

Almaz5200:
Strings here should actually remain lowercase for those types. That is because this strings are used to compare them against __typename field when parsing, currently they would not parse


AnthonyMDev:
Thank you for this fix! It needs updates to the tests to pass CI, but I can make those fixes and push a new PR this afternoon. If you want to make the test updates so we can get this PR merged in, I'd be happy to merge it. It's tests for the `ApolloCodegenLib`

Almaz5200:
Thanks for the tip! I've fixed tests, should be all right now :)

AnthonyMDev (approved):
Thanks for this! Looks great

PR #2467, opened 2022-08-26
feat: Add `--verbose` parameter to CLI

calvincestari:
Part of #2381 

Adds a command line parameter to change the level of verbosity in the CLI output for `fetch-schema` and `generate`; the default remains `.warning` but the user can change the level to include debug output (`.debug`)
```
-v, --verbose            Increase verbosity to include debug output
```

### DIFF
--- a/Sources/CodegenCLI/Protocols/LogLevelSetter.swift
+++ b/Sources/CodegenCLI/Protocols/LogLevelSetter.swift
@@ -1,0 +1,18 @@
+import ApolloCodegenLib
+
+public protocol LogLevelSetter {
+  static func SetLoggingLevel(verbose: Bool)
+  static func SetLoggingLevel(_ level: CodegenLogger.LogLevel)
+}
+
+extension LogLevelSetter {
+  public static func SetLoggingLevel(verbose: Bool) {
+    SetLoggingLevel(verbose ? .debug : .warning)
+  }
+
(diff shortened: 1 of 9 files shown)


Sources/CodegenCLI/Protocols/LogLevelSetter.swift
```diff
@@ -0,0 +1,18 @@
+import ApolloCodegenLib
+
+public protocol LogLevelSetter {
```
calvincestari, line 3: This exists as a way to stub and mock the global state of `CodegenLogger` vs. implementing a protocol for the functionality of the logger itself.

Tests/CodegenCLITests/Support/MockLogLevelSetter.swift
```diff
@@ -0,0 +1,21 @@
...
+import CodegenCLI
+import ApolloCodegenLib
+
+extension LogLevelSetter {
+  static var mock: LogLevelSetter.Type {
+    MockLogLevelSetter.self
+  }
+}
+
+struct MockLogLevelSetter: LogLevelSetter {
```
calvincestari, line 11: There is intentionally no stub for `SetLoggingLevel(Bool)` which relies on the concrete extension of `LogLevelSetter` to function correctly.

AnthonyMDev (commented):
LGTM

PR #2468, opened 2022-08-26
feat: Trap failed termination status and print permissions message

calvincestari:
Trap failed exits and print a helpful message to the user about possibly needing permission enabling-flags when creating files.


calvincestari:
@AnthonyMDev - this isn't the solution we'd briefly spoken about; using custom error codes to show the message only in the case of failed file creation but I feel this simple change delivers what's needed to the user.

I have another [branch](https://github.com/apollographql/apollo-ios/compare/1.0/custom-exit-codes...1.0/sysexit-exit-codes) that implements the custom exit codes but I do not want to merge it. I really dislike the way it makes the CLI the source of the error by catching it, converting, and then throwing the sysexit exit code. I think it's going to make our lives very difficult if we need to debug some obscure message deep within ApolloCodegenLib.

Plugins/SharedPackagePluginExtensions/Process+TerminationHandler.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension Process {
+  static func HandleErrorTermination(_ process: Process) {
+    if process.terminationStatus != 0 {
+      print("""
+      Hint: You may need to use the --disable-sandbox or --allow-writing-to-package-directory flag
```
AnthonyMDev, line 7: It wouldn't be `--allow-write-to-package-directory` it would be `-allow-writing-to-directory ${OUTPUT_DIRECTORY}`

AnthonyMDev (approved):
Other than changing up that wording in the warning I'm good with this approach.

AnthonyMDev:
Me might also want to say "If you are using `--disable-sandbox` and code generation still fails, please make a bug report at
https://github.com/apollographql/apollo-ios".

AnthonyMDev:
Fixed merge conflicts and updated the message here on https://github.com/apollographql/apollo-ios/pull/2473. Merging that in instead.

PR #2470, opened 2022-08-29
docs: change command to run-codegen.sh

sinkcup:
there is no `run-bundled-codegen.sh` but `run-codegen.sh` in v1:

https://github.com/apollographql/apollo-ios/blob/1.0.0-beta.2/scripts/run-codegen.sh

### DIFF
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -132,7 +132,7 @@ If you're using a tool like Interface Builder or SwiftUI to talk to a module wit
 For example, if you're using something like this to run your code generation for a target called `YourTarget`:
 
 ```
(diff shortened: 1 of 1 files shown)


AnthonyMDev:
For v1 we haven't updated the documentation just yet on this, but that script is actually not for running the codegen engine on your code. That just generates the code for the examples we have in the repo for our internal testing purposes. Thank you for the PR, but we'll get the correct documentation for this up in the coming weeks.

The Beta 3 version is coming out today, which makes improvements to the new `apollo-ios-cli`, including allowing you to use it via an SPM plugin. The codegen cli is the new recommended way to run codegen. There is some starting usages docs on the cli here. https://github.com/apollographql/apollo-ios/tree/release/1.0/Sources/apollo-ios-cli

PR #2472, opened 2022-08-29
Update SQLiteSerialization to be an enum

Iron-Ham:
The most minor & nittiest of changes, but a `class` with only static properties should be an `enum` so as to prevent un-intended instantiation .


AnthonyMDev:
Yes it absolutely should! Thanks!

Issue #2437, opened 2022-08-15
Retain Cycle in 1.0.0 Beta GraphQLExcecutor <-> ReadTransaction


lorraine-hatch:
## Bug report

After sending a few queries, there are a lot of retain cycles between `GraphQLExcecutor` and `ReadTransaction`. See screen shots below. 👇 

Help would be appreciated! Thank you 🙏

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: beta 1.0.0
- Xcode version: 13.4.1
- Swift version: 5
- Package manager: SPM

## Steps to reproduce

- Send some queries through the Apollo client (I only made about 5-6 queries)
- Go to memory debugger
- Find memory warnings by `GraphQLExcecutor` and `ReadTransaction`. Is there supposed to be 28 instances of each? 

## Further details

<img width="200" alt="CleanShot 2022-08-15 at 14 21 51@2x" src="https://user-images.githubusercontent.com/105568068/184711949-bb676449-2544-4186-8c77-463fb189a36f.png">
<img width="200" alt="CleanShot 2022-08-15 at 14 22 21@2x" src="https://user-images.githubusercontent.com/105568068/184711966-25194909-aff3-4381-89d8-d954650ed3a9.png">
<img width="200" alt="CleanShot 2022-08-15 at 14 22 00@2x" src="https://user-images.githubusercontent.com/105568068/184711976-b86a292f-986e-44de-8476-c83218298ecd.png">

calvincestari:
Thanks for the bug report @lorraine-hatch, we'll take a look into this asap.

lorraine-hatch:
Thanks!

lorraine-hatch:
@calvincestari I think I've found the solution to this, can I make a pull request?

calvincestari:
@lorraine-hatch yes certainly! We'll gladly review your PR, thanks.

lorraine-hatch:
@calvincestari Here ya go! https://github.com/apollographql/apollo-ios/pull/2475

AnthonyMDev:
Thank you! Fixed in #2476. More details at this comment: https://github.com/apollographql/apollo-ios/pull/2475#issuecomment-1232157290

calvincestari:
Thanks for the contribution @lorraine-hatch, glad you managed to track that down. I was away last week, thanks @AnthonyMDev for reviewing and merging.

PR #2475, opened 2022-08-30
Fix retain cycle

lorraine-hatch:
This is a fix that I implemented to address this issue that I opened: https://github.com/apollographql/apollo-ios/issues/2437

### DIFF
--- a/Sources/Apollo/ApolloStore.swift
+++ b/Sources/Apollo/ApolloStore.swift
@@ -180,11 +180,6 @@ public class ApolloStore {
     fileprivate let cache: NormalizedCache
 
     fileprivate lazy var loader: DataLoader<CacheKey, Record> = DataLoader(self.cache.loadRecords)
-    fileprivate lazy var executor = GraphQLExecutor { object, info in
-        return object[info.cacheKeyForField]
-      } resolveReference: { reference in
-        self.loadObject(forKey: reference.key)
-      }
 
     fileprivate init(store: ApolloStore) {
       self.cache = store.cache
@@ -218,6 +213,12 @@ public class ApolloStore {
       accumulator: Accumulator
     ) throws -> Accumulator.FinalResult {
       let object = try loadObject(forKey: key).get()
+      
+      let executor = GraphQLExecutor { object, info in
+        return object[info.cacheKeyForField]
+      } resolveReference: { reference in
+        self.loadObject(forKey: reference.key)
+      }
 
       return try executor.execute(
         selectionSet: type,


Sources/Apollo/ApolloStore.swift
```diff
@@ -218,6 +213,12 @@ public class ApolloStore {
       accumulator: Accumulator
     ) throws -> Accumulator.FinalResult {
       let object = try loadObject(forKey: key).get()
+      
+      let executor = GraphQLExecutor { object, info in
+        return object[info.cacheKeyForField]
+      } resolveReference: { reference in
+        self.loadObject(forKey: reference.key)
+      }
```
lorraine-hatch, line 222: This is how it was in your main branch. That's where I got the idea to try reverting it. Cleans up the retain cycles 🙌

AnthonyMDev:
Thanks so much for looking into this! Your PR definitely identified where the retain cycle was, which was incredibly helpful and saved us a ton of time. I appreciate it!

In https://github.com/apollographql/apollo-ios/pull/2476, I've gone ahead with an approach to this that uses a `[weak self]` capture list because it should be more performant (keeping a single `executor` instead of creating a new one for each read in the transaction). I also added a unit test to make sure that this fixes the memory leak and we don't let it regress and come back in the future. I'll be merging that PR in shortly.

Once it's merged, if you could point your codebase at the `release/1.0` branch and verify that the fix does indeed remove the retain cycle you are experiencing, that would be great!

lorraine-hatch:
@AnthonyMDev great! I will try it out later today and let you know.

lorraine-hatch:
@AnthonyMDev I've pulled in the `release/1.0` branch, but I'm running into build errors in the generated code: 

CleanShot 2022-08-31 at 10 45 40@2x

Any ideas what's happening here?

AnthonyMDev:
The API for the generated code has changed a bit. Please re-run codegen! You'll also probably need to remove the `Schema.swift` and Schema+CacheKeyResolution.swift` files from your project.

More info about this can be found in the [Beta 3 release notes](https://github.com/apollographql/apollo-ios/releases/tag/1.0.0-beta.3)!

lorraine-hatch:
oh awesome, thank you!

lorraine-hatch:
Ok it worked 🥳, no more retain cycle!

lorraine-hatch:
Does the latest beta release include this?

AnthonyMDev:
No, we’ll get this into beta 4!

lorraine-hatch:
Sounds good! Is there an anticipated release date?

calvincestari:
I think we can tentatively schedule a beta 4 release for the end of the week; I'll confirm with @AnthonyMDev later today. Here's the current [diff](https://github.com/apollographql/apollo-ios/compare/1.0.0-beta.3...release/1.0) of what would be included with a beta 4 release.

lorraine-hatch:
Sounds great, thanks @calvincestari!

PR #2477, opened 2022-09-01
Escape swift reserved words for generated enum case name

bannzai:
This pull request to `release/1.0` branch.

## Abstract
I use apollo-ios codegen version of v1-beta3 and got compile error when genarated enum case name samely swift reserved words. This PR escape swift reserved word for enum case name

## Test
I confirmed to behavior with executing codegen script in my local. And Run UnitTests are passed.


AnthonyMDev:
Thanks for this one! Merged into release branch!

PR #2478, opened 2022-09-01
Add codegen option about convert enum case name strategy

bannzai:
This pull request to `release/1.0` branch. And related  and depend on this PR: https://github.com/apollographql/apollo-ios/pull/2477

## 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 
See #2477 before this PR.

## Abstract
I use apollo-ios codegen version of v1-beta3. My service provide UPPER_CASE enum schema. It will genarate to `case UPPER_CASE`, but It was no good syntax in swift.
Therefore, I create about proposal PR of add to codegen option about enum case convert strategy. I would find a good form through this PR for generating enum case name.

## Test
I confirmed to behavior with executing codegen script in my local. And Run UnitTests are passed.


AnthonyMDev:
Thanks so much for the PRs! I like this approach a lot and we've actually discussed doing something similar. I think we may want to consider utilizing this case conversion strategy option for other things like fields and type names, not just enum cases.

We'll have a conversation about this next week and decide on a direction here.

AnthonyMDev:
We've decided that we are going to go ahead with this approach with a few minor changes! Thanks for making the PR @bannzai.

We want to change the naming from `ConvertStrategy` to `ConversionStrategies`.

Additionally, to support future use cases where we may add conversion strategies for other things, we want to refactor the configuration option to a struct that has a property for `enumCases`.

```swift
  public enum CaseConversionStrategy: String, Codable, Equatable {
    /// Default. Nothing different from the definition of a schema.
    case none
    /// Convert to lower camel case from `snake_case`, `UpperCamelCase`,`UPPERCASE`.
    case camelCase
  }

  public struct ConversionStrategies: Codable, Equatable {
    public let enumCases: CaseConversionStrategy = .camelCase
  }
```

I'm leaning towards defaulting to `.camelCase` instead of `.none`, here. @calvincestari what do you think?

@bannzai Do you want to make these changes to the PR so we can merge it in, or would you like me to do so and make a new PR?

bannzai:
@AnthonyMDev I will try it!

calvincestari:
> I'm leaning towards defaulting to .camelCase instead of .none, here. @calvincestari what do you think?

`.camelCase` seems the most likely output that people would want in Swift code, I'm happy with that as the default.

bannzai:
I'm done to replace new strctures and testing it. Ready for review！

AnthonyMDev (approved):
Thanks so much! This looks great!
Going to merge this in and then clean up the documentation a little bit.

PR #2480, opened 2022-09-07
Proposal PR: Add codegen option of EmbedSwiftOptionalInitializer

bannzai:
This pull request to release/1.0 branch. 
And this PR for proposal feature for codegen. We would like to discuss better options and changes.

## Abstract
I use apollo-ios codegen version of v1-beta3 and faced GraphQLNullable. GraphQLNullable is a good way for following GraphQL specification. 
`.none`,`.null` are helping to clearly separate meaning in the Swift code. However, my service does not require a clear separation between `.none` and `.null`. This means that I want the old way of writing Apollo using the standard Swift Optional. And I guess that many developer want that as an option.

## Problem
Call mutation named UpdateUserProfile with UpdateUserProfileInput. UpdateUserProfileInput has nullable field named `displayName: String`. That means to use GraphQLNulable<String> type on Swift code. Schema is like this.

```graphql
mutation UpdateUserProfile($input: UpdateUserProfileInput!) {
  ...
}

input UpdateUserProfileInput {
  displayName: String
}
```

Call this mutation in Swift code to fill displayName field with optional variable. I want to set `displayName` variable as it is.  But I got compile error. So, it is necessary to instantiate to use GraphQLNullable or `??` operations. 

```swift
var displayName: String?
func performUpdateUserProfile() {
  apolloClient.perform(Mutation(input: .init(displayName: displayName)) // Compile error
}
```

However, my service does not always have to choose between using `.none` or `null`, and I prefer to use `.null`. Also, when I use the `nil` literal, we would rather use `.null`.

## How
My approach is like this.

- Add type of ApolloCodegenConfiguration.EmbedSwiftOptionalInitializer, and defined as option to ApolloCodegenConfiguration.
  - `EmbedSwiftOptionalInitializer` enable to set  `alwaysFillNull` and `alwaysUndefinedField`.
- If developer set EmbedSwiftOptionalInitializer, generating convenience initializer interface using Swift Optional instead of GraphQLNullable. And fill argument with value for selected option of EmbedSwiftOptionalInitializer in convenience initializer in delegate to required init part. 
  - Type inference problems when a field is set to `nil` literal are solved by adding the `@_disfavoredOverload` compiler attribute to required initializer.
  
## Test
- Add test case to OperationDefinitionTemplateTests using EmbedSwiftOptionalInitializer.
- I exec codegen manually, I got expected generated code.
- I run my app and I confirm to pass to new feature convenience initializer.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -314,7 +314,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     public let enumCaseConvertStrategy: EnumCaseConvertStrategy
 
     /// Option to embed convenience initializer about using `GraphQLNullable`.
-    public let embedNullableVariableConvenienceInitializer: EmbedNullableVariableConvenienceInitializer
+    public let embedSwiftOptionalInitializer: EmbedSwiftOptionalInitializer
 
 >>>>>>> 214fb6f6 (Add option)
     /// Designated initializer.
@@ -337,8 +337,14 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///  - enumCaseConvertStrategy: How convert strategy to Swift enum from GraphQL schema.
 =======
     ///  - enumCaseConvertStrategy: How convert strategy to Swift enum from GraphQL schema.
+<<<<<<< HEAD
     ///  - embedNullableVariableConvenienceInitializer: Embed convenience initializer about `GraphQLNullable`
 >>>>>>> 214fb6f6 (Add option)
+||||||| parent of 50f4f75d (Replace to embedSwiftOptionalInitializer)
+    ///  - embedNullableVariableConvenienceInitializer: Embed convenience initializer about `GraphQLNullable`
+=======
+    ///  - embedSwiftOptionalInitializer: Embed convenience initializer about `GraphQLNullable`
+>>>>>>> 50f4f75d (Replace to embedSwiftOptionalInitializer)
     public init(
       additionalInflectionRules: [InflectionRule] = [],
       queryStringLiteralFormat: QueryStringLiteralFormat = .multiline,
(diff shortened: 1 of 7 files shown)


bannzai:
I force pushed it because it contained extra changes. I would appreciate it if you could review it. Thank you.

calvincestari:
Thanks for the contribution @bannzai. @AnthonyMDev and I will discuss but I doubt we want to support both ways of doing nullable. For both maintainability and clarity of function I'd rather just pick one and move forward with it.

AnthonyMDev:
Thank you for all the contributions @bannzai! @calvincestari and I had a _long_ conversation about this one today. We've decided not to merge this one in though.

We very much understand the reasoning behind doing this, but we are hesitant to more and more configuration options here. There are a lot of very custom and specific things that users may want to do and we don't want to end up on the hook for maintaining support for all of them in the future.

We see why this makes things easier for you in the situation where your server does not distinguish between `null` and undefined (`nil`/`none`). But we don't recommend this approach for a few reasons. 

1. If you can use `null` or `nil`, we always recommend `nil`, as it results in less data sent via network transport. Including an option to default to `null` is counter to best practices.

2. Providing these options creates a situation in which your code is actually harder to read and understand. For a team of 1 developer, this might be okay, but working with a larger team, it is unclear what happens when you pass a `nil` optional value. And that behavior depends upon a configuration option that is far removed from the code you are looking at. Configuration options that change the generated code in ways that does not affect the usage of them at the call site, but only the behavior concealed in implementation details is worrisome to us. This isn't a configuration of the naming, shape, or location of your generated models, but rather of the behavior of the Apollo iOS library in a way that is not immediately evident to consumers of the generated models.

3. The current solution to this is to use a `??` nil coalescing operator at your call sites. In your example above it would look like:
```swift
var displayName: String?
func performUpdateUserProfile() {
  apolloClient.perform(Mutation(input: .init(displayName: displayName ?? .null))
}
```
This requires you to make the decision at the call site. In a situation where you _always_ want to default to a specific value, that may be slightly more code at each call site. But it forces developers to make that decision directly each time, which we actually see as a positive. It also makes the behavior completely clear while reading the code. You don't have to go look in another place to understand why your field is getting a `.null` value when you pass `nil` to it.

4. This becomes even more problematic if you consider the situation where someone turns this configuration on in an already existing codebase. For example, if your application had code that used a `nil` Literal like this:
```swift
let mutation = Mutation(input: .init(displayName: nil))
```
 Once someone turns on that configuration option and re-runs code generation, this call is now pointing to a _different_ initializer than it was before (your convenience one). In the past, it was explicitly setting the field to `.none` but now, it could be setting it to `.null`, and you would never be alerted until it someday possibly showed up as a bug (or you had very extensive unit tests).

For these reasons, we don't feel like this is something that we want to formally support in the Apollo iOS Client library. While we recognize the convenience that it provides to you, and perhaps others, we think that it's better left out. 

If you (or anyone else) want to maintain this code templating in your own fork, please feel free to do so. In the future, we hope to extend the templating engine to allow you to more easily create customizations to the generated code like this without having to maintain a fork of the repo. But we want those things to be clearly customization that you are making and not something that we are committed to supporting.

---

That said, we do really appreciate your contributions to the library. We hope that you will continue to submit PRs for bug fixes and additional functionality that you believe provide value to the library. Thank you!

bannzai:
@AnthonyMDev Thank you for your response . I understood some reasons about my changes and agree all for each reasons.

And it is glad to be able to confirm that they are considering the functions that I expected in the future.

>  In the future, we hope to extend the templating engine to allow you to more easily create customizations to the generated code like this without having to maintain a fork of the repo. 

Thank you

PR #2486, opened 2022-09-09
Autodelete generated files

AnthonyMDev:
- Create globs for each of the search paths the generated files would be generated into based on config
- Match those globs for all generated files
- Keep track of file paths that were written during codegen execution
- Delete all existing files that were not re-written during this codegen execution process


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -266,6 +304,67 @@ public class ApolloCodegen {
...
+
+    case let .absolute(operationsPath):
+      globs.append(Glob(
+        ["\(operationsPath)/**/*.graphql.swift"],
+        relativeTo: config.rootURL
+      ))
+
+    case let .relative(subpath):
+      let searchPaths = config.input.operationSearchPaths.map { searchPath -> String in
+        let startOfLastPathComponent = searchPath.lastIndex(of: "/") ?? searchPath.firstIndex(of: ".")!
```
calvincestari, line 329: I'm assuming we trust in using `!` here because of the new validation on input search paths?

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -1096,15 +1864,56 @@ class ApolloCodegenTests: XCTestCase {
   func test_validation_givenTestMockConfiguration_asSwiftPackage_withSchemaTypesModule_asSwiftPackage_shouldNotThrow() throws {
     // given
     let configContext = ApolloCodegen.ConfigurationContext(config: .mock(
-      input: .init(schemaPath: "path"),
-      output: .mock(
-        moduleType: .swiftPackageManager,
-        testMocks: .swiftPackage(targetName: nil)
-      )
+      input: .init(schemaPath: "path.graphqls")
```
calvincestari, line 1867: Should this initializer test include `operationSearchPaths` too so we know it's valid with an extension? The new validation tests only test for throwing on `operationSearchPaths` without an extension.

calvincestari (approved):
Approved - had a question about naming of some function parameters and a test otherwise looks good to me.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -35,6 +38,14 @@ public class ApolloCodegen {
   public static func build(
     with configuration: ApolloCodegenConfiguration,
```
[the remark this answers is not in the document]
calvincestari: Not necessary in this PR. Just on my wish list before 1.0

Sources/ApolloCodegenLib/ApolloCodegen.swift
AnthonyMDev: That's right!

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -52,11 +63,25 @@ public class ApolloCodegen {
...
       ir: ir,
-      config: configContext
+      config: configContext,
+      fileManager: fileManager
     )
+
+    if configuration.options.pruneGeneratedFiles {
+      try deleteExtraneousGeneratedFiles(
+        from: &existingGeneratedFilePaths,
+        afterCodeGenerationUsing: fileManager
```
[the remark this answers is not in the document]
AnthonyMDev: It was just to make it clear that this is very much dependent on you doing this AFTER generating code and with the same file manager you used for generation.

If you moved this function up above `generateFiles`, it would run just fine but wouldn't do what we want it to exactly. (You'd actually end up with the same outcome I think, just all files would get deleted then re-created.)

I'm not super hung up on the name, but I think it's more descriptive of a piece of implicit knowledge and implementation details than `fileManager`.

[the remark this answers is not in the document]
calvincestari: I get why you named it that, but disagree on the value of it. I'm not concerned enough about it to request a change, this is fine.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: I'm saying we don't have a happy path test with the inclusion of `operationSearchPaths`, only throwing tests. It works now because the default value of `operationSearchPaths` in this convenience initializer is `["**/*.graphql"]`. I was suggesting this line become `input: .init(schemaPath: "path.graphqls", operationSearchPaths: ["**/*.graphql"])` to become more explicit about it validating when both `schemaPath` and `operationSearchPaths` include an extension.

calvincestari: The default value proves the test though and if that ever changed this test would likely break. I'm OK with that.

PR #2492, opened 2022-09-13
Add SwiftPM Plugin Permission Definitions

calvincestari:
I came across `permissions` while reading a bit more about build plugins today and it looks like if we specify these the user will be prompted to allow the permission on execution. We have the [hint](https://github.com/apollographql/apollo-ios/blob/release/1.0/Plugins/SharedPackagePluginExtensions/Process%2BTerminationHandler.swift#L7) to help users but this is a nice addition too.


AnthonyMDev:
Awesome! My question is, does using `--disable-sandbox` make it so that this doesn't require the extra step of typing yes? For CI platforms there needs to be some mechanism for doing this in a non-interactive way.

AnthonyMDev:
And if we have a non-interactive method of doing this, we can remove the hint on failure. I agree with your reasoning there.

calvincestari:
OK, here's the automated way to skip that permissions request.

```
❯ swift package plugin apollo-initialize-codegen-config
Plugin ‘ApolloCodegenPlugin-Initialize’ wants permission to write to the package directory.
Stated reason: “Adds a codegen JSON configuration file.”.
Allow this plugin to write to the package directory? (yes/no) no
error: Plugin was denied permission to write to the package directory.
Usage: swift package <options> <subcommand>
  See 'package -help' for more information.
❯ swift package plugin --allow-writing-to-package-directory apollo-initialize-codegen-config
Building for debugging...
[197/197] Linking apollo-ios-cli
Build complete! (4.37s)
New configuration output to ./apollo-codegen-config.json.
❯ swift package plugin apollo-initialize-codegen-config
Plugin ‘ApolloCodegenPlugin-Initialize’ wants permission to write to the package directory.
Stated reason: “Adds a codegen JSON configuration file.”.
Allow this plugin to write to the package directory? (yes/no) no
error: Plugin was denied permission to write to the package directory.
Usage: swift package <options> <subcommand>
  See 'package -help' for more information.
```

Note that the command is `--allow-writing-to-package-directory` not `--allow-writing-to-directory`. The [SwiftPackageTool source has both](https://github.com/apple/swift-package-manager/blob/main/Sources/Commands/SwiftPackageTool.swift#L947) but the first one is what we want to tell people to use. [Here is where that flag is used](https://github.com/apple/swift-package-manager/blob/main/Sources/Commands/SwiftPackageTool.swift#L1032).

calvincestari:
I cannot get `--disable-sandbox` to work in an automated way so I'm thinking we should remove it the hint entirely and document `--allow-writing-to-package-directory` in our CLI docs.

AnthonyMDev:
Perfect. This is good to go then!

calvincestari:
I'll remove that hint first before merging though.

Issue #2439, opened 2022-08-16
Custom Codable for ApolloCodegenConfiguration


calvincestari:
The default Codable behaviour is quite brittle when it comes to adding new properties to structs that get synthesized, i.e.: we can't add a new one even with a default value without Codable requiring it to be present in the JSON file. As we move into GA we need to ensure that once users have their configuration JSON they're not forced to add new parameters as we release them.

calvincestari:
Notes from a conversation yesterday about how we want to improve on this:
* only have required and common properties output in the CodegenCLI `init` JSON
* enums cases without associated values should not be generated as objects; `"moduleType": "other"` vs. `"moduleType": { "other": { } }`

calvincestari:
This is complete with the merging of #2501.

PR #2501, opened 2022-09-15
feat: [cli] Reduce initialized JSON config file to only required parameters

calvincestari:
Finishes #2439.

`ApolloCodegenConfiguration` has quite a few parameters but not all of them are required. With this change only the required parameters are output when a new JSON configuration file is created. The optional parameters will be detailed in documentation along with easy-to-add snippets for as much visibility as possible.

### DIFF
--- a/Sources/CodegenCLI/Commands/Initialize.swift
+++ b/Sources/CodegenCLI/Commands/Initialize.swift
@@ -44,8 +44,8 @@ public struct Initialize: ParsableCommand {
 
   func _run(fileManager: ApolloFileManager = .default, output: OutputClosure? = nil) throws {
     let encoded = try ApolloCodegenConfiguration
-      .default
-      .encoded()
+      .minimalJSON
+      .asData()
 
     if print {
       try print(data: encoded, output: output)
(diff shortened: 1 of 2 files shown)


Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -106,36 +106,52 @@ public struct Initialize: ParsableCommand {
...
     #endif
   }
 
-  func encoded() throws -> Data {
-    let encoder = JSONEncoder()
-    encoder.outputFormatting = [.prettyPrinted]
-
-    return try encoder.encode(self)
+  static func minimalJSON(supportCocoaPods: Bool) -> String {
+    let cocoaPodsOption = supportCocoaPods ? """
```
AnthonyMDev, line 119: lol I love this hack. If `TemplateString` was a separate library, we could use that here ;) but this works fine!

AnthonyMDev (approved):
This looks great! Nice job on the tests. Was worried they wouldn't be able to test both cases because of the `#if COCOAPODS`, but you got it just right!

PR #2502, opened 2022-09-15
Escape input params using reserved keywords

### LINKED ISSUE #2484: Keyword "where" is not escaped
## Bug report

When targeting an API that uses the word "where", this keyword is not escaped with backticks in the generated Swift code. This happens when using the codegen library provided by the `swift package --disable-sandbox plugin apollo-generate` plugin.

For example, the operation below:

```graphql
mutation DeleteProperties(
  $where: PropertyFilterInput
) {
  deleteProperties(where: $where) {
    ID
  }
}
```

Generates this code that does not compile:

Screen Shot 2022-09-08 at 3 07 31 PM

If I manually escape "where", the code compiles as expected:

Screen Shot 2022-09-08 at 3 07 58 PM

## Versions

- `apollo-ios` SDK version: 1.0.0-beta.3
- Xcode version: 13.3.1
- Swift version: swiftlang-5.6.0.323.62, clang-1316.0.20.8
- Package manager: 5.6.0

## Steps to reproduce

Using a schema that has the keyword "where", generate an operation. The generated code will fail to compile. I'm not sure if this is related to the parameter name I am using (which is also "where") or the actual schema.

AnthonyMDev:
Closes #2484


calvincestari (approved):
LGTM!

PR #2506, opened 2022-09-16
Remove Selection.Conditions ExpressibleByLiterals

### LINKED ISSUE #2500: Ambiguous Use Error when using two || joined inclusion conditions
The `||` operators for `Condition` and `Conditions` are colliding here.

```swift
.include(if: "includeSpecies" || "includeA", .field("species", String.self)),  // **Ambiguous use of operator '||'**
```

AnthonyMDev:
Closes #2500 

This removes ambiguity in initializing either a `Condition` or a `Conditions` from a string literal. String literal should always create a `Condition` and the `include(if:selection:)` function overloads are used to wrap those up into a `Conditions` object properly.

Also clean up some formatting


calvincestari (approved):
Nicer work on the fix for this!

Issue #2497, opened 2022-09-15
Audit all _ prefixed things to see if they should be available using ApolloExtension.


AnthonyMDev:
- [x] `SelectionSet.schema` -> `__schema`
- [x] `SelectionSet.selection` -> `__selection`
- [x] `JSONEncodable.jsonValue` -> `_jsonValue`
  - [x] Add public `jsonValue` to `CustomScalar` with extension to convert to `_jsonValue`
  - [x] `JSONEncodable.jsonValue` -> `_jsonEncodableValue`
- [x] `JSONDecodable.init(jsonValue:)` -> init(_jsonValue:)
- [x] `AnyHashableConvertible.asAnyHashable` -> `AnyHashableConvertible._asAnyHashable`
- [x] `OutputTypeConvertible.asOutputType` -> `OutputTypeConvertible._asOutputType`
- [x] `GraphQLOperation.variables` -> `GraphQLOperation._variables`
  - [x] Same for `LocalCacheMutation.variables`

calvincestari:
This is complete with the merging of #2509

PR #2509, opened 2022-09-19
feat: Mark operation `variables` as private

calvincestari:
_PR 3 of 3 for #2497 - I was going to submit all the 'underscore' changes as a single PR but it's going to get too big so I'm splitting them up into similar parts._

This PR marks the `variables` property on `GraphQLOperation` and `LocalCacheMutation` as private. They will remain accessible to developers because the access modifiers aren't changing but the underscore serves as enough of a hint of "use at your own risk, could change without warning".


calvincestari:
@AnthonyMDev let's discuss whether we should mark `GraphQLResponse.variables` with an underscore too.

PR #2511, opened 2022-09-20
feat: Custom `Codable` for `ApolloSchemaDownloadConfiguration`

calvincestari:
This is to support optional properties for `ApolloSchemaDownloadConfiguration` and it's subtypes.


AnthonyMDev (approved):
Thanks for this! LGTM

PR #2519, opened 2022-09-27
Updated CodegenCLI defaults

### LINKED ISSUE #2518: ApolloCodegenConfiguration defaults changes
After building out some of the documentation @AnthonyMDev noted that we were directing users to change settings after initialization when it could be done along with initialization.

1. Change the codegen CLI to require a command line option for schema name
2. If the JSON file is being initialized to support CocoaPods then the default schema module type should be `.other`.

calvincestari:
Closes #2518 

In this PR:
1. Changes the codegen CLI `init` command to require the user to supply a schema name
2. If the JSON file is being initialized to support CocoaPods then the default schema module type is `.other`


AnthonyMDev (approved):
Thanks for getting this done so quickly!

PR #2529, opened 2022-09-29
Default the `RequestChain` to run on background thread

engali94:
RequestChain causing performance regression because it is called on the main thread


calvincestari:
I'm closing this PR due to inactivity. If the rest of the solution is added at any point we're happy to reconsider it - thank you.

PR #2533, opened 2022-09-30
feat: Removes Carthage

calvincestari:
Removes all mention of the Carthage dependency manager from source code, process and documentation.


AnthonyMDev:
Love this so much

PR #2544, opened 2022-10-03
fix: Support multiline deprecation messages

TizianoCoroneo:
Hello everyone, 

I was trying out the new version of the library in our codebase, and I found a small issue: our schema defines a deprecation warning that spans multiple lines. 
This causes the current version of the codegen to output stuff like this: 
```swift
@available(*, deprecated, message: "Very long deprecation message that 
                         spans multiple lines but 
                         without a triple quote, resulting in a \"Unterminated literal\" error.")
```

This PR adds a test case that reproduces the error, and a new string interpolation method on `TemplateString` that makes it easier to add deprecation warnings wherever they're needed.

The new result looks like this: 
```swift
@available(*, deprecated, message: """
    Very long deprecation message that 
    spans multiple lines but 
    without a triple quote, resulting in a \"Unterminated literal\" error.
    """)
```


AnthonyMDev:
Thanks for this @TizianoCoroneo. I think I'm going to do just a little bit of clean-up to use some of the infra we already have in place better, but this looks great. I'll get it in ASAP.

PR #2556, opened 2022-10-08
Apply escaping to input object parameter names

puls:
I'm not sure if this is sufficient, but I have some input objects with fields named `in` and this change allows my generated code to compile.


AnthonyMDev:
Thanks for this! I'm going to write some unit tests for this before merging in. Be on the lookout for version 1.0.1 very soon with the fix!

AnthonyMDev:
Closing in favor of #2561! Thanks so much for this fix!

PR #2561, opened 2022-10-11
Fix escaping input params

AnthonyMDev:
@puls Thanks for the initial fix. This just adds tests and escapes the reserved names in the getters and setters as well.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
@@ -130,7 +130,7 @@ enum SwiftKeywords {
     "try",
   ]
 
-  fileprivate static let InputParameterNamesToEscape: Set<String> = FieldAccessorNamesToEscape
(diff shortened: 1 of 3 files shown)


Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
```diff
@@ -130,7 +130,7 @@ enum SwiftKeywords {
     "try",
   ]
 
-  fileprivate static let InputParameterNamesToEscape: Set<String> = FieldAccessorNamesToEscape
+  static let InputParameterNamesToEscape: Set<String> = FieldAccessorNamesToEscape
```
calvincestari, line 133: Does the access level on this need to change? It doesn't appear to be used anywhere new in this PR.

calvincestari (approved):
Looks good, just one question. Thanks for the contribution @puls and the additional tests @AnthonyMDev

Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
AnthonyMDev: Fixed

PR #2585, opened 2022-10-14
Fix document type namespace

### LINKED ISSUE #2581: Add Namespace for ApolloAPI types in Generated Code
## Feature request

There is a high probability of naming clashes in client app codebase and Apollo types. Name-spacing or prefixing would solve this issue.

## Motivation

There is an actual problem that I'm facing.
I wanted to have `DocumentType` as enum in my app, however all generated operation files contain the following line:
`public static let document: DocumentType = .notPersisted( ...` that causes a build error.

## Proposed solution

There are a few solutions that I can think of:
1. Specifying the module in this specific case: `public static let document: Apollo.DocumentType = .notPersisted( ...`. This one is more of a quick fix, but it would solve this particular problem.
2. Using a namespace to prevent other potential clashes.

## Outstanding Questions

Perhaps there's another way of getting around this issue. If I'm missing something I'd appreciate further instructions 🙂

AnthonyMDev:
Closes #2581

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -649,6 +649,10 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     public static let schemaDownloadConfiguration: ApolloSchemaDownloadConfiguration? = nil
   }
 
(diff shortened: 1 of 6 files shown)


calvincestari (approved):
👍🏻

Issue #2546, opened 2022-10-04
Problem with `schemaName`


pchmelar:
## Bug report
- Thanks for the 1.0.0 release! ❤️  I just integrated it into our testing project, and I have one particular problem with `schemaName`

## Steps to reproduce
- I'm using [`ModuleType.embeddedInTarget`](https://www.apollographql.com/docs/ios/code-generation/codegen-configuration#embedded-in-target) and [`TestMockFileOutput.absolute`](https://www.apollographql.com/docs/ios/code-generation/codegen-configuration#absolute-test-mocks-output) options.
- When `schemaName` is in lowercase (`"schemaName": "rocket"` for example), the `rocket.graphql.swift` with the `Rocket` caseless namespace enum is created. However, the generated Mocks are not working, because there is `rocket` instead of `Rocket`.
- When I change the `schemaName` to start with the capitalized letter (`"schemaName": "Rocket"` for example), the file with caseless namespace enum isn't created at all. 😔 

![New Project](https://user-images.githubusercontent.com/3842139/193872790-99b2f263-9433-4552-8506-1d7367367d3f.png)
<img width="1049" alt="Screenshot 2022-10-04 at 14 21 02" src="https://user-images.githubusercontent.com/3842139/193872686-8625e0e3-fdae-4098-9ac3-556b7ff9192b.png">

## Versions
- `apollo-ios` SDK version: 1.0.0
- Xcode version: 14.0
- Swift version: 5.7
- Package manager: SPM

AnthonyMDev:
Thanks for alerting us to this one. Supports a lowercase named schema here should be an easy fix, we'll get a PR up for that ASAP and release it as part of the first patch version. (We're at GraphQL summit this week, so that will probably be on Monday when we're back!)

> When I change the schemaName to start with the capitalized letter ("schemaName": "Rocket" for example), the file with caseless namespace enum isn't created at all. 😔

This definitely shouldn't be happening. I'm not sure why it would not create the enum. I'll see if we can reproduce that...

calvincestari:
@SilverTab are you able to share with me a project (schema + operations + config) that demonstrates the second issue - missing file? I can replicate the first issue of test mocks using the incorrectly cased type but not the second missing file bug.

AnthonyMDev:
@calvincestari so this wasn't due to having a type named `Rocket` then?

SilverTab:
[...]

PR #2586, opened 2022-10-14
fix: Uses correct casing for schema name in generated namespace

calvincestari:
PR 1 of 3 for #2546 

This PR:
* Correctly cases `schemaName` when used in namespacing
* Refactored logic to use the config schema name and not the IR schema name
* Removes schema name from function declarations when config is already present


AnthonyMDev (approved):

AnthonyMDev:
Thanks so much for this! Approved!

AnthonyMDev:
And nice job cleaning up the unneeded `moduleName` parameters

PR #2589, opened 2022-10-18
fix: Schema name conflict errors

calvincestari:
PR 2 of 3 for #2546

This PR ensures that the user-given schema name:
- does not conflict with any of the generated file/type names by disallowing the schema name to match any of the schema referenced types
- does not conflict with any of the generation selection set names by disallowing a schema name matching any entity field name and any pluralized version of an entity list name
- these checks are done during the compilation stage because that is where the code still has access to the schema and can check field definition types
- ValidationRule checks have been reduced to only reserved words


Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -1912,8 +1914,172 @@ class ApolloCodegenTests: XCTestCase {
     ), rootURL: nil)
 
     // then
-    expect(try ApolloCodegen.validate(config: configContext))
+    expect(try ApolloCodegen.validate(config: configContext, compilationResult: .mock()))
       .to(throwError(ApolloCodegen.Error.inputSearchPathInvalid(path: "schema/*.")))
   }
 
+  // MARK: Validation Tests (conflicting schema name)
```
calvincestari, line 1921: @AnthonyMDev these tests do the job but they rely on knowing the implementation detail of how we name the generated files, i.e.: there is currently no test for mutations because we append "Mutation" or "Local CacheMutation" to the filename.

I wonder if these tests might be better suited to going in the file generators?

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -24,6 +25,8 @@ public class ApolloCodegen {
         return "Schema Types must be generated with module type 'swiftPackageManager' to generate a swift package for test mocks."
       case let .inputSearchPathInvalid(path):
         return "Input search path '\(path)' is invalid. Input search paths must include a file extension component. (eg. '.graphql')"
+      case let .schemaNameConflict(name):
+        return "Schema name \(name) conflicts with a GraphQL schema type. Please choose a difference schema name."
```
AnthonyMDev, line 29:
```suggestion
        return "Schema name \(name) conflicts with name of a type in your GraphQL schema. Please choose a different schema name. Suggestions: \(name)Schema, \(name)GraphQL, \(name)API"
```

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
AnthonyMDev: I think this is appropriate here. Unit tests define an expected behavior. Because they are unit tests, _not_ integration tests, that sometimes relies on implementation details. As long a the production code doesn't rely on implementation details, I think this is where we verify that we aren't violating the behaviors as implementation details change in the future.

Just clarifying, there is also no validation for queries and subscriptions, just for schema types right? You specifically called out mutations which threw me off there.

AnthonyMDev (approved):

AnthonyMDev:
Just a type and minor addition to the schema conflict error message. But otherwise this LGTM!

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
calvincestari: > Just clarifying, there is also no validation for queries and subscriptions, just for schema types right? You specifically called out mutations which threw me off there.

Schema types and fragments because we don't always append "Fragment" to the fragment filename. Operations should always have their _type_ appended to the filename.

calvincestari:
> Just a type and minor addition to the schema conflict error message. But otherwise this LGTM!

Still wanting to add the entity vs. list behaviours to this PR too.

Sources/ApolloCodegenLib/Frontend/JavaScript/src/compiler/index.ts
```diff
@@ -292,6 +297,26 @@ export function compileToIR(
...
+                { nodes: node }
+              );
+            }
+          }
+        }
+
+        if (isListType(fieldType) || (isNonNullType(fieldType) && isListType(fieldType.ofType))) {
+          validateFieldName(selectionNode, validationOptions.disallowedEntityListFieldNames)
+        } else if (isCompositeType(unwrappedFieldType)) {
+          validateFieldName(selectionNode, validationOptions.disallowedEntityFieldNames)
```
calvincestari, line 317: Arguably this check could go into the below `if (isCompositeType(unwrappedFieldType)) {` statement, let me know if you feel strongly about it. I prefer to have the `if else` and keep the logic together.

calvincestari: Also, if we move the check into the if statement below then the validation will be done twice because a field can be both a list type and and a composite type.

Tests/ApolloCodegenTests/Frontend/DocumentParsingAndValidationTests.swift
```diff
@@ -214,45 +214,7 @@ class DocumentParsingAndValidationTests: XCTestCase {
...
-      Field name "\(field)" is not allowed because it conflicts with generated \
-      object APIs. Please use an alias to change the field name.
-      """,
-      ])
-
-      XCTAssertEqual(document.filePath, "TestQuery.graphql")
-    }
-  }
-
-  func test__validateDocument__givenFieldNameIsSchemaName_throwsError() throws {
```
calvincestari, line 231: This test now becomes part of the compilation stage tests.

Sources/ApolloCodegenLib/Frontend/ValidationOptions.swift
```diff
@@ -3,12 +3,44 @@ import JavaScriptCore
 
 public struct ValidationOptions {
 
-  let disallowedFieldNames: Set<String>
+  struct DisallowedFieldNames {
+    let scalar: Set<String>
```
AnthonyMDev, line 7: This should be `allFields` not `scalar`. It does (and should) also apply to entity fields, so calling it `scalar` is misleading.

AnthonyMDev (approved):
Thanks so much @calvincestari, I know this one was a lot harder than we expected!

Sources/ApolloCodegenLib/Frontend/ValidationOptions.swift
calvincestari: Fair point, I'll change that up now - thanks.

Issue #2596, opened 2022-10-20
`GraphQLNullable` nil coalescing


calvincestari:
From community topic https://community.apollographql.com/t/ios-1-0-graphqlnullable-null-coalescing/4681/1.

For something like:
```
let cursor: String? = "blah"

let query: Query = Query(pagination: Pagination(before: cursor ?? .none))
```

I keep getting the following error:
```
Cannot convert value of type ‘String?’ to expected argument type ‘GraphQLNullable’
```

calvincestari:
See https://github.com/Pocket/pocket-ios/pull/277 for a reproducible sample.

PR #2600, opened 2022-10-21
Exported Import all of Apollo API in generated operations

AnthonyMDev:
This fixes https://github.com/apollographql/apollo-ios/issues/2596 by ensuring the operator overload is imported wherever you are using the generated models.


calvincestari (approved):
Awesome, thanks for the quick fix!

PR #2607, opened 2022-10-25
fix: InputDict `dynamicMember` operator conflict with `hash` function

### LINKED ISSUE #2603: Generated code produces compile error when accessing "data" dictionary in the "InputDict" struct if the name of the accessed property is "hash"
## Bug report

Currently there is a compiler error in generated code when accessing "data" dictionary in the "InputDict" struct if the name of the accessed property is "hash".

## Versions

- `apollo-ios` SDK version: 1.0.2
- Xcode version: 14.1 RC
- Swift version: 5.7.1

## Steps to reproduce

1. execute generate code command.

## Further details
Here is a screenshot of errors produced by the compiler:

Screenshot 2022-10-24 at 15 38 55

I think the reason why this fails is because InputDict has a function:

```swift
  public func hash(into hasher: inout Hasher) {
    hasher.combine(data._jsonEncodableValue?._jsonValue)
  }
```

and that causes the compiler to not be able to find "hash" field through dynamic member lookup.

changing generated code to this removes the compile error:

Screenshot 2022-10-24 at 15 44 13

If there is need for any additional info for solving this issue I will try to answer it here.

calvincestari:
Fixes #2603 

If you have an `InputObject` with a parameter name of `hash` it will conflict with the Hashable protocol function named `hash`. We can fix this by using the fully qualified operator name for the `dynamicMember` subscript.


AnthonyMDev:
I assume this will be the same issue for `SelectionSet` if you have a field named `hash` then? I don't love the verbosity of adding this qualified member name to every field accessor though.

I'd rather special case the keyword `hash` and just use `data[dynamicMember: "hash"]` in the specific case of this. It may be that there are other keywords which will cause this problem also that we have not encountered yet (or that will become issues in the future if we add to the `SelectionSet/InputObject` APIs). And we should have a general solution for this built into `SwiftKeywords` for this.

calvincestari:
> I assume this will be the same issue for SelectionSet if you have a field named hash then? I don't love the verbosity of adding this qualified member name to every field accessor though.

I checked and no this issue doesn't exist for `SelectionSet` because `DataDict` does not offer `@dynamicMemberLookup`. It instead implements a custom subscript and avoids this issue.

That gave me the idea of simply making `InputDict` more like `DataDict` and to remove `@dynamicMemberLookup`, and from what I can tell we don't lose anything and manage to avoid needing to use `dynamicMember:`.

I've added a field named `hash` to AnimalKingdomAPI in both an input and selection set and with all generated code updated it all builds and functions as expected. Can you think of any reason we'd need to keep `@dynamicMemberLookup`?

AnthonyMDev (approved):
This is definitely the right way to handle this. Nice!

Nit: I'm not a fan of adding properties to test edge cases to our example generated models. I add things to them while I'm testing things out, but then I write a unit test and remove the test properties from the main generated models. This really isn't a big deal, and if you want to keep it, I'm not going to argue it. But I would hesitate to make this a common practice. These are meant to be an example of the basic structure, so that our users can see an example of what good generated models look like. But if we start using them to maintain regression testing of edge cases, they will become ugly and not a good quality example for people. And they also conflict with the point of unit tests. I don't want us getting into a habit of relying on "if the example models still build, then it's okay." because that logic leads to lazier unit testing habits.

calvincestari:
I had this test which I removed - https://github.com/apollographql/apollo-ios/pull/2607/commits/d766a297b65e6cd826bc4b664058f6ba476b4554.

PR #2609, opened 2022-10-26
Download schema relative to root url

Anteo95:
This PR fixes downloading schema when "rootURL" is provided. 

Currently, even though rootURL can be provided as parameter to function

```swift
  public static func fetch(
    configuration: ApolloSchemaDownloadConfiguration,
    withRootURL rootURL: URL? = nil
  )
```

it is actually not used internally on all places where it should be. This causes schema to be downloaded into incorrect folder.


calvincestari (approved):

calvincestari:
Codegen CLI Integration Tests are failing because of a CocoaPods error with forks - merging anyways..

PR #2618, opened 2022-10-27
fix: Throw error when no schema nor operation files are found

### LINKED ISSUE #2567: Return error if no matches to schema or operation search paths
https://github.com/apollographql/apollo-ios/issues/2543#issuecomment-1272558850

> Error: Cannot query field "company" on type "undefined"

This error is indicative of a schema that could not be loaded. I think we used to have this behaviour and it was removed when we changed the configuration validation. Clearly there weren't good enough tests around the behaviour, we'll fix that now.

calvincestari:
Closes #2567 

When a schema file could not be found graphql-js would emit errors but they were not indicative of the underlying problem. I've added validation to ensure that at least one match of the schema/operation search paths is found otherwise an error is thrown.


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -56,12 +62,14 @@ public class ApolloCodegen {
       rootURL: rootURL
     )
 
+    try validate(config: configContext)
```
calvincestari, line 65: I've separated these two `validate` calls because the first one includes path syntax validation and should be done before compilation of the schema + operations. The second validation ensures no schema name conflict.

I don't like that they're now separate, some ideas to improve:
* rename the second one from `validate` to something else
* move the schema name logic somewhere else

Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -191,7 +191,7 @@ public struct Glob {
       globfree(&globT)
     }
 
-    CodegenLogger.log("Matching \(pattern)", logLevel: .debug)
+    CodegenLogger.log("Evaluating \(pattern)", logLevel: .debug)
```
calvincestari, line 194: `Matching` was too close to `Matched`. `Evaluating` is more indicative of what the logic is actually doing.

PR #2622, opened 2022-10-28
Added fieldInfo to GraphQLExecutionError, to enable more robust error messaging around missing values.

danieltiger:
Right now the `GraphQLExecutionError` only includes the `path` of the error. But, in the case of an underlying `JSONDecodingError.missingValue` error, the error description doesn't provide enough information to understand what the problem actually was. You know that there was an object and a value was missing, but no idea what key wasn't found.

This PR is intended to start a discussion on how to improve this. By adding `fieldInfo` to `GraphQLExecutionError` it's possible to print out `fieldInfo.cacheKeyForField` and figure out what the key was.

I ended up taking this approach because it was the least invasive. No current consumer of `GraphQLExecutionError` has to change anything, and all the tests continue to pass. I considered adding an associated value to `JSONDecodingError.missingValue` but that would have been a much larger change. Also, there are already situations where the `FieldExecutionInfo` wouldn't make sense for `JSONDecodingError`. 

Anyway, I'm happy to change any/all of this. Like I said, I just want to start a process of improving this.

### DIFF
--- a/Sources/Apollo/GraphQLExecutor.swift
+++ b/Sources/Apollo/GraphQLExecutor.swift
@@ -128,8 +128,9 @@ fileprivate struct FieldSelectionGrouping: Sequence {
 
 /// An error which has occurred during GraphQL execution.
 public struct GraphQLExecutionError: Error, LocalizedError {
-  let path: ResponsePath
+  let fieldInfo: FieldExecutionInfo
 
+  var path: ResponsePath { fieldInfo.responsePath }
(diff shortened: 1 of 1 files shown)


AnthonyMDev:
Thanks for bringing this to our attention. I'm going to do a little more thinking about this to figure out what the appropriate format for exposing this information is. I'll get back to you when we get to this one!

calvincestari:
Closing due to inactivity.

PR #2624, opened 2022-10-29
Deletes out-of-date generated test code

calvincestari:
@AnthonyMDev - is there any reason we still need this generated code? Doesn't seem to be used in any tests and we have the other AnimalKingdomAPI generated sample which is actually maintained.


AnthonyMDev:
Nope, go ahead and kill these!

PR #2634, opened 2022-11-02
Do not generate a convience initializer for objects without fields

Gois:
We encountered a problem during testing with the mocks, some objects don't have fields except for their __typename. In that case the cli would still generate a convenience initializer that would call `self.init()` which causes infinite recursion.

This has been fixed by checking if the fields are empty or not.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
@@ -49,17 +49,21 @@ struct MockObjectTemplate: TemplateRenderer {
         }, separator: "\n")
       }
     }
-
-    public extension Mock where O == \(objectName) {
-      convenience init(
-        \(fields.map { """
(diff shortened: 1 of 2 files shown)


calvincestari:
Hi @Gois 👋🏻 - thanks for the contribution.

Do you have a GraphQL example you can share, I'm intrigued at the practical use of an object without fields?

Gois:
Hi @calvincestari, sure I can elaborate a bit more. In our case we use it for example like this in a query:

```
... on CategoryPicker {
    __typename
}
```

Based on that type (if we get it back) in our response we know that we have to build up a certain UI element. So in Swift we check it like this:

```swift
if child?.asCategoryPicker != nil {
    // code for adding UI element
}
```

In this case we don't have any fields in our schema for this type, so that's why it only has the typename field and we ran into the problem with the convenience initializer.

calvincestari:
Got it, thanks for sharing! I'll get this PR merged today.

calvincestari (approved):
LGTM!

calvincestari:
Codegen CLI Integration Tests are failing because of a CocoaPods error with forks - merging anyways..

PR #2638, opened 2022-11-04
Generate library for test mock target

Gois:
Currently when you generate the test mocks for a Swift package it won't create a library for it, so you can't import the test mocks properly into your unit tests.

To fix this I've updated the SwiftPackageManagerModuleTemplate to also create a library for the test mocks if there is a testMockTarget.


AnthonyMDev:
Thanks for this! We definitely should get this added. I'll do a full review ASAP to make sure this is good to go before merging. And you can ignore that integration test failure. It's a known issue and we're working on it.

AnthonyMDev (approved):
Thanks so much for this PR! Approved and merging in, this will be in the next patch release!

PR #2658, opened 2022-11-10
Create TestIsolatedFileManager

AnthonyMDev:
I'm going to need to create temporary files again for my next PR, so I broke this functionality into a test infra piece!

```swift
/// `TestIsolatedFileManager` is a test helper object that manages creation and deletion of files in a temporary directory
/// that ensures test isolation.
///
/// All files and directories created by this class will be automatically deleted upon test
/// completion prior to the test case's `tearDown()` function being called.
///
/// You can create a file manager from within a specific unit test with the
/// `testIsolatedFileManager()` function on `XCTestCase`.
```


Sources/ApolloCodegenLib/Glob.swift
```diff
@@ -206,7 +206,8 @@ public struct Glob {
 
     CodegenLogger.log("Evaluating \(pattern)", logLevel: .debug)
 
-    let response = glob(pattern, flags, nil, &globT)
+    let encodedPattern = pattern
```
calvincestari, line 209: Any reason for the new variable assignment here?

calvincestari (approved):
Thanks for this new infrastructure, a win for tests!

Sources/ApolloCodegenLib/Glob.swift
AnthonyMDev: Oops, was doing some debugging and forgot to fix that. I'll fix it before merging.

PR #2659, opened 2022-11-11
Implement CLI Version Checker

### LINKED ISSUE #2562: Investigate how to distribute CLI outside of SPM
Using the CLI through SPM plugins is not the best experience and requires the use of options such as `--disable-sandbox` in order to unlock the full functionality. Is there a way we can distribute the CLI outside of SPM and maintain a strong link to the correct version of apollo-ios.

AnthonyMDev:
This implements the feature as described here: https://github.com/apollographql/apollo-ios/issues/2562#issuecomment-1287486458

Closes #2562


Sources/CodegenCLI/Commands/Generate.swift
```diff
@@ -53,6 +61,36 @@ public struct Generate: ParsableCommand {
...
+      if ignoreVersionMismatch {
+        print("""
+          Warning: \(errorMessage)
+          """)
+      } else {
+
+        throw Error(errorDescription: """
+          Error: \(errorMessage)
+
+          To ignore this error and run the CLI anyways, use the argument: --ignore-version-mismatch.
```
calvincestari, line 88:
```suggestion
          To ignore this error during code generation, use the argument: --ignore-version-mismatch.
```
nit: Slightly clearer message of which command/workflow will ignore the error.

Tests/CodegenCLITests/VersionCheckerTests.swift
```diff
@@ -0,0 +1,231 @@
...
+    // when
+    let result = try VersionChecker.matchCLIVersionToApolloVersion(
+      projectRootURL: nil
+    )
+
+    // then
+    expect(result).to(equal(.versionMismatch(cliVersion: Constants.CLIVersion, apolloVersion: apolloVersion)))
+  }
+
+  func test__matchCLIVersionToApolloVersion__givenPackageResolvedFileInXcodeProject_withVersion2FileFormat_hasMatchingVersion_returns_versionMatch() throws {
```
calvincestari, line 189: Is only using version 2 of the package file format linked to the change in swift tools version of 5.6? There is a chance that a project doesn't regenerate their schema package and still uses an older version of Xcode; do we know how far back it began to use v2 of the package format?

Tests/CodegenCLITests/Commands/GenerateTests.swift
```diff
@@ -445,4 +446,106 @@ class GenerateTests: XCTestCase {
     // then
     expect(level).toEventually(equal(.debug))
   }
+
+  // MARK: Version Checking Tests
```
calvincestari, line 450: Should we have another test here to simulate a CocoaPods environment, i.e.: any other dependency manager where `.noApolloVersionFound` would be returned because there is no package file. I see there's a test for that in the version checker but we have no validation for how the generate command responds in that case.

calvincestari (approved):
A couple minor questions, otherwise looks good!

Tests/CodegenCLITests/VersionCheckerTests.swift
AnthonyMDev: Yeah v2 is linked to the swift tools version 5.6.

Yes, the Version Checker does support both v1 and v2. There are tests that it parses a v1 package. But there is no reason to add another test for that here. We have tests for the behavior, and writing a new test that is never going to fail and doesn't require any new code to be written to pass violates TDD principles and is honestly a waste of time. Unless there is actually some new behavior to test, having multiple tests that test the same lines of code is not necessary.

Tests/CodegenCLITests/Commands/GenerateTests.swift
AnthonyMDev: Sure, test added!

Sources/CodegenCLI/Commands/Generate.swift
AnthonyMDev: Done

Sources/CodegenCLI/Commands/Generate.swift
```diff
@@ -19,6 +19,12 @@ public struct Generate: ParsableCommand {
   )
   var fetchSchema: Bool = false
 
+  @Flag(
+    name: .long,
+    help: "Ignore Apollo version mismatch errors. This may lead to incompatible generated objects."
```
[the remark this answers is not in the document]
AnthonyMDev: Done

Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
```diff
@@ -16,7 +16,7 @@ struct SwiftPackageManagerModuleTemplate: TemplateRenderer {
     let casedSchemaName = config.schemaName.firstUppercased
 
     return TemplateString("""
-    // swift-tools-version:5.3
+    // swift-tools-version:5.6
```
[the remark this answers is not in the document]
AnthonyMDev: Yes, something to add to the neverending list hahah ;)

calvincestari (approved):

PR #2673, opened 2022-11-16
update: CLI defaults

### LINKED ISSUE #2595: CodegenCLI - Improve first experience defaults
This is trying to improve the first experience wit codegen based on what are the common issues being experienced since the release of 1.0.0.

1. Require `moduleType` as input on `init`
  a. `embeddedInTarget` require target name
2. Operation changed to `inSchemaModule`
3. Documentation changes needed
  a. Everywhere we link to `init` command should be updated
  b. Expand details on module type and what is required to be done by them, i.e.: `other` would require a podspec to be created.

calvincestari:
Closes #2595 

* Updates the `swift-argument-parser` dependency to the latest version, `1.2.0` (both SPM and Xcode)
* Changes `output.operations` default to `.inSchemaModule`
* Requires the user to specify `--module-type` on CLI `init` command
* Requires the user to specify `--target-name` when the module type is `.embeddedInTarget`
* Updates all documentation referencing the `init` command to detail the new parameters


Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -15,7 +15,26 @@ public struct Initialize: ParsableCommand {
...
     help: "Name used to scope the generated schema type files."
   )
-  var schemaName: String = ""
+  var schemaName: String
+
+  @Option(
+    name: [.long, .customShort("m")],
+    help: """
+      How to package the schema types for dependency management. Possible types: \
+      \(ModuleTypeExpressibleByArgument.allValueStrings.joined(separator: ", ")).
```
calvincestari, line 24: I'm joining them here with `,` because `allValueStrings` printed output contains the array braces `[]`.

Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -15,7 +15,26 @@ public struct Initialize: ParsableCommand {
     name: [.long, .customShort("n")],
     help: "Name used to scope the generated schema type files."
   )
-  var schemaName: String = ""
```
calvincestari, line 18: This isn't needed - swift argument parser will give a meaningful error to the user when a parameter isn't supplied when it's definition isn't initialized.

Tests/CodegenCLITests/Commands/InitializeTests.swift
```diff
@@ -191,7 +338,12 @@ class InitializeTests: XCTestCase {
 
       expect(actualPath).to(equal(expectedPath))
       expect(data?.asString).to(equal(
-        ApolloCodegenConfiguration.minimalJSON(schemaName: "MockSchema")))
+        ApolloCodegenConfiguration.minimalJSON(
```
AnthonyMDev, line 341: These tests are testing that you are calling `minimalJSON` with the correct parameters given the arguments you are parsing. But is there anything testing that `minimalJSON` is actually correct?

docs/shared/pods-setup-codegen-panel.mdx
```diff
@@ -12,10 +12,11 @@ The Codegen CLI is located in your project at `Pods/Apollo/apollo-ios-cli`
 The Codegen CLI uses a JSON file to configure the code generation engine. Use the Codegen CLI `init` command to create this file with default values. From the directory of your `Podfile`, run:
 
 ```bash
-./Pods/Apollo/apollo-ios-cli init --schema-name ${MySchemaName}
+./Pods/Apollo/apollo-ios-cli init --schema-name ${MySchemaName} --module-type other
 ```
 
-`${MySchemaName}` should be the name you want for the namespace of your generated schema files.
+* `${MySchemaName}` should be the name you want for the namespace of your generated schema files.
+* `other` is the module type to use when your dependency manager for Apollo iOS is CocoaPods.
```
AnthonyMDev, line 19: I think I have a problem with this. Let's talk about it more.

I don't think that telling them it's a no-brainer and just to use `other` if they are using Cocoapods is accurate. 

`other` is used if you want to package your generated schema as a separate module and distribute that as a Pod. But what we've found is that _most_ users actually just want to embed the generated models right into their application target. They are just including `Apollo` via Cocoapods. 

Even though we don't believe that is the best choice, it does seem to be the default expectation of users. So I think `embeddedInTarget` is still the default module type people would want to use. Either way, we should be very clear in explaining to people that this is a choice to be made and they should understand what the implications of each possible choice is. That way we aren't making `embeddedInTarget` the path of least resistance, but we also aren't confusing people who assumed that it would work that way, when their generated files don't link to their project and build properly.

docs/shared/spm-setup-codegen-panel.mdx
```diff
@@ -19,10 +19,11 @@ import SPMInstallCLI from "../shared/spm-install-cli.mdx"
 The Codegen CLI uses a JSON file to configure the code generation engine. Use the Codegen CLI `init` command to create this file with default values. From your project's root directory, run:
 
 ```bash
-./apollo-ios-cli init --schema-name ${MySchemaName}
+./apollo-ios-cli init --schema-name ${MySchemaName} --module-type swiftPackageManager
 ```
 
-`${MySchemaName}` should be the name you want for the namespace of your generated schema files.
+* `${MySchemaName}` should be the name you want for the namespace of your generated schema files.
+* `swiftPackageManager` is the module type to use when your dependency manager for Apollo iOS is Swift Package Manager.
```
AnthonyMDev, line 26: Same as above

Tests/CodegenCLITests/Commands/InitializeTests.swift
```diff
@@ -71,11 +94,69 @@ class InitializeTests: XCTestCase {
     expect(command.schemaName).to(equal("ShortFormatSchemaName"))
   }
 
+  func test__parsing__givenParameters_moduleNameLongFormat_shouldParse() throws {
```
AnthonyMDev, line 97: While these tests are awesome and super comprehensive, remind me to have a conversation with you about my views on testing behavior vs configuration! Would love to save you some time in the future possibly.

Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -154,16 +217,25 @@ extension ApolloCodegenConfiguration {
         "schemaTypes" : {
           "path" : "./\(schemaName)",
           "moduleType" : {
-            \(supportCocoaPods ? "\"other\"" : "\"swiftPackageManager\"") : {
-            }
+            "\(moduleType)" : {
+            \(moduleTarget)
```
AnthonyMDev, line 221: I could be wrong, but this looks like the indentation is off, and that it may be missing a `}`? Not sure. Do we have unit tests for this? (I have another comment in the `InitializeTests` asking about testing this as well.)

AnthonyMDev (changes requested):
Looks really good, just a couple comments around tests and documentation mostly!

I want to do a more comprehensive review of the docs because this only shows the files you _did_ change, and I think there should be some additions to docs in the getting started guide (and maybe other areas?) to really explain that you need to pick a module type and link to the section of the docs that explains what each of those options really does.

I think the docs need to clearly explain that this is a decision that affects your project structure, and depending on what you select here, how you include your generated files is going to be significantly different.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -169,7 +169,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
 
     /// Default property values
     public struct Default {
-      public static let operations: OperationsFileOutput = .relative(subpath: nil)
+      public static let operations: OperationsFileOutput = .inSchemaModule
```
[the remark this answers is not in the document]
calvincestari: Yes, good catch!

Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -45,8 +64,21 @@ public struct Initialize: ParsableCommand {
...
+
+    let targetNameValiationError = ValidationError("""
+      Target name is required when using \"embeddedInTarget\" module type. Use --target-name \
+      to specify.
+      """
+    )
+
+    switch (moduleType, targetName?.isEmpty) {
+    case (.embeddedInTarget, nil), (.embeddedInTarget, true):
+      throw targetNameValiationError
```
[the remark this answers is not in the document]
calvincestari: This was from when I had the code structured differently. I refactored the case statement into better form but completely forgot to get rid of that - thanks.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
```diff
@@ -261,12 +261,12 @@ class InputObjectTemplateTests: XCTestCase {
         intField: GraphQLNullable<Int> = nil,
         boolField: GraphQLNullable<Bool> = nil,
         floatField: GraphQLNullable<Double> = nil,
-        customScalarField: GraphQLNullable<TestSchema.CustomScalar> = nil,
-        lowercaseCustomScalarField: GraphQLNullable<TestSchema.LowercaseCustomScalar> = nil,
-        enumField: GraphQLNullable<GraphQLEnum<TestSchema.EnumType>> = nil,
-        lowercaseEnumField: GraphQLNullable<GraphQLEnum<TestSchema.LowercaseEnumType>> = nil,
-        inputField: GraphQLNullable<TestSchema.InnerInputObject> = nil,
-        lowercaseInputField: GraphQLNullable<TestSchema.LowercaseInnerInputObject> = nil,
+        customScalarField: GraphQLNullable<CustomScalar> = nil,
```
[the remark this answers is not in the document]
calvincestari: Yes - this test that changed is to validate the parameters and the initializers, namespacing is not the focus of the test. The next test in this file is designed to test namespacing, see `test__render__givenSchemaModuleInputFieldTypes__generatesCorrectParametersAndInitializer_withNamespaceWhenRequired`

Tests/CodegenCLITests/Commands/InitializeTests.swift
```diff
@@ -7,7 +7,10 @@ import ApolloCodegenLib
 class InitializeTests: XCTestCase {
 
   var mockFileManager: MockApolloFileManager!
-  let baseOptions = ["--schema-name=MockSchema"]
+  let requiredOptions = [
+    "--schema-name=MockSchema",
```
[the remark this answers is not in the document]
calvincestari: It works with both formats, but for consistency we should pick one and use it throughout the documentation.

> Keys start with one or two dashes (- or --), and a user can separate the key and value with an equal sign (=) or a space

From _Options_ in the [Overview](https://apple.github.io/swift-argument-parser/documentation/argumentparser/declaringarguments/#Overview) guide.

Also, the default `SingleValueParsingStrategy` for `@Option` wrapped properties is `.next` - [docs](https://apple.github.io/swift-argument-parser/documentation/argumentparser/singlevalueparsingstrategy/next). So the value directly after the option is interpreted as the value.

Tests/CodegenCLITests/Commands/InitializeTests.swift
calvincestari: This test at line 341 is to validate that the correct custom path is used when creating the initialization file and that the data to be written matches what is expected. Tests specifically for minimalJSON 'correctness' are down from line 478 onwards.

`minimalJSON` is exactly that, the minimal amount of JSON (`String`) to be able to decode an `ApolloCodegenConfiguration` instance and the rest are filled in by the defaults we have in the `ApolloCodegenConfiguration` decoding functions.

The tests are premised on:
* successful decoding proves the JSON is correct and valid (not malformed) otherwise `JSONDecoder().decode` would fail. This would also fail if a required parameter were added to `ApolloCodegenConfiguration` but not added to `minimalJSON`.
* each test is testing a value fed into `minimalJSON` to validate the logic for `supportCocoaPods` and `targetName`.

docs/shared/pods-setup-codegen-panel.mdx
calvincestari: This is in a CocoaPods focused panel, which is why it says to use `other` - it's contextual.

If they added Apollo to their project via CocoaPods the CLI is built with the `COCOAPODS` flag and will therefore always add `cocoapodsCompatibleImportStatements=true` to the initialization file. This happens without them knowing and if they used anything other than `other` the generated code wouldn't build because the `import` statements would be incorrect.

There is nuance and edge cases but the overwhelming majority of users that use CocoaPods should be using `other` as the module type here.

Sources/CodegenCLI/Commands/Initialize.swift
calvincestari: The JSON is correct, otherwise none of the decoding from the return string of `minimalJSON` into an `ApolloCodegenConfiguration` instance would succeed.

calvincestari:
> I think there should be some additions to docs in the getting started guide (and maybe other areas?) to really explain that you need to pick a module type and link to the section of the docs that explains what each of those options really does.
> 
> I think the docs need to clearly explain that this is a decision that affects your project structure, and depending on what you select here, how you include your generated files is going to be significantly different.

Yes, I agree with this. More thorough explanation of the modules types will save us a bunch of issues being created and help users to be less confused about the output they get.

Tests/CodegenCLITests/Commands/InitializeTests.swift
AnthonyMDev: > successful decoding proves the JSON is correct and valid (not malformed) otherwise JSONDecoder().decode would fail.

This is what I was looking for. Didn't see that we were actually decoding the JSON in the tests. Thanks!

AnthonyMDev (approved):

PR #2683, opened 2022-11-21
fix: Add missing `InstallCLI` plugin product

### LINKED ISSUE #2682: `InstallCLI` plugin is missing
The new `InstallCLI` plugin is missing from SPM and Xcode.

Version: `1.0.4`

calvincestari:
Fixes #2682 

The `.plugin` product is missing, effectively 'hiding' the new plugin from SPM and Xcode.
* adds the missing `.plugin` product
* uses a custom path in the target definition so the menu name can get a space between the two words

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,6 +20,7 @@ let package = Package(
     .library(name: "ApolloWebSocket", targets: ["ApolloWebSocket"]),
     .library(name: "ApolloTestSupport", targets: ["ApolloTestSupport"]),
     .executable(name: "apollo-ios-cli", targets: ["apollo-ios-cli"]),
+    .plugin(name: "InstallCLI", targets: ["Install CLI"]),
   ],
   dependencies: [
     .package(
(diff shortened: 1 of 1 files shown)


cwjames19:
@calvincestari After installing v1.0.5 with SPM and Xcode, I still don't see the "Install CLI" option in the menu after I right-click "Apollo" under package dependencies as described in the [CLI installation docs](https://www.apollographql.com/docs/ios/code-generation/codegen-cli#installation).

I've tried selecting different combinations of package products when adding the Apollo package dependency but the result is always the same. I can see the file `Apollo/Plugins/InstallCLI/InstallCLIPluginCommand` is present, though.

calvincestari:
That screenshot in the docs is misleading; you need to right-click on your project, not the dependency. It looks like the dependency in the screenshot because of what the name is. I think that screenshot was taken while working in Apollo iOS and not a sample project. I'll get the screenshot updated.

See below:
Screen Shot 2022-12-14 at 9 58 06 PM

cwjames19:
Thank you!

PR #2692, opened 2022-11-29
cleanup: Remove unnecessary versioning files

### LINKED ISSUE #2689: GitHubAPI exports Star Wars API version variables instead of its own version
## Bug report

It looks like GitHubAPI.h file was copy-pasted from StarWarsAPI.h:

https://github.com/apollographql/apollo-ios/blob/014660b14262df15f0d2c7b7e973d0a53be27b7e/Sources/GitHubAPI/GitHubAPI.h#L3-L7

calvincestari:
Closes #2689 

@AnthonyMDev I can't see any need for these files. Maybe they were part of a previous method of doing versioning but they don't seem to have any use now. Is there anything I'm missing here?


AnthonyMDev (approved):
Yupp, we definitely don't need these. Thanks!

PR #2695, opened 2022-11-30
update: Raise minimum Swift tools version to `5.7`

### LINKED ISSUE #2688: Pod install fails for versions after 1.0.3 due to VersionChecker.swift
## Bug report

Installing Apollo via Cocoapods for versions after 1.0.4+ fails due to an issue with a file `VersionChecker.swift`. I checked the commits and it looks like this file was added 8 days ago before v1.0.4 was released. Installing a previous ver worked fine.

This is the error message:

```
/Users/y/Library/Caches/CocoaPods/Pods/Release/Apollo/1.0.4-7c977/Sources/CodegenCLI/Extensions/VersionChecker.swift:34:29: error: variable binding in a condition requires an initializer
      if let projectRootURL {
                            ^
/Users/y/Library/Caches/CocoaPods/Pods/Release/Apollo/1.0.4-7c977/Sources/CodegenCLI/Extensions/VersionChecker.swift:52:29: error: variable binding in a condition requires an initializer
      if let projectRootURL {
                            ^
make: *** [build-cli-for-cocoapods] Error 1
```

## Versions

- `apollo-ios` SDK version: 1.0.4+
- Xcode version: 13.4
- Swift version: 5.6
- Package manager: Cocoapods

## Steps to reproduce

Install Apollo 1.0.4 or 1.0.5 using Cocoapods

calvincestari:
Closes #2688 

Apollo iOS contains code that will only build with Swift `5.7`, such as the new `if let` shorthand from [SE-0435](https://github.com/apple/swift-evolution/blob/main/proposals/0345-if-let-shorthand.md). It's not a great experience to have the project load only to result a build error if you're using a version of Xcode that doesn't support Swift `5.7`; < `14.0`.

This PR simply raises the required version of Swift to build the project making the user aware earlier of their tool versioning mismatch.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
@@ -16,7 +16,7 @@ struct SwiftPackageManagerModuleTemplate: TemplateRenderer {
     let casedSchemaName = config.schemaName.firstUppercased
 
     return TemplateString("""
-    // swift-tools-version:5.6
+    // swift-tools-version:5.7
 
     import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.6
+// swift-tools-version:5.7
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/SwiftPackageManagerModuleTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/SwiftPackageManagerModuleTemplateTests.swift
@@ -33,7 +33,7 @@ class SwiftPackageManagerModuleTemplateTests: XCTestCase {
     buildSubject()
 
     let expected = """
-    // swift-tools-version:5.6
+    // swift-tools-version:5.7
     """
 
     // when


Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
```diff
@@ -16,7 +16,7 @@ struct SwiftPackageManagerModuleTemplate: TemplateRenderer {
     let casedSchemaName = config.schemaName.firstUppercased
 
     return TemplateString("""
-    // swift-tools-version:5.6
+    // swift-tools-version:5.7
```
calvincestari, line 19: There's nothing in `ApolloAPI` at the moment that would warrant this but there is no reason we couldn't use those new Swift features there.

AnthonyMDev (approved):

corymosiman12:
hmm. This doesn't feel like it should've been part of a patch release IMO. Just got the error that I can no longer build with it...just as an fyi

calvincestari:
That's a fair point @corymosiman12, thanks for the feedback.

It was viewed as a patch release because the current state, at the time, was broken builds on older versions of Xcode. Had this been a minor version release the situation then would have still been to stay on a broken patch release or migrate to the next minor bump.

calvincestari:
In retrospect we maybe should have left the generated package on `5.6`, in `SwiftPackageManagerTemplate.swift`, since the failed builds were in `ApolloCodegenLib` which is not used in production shipping apps. So, my apologies for that.

corymosiman12:
all good, I'm no pro at this either, just wanted to relay my feedback

PR #2715, opened 2022-12-09
fix: Fork PRs failing CocoaPods integration test

calvincestari:
Part of #2563 

* This PR changes the Podfile for the `pod-install-project` test project to build the pod location from CircleCI environment variables based on whether it is a forked PR or not.
* There is also an update to the documentation to inform users that when installing Apollo iOS with CocoaPods from a local development path, the CLI will not be automatically built for them.

There are three PRs to verify the change works as expected in each case:

1.  **PRs on the official repo - `apollographql/apollo-ios`**
* These PRs use a branch as the pod source - [Podfile line 12](https://github.com/apollographql/apollo-ios/compare/main...calvincestari:apollo-ios:fix/fork-pr-pod-test?expand=1#diff-c76ab14169f6e66310ded3bfc77ca59d6dc78eacdb0325898a2c06c72698c26eR12).
* PR #2714 is an example, which results in `pod 'Apollo', :git => 'https://github.com/apollographql/apollo-ios.git', :branch => 'test/ci-env-dump'`.
* CI builds of that PR can be found [here](https://app.circleci.com/pipelines/github/apollographql/apollo-ios?branch=test%2Fci-env-dump).
* Looking at the [CocoaPods Integration Tests](https://app.circleci.com/pipelines/github/apollographql/apollo-ios/4100/workflows/04bd2222-adfb-4dee-93af-a087caeb805a/jobs/32711) job on the last build of that PR, for commit 6e34253, you can see the environment variables that are used in the "Preparing environment variables" step, and the successful "CocoaPods - Install" step.

2. **PRs from forked repos with the same repo name - `calvincestari/apollo-ios`**
* These PRs use a combination of the PR username/reponame and commit hash as the pod source - [Podfile line 10](https://github.com/apollographql/apollo-ios/compare/main...calvincestari:apollo-ios:fix/fork-pr-pod-test?expand=1#diff-c76ab14169f6e66310ded3bfc77ca59d6dc78eacdb0325898a2c06c72698c26eR10).
* PR #2713 is an example, which results in `pod 'Apollo', :git => "https://github.com/calvincestari/apollo-ios.git", :commit => '48f33050a57d2b5cd07a820bc310add2e272ac73'`.
* CI builds of that PR can be found [here](https://app.circleci.com/pipelines/github/apollographql/apollo-ios?branch=pull%2F2713).
[...]

PR #2716, opened 2022-12-09
Fix crash when data contains Null value for a SelectionSet

harry-lush:
# Intended outcome:
Apollo should support null in optional responses from the server (distinct from nil).

# Actual outcome:
The  app would crash with a fatalError on line 55 of DataDict.swift.

```
extension AnySelectionSet {
  @inlinable public init(fieldData: AnyHashable?, variables: GraphQLOperation.Variables?) {
    guard let fieldData = fieldData as? JSONObject else {
      fatalError("\(Self.self) expected data for entity.")
    }
    self.init(data: DataDict(fieldData, variables: variables))
  }

  @inlinable public var _fieldData: AnyHashable { __data._data }
}
```

# How to reproduce the issue
Have a response that has an optional value as a null. Alternatively checkout the commit for the test which precedes the commit for the fix. Running the tests will crash

# Fix
This PR tests for Null value when unwrapping an optional in a selection set.
I don't know the Apollo code base that well so there may well be another way, or better place to fix this.

I've tested that linking directly to this modified version of the SDK fixes the crash in our app.

**Update:** Added an issue here https://github.com/apollographql/apollo-ios/issues/2725 for visibility

### DIFF
--- a/Sources/ApolloAPI/DataDict.swift
+++ b/Sources/ApolloAPI/DataDict.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// A structure that wraps the underlying data dictionary used by `SelectionSet`s.
 public struct DataDict: Hashable {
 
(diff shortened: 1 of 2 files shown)


harry-lush:
I'm pretty sure the failing integration test here is a false negative. I'd be *highly* surprised if it was this code

calvincestari:
Thanks for the PR @harry-lush.

> I'm pretty sure the failing integration test here is a false negative. I'd be highly surprised if it was this code

Correct, I've got PR #2715 up to fix it.

PR #2730, opened 2022-12-14
fix: Use correct schema namespace casing in fragments

### LINKED ISSUE #2705: Cannot find type 'graphQLSchema' in scope.
## Bug report

I followed the new migration guide coming with 1.0.6 and it helped me finally generate all the correct files for our queries and mutations. Thank you for the detailed guide! 

However after dragging the swift package into our Xcode I'm running into some circular dependencies and Xcode errors around not being able to find the schema package in the source files. Are there plans to add more documentation of how to use the new Package integration? 

## Versions

- Xcode version: 14.1
- Swift version: 5
- Package manager: SPM

## Steps to reproduce

generate code with this configuration file. and dragged the package into our Xcode project
```json
{
  "schemaName" : "graphQLSchema",
  "input" : {
    "operationSearchPaths" : [
      ".../Apollo/**/*.graphql"
    ],
    "schemaSearchPaths" : [
      ".../ios/graphQLSchema.graphqls"
    ]
  },
  "output" : {
    "testMocks" : {
      "none" : {
      }
    },
    "schemaTypes" : {
      "path" : ".../Apollo/GraphQLSchema",
      "moduleType" : {
        "swiftPackageManager" : {
        }
      }
    },
    "operations" : {
      "inSchemaModule" : {
      }
    }
  }
}
```
## Further details
Here is the error we are seeing in the generated files: 
Screenshot 2022-12-07 at 9 58 50 AM

calvincestari:
Fixes #2705 

`FragmentTemplate` was a template still using schema name from the IR schema. It now uses schema name from the codegen configuration.

I thought we'd caught all these places in #2586 but it turns out not. So I've now removed schema name from `IR.Schema` so it cannot be had from there anymore and I've removed the IR schema from templates and file generators where it's not used.

_I'd prefer to have a centralized place where `config.schemaName` is first capitalized without having to do it in many different places but I'll handle that in the next round of config changes._

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -267,7 +267,7 @@ public class ApolloCodegen {
     for fragment in compilationResult.fragments {
       try autoreleasepool {
         let irFragment = ir.build(fragment: fragment)
-        try FragmentFileGenerator(irFragment: irFragment, schema: ir.schema, config: config)
+        try FragmentFileGenerator(irFragment: irFragment, config: config)
           .generate(forConfig: config, fileManager: fileManager)
       }
     }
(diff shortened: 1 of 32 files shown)


AnthonyMDev (approved):
This is great clean up! Thanks @calvincestari

PR #2741, opened 2022-12-19
Codegen Documentation Sample Code Fix

philfi:
Resolves issue with misplaced commas


calvincestari (approved):
Thanks for the contribution @philfi!

PR #2745, opened 2022-12-23
fix: GraphQLEnum value camelCase conversion strategy

### LINKED ISSUE #2640: Numeric enums with underscore at beginning get invalid code generated
## Bug report

We recently started migration to new apollo-ios-cli code generation tool and something that was working before, with old code generator, is no longer working.

## Versions

- Xcode version: 14.1
- Swift version: 5.7
- Package manager: Cocoapods

## Steps to reproduce

Following enum in schema:
```
enum NumberOfStoriesEnum {
  _1
  _15
  _2
  _25
  _3
}
```

Generated code:
```
public enum NumberOfStoriesEnum: String, EnumType {
  case 1 = "_1"
  case 15 = "_15"
  case 2 = "_2"
  case 25 = "_25"
  case 3 = "_3"
}
```

As you can see, the case is just a number - which is not valid Swift code. In old code generator, the generated cases still had the underscore as a first character.

Note: If I change conversionStrategies.enumCases to .none - then it works, but it should work correctly with .camelCase as well, since it's mostly preferred strategy.

calvincestari:
This PR refactors the `.camelCase` conversation strategy implementation
* Fixes #2640 
* Fixes #2745 
* Uses the [`_convertFromSnakeCase`](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/JSONDecoder.swift#L103) implementation from `JSONDecoder` in `Foundation` with modifications. _This [question](https://forums.swift.org/t/support-for-string-transformation-from-camelcase-to-snake-case-and-vice-versa/38045) in the Swift forums makes me hopeful that one day we can remove this code and use a Swift standard library implementation, but we're not there yet._
* Adds Swift enum value test cases. _I checked the original PR (#2478) and oddly there weren't any in there, I'm not sure why._

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLEnumValue+Rendered.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLEnumValue+Rendered.swift
@@ -23,64 +23,23 @@ extension GraphQLEnumValue.Name {
     }
   }
 
-  /// Convert to `camelCase` from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
+  /// Convert to `camelCase` from a number of different `snake_case` variants.
   ///
-  /// 1. Returns a first lowercased string
-  /// 1. Capitalizes the word starting after each `_`
-  /// 2. Removes all inner `_`
-  /// 3. Preserves leading and trailing `_`
-  /// 4. Converts an all uppercase string into all lowercase
+  /// All inner `_` characters will be removed, each 'word' will be capitalized, returning a final
+  /// firstLowercased string while preserving original leading and trailing `_` characters.
   private func convertToCamelCase(_ value: String) -> String {
-    // The source for this function is from the JSONDecoder implementation in Swift Foundation,
-    // licensed under Apache License v2.0 with Runtime Library Exception. Modifications were made
-    // to the return when no underscore characters are found.
-    //
-    // See https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/JSONDecoder.swift
-
-    // Find the first non-underscore character
-    guard let firstNonUnderscore = value.firstIndex(where: { $0 != "_" }) else {
-      return value
-    }
-
-    // Find the last non-underscore character
-    var lastNonUnderscore = value.index(before: value.endIndex)
-    while lastNonUnderscore > firstNonUnderscore && value[lastNonUnderscore] == "_" {
-      value.formIndex(before: &lastNonUnderscore)
-    }
-
-    // Cater for leading and trailing underscore characters
-    let valueRange = firstNonUnderscore...lastNonUnderscore
-    let leadingUnderscoreRange = value.startIndex..<firstNonUnderscore
-    let trailingUnderscoreRange = value.index(after: lastNonUnderscore)..<value.endIndex
-
-    // Split inner string into 'words'
-    let components = value[valueRange].split(separator: "_")
-    let joinedString: String
-    if components.count == 1 {
+    guard value.firstIndex(of: "_") != nil else {
       if value.firstIndex(where: { $0.isLowercase }) != nil {
-        joinedString = String(value[valueRange]).firstLowercased
+        return value.firstLowercased
       } else {
-        joinedString = String(value[valueRange]).lowercased()
+        return value.lowercased()
       }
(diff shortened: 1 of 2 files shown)


calvincestari:
I'm putting this back into draft to fix #2749.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLEnumValue+Rendered.swift
```diff
@@ -23,20 +23,64 @@ extension GraphQLEnumValue.Name {
...
+  /// 1. Returns a first lowercased string
+  /// 1. Capitalizes the word starting after each `_`
+  /// 2. Removes all inner `_`
+  /// 3. Preserves leading and trailing `_`
+  /// 4. Converts an all uppercase string into all lowercase
   private func convertToCamelCase(_ value: String) -> String {
-    if value.allSatisfy({ $0.isUppercase }) {
-      // For `UPPERCASE`. e.g) UPPER -> upper, STARWARS -> starwards
-      return value.lowercased()
+    // The source for this function is from the JSONDecoder implementation in Swift Foundation,
```
AnthonyMDev, line 34: @bignimbus @hwillson Is this in-line comment enough to be in compliance with the license? This is coming from Apple's open source swift foundation library. With modifications for our specific needed behaviors.

LGTM, just want to make sure this is the correct way to do this.

AnthonyMDev (approved):
Thanks so much for nailing this @calvincestari! Let's just make sure the license is okay and then move this forward!

Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLEnumValue+Rendered.swift
calvincestari: I took another look at the license requirements and I'm now doubtful that statement is enough. Reading [this](https://fossa.com/blog/open-source-licenses-101-apache-license-2-0/#:~:text=Anyone%20who%20uses%20open%20source,made%20to%20the%20original%20code) we'd need to include a copy of the original copyright notice (not a problem) and a copy of the license itself (I don't want to add another license to the repo).

I'll look at rewriting the function tomorrow from scratch and we can get avoid this.

calvincestari: This is done - https://github.com/apollographql/apollo-ios/pull/2745/commits/e6bc8a6bdf722da612721b975da7d720d9cbf6e0

calvincestari: @AnthonyMDev you can take another pass at this, it's done - https://github.com/apollographql/apollo-ios/commit/e6bc8a6bdf722da612721b975da7d720d9cbf6e0.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLEnumValue+Rendered.swift
```diff
@@ -23,20 +23,23 @@ extension GraphQLEnumValue.Name {
...
-      // For `snake_case`. e.g) snake_case -> snakeCase, UPPER_SNAKE_CASE -> upperSnakeCase
-      return value.split(separator: "_").enumerated().map { $0.offset == 0 ? $0.element.lowercased() : $0.element.capitalized }.joined()
-    }
-    if let firstChar = value.first, firstChar.isUppercase {
-      // For `UpperCamelCase`. e.g) UpperCamelCase -> upperCamelCase
-      return [firstChar.lowercased(), String(value.suffix(from: value.index(value.startIndex, offsetBy: 1)))].joined()
-    }
-    return value
+
+    return value.components(separatedBy: "_")
```
calvincestari, line 39: It might look odd to be separating by `_` and then converting empty elements back to `_` but that's the trick to preserving leading/trailing `_` without having to find the first/last indexes around them.

calvincestari: Components also preserves empty `_` positions vs. split which would just give us the inner 'words'.

AnthonyMDev (approved):
Awesome job on this!

PR #2754, opened 2023-01-05
Fix naming collisions for Selection and ParentType

### LINKED ISSUE #2708: Naming collision with Selection type from apollo
## Bug report

In our schema we have an enum defined as `Selection` and in the generated files the `Selection` type for `__selections` is not namespaced with `Apollo.` This is causing a name conflict and our code does not compile. There is no workaround for this as I cannot request a server change for something like this.

## Versions

- Xcode version: 14.1
- Swift version: 5.7
- Package manager: Cocoapods 1.11.3

## Steps to reproduce

To reproduce this you will have to use a scheme that defines `Selection` as an enum type

## Further details

Nothing more to add here.
Screen Shot 2022-12-07 at 12 47 33 PM

AnthonyMDev:
Fixes #2708


calvincestari (approved):
LGTM! 🚀

PR #2755, opened 2023-01-05
Adds disallowed names to schema namespace validation

### LINKED ISSUE #2664: Namespace collision when using Schema for schemaName
## Bug report

When you pass `schema` for `schemaName` to generate types using the `swiftPackageManager` , the generated code will not compile because the generated module collides with an Apollo `Schema` type. It would be great if we document invalid `schemaName` values somewhere or use explicit namespacing in Apollo to avoid this.

## Versions

- Xcode version: 14.0.1
- Swift version: 5.7
- Package manager: SPM

## Steps to reproduce

1. Generate code using the `swiftPackageManager` module type using `schema` for `schemaName`
2. Add generated Swift Package to target
3. Attempt to build project
4. Observe error

calvincestari:
Fixes #2664 

* Repurposes the `schemaNameConflict` error to be more a more general purpose type collision error
* Adds logic to validate the given schema namespace name against a new list of disallowed names

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -38,8 +38,8 @@ public class ApolloCodegen {
           """
       case let .schemaNameConflict(name):
         return """
-          Schema name \(name) conflicts with name of a type in your GraphQL schema. Please \
-          choose a different schema name. Suggestions: \(name)Schema, \(name)GraphQL, \(name)API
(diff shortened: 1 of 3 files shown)


AnthonyMDev (approved):
Perfect! Thank you!

AnthonyMDev:
Are there any other obvious Schema names we want to disallow? I don't think any others should be problematic off the top of my head?

Actually `Fragments` may or may not be? Kinda an odd name to use, but I think it would cause the same compilation error possibly?

PR #2757, opened 2023-01-06
Fix unqualified scalar collision

### LINKED ISSUE #2691: Naming collision with fragments and scalars.
## Bug report

In our schema, we have a scalar named `Text` which is mapped to String. 

We also have a fragment that looks something like 
```
fragment event on SomeEvent { 
     items {
        title
        text { 
            ... on Type1 {
                ...field1
            } 
            ... on Type2 {
                ..field2
            }
        } 
     }
}
```
where `text` is a field with a Union Type of Type1 & Type2 and `title` is of the scalar type `Text`. 

In the generated code, we're getting both `title` and `text` as type `Text`. And this is because there is a type added on the fragment of `Text` for the type of object that the field `text` is. If you make your `FileOutput.operations` be `absolute`, the scalar is namespaced as `SchemaName.Text` and then it works correctly, however that is not an option for us as it's the incorrect setup for us. We are able to work around this issue by renaming the field from `text` to something else in the fragment. 

## Versions

- Xcode version: 14.1
- Swift version:  5.7
- Package manager: cocoapods 1.11.3

## Steps to reproduce

Create a schema with a type `Text`. Create a union type. Use the union type as a field named `text`.  create a fragment that uses both the scalar type and a the field named `text`. Generate the code using SchemaTypesFileOutput.moduleType.embeddedInTarget, and FileOutput.operations.inSchemaModule

## Further details

As mentioned, you can workaround this by renaming the field of the `Union` type.

AnthonyMDev:
Fixes #2691 

Always qualify names of shared schema types referenced as the types of fields in selections sets. This prevents collisions with names of other generated selection sets for entity type fields whose names are the same as a referenced schema type.

This namespaces enums, custom scalars, and input objects.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
@@ -44,32 +44,37 @@ extension GraphQLType {
     replacingNamedTypeWith newTypeName: String? = nil,
     config: ApolloCodegenConfiguration
   ) -> String {
-
-    lazy var schemaModuleName: String = {
-      !config.output.operations.isInModule ? "\(config.schemaName.firstUppercased)." : ""
-    }()
-
     switch self {
-    case let .entity(type as GraphQLNamedType):
-      let typeName = newTypeName ?? type.swiftName.firstUppercased
+    case
+        .entity(let type as GraphQLNamedType),
+        .scalar(let type as GraphQLNamedType),
+        .enum(let type as GraphQLNamedType),
+        .inputObject(let type as GraphQLNamedType):
+
+      let typeName = type.qualifiedRootTypeName(
+        in: .selectionSetField(),
+        replacingNamedTypeWith: newTypeName,
+        config: config
+      )
+
       return containedInNonNull ? typeName : "\(typeName)?"
 
-    case let .inputObject(type as GraphQLNamedType):
-      let typeName = newTypeName ?? type.swiftName.firstUppercased
-      return TemplateString("\(schemaModuleName)\(typeName)\(if: !containedInNonNull, "?")").description
-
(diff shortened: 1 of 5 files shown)


Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
```diff
@@ -173,4 +177,47 @@ extension GraphQLNamedType {
 
     return swiftName
   }
+
+  fileprivate func qualifiedRootTypeName(
+    in context: GraphQLType.RenderContext,
+    replacingNamedTypeWith newTypeName: String? = nil,
+    config: ApolloCodegenConfiguration
+  ) -> String {
+    lazy var typeName = { newTypeName ?? self.swiftName.firstUppercased }()
```
calvincestari, line 186: _This doesn't need to change, my question is more out of curiosity._

 Is there a benefit to `typeName` and `schemaModuleName` being defined as `lazy`? The return references both of them; it doesn't look like there is a logic path where they wouldn't be used.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
```diff
@@ -173,4 +177,47 @@ extension GraphQLNamedType {
...
+      switch self {
+      case is GraphQLCompositeType:
+        return ""
+
+      case let scalar as GraphQLScalarType where scalar.isSwiftType:
+        return ""
+
+      default:
+        switch context {
+        case .testMockField, .inputValue:
```
calvincestari, line 198: `renderedAsTestMockField` doesn't look like it flows through this new logic - should it?

calvincestari (approved):
Looks good - thanks. Just one question about the path for test mock rendering.

PR #2760, opened 2023-01-09
fix: Schema/Target/Module name with spaces breaks generated code

### LINKED ISSUE #2653: Schema/Target/Module name with spaces in it breaks generated code
## Bug report

When a Xcode project name contains a space character, relative module name contains an `_` character instead.

In Test mocks files (Xxx+Mock.graphql.swift), for a project named `First Project`, imports are like:
```
// @generated

import Apollo
import First Project

public class Xxx: MockObject {
```
but it should be
```
// @generated

import Apollo
import First_Project

public class Xxx: MockObject {
```

## Versions

- Xcode version: 13.4.1
- Swift version: 5.6.1
- CocoaPods: 1.11.3

## Steps to reproduce

Run `apollo-ios-cli generate` onto a Xcode project whose name contains a space.

calvincestari:
Fixes #2653 
Related to #2664 

This PR refactors the flow of configuration validation logic.
* Validation logic that is not specific to only the CLI command line parameters is now all in `ApolloCodegen`.
* There is a public function `_validation` that performs config-only validation, not dependent on the schema nor operations, that can be called from the CLI.
* There is also a fix to the install CLI plugin where it would have failed if the path had spaces in it.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -130,16 +130,16 @@ public class ApolloCodegen {
     }
   }
 
-
-  /// Performs validation against deterministic errors that will cause code generation to fail.
+  /// Validates the configuration against deterministic errors that will cause code generation to
+  /// fail. This validation step does not take into account schema and operation specific types, it
+  /// is only a static analysis of the configuration.
   ///
   /// - Parameter config: Code generation configuration settings.
-  public static func validate(config: ApolloCodegenConfiguration) throws {
+  public static func _validate(config: ApolloCodegenConfiguration) throws {
     try validate(context: ConfigurationContext(config: config))
   }
 
-  /// Performs validation against deterministic errors that will cause code generation to fail.
-  static func validate(context: ConfigurationContext) throws {
+  static private func validate(context: ConfigurationContext) throws {
     guard
       !SwiftKeywords.DisallowedSchemaNamespaceNames.contains(context.schemaName.lowercased())
     else {
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -79,14 +79,14 @@ public class ApolloCodegen {
       rootURL: rootURL
     )
 
-    try validate(context: configContext)
+    try validate(configContext)
 
     let compilationResult = try compileGraphQLResult(
       configContext,
       experimentalFeatures: configuration.experimentalFeatures
     )
 
-    try validate(context: configContext, compilationResult: compilationResult)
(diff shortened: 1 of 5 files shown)


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -130,26 +133,50 @@ public class ApolloCodegen {
...
-  /// Performs validation against deterministic errors that will cause code generation to fail.
-  static func validate(config: ConfigurationContext) throws {
-    if case .swiftPackage = config.output.testMocks,
-        config.output.schemaTypes.moduleType != .swiftPackageManager {
+  /// Validates the configuration against deterministic errors that will cause code generation to
+  /// fail. This validation step does not take into account schema and operation specific types, it
+  /// is only a static analysis of the configuration.
+  ///
+  /// - Parameter config: Code generation configuration settings.
+  public static func _validate(config: ApolloCodegenConfiguration) throws {
```
calvincestari, line 141: The public method prefixed with `_` to indicate that while it's public it is still considered private.

Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -86,11 +82,11 @@ public struct Initialize: ParsableCommand {
...
     let encoded = try ApolloCodegenConfiguration
-      .minimalJSON(
-        schemaName: schemaName,
-        moduleType: moduleType,
-        targetName: targetName
-      ).asData()
+      .minimalJSON(schemaName: schemaName, moduleType: moduleType, targetName: targetName)
+      .asData()
+
+    let decoded = try JSONDecoder().decode(ApolloCodegenConfiguration.self, from: encoded)
```
calvincestari, line 88: I'm not thrilled about decoding the config data but it's a small price to pay considering this is an infrequent workflow, not in the critical path of performance and IMO it's far better than having one single public method that only validates the schema name string.

AnthonyMDev (approved):
Minor suggestion, but looks good to me!

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -76,14 +79,14 @@ public class ApolloCodegen {
       rootURL: rootURL
     )
 
-    try validate(config: configContext)
+    try validate(context: configContext)
```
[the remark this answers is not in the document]
calvincestari: How about this? - https://github.com/apollographql/apollo-ios/pull/2760/commits/a492d548b002f33c62af0ec2ab30633e45b5d0fb
* Takes the suggestion of the nameless parameter
* Keeps `compilationResult` as a second param but using the label `with` because it's still validation of the configuration but dependent on the compilation results rather than validating the compilation results.

Issue #2560, opened 2022-10-10
Implement codegen configuration for field merging


pm-dev:
## Bug report

My team was excited to upgrade to the 1.0 release (thanks for all the hard work), unfortunately we've been unable to get codegen to work. Pre-v1.0 we used `ApolloCLI` from `ApolloCodegenLib` to shell out to the js codegen tool and it would take ~5 seconds to complete. After updating to 1.0 and swift-native codegen, we're able to compile and run, but codegen never finishes. It's able to find our schema and .graphql queries, and about half our queries (20 or so) are generated immediately, but then codegen starts to take minutes per query and the memory the process is using grows to multiple GBs.

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: 1.0
- Xcode version: 14.0
- Swift version: 5.7
- Package manager: SPM

## Steps to reproduce

I would gladly provide our schema and queries if there was a confidential way to share those files so we could debug this together. 

I've tried playing around with the configuration a bit to see if any of these options make a difference, but no luck. Here's an example of what it looks like:
```swift
    let configuration = ApolloCodegenConfiguration(
        schemaName: "GraphQLTypes",
        input: ApolloCodegenConfiguration.FileInput(
            schemaPath: schemaFileURL.relativePath,
            operationSearchPaths: [
                projectDirectoryURL.relativePath + "/iOS/**/*.graphql",
                projectDirectoryURL.relativePath + "/iOS/**/*.graphql"
            ]
        ),
        output: ApolloCodegenConfiguration.FileOutput(
            schemaTypes: ApolloCodegenConfiguration.SchemaTypesFileOutput(
                path: projectDirectoryURL.relativePath + "/iOS/Libraries/GraphQL/GraphQLTypes/Generated Types",
                moduleType: .swiftPackageManager
            ),
            operations: .inSchemaModule,
            testMocks: .none,
            operationIdentifiersPath: operationsURL.relativePath
        ),
        options: ApolloCodegenConfiguration.OutputOptions(
            queryStringLiteralFormat: .multiline,
            deprecatedEnumCases: .include,
            schemaDocumentation: .include,
            apqs: .persistedOperationsOnly,
            cocoapodsCompatibleImportStatements: false,
            warningsOnDeprecatedUsage: .include,
            pruneGeneratedFiles: true
        )
    )
    try ApolloCodegen.build(with: configuration, withRootURL: workspaceRootURL)
```

[...]

PR #2761, opened 2023-01-09
January 2023 roadmap update

bignimbus:
Happy New Year!

Most of this was copied or adapted from https://github.com/apollographql/apollo-ios/issues/2560#issuecomment-1373317272.


calvincestari (approved):
I'm happy adding this, good transparency. @AnthonyMDev - anything changes you had in mind since first writing this list?

AnthonyMDev (commented):
We should also add something stating we will be working in parallel on `@defer` support!

calvincestari:
While we're editing this file should we remove the "Future" section? It's so far off I don't think it's even worth talking about at the moment and there's many other things that should be in that list too if we want to keep it around.

calvincestari (approved):
Good to go!

PR #2762, opened 2023-01-09
fix: Typename collisions in generated test mocks

### LINKED ISSUE #2719: Naming collision with DocumentType in generated mock code
## Bug report

This is similar to https://github.com/apollographql/apollo-ios/issues/2708 but it only concerns mocks, so it seemed worth treating as a separate issue.

I have a enum type `DocumentType` that conflicts with `ApolloAPI.DocumentType` in generated mock code.
Would it be possible to prefix generated types in mocks with the schemaName they're living in to resolve ambiguity?

## Versions

- Xcode version: 14.1
- Swift version: 5.7
- Package manager: Swift Package Manager

## Steps to reproduce

To reproduce this you will have to use a scheme that defines `DocumentType` as an enum type.

## Further details

Screen Shot 2022-12-09 at 6 02 50 PM

https://github.com/apollographql/apollo-ios/pull/2679 seemed to fix this issue for the main generated code, but I'm still experiencing this issue for generated mocks.
If I manually specify the type as coming from my generated package - `GQLTypes`, I can get the mocks compiling successfully:
```swift
public struct MockFields {
    @Field<GraphQLEnum<GQLTypes.DocumentType>>("documentType") public var documentType
}
```

calvincestari:
Fixes #2719 

* Refactors test mock field rendering to use the functions from #2757 
* All schema referenced types within test mocks now use a fully qualified namespace

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLType+Rendered.swift
@@ -59,46 +59,12 @@ extension GraphQLType {
     replacingNamedTypeWith newTypeName: String? = nil,
     config: ApolloCodegenConfiguration
   ) -> String {
-
-    lazy var schemaModuleName: String = {
-      !config.output.schemaTypes.isInModule ? "\(config.schemaName.firstUppercased)." : ""
-    }()
-
-    switch self {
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):
Ah, didn’t realize this was needed for another issue! Looks good! Glad we could reuse the new logic I added here!

Issue #2759, opened 2023-01-09
Subscript GraphQL variable from dictionary crash when Swift modifier used as key.


SzymonMatysik:
### Summary

Hey there, great thanks for hard work put into new release!

Recently I've started to migrate from 0.53 to 1.0.5 and have to say - it's not a very easy task to achieve. Now I've stomped on issue that nobody else mentioned before (based on my research). 

Everything was working fine before the migration - I think it's issue related to the codegen feature.
When I got in schema variable in fragment named after Swift's special words `default`, `public`, `private` or else then in generated code it's ` __data["'public'"] }` (apostrophes on each side to escape the special word)
It leads to a problem in getting the value resulting a crash as in subscript.
part of generated file:
```
    public var `public`: GraphQLNullable<Bool> {
      get { __data["`public`"] }
      set { __data["`public`"] = newValue }
    }
```

Maybe we should stripe the apostrophes which are clearly from the Swift requirements not from a schema.

### Version

1.0.5

### Steps to reproduce the behavior

- Make schema with variable in any model named after Swift's special word for example `public: Bool`.
- Generate schema.
- Check string representation of variable in Swift generated file.

### Logs
```
Could not cast value of type 'Swift.Optional<ApolloAPI.GraphQLOperationVariableValue>' (0x15460b190) to 'ApolloAPI.GraphQLNullable<Swift.Bool>' (0x1510a3608).
(lldb) 
```

### Anything else?
As a workaround I've applied direct dictionary lookup for `public` field as follows:
```let lhsPublic: GraphQLNullable<Bool> = lhs.__data["public"]```

calvincestari:
Thanks for the bug report @SzymonMatysik. We'll get this fixed in a patch release; we're planning on releasing `1.0.6` today and `1.0.7` by the end of the week. It'll be in one of those.

calvincestari:
Note that this bug is only apparent on input types, selection sets do not have this issue.

calvincestari:
Adding this to "planned-next" because fixing #2753 will need this to be fixed too before it's all functional.

calvincestari:
@SzymonMatysik this is fixed, available on `main`, and will go out in the next patch release.

PR #2769, opened 2023-01-11
fix: fields labeled `_`

### LINKED ISSUE #2753: Unnamed fields in schema results in broken generated Swift code
### Summary

In a schema with unnamed fields such as:
```
input MyInput {
  _: Boolean
}
```
the resulting generated code is broken as:
```
__data = InputDict([
  "_": _
])

public var _: GraphQLNullable<Bool> {
  get { __data["_"] }
  set { __data["_"] = newValue }
}
```

### Version

1.0.5

1. Create a schema with an unnamed field.
2. Run CLI to generate code.
3. Observe the generated Swift code does not compile.

### Logs

_No response_

### Anything else?

_No response_

calvincestari:
Fixes #2753 

This PR brings back the `_` character as an allowed field name. Removing it from `DisallowedFieldNames` and adding it to `FieldAccessorNamesToEscape` causes it to be escaped as expected in all property and initializer definitions.

_This functionality is dependent on #2759 being fixed too._

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/String+SwiftNameEscaping.swift
@@ -38,7 +38,7 @@ extension String {
 enum SwiftKeywords {
 
   static let DisallowedFieldNames: Set<String> = [
-    "__data", "fragments", "_"
+    "__data", "fragments"
   ]
 
   static let DisallowedInputParameterNames: Set<String> = [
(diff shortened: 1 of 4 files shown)


calvincestari:
@AnthonyMDev the input objects and mocks work as expected but the initializer provided by autocomplete lacks backticks around the `_`; I'm not aware of any way to 'fix' that for autocompletion.

Let's talk about it tomorrow. I'm not sure how I feel about it since it's not a great user experience but I feel it's better than simply disallowing `_` as a field name.

AnthonyMDev (approved):
LGTM!

PR #2772, opened 2023-01-11
fix: Additional reserved keyword

### LINKED ISSUE #2765: Couldn't build the generated graphql project when using some reserved words in the graphql file
### Summary

Hello there, Thank you for hard work in the new major version of apollo-ios 
But I have an issue with the new code generation.

I couldn't build the generated graphql project that there are some reserved words in the graphql files because the variable name is one of the reserved words in swift language.

For example If I create a schema with the swift reserved words and use it in my .graphql file like this

**Schema File**
```graphql
interface Animal {
  id: ID!
  for: String!
  if: String!
}
```
**GraphQL File**
```graphql
query Animal {
  animal {
    id
    for
    if
  }
}
```

After I run `./apollo-ios-cli generate` and build my xcode project there will be an error in my generated graphql project that shown `Keyword 'for' cannot be used as an identifier here`.

In `AnimalQuery.graphql.swift` You will see in the line `public var for: String { __data["for"] }`  that we couldn't use `for` as a variable name in swift. But for `if` variable it works correctly because it is escaped by the code generation.

**AnimalQuery.graphql.swift**
```swift
public struct Animal: TestApolloAlphaGraphQLGenerated.SelectionSet {
  public let __data: DataDict
  public init(data: DataDict) { __data = data }

  public static var __parentType: ParentType { TestApolloAlphaGraphQLGenerated.Interfaces.Animal }
  public static var __selections: [Selection] { [
    .field("id", ID.self),
    .field("if", String.self),
    .field("for", String.self),
  ] }

  public var id: ID { __data["id"] }
  public var `if`: String { __data["if"] }
  public var for: String { __data["for"] } // There is a problem from this line
}
```

### Version

v1.0.5

1. Create A new schema with some reserved words
2. Create A query file with reserved word fields
3. Run cli to generate .swift files
4. Build the project
5. There will be an error that we can't use a reserved word as a variable name

### Logs

`Keyword 'for' cannot be used as an identifier here`

### Anything else?

I have looked into the reserved words list in the apollo-tooling project and the new one in apollo-ios. I found that list of the reserved words are not the same (there is no `for` keyword in the new code generation list too). I think if the reserved words list is the same, it should works correctly.
Ref: 
[Reserved words in the apollo-tooling project](https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-codegen-swift/src/language.ts)
[Reserved words in the apollo-ios project (v1.0.5)](https://github.com/apollographql/apollo-ios/blob/main/Sources/ApolloCodegenLib/Templates/RenderingHelpers/String%2BSwiftNameEscaping.swift)

The workaround for this problem if I use the alias name for the reserved word variable. it still works really fine.
For example: **Aninal.graphql**
```graphql
query Animal {
  animal {
    id
    if
    f: for # It works correctly if we use the alias instead of the for keyword
  }
}

```

calvincestari:
Fixes #2765 

This PR simply adds `for` to the list of reserved keywords to escape in backticks.

_I looked through the [Apollo Tooling list](https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-codegen-swift/src/language.ts#L43) and both `open` and `for` were missing from 1.0 but `open` is not an issue in the generated code._

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
@@ -129,57 +129,7 @@ class EnumTemplateTests: XCTestCase {
 
         // Escaped keywords
         ("associatedtype", nil, nil),
-        ("class", nil, nil),
-        ("deinit", nil, nil),
-        ("enum", nil, nil),
-        ("extension", nil, nil),
-        ("fileprivate", nil, nil),
-        ("func", nil, nil),
-        ("import", nil, nil),
-        ("init", nil, nil),
-        ("inout", nil, nil),
-        ("internal", nil, nil),
-        ("let", nil, nil),
-        ("operator", nil, nil),
-        ("private", nil, nil),
-        ("precedencegroup", nil, nil),
-        ("protocol", nil, nil),
         ("Protocol", nil, nil),
-        ("public", nil, nil),
-        ("rethrows", nil, nil),
-        ("static", nil, nil),
-        ("struct", nil, nil),
-        ("subscript", nil, nil),
-        ("typealias", nil, nil),
-        ("var", nil, nil),
-        ("break", nil, nil),
-        ("case", nil, nil),
-        ("catch", nil, nil),
-        ("continue", nil, nil),
-        ("default", nil, nil),
-        ("defer", nil, nil),
-        ("do", nil, nil),
-        ("else", nil, nil),
-        ("fallthrough", nil, nil),
-        ("guard", nil, nil),
-        ("if", nil, nil),
-        ("in", nil, nil),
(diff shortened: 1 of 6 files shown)


Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
```diff
@@ -129,57 +129,7 @@ class EnumTemplateTests: XCTestCase {
 
         // Escaped keywords
         ("associatedtype", nil, nil),
-        ("class", nil, nil),
```
calvincestari, line 132: This is test cleanup - I'm not sure why we need all the escaped keywords in this test so I left two with different casing that require backticks.

AnthonyMDev (approved):
Thanks for catching that!

PR #2773, opened 2023-01-12
fix: Removes backticks from subscript keys

### LINKED ISSUE #2759: Subscript GraphQL variable from dictionary crash when Swift modifier used as key.

calvincestari:
Fixes #2759 

Removes the backticks from subscript keys of input objects.

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
@@ -109,8 +109,8 @@ struct InputObjectTemplate: TemplateRenderer {
     \(documentation: field.documentation, config: config)
     \(deprecationReason: field.deprecationReason, config: config)
     public var \(field.name.asInputParameterName): \(field.renderInputValueType(config: config.config)) {
-      get { __data["\(field.name.asInputParameterName)"] }
-      set { __data["\(field.name.asInputParameterName)"] = newValue }
+      get { __data["\(field.name.firstLowercased)"] }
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):
Good catch! Thanks

PR #2778, opened 2023-01-13
Fix selection set value serialization when writing to cache

### LINKED ISSUE #2775: LocalCacheMutation with an enum field fails with `Invalid type in JSON write`
### Summary

executing a LocalCacheMutation with an enum field causes the the `ReadWriteTransaction` to fail with an:
* `Invalid type in JSON write` for `SQLiteNormalizedCache`
* `ApolloAPI.JSONDecodingError.couldNotConvert(value: AnyHashable(ApolloAPI.GraphQLEnum<SomeEnum>.case(someCase)), to: Swift.String))` for `InMemoryNormalizedCache`

The issue seems to spawn from the fact that on cache read, the enum gets created via it's `rawValue` accurately, but on write, in the case of the `SQLiteNormalizedCache` the `JSONSerialization` fails to serialize the data, as it receives an `Enum` type to serialize instead of the raw value of the corresponding enum case.

Adding an  extra case to the `SQLiteSerialization.serialize(fieldValue:)` seems to fix this issue for the `SQLiteNormalizedCache`:

```swift
private static func serialize(fieldValue: Record.Value) throws -> Any {
    switch fieldValue {
    ...
    case let value as JSONEncodable:
        return value._jsonValue
    default:
      ...
    }
  }
```

But I'm not too sure the consequences of this particular change, and it also doesn't fix it for the inMemoryCache either.

### Version

apollo-ios SDK version: 1.0.5

I created a failing unit test in the `ReadWriteFromStoreTests` as follows:

**For simplicity, I did a `noop` in the actual `ReadWriteTransaction` so as to make the issue in question clearer**

```swift
func test_updateCacheMutation_enumFields_identityUpdates() throws {
    // given
    struct GivenSelectionSet: MockMutableRootSelectionSet {
      public var __data: DataDict = DataDict([:], variables: nil)
      init(data: DataDict) { __data = data }
      
      static var __selections: [Selection] { [
        .field("hero", Hero.self)
      ]}
      
      var hero: Hero {
        get { __data["hero"] }
        set { __data["hero"] = newValue }
      }
      
      enum HeroType: String, EnumType {
        case droid
      }
      
      struct Hero: MockMutableRootSelectionSet {
        public var __data: DataDict = DataDict([:], variables: nil)
        init(data: DataDict) { __data = data }
        
        static var __selections: [Selection] { [
          .field("type", GraphQLEnum<HeroType>.self)
        ]}
        
        var type: GraphQLEnum<HeroType> {
          get { __data["type"] }
          set { __data["type"] = newValue }
        }
      }
    }
    
    let cacheMutation = MockLocalCacheMutation<GivenSelectionSet>()
    
    mergeRecordsIntoCache([
      "QUERY_ROOT": ["hero": CacheReference("QUERY_ROOT.hero")],
      "QUERY_ROOT.hero": ["__typename": "Droid", "type": "droid"]
    ])
    
    runActivity("update mutation") { _ in
      let updateCompletedExpectation = expectation(description: "Update completed")
      
      store.withinReadWriteTransaction({ transaction in
        try transaction.update(cacheMutation) { data in
          // noop
        }
      }, completion: { result in
        defer { updateCompletedExpectation.fulfill() }
        XCTAssertSuccessResult(result)
      })
      
      self.wait(for: [updateCompletedExpectation], timeout: Self.defaultWaitTimeout)
    }
    
    let query = MockQuery<GivenSelectionSet>()
    
    loadFromStore(operation: query) { result in
      try XCTAssertSuccessResult(result) { graphQLResult in
        XCTAssertEqual(graphQLResult.source, .cache)
        XCTAssertNil(graphQLResult.errors)
        
        let data = try XCTUnwrap(graphQLResult.data)
        XCTAssertEqual(data.hero.type, .case(.droid))
      }
    }
  }
```

AnthonyMDev:
Fixes #2775

When writing a SelectionSets data back into the cache, custom scalars need to re-encoded into their `_jsonValue`. Since we don't want to have to do this when writing network response data (which is already properly serialized JSON) we created a new normalizer and a factory to return the proper normalizer based on the data source.

### DIFF
--- a/Sources/Apollo/GraphQLResultNormalizer.swift
+++ b/Sources/Apollo/GraphQLResultNormalizer.swift
@@ -3,35 +3,53 @@ import Foundation
 import ApolloAPI
 #endif
 
-final class GraphQLResultNormalizer: GraphQLResultAccumulator {
+struct ResultNormalizerFactory {
+  private init() {}
+
+  static func selectionSetDataNormalizer() -> SelectionSetDataResultNormalizer {
+    SelectionSetDataResultNormalizer()
+  }
+
+  static func networkResponseDataNormalizer() -> RawJSONResultNormalizer {
+    RawJSONResultNormalizer()
+  }
+}
+
+class BaseGraphQLResultNormalizer: GraphQLResultAccumulator {
   private var records: RecordSet = [:]
 
-  func accept(scalar: JSONValue, info: FieldExecutionInfo) -> JSONValue? {
+  fileprivate init() {}
+
+  final func accept(scalar: JSONValue, info: FieldExecutionInfo) -> JSONValue? {
     return scalar
   }
 
-  func acceptNullValue(info: FieldExecutionInfo) -> JSONValue? {
+  func accept(customScalar: JSONValue, info: FieldExecutionInfo) -> JSONValue? {
+    return customScalar
+  }
+
+  final func acceptNullValue(info: FieldExecutionInfo) -> JSONValue? {
     return NSNull()
   }
 
-  func acceptMissingValue(info: FieldExecutionInfo) -> JSONValue? {
+  final func acceptMissingValue(info: FieldExecutionInfo) -> JSONValue? {
     return nil
   }
 
-  func accept(list: [JSONValue?], info: FieldExecutionInfo) -> JSONValue? {
+  final func accept(list: [JSONValue?], info: FieldExecutionInfo) -> JSONValue? {
     return list
   }
 
-  func accept(childObject: CacheReference, info: FieldExecutionInfo) -> JSONValue? {
+  final func accept(childObject: CacheReference, info: FieldExecutionInfo) -> JSONValue? {
     return childObject
   }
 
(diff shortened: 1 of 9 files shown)


Sources/Apollo/GraphQLResultNormalizer.swift
```diff
@@ -3,47 +3,76 @@ import Foundation
 import ApolloAPI
 #endif
 
-final class GraphQLResultNormalizer: GraphQLResultAccumulator {
+struct ResultNormalizerFactory {
+  private init() {}
+
+  static func selectionSetDataNormalizer() -> SelectionSetDataResultNormalizer {
```
AnthonyMDev, line 9: I decided to create separate functions rather than `normalizer(for: DataSource)` because this way, each function returns the concrete type (instead of `BaseGraphQLResultNormalizer`). This allows the generic functions in the `GraphQLExecutor` to still specialize the function calls, using static dispatch instead of dynamic dispatch.

calvincestari (approved):
Nice work, I think we settled on the right solution here.

PR #2787, opened 2023-01-20
fix: Removes links to underscored properties and functions

### LINKED ISSUE #2766: Apollo SDK integrated through SPM emits documentation reference warnings for comments
### Summary

When integrating Apollo 1.0.5+ (1.0.6 has same issue) as a Swift Package, there are some topic references that are causing warnings in Xcode 14. Two copies of each of these warnings are emitted with each build and are preventing us from achieving a clean build, turning on warnings as errors, etc. Because the source is integrated through SPM, options are limited for suppressing the warnings.

<path>/SourcePackages/checkouts/apollo-ios/Sources/ApolloAPI/InputValue.swift:13:7 Topic reference 'GraphQLOperation/variables-swift.property-4o32c' couldn't be resolved. No local documentation matches this reference.

<path>/SourcePackages/checkouts/apollo-ios/Sources/ApolloAPI/JSONDecodingError.swift:5:40 Topic reference 'JSONDecodable/init(jsonValue:)' couldn't be resolved. No local documentation matches this reference.

### Version

1.0.6

- Create an Xcode project
- Integrate Apollo iOS latest as a Swift Package
- Build

Expected: No warnings
Actual: 4 warnings from the 2 files above (InputValue.swift and JSONDecodingError.swift

### Logs

_No response_

### Anything else?

Screenshot 2023-01-10 at 2 00 19 PM
Screenshot 2023-01-10 at 2 00 09 PM

calvincestari:
Fixes #2766 

The reason these documentation links have broken is because the properties they link to were changed to be 'private' by using the `_` prefix in their names. This change was largely done between `1.0.0-beta.4` and `1.0.0-rc.1`; _if you build the documentation between these two versions you'll notice there are many more when building for `1.0.0-rc.1`_.

It looks like Swift `5.7` simply doesn't emit the hidden signatures so docc can never consume them, but [coming in Swift `5.8`](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227/53) there might be the ability to [annotate the hidden signatures](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227/48) as `public`/`internal`/`private`.

To fix this now I've simply reworded the documentation to link to the related part that is public.

### DIFF
--- a/Sources/Apollo/Atomic.swift
+++ b/Sources/Apollo/Atomic.swift
@@ -41,7 +41,7 @@ public class Atomic<T> {
 
 public extension Atomic where T : Numeric {
 
-  /// Increments the wrapped `Int` atomically, adding +1 to the ``value``.
+  /// Increments the wrapped `Int` atomically, adding +1 to the value.
   @discardableResult
   func increment() -> T {
     lock.lock()
--- a/Sources/ApolloAPI/InputValue.swift
+++ b/Sources/ApolloAPI/InputValue.swift
@@ -9,12 +9,12 @@ public indirect enum InputValue {
   /// For enum input values, the enum cases's `rawValue` as a `String` should be used.
   case scalar(any ScalarType)
 
-  /// A variable input value to be evaluated using the operation's
-  /// ``GraphQLOperation/variables-swift.property-4o32c`` dictionary at runtime.
+  /// A variable input value to be evaluated using the operation's variables dictionary at runtime.
+  /// See ``GraphQLOperation``.
   ///
   /// `.variable` should only be used as the value for an argument in a ``Selection/Field``.
-  /// A `.variable` value should not be included in an operation's
-  /// ``GraphQLOperation/variables-swift.property-4o32c`` dictionary.
+  /// A `.variable` value should not be included in an operation's variables dictionary. See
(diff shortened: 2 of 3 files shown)


calvincestari:
I've left this PR in draft because there is still a couple of warnings that I'm not sure how to resolve and they are emitted when building the documentation. They seem to be related to links that cannot be resolved even though they're actually valid in the final docc archive.

> Link in document summary will not be displayed

From [this](https://github.com/apple/swift-docc/issues/341) it looks like you might have gone down this rabbit-hole already @AnthonyMDev. 😄

AnthonyMDev:
Yeah, this looks like all we need to do here for now. I don't mind having the other warnings, as they only appear when you run "Build Documentation" and they are actually invalid right now. This is something we are waiting on the DocC team to fix in their library.

The things you fixed here are actually invalid documentation that would be creating invalid links in our generated API Reference site and Documentation Bundles.

AnthonyMDev (approved):
I think this is ready to be merged in. The other warnings can be tackled when the next version of Xcode comes out with a version of the DocC tooling that (hopefully) fixes that bug.

PR #2811, opened 2023-02-06
fix: ApolloSchemaDownloadConfiguration HTTP headers json

### LINKED ISSUE #2661: No value associated with key CodingKeys when passing authorization in headers in codegen-config file
## Bug report

I am trying to upgrade apollo SDK to 1.0.2 version.

when configuring codegen-config file I need to pass headers and bearer token as dict key value.

I am getting following error

Screenshot 2022-11-14 at 1 05 51 PM

## Versions

- `apollo-ios` SDK version: 1.0.2
- Xcode version: 14.1
- Swift version: 5
- Package manager: SPM

## Steps to reproduce

```
{
  "schemaName" : "My Schema",
  "input" : {
    "operationSearchPaths" : [
      "**/*.graphql"
    ],
    "schemaSearchPaths" : [
      "**/*.graphqls"
    ]
  },
  "output" : {
    "testMocks" : {
      "none" : {
      }
    },
    "schemaTypes" : {
      "path" : "./mySchema",
      "moduleType" : {
        "swiftPackageManager" : {
        }
      }
    },
    "operations" : {
      "relative" : {
      }
    },
    "operationIdentifiersPath": "./Generated/"
  },
  "schemaDownloadConfiguration": {
      "downloadMethod": {
          "introspection": {
              "endpointURL": "my url",
              "httpMethod": {
                  "POST": {}
              },
              "includeDeprecatedInputValues": false,
              "outputFormat": "SDL"
          }
      },
      "downloadTimeout": 60,
      "headers": [{"Authorization":"Bearer  my token"}],
      "outputPath": "./GraphQL/"
  }
}
```

My shell script is

```
./apollo-ios-cli fetch-schema 
    
./apollo-ios-cli generate
```

Try to fetch the schema

calvincestari:
Fixes #2661 

This PR enables the HTTP headers, `"headers"` property within the `"ApolloSchemaDownloadConfiguration"` section of the JSON configuration file, to be written and parsed in a more intuitive format.

An example of the current format is `[{ "key": "Authorization", "value": "<token>"}]` where the new format would allow `{ "Authorization": "<token>"}`.

* The logic allows parsing of both formats.
* The property type has not changed because that would be a breaking change and I didn't want to do that in a patch release. _The config file format needs to be refactored at some point, this property change can happen then._
* The encoding format has not changed either because there might be projects that build the configuration in code, serialize it to JSON and use it in outside tooling which would break if the encoding format changed now too. Adding the ability to simply parse the new format is the least intrusive change here.
* PR #2812 is layered on top of this that updates the documentation to promote use of the new format.

### DIFF
--- a/Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
+++ b/Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
@@ -74,7 +74,7 @@ class ApolloSchemaDownloadConfigurationCodableTests: XCTestCase {
     expect(actual).to(equal(expected))
   }
 
-  func test__decodeApolloSchemaDownloadConfiguration__givenAllParameters_withHTTPHeadersAsArrayOfDictionaries_shouldReturnStruct() throws {
+  func test__decodeApolloSchemaDownloadConfiguration__givenAllParameters_withHTTPHeadersAsArrayOfStructs_shouldReturnStruct() throws {
     // given
     let subject = """
       {
--- a/Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
+++ b/Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
@@ -21,22 +21,27 @@ class ApolloSchemaDownloadConfigurationCodableTests: XCTestCase {
 
   // MARK: - ApolloSchemaDownloadConfiguration Tests
 
-  enum MockApolloSchemaDownloadConfiguration {
-    static var decodedStruct: ApolloSchemaDownloadConfiguration {
-      .init(
-        using: .introspection(
-          endpointURL: URL(string: "http://server.com")!,
-          httpMethod: .POST,
-          outputFormat: .SDL,
-          includeDeprecatedInputValues: true),
-        timeout: 120,
-        headers: [],
-        outputPath: "ServerSchema.graphqls"
-      )
-    }
+  func test__encodeApolloSchemaDownloadConfiguration__givenAllParameters_shouldReturnJSON_withHTTPHeadersAsArrayOfDictionaries() throws {
+    // given
+    let subject = ApolloSchemaDownloadConfiguration(
+      using: .introspection(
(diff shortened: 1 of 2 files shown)


Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
```diff
@@ -21,22 +21,27 @@ class ApolloSchemaDownloadConfigurationCodableTests: XCTestCase {
...
-          endpointURL: URL(string: "http://server.com")!,
-          httpMethod: .POST,
-          outputFormat: .SDL,
-          includeDeprecatedInputValues: true),
-        timeout: 120,
-        headers: [],
-        outputPath: "ServerSchema.graphqls"
-      )
-    }
+  func test__encodeApolloSchemaDownloadConfiguration__givenAllParameters_shouldReturnJSON_withHTTPHeadersAsArrayOfDictionaries() throws {
```
calvincestari, line 24:
```suggestion
  func test__encodeApolloSchemaDownloadConfiguration__givenAllParameters_shouldReturnJSON_withHTTPHeadersAsArrayOfStructs() throws {
```

Tests/ApolloCodegenTests/ApolloSchemaDownloadConfigurationCodableTests.swift
```diff
@@ -52,35 +57,125 @@ class ApolloSchemaDownloadConfigurationCodableTests: XCTestCase {
...
       }
       """
-    }
+
+    // then
+    expect(actual).to(equal(expected))
   }
 
-  func test__encodeApolloSchemaDownloadConfiguration__givenAllParameters_shouldReturnJSON() throws {
+  func test__decodeApolloSchemaDownloadConfiguration__givenAllParameters_withHTTPHeadersAsArrayOfDictionaries_shouldReturnStruct() throws {
```
calvincestari, line 77:
```suggestion
  func test__decodeApolloSchemaDownloadConfiguration__givenAllParameters_withHTTPHeadersAsArrayOfStructs_shouldReturnStruct() throws {
```

AnthonyMDev (approved):
Beautiful :)

PR #2816, opened 2023-02-08
fix: Parse query root interfaces

### LINKED ISSUE #2756: Interfaces implemented by schema root are not generated
### Summary

Interfaces that are implemented on the root note of a schema are not generated. Code generation does not produce any errors or warning but the generated code will not compile due to the missing interface.
I tried to produce a minimal schema and example query to illustrate the problem below.

The previous (non-swift) code generation tool did not have this issue.

### Version

1.0.5

Consider the following schema

```graphql
interface MissingInterface {
  missingInterfaceString: String!
}

type Query implements MissingInterface {
  missingInterfaceString: String!
}
```

We have a one simple query defined in our project:
```graphql
query fetchMyString {
  missingInterfaceString
}
```

After building and running the code generation tool for release `1.0.5` the generated query looks correct but in our Query type we are referencing an interface that was never generated causing a compilation error:

```swift
// @generated

import ApolloAPI

public extension Objects {
  static let Query = Object(
    typename: "Query",
    implementedInterfaces: [Interfaces.MissingInterface.self] // <--- Compilation error since MissingInterface not generated
  )
}
```

calvincestari:
Fixes #2756 

This PR simply pushes the operation root type through the `addReferencedType` function so that all it's references interfaces can be evaluated.


AnthonyMDev (approved):
Nice easy fix! Thanks!

PR #2828, opened 2023-02-15
Fix mock use documentation

vincentisambart:
As I tried following the documentation for test mocks I found a mistake so here's the fix.


calvincestari (approved):
Thanks @vincentisambart!

PR #2838, opened 2023-02-21
Fix docs for cache key configuration

AnthonyMDev:
These docs were incorrect and out of date on where the cache key configuration function should be located.


calvincestari (approved):
Thanks for fixing this.

PR #2840, opened 2023-02-21
docs: Adds API links to codegen configuration properties

### LINKED ISSUE #2837: Link to API documentation from codegen configuration documentation
This is in reference to a comment from docs feedback - https://apollograph.slack.com/archives/C01BMLR2YBG/p1675458457780929.

We do a generally good job of providing examples and listing all values for the configuration options but there are some properties where we don't list out the values. This should solve that need.

calvincestari:
Closes #2837 

Enables users to jump from the codegen configuration properties and long-format description/discussion to the code-specific API documentation.


AnthonyMDev (approved):
Thanks for adding these!

PR #2842, opened 2023-02-22
Proof of Concept: Code generation with @spi import

AnthonyMDev:


calvincestari:
I think we should probably include the removal of the `_` to the properties that `@_spi` helps make `internal` or `private` in this PR so we can ensure through CI that everything continues to work as expected.

AnthonyMDev:
I'm open to removing the `_` but I'm unclear how that has to do with our CI pipeline?

calvincestari:
It's not just removing the `_` though right, the access level would change from `public` to `internal`/`private` too? That's the whole point of @_spi no; not needing `public` level scope just to expose things to outside modules?

I didn't mean anything specific to this PR with CI, just ensuring the whole test suite continues to pass. This is a poc at the moment so any advance notice of something not building, or test not passing is helpful to know the poc is viable.

AnthonyMDev:
Oh! I'll play with that to confirm, but my understanding is that they are still marked public, but they are only publicly available when importing with the `@_spi(ApolloInternal)` import statement. I believe making them internal would actually defeat the purpose of `@_spi`. I'll confirm though.

calvincestari:
Linked to https://github.com/apollographql/apollo-ios/issues/2847.

PR #2849, opened 2023-02-24
Generate Initializers for SelectionSets [1/X]

AnthonyMDev:
This is the core of the implementation of generating initializers for selection sets.

This does not support `@include/skip` directives yet.
There are questions around allowing fields to be initialized with `null` vs `nil` values to be address in a future PR.


Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplate_Initializers_Tests.swift
```diff
@@ -0,0 +1,758 @@
...
+    type Cat implements Animal {
+      species: String!
+      friend: Animal!
+    }
+    """
+
+    document = """
+    query TestOperation {
+      allAnimals {
+        ... on Cat {
```
calvincestari, line 478: How would the fields in this selection be initialized?

calvincestari (approved):
Nice of work in this PR. Tests are looking good too.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -1,17 +1,21 @@
...
 
   let isMutable: Bool
+  let generateInitializers: Bool
   let config: ApolloCodegen.ConfigurationContext
 
   private let nameCache: SelectionSetNameCache
 
   init(
     mutable: Bool = false,
+    generateInitializers: Bool,
```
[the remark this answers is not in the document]
AnthonyMDev: I agree that it feels weird. But other templates are using this in a few different ways. `FragmentTemplate` initializes the `SelectionSetTemplate`, but only renders the `BodyTemplate` portion of it.

Though it feels weird, I actually think that having whatever is initializing a `SelectionSetTemplate` handle the determination of this functionality gives us more room for flexibility in the future.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplate_Initializers_Tests.swift
AnthonyMDev: The generated `Data.AllAnimal.AsCat` object has it's own initializer that can be used for this!

PR #2865, opened 2023-03-07
Add asEntityRootType converter to InlineFragments

AnthonyMDev:
This adds a new property `.asEntityRootType` to all generated inline fragment selection sets. This allows you to convert them back to the root selection set model of the entity directly. This means you never have to hold on to a reference to the parent. It also is necessary for initializing type condition models and passing them as the values to entity fields when using generated initializers for the models.


AnthonyMDev:
The selection set name generation in `SelectionSetTemplate` is getting a bit messy. There are a lot of complex ways we need to compute selection set names in multiple different ways here. The logic is a bit rough, but we should look at this at some point and see if we can create a better API for handing all of these different formats. Similar to the `RenderingContext` for the `GraphQLType` code.

calvincestari (approved):

PR #2868, opened 2023-03-08
Allows `swift-format` and `swiftlint` to ignore generated files

rickpasetto:
Adds lines to generated template renderer to direct swift-format and SwiftLint to ignore generated files.
Also updates tests.

```
    // swift-format-ignore-file
    // swiftlint:disable all
```

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
+++ b/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
@@ -169,6 +169,8 @@ struct HeaderCommentTemplate {
   static let template: StaticString =
     """
     // @generated
+    // swift-format-ignore-file
+    // swiftlint:disable all
     // This file was automatically generated and should not be edited.
     """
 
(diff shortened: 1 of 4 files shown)


calvincestari:
I don't believe we need to add these tool-specific directives to Apollo generated files.
* Exclusion is already possible in swiftlint through [configuration](https://github.com/realm/SwiftLint#configuration); use the `excluded:` property.
* I'm not sure how exclusion is done in swift-format but there is probably a way to achieve that through configuration too.

rickpasetto:
@calvincestari Thanks for the reply. You are correct, `swiftlint` does have the ability to specify exclusion.  I added this as a "nice to have" so I can remove it.  But [swift-format](https://github.com/apple/swift-format) does not, at least not yet, which was the impetus of this PR.  Would you consider it if I only added

```
    // swift-format-ignore-file
```
?

calvincestari:
I still don't believe we should be adding external tooling attributes to Apollo generated source code. The right place to fix this is in swift-format.

PR #2870, opened 2023-03-08
feat: Multipart subscriptions

calvincestari:
_This is both the RFC and implementation. I'm capturing the design decision (RFC) in the PR because the original suggested design differs from what the final solution is. This will provide a high-level overview of the design instead of having to read the code to fully understand the change._

### Goal
Support [GraphQL subscriptions](https://spec.graphql.org/draft/#sec-Subscription-Operation-Definitions) over HTTP using a multipart response (incremental delivery) format similar to that defined in the [`@defer`/`@stream` RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md).

### Existing
GraphQL subscriptions are currently supported in Apollo iOS but only over WebSocket - [documentation](https://www.apollographql.com/docs/ios/fetching/subscriptions#enabling-graphql-subscription-support). Coincidentally a subscription _can_ be sent over HTTP using the [`RequestChainNetworkTransport`](https://www.apollographql.com/docs/ios/networking/request-pipeline) transport but the response would never be returned because the current request chain is designed for a single response; the technical reason is that `urlSession(_:task:didCompleteWithError:)` would not be called because of the multipart data transfer (the request never actually _completes_) and subsequently the completion block of the subscription request is never called.

#### Discarded
The first proposed design was to create an entirely new transport that would operate differently than the current concept of a '[request chain](https://www.apollographql.com/docs/ios/networking/request-pipeline#request-chains)' and implement something closer to [Apollo Link](https://www.apollographql.com/docs/react/api/link/introduction/). The scope of this change is quite large and it doesn't offer an easy path for users to migrate from existing WebSocket-based subscriptions to HTTP-based subscriptions; additional work would be required to let the user choose which transport a particular subscription should use and `SplitNetworkTransport` would need to juggle another transport.

#### Preferred
The implemented design is to adapt the existing request chain to be re-entrant; the re-entrant behaviour is key to making this design possible. Once the request chain becomes re-entrant all the current interceptors can be reused and it fixes the currently broken behaviour of subscriptions over HTTP. The migration path becomes easier too since users would simply need to send a subscription over the `RequestChainNetworkTransport` instead of the `WebSocketTransport`. 

### Technical details
The current limitation of the request chain is that progress through the interceptors is [tracked with an index](https://github.com/apollographql/apollo-ios/blob/127cc6bd16f0456b51bc7df94a67ea4bcb3787ae/Sources/Apollo/RequestChain.swift#L82). This makes it impossible to handle a multipart response where network responses need to move forward at different times from the same interceptor, with no guarantee that the chain has completed before the next response arrives. An index is required because in the `proceedAsync(request:response:completion:)` callback there is no indication of which interceptor has finished processing. This is further complicated knowing that these interfaces/classes/structs are all public and there could be any number of custom interceptors that users have built specific to their data flow.

To make the request chain support multipart responses it has to become re-entrant which then separates each multipart response into its own forward continuation of the request chain. To do this in a way that is backwards compatible with custom interceptors means we need to hide the abstraction and handle it internally. This is done by wrapping each interceptor in the supplied interceptor list in a wrapper that both functions as the request chain the interceptor calls back to and as the 'forwarder' of request chain progress to each wrapped interceptor. On interceptor callback the wrapper will actually use a [_new_ function that requires the interceptor to identify itself](https://github.com/apollographql/apollo-ios/pull/2870/files#diff-aba50d14aacb6c7410fa1214b08e88767c998a0343040b433788f9d55dfc66bbR110); this interceptor identification is what enables independent forward continuation of the request chain.

There is also a new interceptor, `MultipartResponseParsingInterceptor`, that is inserted before the `JSONResponseParsingInterceptor` in the interceptor chain. This new interceptor parses the multipart responses at the given boundary and forwards each chunk on to the next interceptor.

**Will be done in a subsequent PRs:**
- Ensure that SplitNetworkTransport functions correctly for the migration path
- End-to-end tests
- Documentation


calvincestari:
I'm putting this PR into review so that the general approach and implementation can be commented on. Once I'm back next week I'll finish up the tests for the new `MultipartResponseParsingInterceptor` interceptor.

Sources/Apollo/ApolloClient.swift
```diff
@@ -138,7 +138,6 @@ extension ApolloClient: ApolloClientProtocol {
     }
   }
   
-  @discardableResult
```
calvincestari, line 141: Our [documentation](https://www.apollographql.com/docs/ios/fetching/subscriptions#cancelling-a-subscription) states that subscriptions must be cancelled but through `@discardableResult` users were not forced to hold on to the result of the subscription request and therefore had no way to cancel it.

Sources/Apollo/DefaultInterceptorProvider.swift
```diff
@@ -35,6 +35,7 @@ open class DefaultInterceptorProvider: InterceptorProvider {
         CacheReadInterceptor(store: self.store),
         NetworkFetchInterceptor(client: self.client),
         ResponseCodeInterceptor(),
+        MultipartResponseParsingInterceptor(),
```
calvincestari, line 38: This is in the interceptor chain for all operation types. Right now it is applicable only to `Subscription` but supporting `@defer` will require it for `Query` too. There is a short-circuit to check for the `multipart/mixed` header and exit early if it's not present in the response.

Sources/Apollo/URLSessionClient.swift
```diff
@@ -247,21 +253,47 @@ open class URLSessionClient: NSObject, URLSessionDelegate, URLSessionTaskDelegat
...
-        assertionFailure("No data found for task \(dataTask.taskIdentifier), cannot append received data")
+
+    guard let taskData = self.tasks[dataTask.taskIdentifier] else {
+      assertionFailure("No data found for task \(dataTask.taskIdentifier), cannot append received data")
+      return
+    }
+
+    taskData.append(additionalData: data)
+
+    if let httpResponse = dataTask.response as? HTTPURLResponse, httpResponse.isMultipart {
```
calvincestari, line 273: This multipart boundary check is to ensure that only complete chunks are sent to the next interceptor. Any remaining data that was received that should be processed with the next incoming data to form a complete chunk will get held back.

Tests/ApolloTests/Interceptors/MultipartResponseParsingInterceptorTests.swift
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    func returnValueAsync<Operation>(
+      for request: Apollo.HTTPRequest<Operation>,
+      value: Apollo.GraphQLResult<Operation.Data>,
+      completion: @escaping (Result<Apollo.GraphQLResult<Operation.Data>, Error>) -> Void
+    ) {}
+  }
+
+  // MARK: - Error tests
```
calvincestari, line 49: These error tests are done directly on the interceptor because in normal operation there are other safeguards in earlier interceptors that _should_ never make these possible but with optionals we need to handle unintended behaviour. Doing this on only the interceptor ensures we can test all error cases.

Tests/ApolloTests/Interceptors/MultipartResponseParsingInterceptorTests.swift
```diff
@@ -0,0 +1,355 @@
...
+          --graphql
+          """.crlfFormattedData()
+      )
+    ) { result in }
+
+    expect(requestChain.error as? MultipartResponseParsingInterceptor.MultipartResponseParsingError)
+      .to(equal(.cannotParseChunkData))
+  }
+
+  // MARK: Parsing tests
```
calvincestari, line 178: The parsing tests involve more of the request chain mechanics to ensure that the interceptors all work together as expected.

calvincestari:
@AnthonyMDev this PR is now ready for review.

Note - there is a lot of `String` to/from `Data` encoding as the various parsers inspect multipart boundaries and process the separate chunks. I believe this is OK for now, which I'll validate with performance tests to get benchmarks recorded into our test suite. Once those benchmarks are in I'm going to see how much of an improvement we can get from working with the byte data instead of back and forth with `String`.

Sources/Apollo/ApolloClient.swift
AnthonyMDev: I'm all for this, but this is technically an API breaking change. It seems to be a pretty significant error though, so I'm okay with this!

Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
```diff
@@ -0,0 +1,109 @@
...
+class ApolloInterceptorReentrantWrapper: RequestChain {
+  @Atomic var isCancelled: Bool = false
+
+  let wrappedInterceptor: ApolloInterceptor
+  let requestChain: Unmanaged<InterceptorRequestChain>
+  let index: Int
+
+  init(
+    interceptor: ApolloInterceptor,
+    requestChain: Unmanaged<InterceptorRequestChain>,
```
AnthonyMDev, line 18: Love that we are using `Unmanaged` here!

AnthonyMDev: Do we have a unit test to ensure we aren't accidentally creating a zombie when using this?

Sources/Apollo/MultipartResponseParsingInterceptor.swift
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  private enum ChunkedDataLine {
+    case heartbeat
+    case contentHeader(type: String)
+    case json(object: JSONObject)
+    case unknown
+  }
+
+  private let dataLineSeparator: String = "\r\n\r\n"
```
AnthonyMDev, line 42: If these are constants, they should be static, instead of initialized for each instance. 

nit: Also, you can make them `StaticString`, and use the `.description` property to turn them into `String` when using them for better performance.

Sources/Apollo/URLSessionClient.swift
```diff
@@ -247,21 +253,47 @@ open class URLSessionClient: NSObject, URLSessionDelegate, URLSessionTaskDelegat
...
     guard dataTask.state != .canceling else {
       // Task is in the process of cancelling, don't bother handling its data.
       return
     }
-    
-    self.$tasks.mutate {
-      guard let taskData = $0[dataTask.taskIdentifier] else {
-        assertionFailure("No data found for task \(dataTask.taskIdentifier), cannot append received data")
+
+    guard let taskData = self.tasks[dataTask.taskIdentifier] else {
```
AnthonyMDev, line 266: By removing the `self.$tasks.mutate` block, this is no longer atomic. I get why it's not necessary for updating the `TaskData` in place, because `TaskData` is a `class`. But does this mean that this is not guaranteed to happen in a sequentially? Can these create problematic race conditions?

Sources/Apollo/URLSessionClient.swift
AnthonyMDev: Will that chunk get automatically included in the `dataTask` next time this function is called? I'm not seeing anywhere that you are holding onto the `dataTask` to merge them together when the next chunk comes in.

Tests/ApolloTests/Interceptors/MultipartResponseParsingInterceptorTests.swift
AnthonyMDev: So are the safeguards in earlier interceptors ignored for multipart requests?

Tests/ApolloTests/WebSocket/WebSocketTests.swift
```diff
@@ -62,7 +62,7 @@ class WebSocketTests: XCTestCase {
   func testLocalSingleSubscription() throws {
     let expectation = self.expectation(description: "Single subscription")
     
-    client.subscribe(
+    _ = client.subscribe(
```
AnthonyMDev, line 65: We should probably be cancelling the subscriptions at the end of these tests as well!

AnthonyMDev (commented):
This looks awesome! I have a few questions and suggestions, but very minor. Great job on this. It was not a simple task...

Sources/Apollo/ApolloClient.swift
calvincestari: Correct, leans more towards a "bug fix" IMO.

Sources/Apollo/URLSessionClient.swift
calvincestari: `tasks` is annotated with the `@Atomic` property wrapper so [all access to the task list should still be atomic](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/Atomic.swift#L22); we're assured that the task won't be removed at least. We only need `mutate` when we want to alter the task list.

`data` within each `TaskData` is not protected by anything though so yes I suppose there is a chance of a race condition when working with incoming data for a particular task. We shouldn't be using the list atomic lock to ensure consistent data access for a single task though. I suspect it was either a mistake to have `mutate` there in the first place or it was being abused for the purpose of single task data access.

The default callback queue is `main` which, being a serial queue, will ensure there is no race condition. Anyone who cares about a performant UI though will be giving us their own queue, which could be concurrent and introduce the potential for a race condition.

We could use `@Atomic` on `TaskData.data` to prevent this but I'd want a performance benchmark test on that to measure the cost of doing so.

Sources/Apollo/URLSessionClient.swift
calvincestari: All incoming data is appended to any data already related to the task, see line [271](https://github.com/apollographql/apollo-ios/pull/2870/files/2de76f1a53ec80fb1d56107b2852f9a010e9c41e#diff-80dc6761af321b52c5b1108a063a8ad3f0086925f9a813c6eb1fe043ada83c86R271). Then we work with the data sending off the complete chunks to other interceptors and on line [290](https://github.com/apollographql/apollo-ios/pull/2870/files/2de76f1a53ec80fb1d56107b2852f9a010e9c41e#diff-80dc6761af321b52c5b1108a063a8ad3f0086925f9a813c6eb1fe043ada83c86R290) is where the task data is reset with any remaining bytes, ready for the callback block of data to be added to it.

Tests/ApolloTests/WebSocket/WebSocketTests.swift
calvincestari: I added them but I'm less convinced of the value of doing it in these tests since they all exit and get torn down afterwards. There might be some negligible benefit for CI resources.

Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
calvincestari: The watcher tests we fixed for the other day are an indirect way of testing of for a retain cycle but I added a more direct test on the request chain itself - https://github.com/apollographql/apollo-ios/pull/2870/commits/21857e9b8b823ab18179398dfe96f58fca8b5bbb

Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
```diff
@@ -54,8 +54,6 @@ class ApolloInterceptorReentrantWrapper: RequestChain {
     if let cancellableInterceptor = wrappedInterceptor as? Cancellable {
       cancellableInterceptor.cancel()
     }
-
-    requestChain.takeUnretainedValue().cancel()
```
AnthonyMDev: This removal feels very wrong to me. Can you explain what is going on that makes you need to remove this?

calvincestari: I removed it because it's cyclical. Note that this is in the reentrant wrapper so all calls to `cancel()` from all wrapped interceptors will be going back to the same 'parent' request chain but it will simply exit early because it's already been cancelled - [here](https://github.com/apollographql/apollo-ios/blob/feature/multipart-subscriptions/Sources/Apollo/InterceptorRequestChain.swift#L156). A couple lines down in that same function is where the call to cancel each cancellable wrapped interceptor is - [here](https://github.com/apollographql/apollo-ios/blob/feature/multipart-subscriptions/Sources/Apollo/InterceptorRequestChain.swift#L165).

All tests pass including the new retain cycles tests which are both on a query and a subscription, of which the subscription is required to be cancelled in order to release the unmanaged object.

AnthonyMDev: Is there any situation in which the `cancel()` function on the reentrant wrapper would be called directly, rather than via one of the wrapped interceptors? 

If it doesn't cause issues to have it, it feels like it's safer to keep it there for possible future iterations where you call `cancel()` directly here. But I may be thinking about this wrong. I don't have the context you have.

calvincestari: The reentrant wrapper is [internal access](https://github.com/apollographql/apollo-ios/blob/feature/multipart-subscriptions/Sources/Apollo/ApolloInterceptorReentrantWrapper.swift#L9) only and the [list of wrapped interceptors on the request chain is private](https://github.com/apollographql/apollo-ios/blob/feature/multipart-subscriptions/Sources/Apollo/InterceptorRequestChain.swift#L23) so it's not possible publicly. We would have to be doing that internally which I feel goes against the design intent of it and _should_ break tests.

Sources/Apollo/URLSessionClient.swift
AnthonyMDev: Yes, I agree with your assessment here. I do think we should be making the mutation of each individual `TaskData` be atomic though. This is far too likely to cause a race condition for someone in its current state.

Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
AnthonyMDev: Got it. Sounds good!

Sources/Apollo/URLSessionClient.swift
AnthonyMDev: Got it. To me, this makes it even more important that task data is mutated atomically then. Totally good with this implementation in every other regard. But seems very easy for this to get corrupted by parallel threads executing in a weird order.

calvincestari: 👍🏻 - created https://github.com/apollographql/apollo-ios/issues/2900 to handle it.

Sources/Apollo/URLSessionClient.swift
calvincestari: 👍🏻 - created https://github.com/apollographql/apollo-ios/issues/2900 to handle it.

AnthonyMDev (approved):

PR #2877, opened 2023-03-15
update: Deprecate `--schema-name` property

### LINKED ISSUE #2642: Rename `--schema-name`
This is a property on the CLI and ApolloCodegenConfiguration. I think people are thinking this is the filename of their schema file. I believe `--schema-namespace` would lead to less confusion and errors.

calvincestari:
Closes #2642 

This PR deprecates the `--schema-name` property with `@available` annotations including the renamed property/function name for an easier transition in user projects.

There is a second PR branched from this one that makes changes not directly relevant to user code but for internal use where we still have reference to `schemaName`. It's probably find to keep it like that but I'd rather have less confusion and mixed terms for anyone else looking at the code without the context that we have.

~_I'm still looking for a way to tell a CLI user that the `--schema-name` property is deprecated. Will change this PR from draft to review when I've figured that out._~


Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -11,11 +11,21 @@ public struct Initialize: ParsableCommand {
     abstract: "Initialize a new configuration with defaults."
   )
 
+  @Option(
+    name: [.long],
+    help: "DEPRECATED - Use --schema-namespace instead."
+  )
+  // When removing this property also do:
```
calvincestari, line 18: I'm not a huge fan of needing to have this here but some of the changes for the CLI option deprecation warning feel a bit hacky and I want to make sure we get rid of them when we finally remove the deprecated property.

Sources/CodegenCLI/Commands/Initialize.swift
```diff
@@ -11,11 +11,21 @@ public struct Initialize: ParsableCommand {
...
+  // - remove schemaName validation in validate()
+  // - remove mutating keyword from validate() signature
+  var schemaName: String?
+
   @Option(
     name: [.long, .customShort("n")],
     help: "Name used to scope the generated schema type files."
   )
-  var schemaName: String
+  var schemaNamespace: String = ""
```
calvincestari, line 28: Required to let users continue to use the `init` command with the `--schema-name` option for now.

Tests/CodegenCLITests/Commands/InitializeTests.swift
```diff
@@ -113,14 +113,32 @@ class InitializeTests: XCTestCase {
...
-      "--schema-name=MySchemaName",
+      "--schema-namespace=MySchemaName",
       "--module-type=other",
     ]
 
     // then
     expect { try self.parse(options) }.notTo(throwError())
   }
 
+  func test__validation__givenBothSchemaNamespaceOptions_shouldThrow() throws {
```
calvincestari, line 124: I would have liked to have a test to ensure that `--schema-name` can be used but it seems that swift-argument-parser actually injects options into the command line for properties with initialized values, like `schemaNamespace` now is. So when I wrote that test it fails with the validation that you cannot use both options at the same time. 😢

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -679,7 +679,9 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
   // MARK: - Properties
 
   /// Name used to scope the generated schema type files.
-  public let schemaName: String
+  @available(*, deprecated, renamed: "schemaNamespace")
+  public var schemaName: String { schemaNamespace }
```
AnthonyMDev, line 683: Can we move deprecated properties and initializers into an extension with `// MARK: - Deprecations` so they aren't in the way as much? It also give us an easy place to keep them so we can keep track of stuff to kill with the next major (breaking) version.

AnthonyMDev (approved):
This looks great! Thanks!. Only one minor suggestion.

PR #2887, opened 2023-03-22
Refactor SelectionSet models to track Fulfilled Fragments

AnthonyMDev:
This is a significant change in our approach to structuring model data.

This PR removes `_objectType` and `_variables` from `DataDict`. These were used to check if a type conversion for a `SelectionSet` model was valid. Instead, the logic for determining if a type conversion can be done happens during GraphQL Execution. The `GraphQLExecutor` tracks all fragments that should be fulfilled during execution. 

Both the `GraphQLExecutor` and the generated `SelectionSet` initializers will initialize the `DataDict._data` with an included `"__fulfilled"` metadata object. All type conversions for `SelectionSet` models will now use this to determine if they can be completed.

This PR also re-enables the existing performance tests as well as adds some new ones.

### Broken Tests

Currently, it has tests that are failing for writing manually initialized selection sets to the cache. These will be fixed in my next PR.

### Test Mocks Broken

This PR also breaks Test Mock usage. Though the new initializers are  now an alternative for creating mocks, TestMocks are not being obsoleted as they still do serve a useful purpose currently (I also don't want to be removing features people are using in a minor version release!). Test Mocks will be fixed in a future PR on this branch.


AnthonyMDev:
Closing because this is merged into #2893

Issue #2386, opened 2022-07-21
Update internal testing APIs to use 1.0 code generation


calvincestari:
Some of the internal testing APIs are out of date with the latest changes to code generation and some still have the legacy codegen sources.
* StarWarsAPI
* UploadAPI
* SubscriptionAPI
* GitHubAPI

calvincestari:
This is done with the merging of #2385.

PR #2888, opened 2023-03-22
Enable ApolloPerformanceTests

### LINKED ISSUE #2387: Enable ApolloPerformanceTests
These were disabled during codegen development. When #2386 is done the source to support the performance tests will be up-to-date.

jawwad:
Closes #2387

The issue mentioned that re-enabling the tests was waiting on #2386 (Update internal testing APIs to use 1.0 code generation). Since that was already completed I uncommented the tests to see if they would run, and they do run with no issues.

I only ran into one compile error in which I had to update `query.operationName` to `Query.operationName` due to the following error: `Static member 'operationName' cannot be used on instance of type 'Query'`

Screen Shot 2023-03-21 at 10 47 25 PM

### DIFF
--- a/Tests/ApolloPerformanceTests/ParsingPerformanceTests.swift
+++ b/Tests/ApolloPerformanceTests/ParsingPerformanceTests.swift
@@ -1,53 +1,52 @@
 import XCTest
 @testable import Apollo
 import ApolloInternalTestHelpers
-//import GitHubAPI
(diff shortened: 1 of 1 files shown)


calvincestari:
Thanks for the contribution @jawwad but this PR may not be needed, I think @AnthonyMDev has enabled them in the work for 1.1.

PR #2891, opened 2023-03-23
feat(error): add `path` field to `GraphQLError`

### LINKED ISSUE #2890: Add `path` field to `GraphQLError`
### Use case

The `GraphQLError` currently has all necessary information according to the [GraphQL Spec](https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format) except for one: the path.

The path field is extremely useful for client applications to know whether a variable that was returned as null in the data response is expected to be null or there was an error which prevented it from returning the correct value. Missing out on this field is missing out on important information about the error.

This field is even shown in the [Apollo documentation](https://www.apollographql.com/docs/ios/fetching/error-handling#graphql-errors), but, for some reason, it's not implemented into the actual type-safe type (`GraphQLError`)

The field should be added alongside `locations` and such. Since a path entry is a list that can have String (field name) or Int (list index) values, we can create a simple enum to represent this:
```swift
enum PathEntry {
   case field(String)
   case index(Int)
}
```
and return an array in the `path` variable:

```swift
public var path: [PathEntry]? {
...
```

This maintains the type-safeness of the `GraphQLError` type and provides the necessary information about the path. Further parsing (concatenating the variables as a full path string, for example) may be done outside of this type, by users of the API. The raw type should be a concrete representation of the GQL spec, so the enum will suffice.

AmadeusK525:
This field returns an array of `PathEntries`, an enum that represents the list that may be returned by the GraphQL server (it can be either a `field` - String - or an index - Int).

Tests were added to secure the behavior

- Closes #2890


AnthonyMDev (approved):
Thank you for the contribution @AmadeusK525! We're going to accept this and merge it in! This will be released with the 1.1 GA release.

PR #2895, opened 2023-03-24
Re-enable skipped tests. Closes #2374

### LINKED ISSUE #2374: ci: Reassess skipped tests
There are a number of tests being skipped in the ApolloTests. We need to either rewrite them if they're incorrect or remove them if they're no longer useful.

jawwad:
I re-enabled the skipped tests tests and they all seem to be working without any issues. I ran each of the disabled ones individually as well as the default test suite.

There were 3 test classes that were skipped:
GraphQLFileTests
MultipartFormDataTests
StoreConcurrencyTests

As well as 2 specific tests that were skipped:
URLDownloaderTests.testDownloader_withCorrectResponse_shouldNotThrow() StarWarsSubscriptionTests.testConcurrentSubscribing()

Maybe these were initially disabled since they were flakey, or maybe whatever the issue was has been resolved.

### DIFF
--- a/Tests/TestPlans/Apollo-Codegen-CITestPlan.xctestplan
+++ b/Tests/TestPlans/Apollo-Codegen-CITestPlan.xctestplan
@@ -14,9 +14,6 @@
   "testTargets" : [
     {
       "parallelizable" : true,
-      "skippedTests" : [
-        "URLDownloaderTests\/testDownloader_withCorrectResponse_shouldNotThrow()"
-      ],
       "target" : {
         "containerPath" : "container:Apollo.xcodeproj",
         "identifier" : "9BAEEBFB234BB8FD00808306",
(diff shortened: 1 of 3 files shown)


calvincestari:
Thanks for the contribution @jawwad. It's worth noting that during development of 1.0 we disabled/skipped tests that were failing or needed a review of what they were testing, given the large scope of change in 1.0.

tl;dr - We need more rigour around the disabled/skipped tests instead of simply re-enabling them.

jawwad:
Aah ok. So these will require more of a manual audit to ensure that they are testing the correct thing. The description on the original task now makes more sense to me, where you said: "We need to either rewrite them if they're incorrect...". For some reason, I took _incorrect_ to mean not passing, but you meant if they are not testing the correct thing. Since I'm not familiar with the scope of changes that need to be tested I'll close this out for now.

PR #2897, opened 2023-03-24
Fix creating of TestMocks to compute fulfilledFragments

AnthonyMDev:
Test Mocks need to compute the fulfilledFragments when converting to SelectionSets now. This means they need to go through the executor.

This configures an executor that will allow missing values for use with test mocks.

### DIFF
--- a/Sources/Apollo/GraphQLExecutor.swift
+++ b/Sources/Apollo/GraphQLExecutor.swift
@@ -245,6 +245,7 @@ final class GraphQLExecutor<FieldCollector: FieldSelectionCollector> {
   ) throws -> FieldSelectionGrouping {
     var grouping = FieldSelectionGrouping(info: info)
 
+    #warning("TODO: When executing on a selection set that isn't operation root. This will be wrong.")
     // Add __typename field to all selection sets other than the root of the operation.
     if !info.responsePath.isEmpty {
       grouping.append(field: .init("__typename", type: .scalar(String.self)), withInfo: info)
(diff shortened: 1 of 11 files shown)


Sources/Apollo/GraphQLExecutor.swift
```diff
@@ -245,6 +245,7 @@ final class GraphQLExecutor<FieldCollector: FieldSelectionCollector> {
   ) throws -> FieldSelectionGrouping {
     var grouping = FieldSelectionGrouping(info: info)
 
+    #warning("TODO: When executing on a selection set that isn't operation root. This will be wrong.")
```
calvincestari, line 248: Will this be handled in a subsequent PR? I think that's what we discussed in our last conversation.

Sources/Apollo/GraphQLSelectionSetMapper.swift
```diff
@@ -8,14 +8,22 @@ import Foundation
 final class GraphQLSelectionSetMapper<T: SelectionSet>: GraphQLResultAccumulator {
 
   let stripNullValues: Bool
-  let allowMissingValuesForOptionalFields: Bool
+  let handleMissingValues: HandleMissingValues
+
+  enum HandleMissingValues {
```
calvincestari, line 13: 👍🏻

calvincestari (approved):

Sources/Apollo/GraphQLExecutor.swift
AnthonyMDev: Yupp!

PR #2899, opened 2023-03-25
Update usage docs for 1.1

AnthonyMDev:
Docs for:
- SelectionSetInitializer config
- Cache key resolution limitations
- Fix for JSONParsingInterceptor initializer in example

### DIFF
--- a/docs/source/caching/cache-key-resolution.mdx
+++ b/docs/source/caching/cache-key-resolution.mdx
@@ -176,3 +176,21 @@ public enum SchemaConfiguration: ApolloAPI.SchemaConfiguration {
   }
 }
 ```
+
+## Caveats & limitations
+
+Because cache key resolution is performed both on raw JSON (from a network response or cache hit) and `SelectionSet` model data (when writing to the cache directly), the underlying `object` will have different formats.
+
+This means that cache key resolution has a few notable limitations:
+
+1. Computing cache keys from fields on nested objects is only allowed if the nested object
+does not have its own cache key. If the nested object has its own cache key, it will be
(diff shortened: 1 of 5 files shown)


docs/source/caching/cache-key-resolution.mdx
```diff
@@ -175,4 +175,22 @@ public enum SchemaConfiguration: ApolloAPI.SchemaConfiguration {
...
+does not have its own cache key. If the nested object has its own cache key, it will be
+normalized as a seperate cache entity. Cache keys for entities cannot be dependent on
+other entities.
+
+2. When computing a cache key using a field of a `CustomScalarType`, the underlying type of
+the value in the `object` dictionary will vary. It may be the raw JSON value for the scalar
+(when the source is a network response or cache hit) or the specific custom scalar type for
+the field (when the source is a `SelectionSet` model to write to the cache). When using a
+custom scalar field to compute a cache key, make sure to check the type and handle both of
+these possibilities
```
calvincestari, line 196: This is a good start to the docs. I think it would be helpful to add examples here of how to work with the cache data from the various sources and how to handle the different formats.

calvincestari (approved):
Nice! Is there anything worth mentioning about these initializers in the migration guide? Users coming from 0.x models may have expectations about the initializers behaviour that we can explain here rather than in new issues/questions.

docs/source/caching/cache-key-resolution.mdx
AnthonyMDev: Definitely agree! This is just the bare minimum for Beta. I'm going to add more details and usage examples for this as well as for the new initializers next week.

PR #2903, opened 2023-03-25
Add test for when JSONResponseParsingError.noResponseToParse is called

tokizuoh:
Add test where an unintended call to the JSONResponseParsingInterceptor is made and verify that the error is `.noResponseToParse` in the completion handler.


calvincestari:
Thanks for the PR @tokizuoh. You mentioned "unintended call", did you encounter this in general operation of the interceptor chain?

tokizuoh:
Thank you for replying @calvincestari !

No, I did not encounter it during the general operation of the interceptor chain. I only added it because there was no test code for the defined error. I have not actually encountered it in practice.

calvincestari:
We now have the [`MockInterceptorProvider`](https://github.com/apollographql/apollo-ios/blob/main/Tests/ApolloInternalTestHelpers/MockInterceptorProvider.swift) class so we can probably cleanup the tests here and reuse some of the more recently added test helpers.

Would you like to do that, I'm happy to take it on if not.

tokizuoh:
MockInterceptorProvider class is good! I tried to use it.

043459b

calvincestari:
> I would like to update other tests with another Pull-Request if https://github.com/apollographql/apollo-ios/commit/043459b8685e70fc595465af61f412ae13d4a292's policy looks good.

Yes that's what I meant; cleaning up this whole test file. You can do it all in this PR, I don't think we need another one.

tokizuoh:
OK! I agree.
I changed it in f994966.

Tests/ApolloTests/Interceptors/JSONResponseParsingInterceptorTests.swift
```diff
@@ -4,28 +4,50 @@ import ApolloInternalTestHelpers
 import XCTest
 
 class JSONResponseParsingInterceptorTests: XCTestCase {
+  func testJSONResponseParsingInterceptorFailsWithImproperlyOrderedCalls() {
```
calvincestari, line 7:
```suggestion
  func testJSONResponseParsingInterceptorFailsWhenNoResponse() {
```

calvincestari (approved):
I renamed the new test but looks good otherwise. I'll get this merged now, thanks for the contribution @tokizuoh.

tokizuoh:
The renaming makes it easier to understand. Thank you!

PR #2905, opened 2023-03-27
docs: Update 'Install the Apollo Frameworks' Xcode SPM section

### LINKED ISSUE #2885: docs: Update 'Install the Apollo Frameworks' section
Update the image to include 'apollo-ios-cli' product with warning not to include it as SPM dependency.

https://www.apollographql.com/docs/ios/get-started/#1-install-the-apollo-frameworks

calvincestari:
Closes #2885 

This is a change to update the screenshots and copy to warn users not to install the `apollo-ios-cli` product as a project dependency. I've also cleaned up some screenshot resources that were not referenced in any documentation.


docs/shared/spm-xcode-installation-panel.mdx
```diff
@@ -7,14 +7,14 @@ If using Swift Package Manager within an Xcode project or workspace, use the Xco
...
 Go to **File > Add Packages...**
 
-<img class="screenshot" src="../source/screenshot/spm_packages_add_package.jpg" alt="Adding an SPM package" width="300" />
+<img class="screenshot" src="../source/screenshot/spm_packages_add_package.png" alt="Adding an SPM package" width="560" />
 
 </ExpansionPanelListItem>
 <ExpansionPanelListItem>
 
-In the dialog that appears, paste the URL of the Apollo iOS GitHub repo (`https://github.com/apollographql/apollo-ios.git`) into the search bar, then select the `apollo-ios` package that appears:
+In the dialog that appears, type `apollo-ios` into the search bar, then select the `apollo-ios` package that appears:
```
AnthonyMDev, line 15: Are we sure this is true? SPM keeps a cache of used packages, but there isn't a central repository. If you haven't already got a local clone of apollo-ios in your SPM module cache, will it show up from a search this way?

AnthonyMDev (approved):
Thanks for making these improvements! I'm unsure about replacing the project URL, but the rest of this looks awesome!

docs/shared/spm-xcode-installation-panel.mdx
calvincestari: I added this because I was able to replicate the issue where the readme doesn't load when using the full repo URL but not with the package name. It could just be a quirk or something else. I'm happy to revert this particular change given we don't know how Xcode package searching works.

calvincestari, line 15:
```suggestion
In the dialog that appears, paste the URL of the Apollo iOS GitHub repo (`https://github.com/apollographql/apollo-ios.git`) into the search bar, then select the `apollo-ios` package that appears:
```

AnthonyMDev: Ah interesting! That makes me thing that the Xcode bug has to do with the fact that when you use the git URL it isn't fetching properly there (since it's probably not running the entire SPM package resolution step), but if the module is in the local cache, it can find the README there?

Thanks for reverting that change. The rest looks great!

PR #2906, opened 2023-03-28
Generated inline fragment names now always include inclusion conditions

### LINKED ISSUE #2792: Fix Combined TypeCase/InclusionCondition Code Generation
When selecting a type case and a fragment on that type case with an inclusion condition, you get conflicting generated selection sets.

```graphql
  query Test($a: Boolean!) {
      allAnimals {
        ... on Bird {
          wingspan
        }
        ...FragmentA @include(if: $a)
      }
    }

    fragment FragmentA on Bird {
      a
    }
```
This generates a model with the inline fragments:
```swift
public var asBird: AsBird? { _asInlineFragment() }
public var asBird: AsBird? { _asInlineFragment(if: "getWarmBlooded") }
```

The `ifGetWarmBlooded` should be nested inside of `AsBird`. We've discussed this and the best way forward we believe is to **always** separate inclusion conditions and type cases instead of creating a compound conditional selection set.

So a `SelectionSet` for "As Bird If getWarmBlooded" should be an "As Bird" with a nested "If getWarmBlooded". This ensures that changes to other parts of an operation do not cause breaking changes to your generated models.

This will be a minor breaking change, as the accessors for inline fragments with inclusion conditions will change in the generated models.

AnthonyMDev:
Closes #2792

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
@@ -137,11 +137,11 @@ struct SelectionSetTemplate {
   private func ParentTypeTemplate(_ type: GraphQLCompositeType) -> String {
     """
     public static var __parentType: \(config.ApolloAPITargetName).ParentType { \
-    \(GeneratedTypeReference(type)) }
+    \(GeneratedSchemaTypeReference(type)) }
     """
   }
 
-  private func GeneratedTypeReference(_ type: GraphQLCompositeType) -> TemplateString {
+  private func GeneratedSchemaTypeReference(_ type: GraphQLCompositeType) -> TemplateString {
     "\(config.schemaNamespace.firstUppercased).\(type.schemaTypesNamespace).\(type.name.firstUppercased)"
   }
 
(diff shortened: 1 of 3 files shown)


Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -4148,6 +4148,56 @@ class SelectionSetTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected, atLine: 12, ignoringExtraLines: true))
   }
 
+  func test__render_inlineFragmentAccessor__givenInlineFragmentAndNamedFragmentOnSameTypeWithInclusionCondition_rendersBothInlineFragments() throws {
```
calvincestari, line 4151: 👍🏻

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -748,15 +748,10 @@ fileprivate struct SelectionSetNameGenerator {
...
   var selectionSetNameComponent: String {
-    if let type = type {
-      return "As\(type.name.firstUppercased)"
-    }
-
-    if let conditions = conditions {
-      return "If\(conditions.typeNameComponents)"
-    }
-
-    fatalError("ScopeCondition is empty!")
```
calvincestari, line 759: Would the missing scope condition have been an internal error? If this is encountered again the generated code probably wouldn't build but how would we be able to narrow down on this being the source of that?

calvincestari (approved):
An easier fix than I expected, nice! LGTM!

PR #2908, opened 2023-03-29
fix nil AnyHashable

### LINKED ISSUE #2907: Crash on nil property
### Summary

We've been seeing a crash after using the generated inits and then accessing a nil property. It occurs within DataDict.swift
The crash is on line 78 but I believe the issue is from line 91, where an `AnyHashable?` is unwrapped unsuccessfully and is then passed along as AnyHashable with a value of nil on line 95. I don't know if it's an issue in the inits not managing nil values properly or a swift dynamic casting issue with the functions here.

Screenshot 2023-03-29 at 3 07 06 pm

lldb on line 95:
```
(lldb) po data
▿ AnyHashable(nil)
  - value : nil
```

Here is the full generated model we're trying to initialise. A bunch of properties have been removed to simplify it:

```swift
public struct AppPageHeroSectionData: ShopAPI.SelectionSet, Fragment {
  public static var fragmentDefinition: StaticString { """
    fragment AppPageHeroSectionData on AppPageHeroSection {
      __typename
      id
      personalizedMessaging {
        __typename
        id
        title
      }
    }
    """ }

  public let __data: DataDict
  public init(_dataDict: DataDict) { __data = _dataDict }

  public static var __parentType: ApolloAPI.ParentType { ShopAPI.Objects.AppPageHeroSection }
  public static var __selections: [ApolloAPI.Selection] { [
    .field("__typename", String.self),
    .field("id", ShopAPI.ID.self),
    .field("personalizedMessaging", PersonalizedMessaging?.self),
  ] }

  public var id: ShopAPI.ID { __data["id"] }
  public var personalizedMessaging: PersonalizedMessaging? { __data["personalizedMessaging"] }

  public init(
    id: ShopAPI.ID,
    personalizedMessaging: PersonalizedMessaging? = nil
  ) {
    self.init(_dataDict: DataDict(data: [
      "__typename": ShopAPI.Objects.AppPageHeroSection.typename,
      "id": id,
      "personalizedMessaging": personalizedMessaging._fieldData,
      "__fulfilled": Set([
        ObjectIdentifier(Self.self)
      ])
    ]))
  }

  /// PersonalizedMessaging
  ///
  /// Parent Type: `PersonalizedMessagingTextBlock`
  public struct PersonalizedMessaging: ShopAPI.SelectionSet {
    public let __data: DataDict
    public init(_dataDict: DataDict) { __data = _dataDict }

    public static var __parentType: ApolloAPI.ParentType { ShopAPI.Objects.PersonalizedMessagingTextBlock }
    public static var __selections: [ApolloAPI.Selection] { [
      .field("__typename", String.self),
      .field("id", ShopAPI.ID.self),
      .field("title", String.self),
    ] }

    public var id: ShopAPI.ID { __data["id"] }
    public var title: String { __data["title"] }

    public init(
      id: ShopAPI.ID,
      title: String
    ) {
      self.init(_dataDict: DataDict(data: [
        "__typename": ShopAPI.Objects.PersonalizedMessagingTextBlock.typename,
        "id": id,
        "title": title,
        "__fulfilled": Set([
          ObjectIdentifier(Self.self)
        ])
      ]))
    }
  }
}
```

### Version

Apollo: 1.1.0-beta
Xcode: 14.2
Swift: 5.7.2
iOS: 16.2

Create a SelectionSet that has an optional SelectionSet property, initialise that to nil and access it.
```swift
let section = AppPageHeroSectionData(
  id: "1",
  personalizedMessaging: nil
)
let message = section.personalizedMessaging // expect nil but it crashes
```

### Logs

_No response_

### Anything else?

This may be related to the following 2 issues:
- https://github.com/apollographql/apollo-ios/issues/2668 (Not this is on iOS 16.2)
- https://github.com/apollographql/apollo-ios/issues/2861

yonaskolb:
This attempts to fix #2907, which does the job in our use case, though I don't know if there is a better way to tackle this, as this could probably be fixed up the stack somewhere.

This approach unwraps the data from a double optional if required.

### DIFF
--- a/Sources/ApolloAPI/DataDict.swift
+++ b/Sources/ApolloAPI/DataDict.swift
@@ -88,11 +88,16 @@ extension Optional: SelectionSetEntityValue where Wrapped: SelectionSetEntityVal
   /// - Warning: This function is not supported for external use.
   /// Unsupported usage may result in unintended consequences including crashes.
   @inlinable public init(_fieldData data: AnyHashable?) {
-    guard case let .some(data) = data else {
-      self = .none
-      return
+    switch data {
+      case .none:
+        self = .none
+      case .some(let hashable):
+        if let optional = hashable.base as? Optional<AnyHashable>, optional == nil {
+          self = .none
+          return
+        }
+        self = .some(Wrapped.init(_fieldData: data))
     }
-    self = .some(Wrapped.init(_fieldData: data))
   }
 
   @inlinable public var _fieldData: AnyHashable { map(\._fieldData) }


AnthonyMDev:
Thanks for the fix. After a little investigation, this seems like an appropriate fix! I am going to add another commit on top `main` after merging this in that adds a unit test for this behavior. But I've already tested it on my machine and this code passes my new unit test.

Issue #2861, opened 2023-03-02
Retrieving optional field containing null in cache during ReadWriteTransaction causes crash


dabby-wombo:
### Summary

executing a LocalCacheMutation with an optional field containing `nil` and fetching said optional field during readWriteTransaction causes the the ReadWriteTransaction to crash with:

`Could not cast value of type 'NSNull' (0x1b9055fa8) to 'NSString' (0x1b90518c0).` (In this case the optional field is a `String?`)

This seems to be an issue that was introduced by https://github.com/apollographql/apollo-ios/pull/2784 in an effort to fix the previous crashing [issue](https://github.com/apollographql/apollo-ios/issues/2668) that arose for devices iOS 14.4 and lower.

I can confirm this is the case as reverting this piece of code back to what it was before the PR resolves the crash

```swift
  @inlinable public subscript<T: AnyScalarType & Hashable>(_ key: String) -> T {
    get { _data[key]?.base as! T } (reverting back to _data[key] as! T fixes things)
    set { _data[key] = newValue }
    _modify {
      var value = _data[key] as! T
      defer { _data[key] = value }
      yield &value
    }
  }
```

But seeing as that code was introduces to fix a crash, I dug a bit deeper into understanding what was happening, and it seems that specifically in the case of the `ReadWriteTransaction.updateObject`, we need to set the `GraphQLSelectionSetMapper` to `stripNullValues`.

```swift
accumulator: GraphQLSelectionSetMapper<SelectionSet>(
          stripNullValues: true,
          allowMissingValuesForOptionalFields: true
        )
      )
```

My reasoning here is that since the role of `stripNullValues` is to replace instances of `NSNull` with the scalar `nil` type, this is exactly what we want as the original issue for the crash seems to arise from the fact that Swift isn't able to rightly convert an `NSNull` to an `NSString`, but it can convert a `nil` scalar to an `Optional<String>`

### Version

apollo-ios SDK version: 1.0.7

### Steps to reproduce the behavior

I created a failing unit test in the ReadWriteFromStoreTests as follows:

```
func test_updateCacheMutationWithOptionalField_containiningNull_retrievingOptionalField_returns_nil() {
    // given
    struct GivenSelectionSet: MockMutableRootSelectionSet {
      public var __data: DataDict = DataDict([:], variables: nil)
      init(data: DataDict) { __data = data }
      
      static var __selections: [Selection] { [
        .field("hero", Hero.self)
      ]}
      
      var hero: Hero {
[...]

PR #2913, opened 2023-03-30
fix data dict lookup

yonaskolb:
We're seeing the same crash caused by https://github.com/apollographql/apollo-ios/pull/2784 as noted in https://github.com/apollographql/apollo-ios/issues/2861.

I haven't had the time to delve into why this is occurring, but we were hydrating a model with a mock json string.

This puts in a swift version check to maintain the iOS <= 14.4 fix.

### DIFF
--- a/Sources/ApolloAPI/DataDict.swift
+++ b/Sources/ApolloAPI/DataDict.swift
@@ -24,7 +24,13 @@ public struct DataDict: Hashable {
   }
 
   @inlinable public subscript<T: AnyScalarType & Hashable>(_ key: String) -> T {
-    get { _data[key]?.base as! T }
+    get {
+#if swift(>=5.4)
+        _data[key] as! T
+#else
+        _data[key]?.base as! T
+#endif
+    }
     set { _data[key] = newValue }
     _modify {
       var value = _data[key] as! T
@@ -110,7 +116,13 @@ extension Array: SelectionSetEntityValue where Element: SelectionSetEntityValue 
     guard let data = data as? [AnyHashable?] else {
       fatalError("\(Self.self) expected list of data for entity.")
     }
-    self = data.map { Element.init(_fieldData:$0?.base as? AnyHashable) }
+    self = data.map {
+#if swift(>=5.4)
+        Element.init(_fieldData:$0)
+#else
+        Element.init(_fieldData:$0?.base as? AnyHashable)
+#endif
+    }
   }
 
   @inlinable public var _fieldData: AnyHashable { map(\._fieldData) }


AnthonyMDev:
I merged in #2912 to `main` yesterday that should fix this issue. I'm not sure what issue you're exactly having here. I'm going to need more information, or preferably, a reproduction case here to understand what's going on.

yonaskolb:
#2912 looks like it fixed this crash further up the call stack, but the possible crash in this function still remains. And in our case we can make it crash here via a different path.
I unfortunately don't have time to get a clean reproduction before your planned release of 1.1 tomorrow.
I'd suggest accepting this backwards compatible revert, and then looking into the issue after the release.

AnthonyMDev:
Sure, but my concern there is that it won’t actually be fixed for devices on the earlier versions of Swift. I’d like to try to find the underlying issue.

Can you just describe to me the situation in which this is occurring? Is it when reading models from the cache, initializing them directly with JSON, using the generated initializers, etc.? Anything specific about the shape of the models that this is being seen on, or is it on all models? 

Any info you can provide makes it more likely I can track down the root cause tomorrow! Thanks so much.

yonaskolb:
Yep understand. Removed our fix locally and managed to identify the cause I think - custom scalars. The failure steps for us are:
- initialise a query Data with json using `.init(_data: DataDict)`. (We have to use this one instead of the throwing init, as that fails with missing typename errors as mentioned here https://github.com/apollographql/apollo-ios/issues/2916. Was working fine for that preview including looking up this deeply nested value before 1.1. This may contribute to the issue I'm not sure, but would be a breaking change in behaviour over 1.0.
- reach deep into the tree for a value which works fine
- reach into a `.fragments.ourFragment` which works fine
- reach into a property on that fragment which also happens to be an optional custom scalar: `public typealias Cents = Int`
   - I tried accessing a different non-nil optional and that worked fine, so guessing the issue lies in custom scalars
- crashes within the function edited here on this line `_data[key]?.base as! T`

LLDB reports back the following on the stack frame:
```
po _data[key]
▿ Optional<AnyHashable>
  ▿ some : AnyHashable(4800)
    - value : 4800
  
po _data[key]?.base
▿ Optional<Any>
  - some : 4800
```

AnthonyMDev:
In my testing, I have not been able to reproduce this issue, but I think it's likely due to using the `DataDict` initializer. I'm wondering if the JSON you are passing in is inferring the type of that number as something other than an `Int`? Or what is going on, but part of what happens when you use the safe JSON initializer is that values are coerced to the correct types. Particularly, numbers are converted to the right type (Int, Float, Double, etc.)

I strongly recommend that you adjust your JSON data to include the `__typename` fields so you can start using the proper initializers here. If that is not possible for your use case I'm still unclear why that is.

I'm not against merging this PR in, because for Swift 5.4+, it probably is actually better for performance to not call `base`. I'd like to get the opinion of @calvincestari on this first though.

But if this is an actual bug, this PR doesn't fix it for versions prior to Swift 5.4. If the bug you are seeing is because you are using the unsafe initializer, we then I would say it's not really a bug. If using the supported JSON initializer you are still seeing this behavior, I would be very very interested in reproducing and correcting that.

AnthonyMDev (approved):
This is probably better for performance so we'll merge it in.

yonaskolb:
All makes sense. We’ll try and reproduce once we’ve updated all our mocks. Thanks for the merge.

Strange that the mocks used to work, even though they weren’t supposed to 🤷‍♂️

Issue #2302, opened 2022-06-08
[Swift Codegen] generates 'public' modifier is redundant for type alias declared in a public extension


kimdv:
## Bug report

There is a lot of `'public' modifier is redundant for type alias declared in a public extension` warning which makes it noisy. 

```swift
public extension CustomerGraphQL {
  public typealias ID = String // warning here

  public typealias SelectionSet = CustomerGraphQL_SelectionSet // warning here

  public typealias InlineFragment = CustomerGraphQL_InlineFragment // warning here

  public enum Schema: SchemaConfiguration { // warning here
    public static func objectType(forTypename __typename: String) -> Object.Type? {
      switch __typename {
      case "Query": return CustomerGraphQL.Query.self
      default: return nil
      }
    }
  }
}
```

## Versions

Please fill in the versions you're currently using: 

- `apollo-ios` SDK version: [1.0.0 Alpha 6](https://github.com/apollographql/apollo-ios/releases/tag/1.0.0-alpha.6)
- Xcode version: 13.4.1
- Swift version: 5.6.1
- Package manager: 5.6.1

## Steps to reproduce

Our generation command

```swift
  struct GenerateCode: ParsableCommand {
    static var configuration = CommandConfiguration(
      commandName: "generate",
      abstract: "Generates swift code from your schema + your operations based on information set up in the `GenerateCode` command.")

    func run() throws {
      /// The root of the target for which you want to generate code.
      let targetRootURL = SourceRootURL
        .apollo.childFolderURL(folderName: "customer-ios/graphql")

      /// The URL where the generated schema files will be written to.
      let schemaModuleURL = SourceRootURL
        .apollo.childFolderURL(folderName: "customer-ios/graphql")

      // Make sure the folders exists before trying to generate code.
      try FileManager.default.apollo.createDirectoryIfNeeded(atPath: targetRootURL.path)
      try FileManager.default.apollo.createDirectoryIfNeeded(atPath: schemaModuleURL.path)

      // Create the Codegen configuration object. For all configuration parameters see: https://www.apollographql.com/docs/ios/api/ApolloCodegenLib/structs/ApolloCodegenConfiguration/
      let codegenConfiguration = ApolloCodegenConfiguration(
        schemaName: "GraphQLApi", input: ApolloCodegenConfiguration.FileInput(schemaPath: SchemaOutputURL.path, searchPaths: ["\(targetRootURL.path)/**/*.graphql"]),
[...]

PR #2917, opened 2023-03-30
feat: Codegen config access modifier

### LINKED ISSUE #2630: Add option to generate objects with internal access modifier
## Feature request

Add option to generate objects with internal access modifier.

## Motivation

All of the objects generated by Apollo codegen currently have the `public` access modifier. This setup is great if you need to use generated objects across module boundaries, but if you do not then all of that public code can negatively affect both compile times and app size. Adding a codegen option to make these objects `internal` instead of `public` would fix these problems.

I believe this request is the same as #149 which was closed and completed, but the functionality does not yet exist in 1.x.

calvincestari:
Closes #2630 

Introduces an `accessModifier` associated value to `ApolloCodegenConfiguration` properties where the generated Swift code can have the access level specified by the user.

* Built with defaults that match the generated code before this PR
* Built with optional decoding so that users can use their existing JSON configuration files without error


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -250,7 +255,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       /// - Note: Generated files must be manually added to your application target. The generated
       /// schema types files will be namespaced with the value of your configuration's
       /// `schemaNamespace` to prevent naming conflicts.
-      case embeddedInTarget(name: String)
+      case embeddedInTarget(name: String, accessModifier: AccessModifier = .internal)
```
calvincestari, line 258: It might seem odd that the default is `.internal` but if you look at the tests that have changed there were tests that named `xyz_withPublicModifier` and `xyz_noPublicModifier`. `withPublicModifier` was used for the tests related to a `.swiftPackageManager` module type and `noPublicModifier` was used for tests related to `.embeddedInTarget` module type. Since we weren't specifying any access modifier all the generated code would have implicitly had `internal` access level.

The caveat to this is the namespace enum, which we used to always generate as `public` - that was an error though and not aligned with the rest of the generated code for a `.embeddedInTarget` module type.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -103,7 +103,7 @@ extension TemplateRenderer {
     \(ImportStatementTemplate.SchemaType.template(for: config))
 
     \(ifLet: detachedTemplate, { "\($0)\n" })
-    \(ifLet: namespace, template.wrappedInNamespace(_:), else: template)
+    \(ifLet: namespace, { template.wrappedInNamespace($0, accessModifier: embeddedAccessControlModifier(target: target)) }, else: template)
```
calvincestari, line 106: Not the prettiest but `wrappedInNamespace` requires to know the access modifier too now.

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
```diff
@@ -744,4 +746,57 @@ class ApolloCodegenConfigurationCodableTests: XCTestCase {
       try JSONDecoder().decode(ApolloCodegenConfiguration.APQConfig.self, from: subject)
     ).to(throwError())
   }
+
+  // MARK: - Optional Tests
```
calvincestari, line 750: I would like to add more tests for optional properties of the JSON decoding - will do in a follow-up PR.

Tests/ApolloCodegenTests/CodeGeneration/Templates/InterfaceTemplateTests.swift
```diff
@@ -51,23 +51,6 @@ class InterfaceTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected))
   }
 
-  // MARK: Class Definition Tests
-
-  func test_render_givenModuleType_swiftPackageManager_generatesSwiftClassDefinition_withPublicModifier() {
```
calvincestari, line 56: This test was no longer doing what it was originally intended for.

calvincestari:
Fixing the inner access control modifiers was brutal.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -6,17 +6,20 @@ struct SelectionSetTemplate {
   let isMutable: Bool
   let generateInitializers: Bool
   let config: ApolloCodegen.ConfigurationContext
+  let accessControlRenderer: () -> String
```
calvincestari, line 9: I'm not a huge fan of this and would appreciate any other ideas on a solution.

tl;dr - `TemplateRenderer` holds the information necessary to determine what the correct access control string should be. `SelectionSetTemplate` does not conform to `TemplateRenderer` though so it needs a way to have it's 'parent' template provide the access control string when needed. You'll find the same pattern on `OperationDefinitionTemplate`.

calvincestari:
Putting this back into draft because there is a regression re. https://github.com/apollographql/apollo-ios/issues/2302.

Sources/ApolloCodegenLib/Templates/OperationDefinitionTemplate.swift
```diff
@@ -55,13 +58,15 @@ struct OperationDefinitionTemplate: OperationTemplateRenderer {
       _ operation: CompilationResult.OperationDefinition,
       identifier: @autoclosure () -> String,
       fragments: OrderedSet<IR.NamedFragment>,
-      config: ApolloCodegen.ConfigurationContext
+      config: ApolloCodegen.ConfigurationContext,
+      accessControlRenderer: @autoclosure () -> String
```
calvincestari, line 62: This is a static function and therefore cannot get access to any of the state from the caller so the function to generate the access control modifier has to be passed in as a closure so it can be called when it's needed.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -141,20 +143,99 @@ extension TemplateRenderer {
...
+fileprivate extension ApolloCodegenConfiguration.AccessModifier {
+  var swiftString: String {
+    switch self {
+    case .public: return "public " // there should be no spaces in these strings
+    case .internal: return ""
+    }
+  }
+}
+
+enum AccessControlModifierTypeDefinition {
```
calvincestari, line 159: I'm open to renaming suggestions for these cases. They represent this structure:
```
<namespace> {
  <parent> {
    <member>
    <member>
  }
}
```
When there is a `namespace` (`enum` and it's `extension`) that is `public` the `parent` cannot have the `public` modifier whereas the `member` should all get the `public` modifier.

Tests/TestCodeGenConfigurations/EmbeddedInTarget-RelativeAbsolute/PackageTwo/Package.swift
```diff
@@ -22,15 +22,15 @@ let package = Package(
       name: "PackageTwo",
       dependencies: [
         .product(name: "ApolloAPI", package: "apollo-ios"),
+      ],
+      swiftSettings: [
+        .unsafeFlags(["-warnings-as-errors"])
```
calvincestari, line 27: Needed to be moved into the shared package build settings, not the test target build settings.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -103,7 +103,7 @@ extension TemplateRenderer {
     \(ImportStatementTemplate.SchemaType.template(for: config))
 
     \(ifLet: detachedTemplate, { "\($0)\n" })
-    \(ifLet: namespace, template.wrappedInNamespace(_:), else: template)
+    \(ifLet: namespace, { template.wrappedInNamespace($0, accessModifier: accessControlModifier(target: target, definition: .namespace)) }, else: template)
```
calvincestari, line 106: Not the prettiest but wrappedInNamespace requires to know the access modifier too now.

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
BobaFetters: Honestly I think these cases make sense, `namespace` and `member` especially I think reflect what they represent well, `parent` would be the only one that could _maybe_ change but I'm struggling to think of what would maybe better represent that.

calvincestari: Change `AccessControlModifierTypeDefinition` to `AccessControlScope`

Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -141,20 +143,99 @@ extension TemplateRenderer {
...
+  case namespace
+  case parent
+  case member
+}
+
+extension TemplateRenderer {
+  func accessControlModifier(
+    target: TemplateTarget,
+    definition: AccessControlModifierTypeDefinition
+  ) -> String {
```
AnthonyMDev, line 169: I'd recommend inverting the parameter order and renaming this: 
```swift
accessControlModifier(
  for scope: AccessControlScope,
  in target: TemplateTarget
) -> String
```

Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
```diff
@@ -12,20 +12,21 @@ struct InputObjectTemplate: TemplateRenderer {
 
   var template: TemplateString {
     let (validFields, deprecatedFields) = filterFields(graphqlInputObject.fields)
+    let memberAccessControl = accessControlModifier(target: target, definition: .member)
```
calvincestari, line 15: Reduce into a function on `TemplateRenderer` to only accept scope and take in target from each template.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
calvincestari: Consider renaming to `renderAccessControl()`

lhunath:
Please see https://github.com/apollographql/apollo-ios/issues/3066 for a regression likely related to this change.

PR #2920, opened 2023-03-31
tests: Regenerate test projects

calvincestari:
This runs the internal test projects through the latest codegen engine.


calvincestari:
Superceeded by https://github.com/apollographql/apollo-ios/pull/2991.

PR #2924, opened 2023-04-03
update: HTTP subscriptions protocol update

calvincestari:
Changes per the protocol spec to:
- remove `done` from JSON test data
- handle null `payload` data
- end processing when a fatal error is received
- fix crash at end of steam terminator

### DIFF
--- a/Sources/Apollo/MultipartResponseParsingInterceptor.swift
+++ b/Sources/Apollo/MultipartResponseParsingInterceptor.swift
@@ -113,12 +113,6 @@ public struct MultipartResponseParsingInterceptor: ApolloInterceptor {
             )
           }
 
-          if let done = object["done"] as? Bool, done {
-            // Exit at this point because the router will close the connection; errors would have
-            // been reported or the subscription is complete.
-            return
-          }
-
           guard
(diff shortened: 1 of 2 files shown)


Sources/Apollo/MultipartResponseParsingInterceptor.swift
```diff
@@ -111,44 +110,37 @@ public struct MultipartResponseParsingInterceptor: ApolloInterceptor {
...
-          guard
-            let payload = object["payload"] as? JSONObject,
-            let data: Data = try? JSONSerializationFormat.serialize(value: payload)
-          else {
-            chain.handleErrorAsync(
-              MultipartResponseParsingError.cannotParsePayloadData,
-              request: request,
-              response: response,
-              completion: completion
+          if let payload = object["payload"] as? JSONObject {
```
AnthonyMDev, line 118: It looks here like if there is no `payload` that we don't throw an error or proceed onto the next part of the chain. This seems wrong to me. Is that behavior correct?

AnthonyMDev (approved):
Just uncertain if we are missing handling something when there is no `payload` but if that is the right thing to do, then this LGTM.

Issue #2928, opened 2023-04-04
In 1.1.0, passing custom scalars or GraphQLEnum to mocks fails


vincentisambart:
### Summary

In Apollo iOS 1.1.0, when you pass enum or custom scalar values to create test mocks, you get a JSON decoding error:

```
ApolloTestSupport/TestMock.swift:93: Fatal error: 'try!' expression unexpectedly raised an error: Apollo.GraphQLExecutionError(path: topFrameContent.align, underlying: ApolloAPI.JSONDecodingError.couldNotConvert(value: AnyHashable(ApolloAPI.GraphQLEnum<Api4.BannerAlign>.case(Api4.BannerAlign.center)), to: Swift.String))
```

For custom scalars, I could work around this problem by modifying their `init(_jsonValue value: JSONValue) throws` to allow `value` being of their own type(`if let myCustomScalar = value as? MyCustomScalar`), but I cannot do that to `ApolloAPI.GraphQLEnum`.

At first thought it was me incorrectly use the Apollo iOS API, but the `convenience init` is expecting concrete values of the custom scalars and GraphQLEnum, so I looks like a bug to me.

### Version

1.1.0

### Steps to reproduce the behavior

You need a type that has fields with enums and/or custom scalars.

```graphql
enum Alignment {
  LEFT
  CENTER
  RIGHT
} 
type Banner {
  alignment: Alignment
}
```

And in Swift the following test code will work on Apollo iOS 1.0 but will end up with a `JSONDecodingError` in 1.1.
```swift
let mock = Mock<Banner>(alignment: GraphQLEnum(Banner.center))
```

Sorry I don't have the time to make a full reproduction project so the code above was not tested, but should fail seeing the behavior on our project.

AnthonyMDev:
Thanks so much for the report! I'll look into this and get a fix out!

AnthonyMDev:
I've got a unit test reproducing this now and I'm working on a fix!

vincentisambart:
Thank you for the quick fix!

RussellToon:
And I still seem to have this issue in 1.1.3

calvincestari:
@alexjameslittle, @RussellToon - it would help if either of you are able to provide a sample application that demonstrates the issue you're still experiencing.

alexjameslittle:
I'm not sure I have the project available to provide an example anymore, however we added this based on some inspiration from this repo and it solved our issues. We now no longer have these issues

```swift
@testable import Apollo
[...]

PR #2939, opened 2023-04-06
Fix test mock custom scalar and double nested array execution

### LINKED ISSUE #2809: Crash when using mocks for Double Nested Arrays
### Summary

I have a schema that looks something like this: 

```
type InnerObject { 
    propertyOne: Int!
    propertyTwo: Int!
}

type OuterObject { 
    propertyOne: Int!
    propertyTwo: [[InnerObject!]!]!
}
```

And I have ran codegen for this and created the mocks. When using the mocks in tests, everything compiles fine. However when i set the value of "propertyTwo" two in mocks, and then convert it to a (non-mock) graphQL object using .from(Mock<OuterObject>)  and then try to access "propertyTwo" in code, it crashes with the error
`fatalError("\(Self.self) expected data for entity.")`

### Version

1.0.5

- Use a schema with a property of type:
```
[[SomeObject!]!]!
```
- Generate mocks / code from it. 
- Initialize an object for the containing type use `.from(Mock<>)`
- try to access the property of double nested array type.
- Notices the crash

### Logs

_No response_

### Anything else?

Screenshot 2023-02-03 at 8 35 41 PM

Here's what the call stack looks like.

AnthonyMDev:
Fixes #2809 and #2928

### DIFF
--- a/Sources/ApolloTestSupport/TestMockSelectionSetMapper.swift
+++ b/Sources/ApolloTestSupport/TestMockSelectionSetMapper.swift
@@ -1,0 +1,50 @@
+#if !COCOAPODS
+@testable import ApolloAPI
+#endif
+@testable import Apollo
+import Foundation
+
+/// An accumulator that converts data from a `Mock` to the correct values to create a `SelectionSet`.
+final class TestMockSelectionSetMapper<T: SelectionSet>: GraphQLResultAccumulator {
+
+  let underlyingMapper = GraphQLSelectionSetMapper<T>(handleMissingValues: .allowForAllFields)
+
+  func accept(scalar: AnyHashable, info: FieldExecutionInfo) throws -> AnyHashable? {
+    return scalar
+  }
+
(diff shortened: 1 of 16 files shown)


Sources/ApolloTestSupport/TestMockSelectionSetMapper.swift
```diff
@@ -0,0 +1,50 @@
+#if !COCOAPODS
+@testable import ApolloAPI
+#endif
+@testable import Apollo
```
calvincestari, line 4: This needs to be moved up into the `#if !COCOAPODS` directive otherwise the cocoapods spec linting will fail with a warning that the file is already part of the `Apollo` module. [Part of what was fixed](https://github.com/apollographql/apollo-ios/pull/2937/files#diff-b3d3d639f815a2b3dcee78379825c377b4036869559a53c6eeb302aa1587fccfL3) for #2937.

calvincestari (changes requested):

PR #2948, opened 2023-04-11
Add injecting additionalErrorHandler for upload operations to RequestChainNetworkTransport

RobertDresler:
Currently `additionalErrorHandler ` is not being injected to `chain` for `Upload` operations. This PR fixes it and creates `makeChain(operation:callbackQueue:)` method which creates `chain` for `send` and `upload` the same way.


calvincestari (approved):
Looks good, thanks for the contribution @RobertDresler.

calvincestari:
I added a test to validate that the new error handler is being called for `upload` requests.

PR #2952, opened 2023-04-12
fix: Apollo client operation cache test

### LINKED ISSUE #2929: test: Fix `test__performMutation_givenPublishResultToStore_false_doesNotPublishResultsToStore`
### Summary

`ApolloClientOperationTests.test__performMutation_givenPublishResultToStore_false_doesNotPublishResultsToStore()` is not a valid test for the following reasons and needs to be fixed.

- `NoCache()` is called in the initializer despite the test to check for the presence of cache.
  - https://github.com/apollographql/apollo-ios/blob/10bc14b66ac677a70ad4ec9fe3bd04c82461beed/Tests/ApolloTests/ApolloClientOperationTests.swift#L36
- If I change `publishResultToStore` to true, the test still passes. Essentially, the test should fail because it is cached.
  - https://github.com/apollographql/apollo-ios/blob/10bc14b66ac677a70ad4ec9fe3bd04c82461beed/Tests/ApolloTests/ApolloClientOperationTests.swift#L80
  - I have confirmed that it is failing with the `JSONResponseParsingInterceptor`, but I do not know the detailed cause.
     - https://github.com/apollographql/apollo-ios/blob/10bc14b66ac677a70ad4ec9fe3bd04c82461beed/Sources/Apollo/JSONResponseParsingInterceptor.swift#L64-L69

### Version

1.1.0

test `test__performMutation_givenPublishResultToStore_false_doesNotPublishResultsToStore()`.

### Logs

```shell
Error caught by the `JSONResponseParsingInterceptor`.


po error
▿ Variable "review" was not provided.
     ▿ object : 1 element
       ▿ 0 : 2 elements
         - key : "message"
         ▿ value : AnyHashable("Variable \"review\" was not provided.")
           - value : "Variable \"review\" was not provided."
```
```


### Anything else?

_No response_

calvincestari:
Fixes #2929 

* Remove arguments on the mocked selection set; not relevant to the test and were causing hidden failure because the argument values were not provided when executed.
* Fail on `.failure` result.

### DIFF
--- a/Tests/ApolloTests/ApolloClientOperationTests.swift
+++ b/Tests/ApolloTests/ApolloClientOperationTests.swift
@@ -45,8 +45,7 @@ final class ApolloClientOperationTests: XCTestCase {
     // given
     class GivenSelectionSet: MockSelectionSet {
       override class var __selections: [Selection] { [
-        .field("createReview", CreateReview.self,
-               arguments: ["episode": .variable("episode"), "review": .variable("review")])
+        .field("createReview", CreateReview.self)
       ] }
 
       class CreateReview: MockSelectionSet {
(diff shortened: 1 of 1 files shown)


calvincestari:
@tokizuoh - please confirm that the test behaves how you expect and I'll merge this PR.

tokizuoh:
@calvincestari 
Thanks for the fix.
This is the test I had hoped for!

PR #2953, opened 2023-04-13
test: Apollo client operation cache test when `publishResultToStore` is true

tokizuoh:
ref: #2952, #2929

- Add a test if the `publishResultToStore` argument to `ApolloClient.perform` is true, i.e. cached after perform, and `publishedRecordSets` in `MockStore` is not empty.

### DIFF
--- a/Tests/ApolloTests/ApolloClientOperationTests.swift
+++ b/Tests/ApolloTests/ApolloClientOperationTests.swift
@@ -92,7 +92,21 @@ final class ApolloClientOperationTests: XCTestCase {
     self.wait(for: [serverRequestExpectation, performResultFromServerExpectation], timeout: 0.2)
 
     // then
-    expect(self.store.publishedRecordSets).notTo(beEmpty())
+    expect(self.store.publishedRecordSets.count).to(equal(1))
+    
+    let actual = self.store.publishedRecordSets[0]
+    expect(actual["MUTATION_ROOT"]).to(equal(
+      Record(key: "MUTATION_ROOT", [
+        "createReview": CacheReference("MUTATION_ROOT.createReview")
+      ])
+    ))
+    expect(actual["MUTATION_ROOT.createReview"]).to(equal(
+      Record(key: "MUTATION_ROOT.createReview", [
+        "__typename": "Review",
+        "stars": 3,
+        "commentary": ""
+      ])
+    ))
   }
 
   func test__performMutation_givenPublishResultToStore_false_doesNotPublishResultsToStore() throws {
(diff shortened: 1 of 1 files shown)


Tests/ApolloTests/ApolloClientOperationTests.swift
```diff
@@ -41,21 +41,59 @@ final class ApolloClientOperationTests: XCTestCase {
...
-          .field("stars", Int.self),
-          .field("commentary", String?.self)
-        ] }
-      }
+  func test__performMutation_givenPublishResultToStore_true_publishResultsToStore() throws {
+    let mutation = MockMutation<GivenSelectionSet>()
+    let resultObserver = self.makeResultObserver(for: mutation)
+
+    let serverRequestExpectation = self.server.expect(MockMutation<GivenSelectionSet>.self) { _ in
+      [
```
calvincestari, line 64: This looks like common data between both tests and could be pulled out into the test class.

Tests/ApolloTests/ApolloClientOperationTests.swift
```diff
@@ -41,21 +41,59 @@ final class ApolloClientOperationTests: XCTestCase {
...
+
+    // when
+    self.client.perform(mutation: mutation,
+                        publishResultToStore: true,
+                        resultHandler: resultObserver.handler)
+
+    self.wait(for: [serverRequestExpectation, performResultFromServerExpectation], timeout: 0.2)
+
+    // then
+    expect(self.store.publishedRecordSets).notTo(beEmpty())
```
calvincestari, line 93: I think it would be better to test specifically for a single entry into the store and for the record that was added. This ensures that if somehow there is a different record written to the store this test will then fail.

calvincestari (changes requested):
Thanks for the contribution @tokizuoh! Some minor cleanup still to do but great idea to add the test - thank you.

Tests/ApolloTests/ApolloClientOperationTests.swift
tokizuoh: Indeed.
fixed: [92c233a](https://github.com/apollographql/apollo-ios/pull/2953/commits/92c233a940ad1236e2c9d6e71055e86e1a30e2ba)

Tests/ApolloTests/ApolloClientOperationTests.swift
tokizuoh: Thanks for your suggestion!
fixed: [5b9df5f](https://github.com/apollographql/apollo-ios/pull/2953/commits/5b9df5ff6c1bc1389c7d8a3cb5fa85e16b800708)

PR #2956, opened 2023-04-14
Fully Qualify name of RootEntityType and mergedSources

### LINKED ISSUE #2949: RootEntityType of some inline fragments has a wrong type
### Summary

I've tried migrating from 1.0.7 to 1.1.1, but stumbled upon an issue with generated models

Some inline fragments generated using `ApolloCodegen` 1.1.1 point to a type that doesn't exist in that context resulting in a `'...' is not a member type of struct 'GQLClient.SomeQuery.Data....'` compile-time error.

### Version

1.1.1

The issue does not exhibit with all models, so far it seems like its only inline union types:

```graphql
query Article($id: ID!) {
  article(where: { id: $id }) {
    article {
      ... on SomeArticle {
        id
        title
      }
    }
  }
}
```
(made up query that's not far off the one we're having issue with, happy to provide the concrete query in private channels)

Maybe worth mentioning that some root fields share the same name as our query types, maybe that's confusing the code gen?

Providing the full path to the type fixes the issue, eg. replacing `public typealias RootEntityType = Article.Article` with `public typealias RootEntityType = ArticleQuery.Data.Article`.

### Logs

_No response_

### Anything else?

In one query we have the same issue with a type in `__mergedSources`.

AnthonyMDev:
Fix #2949


calvincestari (approved):
Glad we could get this fixed so quickly.

PR #2957, opened 2023-04-14
[WIP] Update codegen module to support building on linux 

eklipse2k8:
I realize this isn't a very common workflow for iOS development, but as part of our build, we rely on Docker for quite a large portion of it. Although I think there's probably a day or more work to get this building, I wanted to throw this up to get any early feedback and thoughts. 

## Goal

command line code-gen tool can be moved over to linux build pipeline since it generates code as part of a Docker build pipeline

## Example

```sh
docker buildx build . -t ghcr.io/apollographql/apollo-ios-cli:1.1.1 --load
```

Then instead of needing to build the tool, or include it via any other way, one can simply add this to their build flow to grab exact versions or "latest"

```sh
docker run ghcr.io/apollographql/apollo-ios-cli:1.1.1 
```

## Current Status

- [x] InflectorKit -> pluralize
- [x] CryptoKit -> apple-swift-crypto
- [ ] JavaScriptCore -> JXCore

Note: JavaScriptCore bits seem the most finicky so I wanted to stop, get some feedback before diving further down that road.

There is no objective-c runtime on linux and so some of that would have to be refactored.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM swift:latest as builder
+WORKDIR /app
+COPY . .
+
+RUN apt-get update && apt-get install -y libjavascriptcoregtk-4.0-dev \
+    && pkg-config --libs javascriptcoregtk-4.0
+
+RUN swift build --product apollo-ios-cli -c release
+
+CMD ["/app/.build/release/apollo-ios-cli"]
--- a/Package.swift
+++ b/Package.swift
@@ -35,6 +35,9 @@ let package = Package(
     .package(
       url: "https://github.com/apple/swift-argument-parser.git",
       .upToNextMajor(from: "1.2.0")),
+    .package(
+      url: "https://github.com/apple/swift-crypto.git",
+      "1.0.0"..<"3.0.0"),
   ],
   targets: [
     .target(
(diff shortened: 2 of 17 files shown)


AnthonyMDev:
This is something we've been wanting to do eventually, it just hasn't been a priority. So if you want to continue working on it, we are definitely happy to support you on this and pull it in when it's finished!

Regarding the JavaScriptCore issue, we've been hoping for a while to get the time to migrate the entire codegen engine to use the [GraphQLSwift compiler](https://github.com/GraphQLSwift/GraphQL) eventually. That might be a really good step towards making this work. But that is a pretty big lift on its own.

AnthonyMDev:
We also have some plans to pull the codegen lib into it's own repo and Swift Package at some point this year. So just a heads up that this work might have to migrate over to that repo. Let's keep in contact as you continue to work on this so we can make the path forward as painless as possible.

calvincestari:
I'm going to convert this PR to a draft so there is no mistake on when to review/merge it. If anyone strongly objects let me know and we can discuss it - thanks.

AnthonyMDev:
We're going to be closing this PR due to inactivity. At some point, we would love to have this implemented, but there are some changes we are making to codegen right now anyways, so we may want to re start this work from a clean point in the future anyways.

As for the Javascript bridging, that seems to be the biggest blocker here. I've been in chats with the team who maintain the [Swift GraphQL library](https://github.com/GraphQLSwift/GraphQL) and I'm hoping that in the near future it will have the necessary capabilities for us to migrate over to that. This would remove the JS issues.

PR #2961, opened 2023-04-16
test: use XCTAssertThrowsError to test that an error is thrown

tokizuoh:
Change to use `XCTAssertThrowsError` to ensure that the `GraphQLFile` initializer throws an error.


calvincestari:
@tokizuoh - what is the benefit to this change? I've tested both of the current test outcomes and they appear to work as expected.

tokizuoh:
@calvincestari 
As you say, it is well tested in its current form.

However, as it stands now, if the behavior of the GraphQLFile initializer changes in the future and errors are no longer thrown, they will not be detected.

Using XCTAssertThrowsError to check the error thrown itself also makes the test more robust.

calvincestari:
Fair enough, thanks for taking the time to improve the test.

calvincestari (approved):

PR #2965, opened 2023-04-17
fix: `@dynamicMember` conflicting field name

### LINKED ISSUE #2950: Generated test Mock code produces compilation error when object contains property named "hash"
### Bug report
Currently, there's a compilation error in generated mock code when the object contains a property named "hash". The compiler throws an error inside the convenience initializer of the mock object, when it tries to assign a value to the "hash" property.

This is the the error message: `'let' property 'hash' may not be initialized directly; use "self.init(...)" or "self = ..." instead`

### Versions
- `apollo-ios` SDK version: 1.1.1
- `apollo-ios-cli` version: 1.1.1
- Xcode version: 14.3
- Swift version: 5.8

### Steps to reproduce
1. Execute code generation command using `apollo-ios-cli` using a schema with a response object that has a property named `hash`.

### Further details
Here's the generated code (all other properties have been redacted) that it generates to mock a response object of an operation:

```swift
// @generated

import ApolloTestSupport
import ApplinGraphQL

public class SomeGenerated__ApolloType: MockObject {
  public static let objectType: Object = ApplinGraphQL.Objects.SomeGenerated__ApolloType
  public typealias MockValueCollectionType = Array<Mock<SomeGenerated__ApolloType>>

  public struct MockFields {
    // beginning of properties...

    @Field<String>("hash") public var hash

    // rest of the properies...
  }
}

public extension Mock where O == SomeGenerated__ApolloType {
  convenience init(
    // params...

    hash: String? = nil,

    // rest of params...
  ) {
    self.init()
    // properties...

    self.hash = hash  // This causes "'let' property 'hash' may not be initialized directly; use "self.init(...)" or "self = ..." instead"

    // rest of properties...
  }
}
```

Manually editing and renaming this property to anything -- say, `_hash` -- already solves the problem. But this is obviously not a solution as the file will be overwritten every time you run the code generation script.

### Version

1.1.1

1. Execute code generation command using `apollo-ios-cli` using a schema with a response object that has a property named `hash`

### Logs

_No response_

### Anything else?

_No response_

calvincestari:
Fixes #2950 

* Introduces a new set of field names for which properties are generated, overriding the `@dynamicMember` subscript.
* Changes `Mock` subscript setters to use a common private `_set` function that can be called from initializers too.


Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -68,6 +69,19 @@ struct MockObjectTemplate: TemplateRenderer {
     """
   }
 
+  private func conflictingFieldNameProperties(_ fields: [TemplateField]) -> TemplateString {
+    """
+    \(fields.map { """
```
AnthonyMDev, line 74: Nit: might be more performant to compute an intersection with the `TestMockConflictingFieldNames` set and then just map over those? I could be wrong, but I believe that set intersection computation is very fast b/c of hashing algorithms. This is so so minor though, I'm not really worried about making the change.

AnthonyMDev (approved):
Thanks so much for getting this one out @calvincestari!

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
calvincestari: Nothing about `fields` nor the `FieldCollector` is a `Set` unfortunately, I don't think we're going to get much benefit out of going that route here.

Issue #2958, opened 2023-04-15
Disallow a target name of `apollo` when using a module type of `embeddedInTarget(name:)`


moopoints:
### Summary

I'm running into the same issue as previously reported [here](https://github.com/apollographql/apollo-ios/issues/2859): 

The suggestion offered by Calvin in issue #2859 isn't working for me - the SchemaNamespace isn't accessible either. 

**Environment**
- Xcode 14.2
- Apollo iOS 1.1.2

### Version

1.1.2

### Steps to reproduce the behavior

- Brand new SwiftUI project using Xcode 14.2
- Add Apollo SDK via SPM
- Add schema.graphls, set up config, generate
- Add files to Xcode project
- Attempt to access either TestAPI (SchemaNamespace) or any Query

Here is [a link](https://www.dropbox.com/s/m63p140xhroolkp/20230415%20apollo.zip?dl=0) to a sample project very closely following the setup instructions provided [here](https://www.apollographql.com/docs/ios/get-started/#5-create-an-apolloclient). 

### Anything else?

https://share.cleanshot.com/WZ52Fdkw

```[tasklist]
```

AnthonyMDev:
> @AnthonyMDev - we may need to consider adding this to our list of reserved words for schema/target to prevent this in the future.

Yes, that seems like a good idea. Thanks for taking a look at this @calvincestari

moopoints:
Appreciate it @calvincestari.

> re-added it and now all the generated types will be compiled into the target

Is there a trick to how you're adding these files to make sure they're added to the compile sources and how do you recommend handling this each time one regenerates the files after updating queries?

calvincestari:
> Is there a trick to how you're adding these files to make sure they're added to the compile sources

No, nothing special in how they're added to the project: right-click on the project folder that I want them added to, select "Add files to <project>...", select the folder to be added, and ensure that the correct target is selected (this is what adds them to the "Compile Sources" build phase).

> how do you recommend handling this each time one regenerates the files after updating queries?

This is where careful selection of the [`moduleType` property](https://www.apollographql.com/docs/ios/code-generation/codegen-configuration#module-type) for code generation comes into play.
* By choosing `embeddedInTarget` you're stating that you will be responsible for management of the generated files. So if you have already added a generated file to your project it will simply be updated upon regeneration but still a part of the project. However if a new file is generated you will need to manually add it to the target before it can be compiled into your project.
[...]

PR #2972, opened 2023-04-20
fix: Disallow certain targetNames in code generation (#2958)

BobaFetters:
Added new validation check to ensure the `targetName` given when using the `embeddedInTarget` module type does not conflict with existing libraries.


calvincestari (approved):
This looks good, thanks for the quick fix!

I questioned whether we should have another test in `CodegenCLITests::InitializeTests` to validate the error is thrown to the command line during initialization but that validation is in another module with it's own test so I'm fine with it not being in the CLI tests.

PR #2973, opened 2023-04-21
test: Measure `MultipartResponseParsingInterceptor` performance

### LINKED ISSUE #2971: test: Add performance test for multipart response parsing
Getting this in place to measure improvements to the parsing logic and ensure no regression in parsing performance when `@defer` support is implemented.

calvincestari:
Closes #2971 

* Adds a performance test focused on the response data parsing of `MultipartResponseParsingInterceptor`. This gives us a benchmark to improve the parsing performance of multipart responses and test the impact of `@defer` implementation to the multipart processing.
* Changes performance test configuration to fail when no benchmark is found. I believe this is better default behaviour rather than letting an unaware developer believe the test passed within thresholds.
* Refactors a mock request chain for better general usage

### DIFF
--- a/Apollo.xcodeproj/xcshareddata/xcbaselines/9F54C8B3255D760B0065AFD6.xcbaseline/B07DC9C5-A477-43FC-8906-C01C8B98086A.plist
+++ b/Apollo.xcodeproj/xcshareddata/xcbaselines/9F54C8B3255D760B0065AFD6.xcbaseline/B07DC9C5-A477-43FC-8906-C01C8B98086A.plist
@@ -11,7 +11,7 @@
 				<key>com.apple.XCTPerformanceMetric_WallClockTime</key>
 				<dict>
 					<key>baselineAverage</key>
-					<real>0.003480</real>
+					<real>0.020300</real>
 					<key>baselineIntegrationDisplayName</key>
 					<string>Local Baseline</string>
 				</dict>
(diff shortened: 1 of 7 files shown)


Apollo.xcodeproj/xcshareddata/xcbaselines/9F54C8B3255D760B0065AFD6.xcbaseline/B07DC9C5-A477-43FC-8906-C01C8B98086A.plist
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
calvincestari, line 1: These benchmarks are for the UUID of my laptop.

BobaFetters (approved):
Think this looks good, like the addition of the `ResponseCaptureRequestChain` and good call on making tests fail if there are no baselines by default.

calvincestari:
Putting this back into draft state; I just realized that the `testMultipartResponseParsingInterceptor` performance test is parsing the 1000 chunks but only checking the response data once at the end. `interceptAsync` is synchronous so it has to be a completion block to validate each of the responses.

calvincestari:
The refactor to make these interceptor work via a callbacks is in https://github.com/apollographql/apollo-ios/pull/2974 which will get merged into this PR and then all together squashed into `main`.

calvincestari:
I merged my other PR (#2974) into this one so it's easier to holistically review the changeset vs. needing the context of this PR to accurately review #2974.

BobaFetters (approved):

PR #2991, opened 2023-05-02
ci: Build codegen test configurations

calvincestari:
This PR adds building the [codegen test configurations](https://github.com/apollographql/apollo-ios/tree/main/Tests/TestCodeGenConfigurations) to CI jobs. Currently the tests are stored in the repo but they are ad-hoc manually run projects.

#### Build time
* The [CI build against the last commit of this PR](https://app.circleci.com/pipelines/github/apollographql/apollo-ios/4500/workflows/47f70f15-8865-43d2-b0ec-1bca195bab73) took 9m17s for the new job. 
* The whole CI build of this PR took 9m22s, which CircleCI claims is 35% larger than the P95 over the past 7 days.
* We could try split the 5 projects out into their own jobs to parallelize and distribute the build time but that is a much more maintenance heavy way to configure this job. The job and scripts are set up in a way now that simply adding a new project into the `TestCodeGenConfigurations` folder will have it automatically included in the new job.

#### Job configuration
* The job is executed by the `run-test-codegen-configurations.sh` script which has been modified to accept a `-t` parameter which will both run codegen and test each project.
* Each project folder is required to have a `test-project.sh` script in the root which performs the test suite specific to that test project configuration. This was needed because some projects are SPM, some are Xcode and some have more than one package.

#### Projects configuration
* All generated code has been deleted from the test projects and the paths added to `.gitignore` so that the code is regenerated each build to ensure that it is using the codegen engine in the branch at the time.
* All projects use the local [AnimalKingdomAPI](https://github.com/apollographql/apollo-ios/tree/main/Sources/AnimalKingdomAPI/animalkingdom-graphql) schema/operations for consistency between projects. Except `EmbeddedInSchema-RelativeAbsolute` because we want the operations generated at a relative path, so it has a copy of the schema/operations.
* All `Package.resolve` and `Podfile.lock` files were deleted so that the latest allowed versions of dependencies are used to ensure they continue to work with built projects.
* All test projects have at least one test that must pass for the job to be successful.
* All test projects have been configured to treat warnings as errors. This was done to catch Swift warnings such as `'public' modifier is redundant for type alias declared in a public extension` which we've seen in the past.


.circleci/config.yml
```diff
@@ -63,15 +63,6 @@ commands:
     description: Commands to run for setup of every set of tests
     steps:
       - checkout
-      - run:
```
calvincestari, line 66: These commands were not doing anything, we don't even use BitBucket. Removing them has had no effect on CI.

.circleci/config.yml
```diff
@@ -124,6 +115,19 @@ commands:
           name: Save test results
           path: ~/TestResults/ResultBundle.zip
           destination: results
+  codegen_test_configurations_generate_and_test:
+    steps:
+      - run:
+          command: git clone -b "$CIRCLE_BRANCH" "$CIRCLE_REPOSITORY_URL"
```
calvincestari, line 121: This job manually clones the repo because the `checkout` step provided by CircleCI, which we use in other jobs, clones the repo into `/Users/distiller/project` and SPM does not like that for a local package dependency. Manually cloning the repo now places it into `/Users/distiller/project/apollo-ios` and we're able to use local packages as expected.

.gitignore
```diff
@@ -67,9 +67,20 @@ SwiftScripts/.build-**
...
+# Generated js files we don't need committed
+Sources/ApolloCodegenLib/Frontend/dist/
+
 # Local generated packages we don't need in the main project
 Sources/AnimalKingdomAPI/Generated/Package.swift
 Sources/AnimalKingdomAPI/Generated/Package.resolved
-
-# Generated js files we don't need committed
-Sources/ApolloCodegenLib/Frontend/dist/
+Tests/TestCodeGenConfigurations/**/AnimalKingdomAPI/
```
calvincestari, line 76: This is to ignore all the generated code so it's not accidentally committed in the future.

.gitignore
```diff
@@ -67,9 +67,20 @@ SwiftScripts/.build-**
...
+Tests/TestCodeGenConfigurations/**/AnimalKingdomAPITestMocks
+Tests/TestCodeGenConfigurations/**/Package.resolved
+Tests/TestCodeGenConfigurations/**/Podfile.lock
+Tests/TestCodeGenConfigurations/**/TestMocks
+Tests/TestCodeGenConfigurations/EmbeddedInTarget-RelativeAbsolute/PackageOne/**/*.graphql.swift
+Tests/TestCodeGenConfigurations/EmbeddedInTarget-RelativeAbsolute/PackageTwo/Sources/PackageTwo
+!Tests/TestCodeGenConfigurations/Other-CustomTarget/AnimalKingdomAPI/AnimalKingdomAPI.h
```
calvincestari, line 83: The `!` files are ones that we specifically want to keep because they either configure the cache keys for the tests to pass of they're project files that are not generated but required to build.

Sources/ApolloTestSupport/TestMock.swift
```diff
@@ -67,6 +67,7 @@ public class Mock<O: MockObject>: AnyMock, Hashable {
     }
   }
 
+  @_disfavoredOverload
```
calvincestari, line 70: This is required to resolve a build error `Ambiguous use of '_set(_:for:)'` when the initializer has both single value and array values.

.gitignore
```diff
@@ -67,9 +67,20 @@ SwiftScripts/.build-**
...
+!Tests/TestCodeGenConfigurations/Other-CocoaPods/CocoaPodsProject/AnimalKingdomAPI/Schema/SchemaConfiguration.swift
```
AnthonyMDev, line 84: The Schema configurations are probably always going to need to be kept. If we do end up configuring cache keys for future tests in the other example projects. Should we change this to one line of
```
!Tests/TestCodeGenConfigurations/**/SchemaConfiguration.swift
```

AnthonyMDev (approved):
I cannot express how thankful I am to have this PR. You are the best @calvincestari!

AnthonyMDev:
> The job is executed by the run-test-codegen-configurations.sh script which has been modified to accept a -t parameter which will both run codegen and test each project.
> Each project folder is required to have a test-project.sh script in the root which performs the test suite specific to that test project configuration. This was needed because some projects are SPM, some are Xcode and some have more than one package.

We might want some sort of documentation of this stuff somewhere?

It might make sense to add some more guidance on things like this to the `CONTRIBUTING.md` file. Anything that could be useful for developers working on the project, but not necessarily developers that are just using Apollo as a dependency could go there. 

We've probably got a lot of other implicit knowledge that @calvincestari and I know that should go into that document as well. @BobaFetters, as you continue ramping up, would you keep a quick note of things that you learn about the project structure that aren't really obvious or would be helpful for any future contributors so we can add them to this list?

Once we re-structure the project into multiple repos, that information is all liable to change, so I don't think we need to formalize and publish it all right now. But a quick note on what those things are would be valuable so that we can make sure we cover all of it after we restructure the project.

AnthonyMDev:
> All test projects have been configured to treat warnings as errors. This was done to catch Swift warnings such as 'public' modifier is redundant for type alias declared in a public extension which we've seen in the past.

.gitignore
calvincestari: Sure, we can commit and track all of the schema configuration files. I only kept the ones that currently had custom logic in them, but yes I agree it's not so obvious to make a change and then wonder why it's not being tracked. Good call!

AnthonyMDev:
> We could try split the 5 projects out into their own jobs to parallelize and distribute the build time but that is a much more maintenance heavy way to configure this job. The job and scripts are set up in a way now that simply adding a new project into the TestCodeGenConfigurations folder will have it automatically included in the new job.

It might be worth experimenting with [Circle CI parallelism](https://circleci.com/docs/parallelism-faster-jobs/) here?

It would be great if we could check out the code and run codegen once, but then build and run tests for each of the projects in parallel. I don't think that's what [Circle CI parallelism](https://circleci.com/docs/parallelism-faster-jobs/) though...

Running the actual codegen itself probably is a lot less time than some of the others build and test running steps of this job. But parallelizing the codegen engine is something I've thought about doing for a long time also.

---

I'd bet that the longest step is resolving package dependencies for each project. I wonder if there is a way to consolidate some of that so they can use cached versions of our external dependencies between runs instead of having to checkout `InflectorKit`, `Nimble`, `SQLIte`, etc. every time. 

I know that Cocoapods and SPM both have a local dependency cache, so once they have been checked out for one test, they should theoretically not need to be redownloaded for each individual test. That is unless Circle CI is doing something to prevent that since they are using shared virtual machines.

I know when I was working at Salesforce, our CI was set up to use a local Artifactory instance to cache versions of our dependencies to speed up CI. But I imagine you have to be running CI on dedicated machines in order to do that... It sounds similar to [Circle CI's Docker Layer Caching](https://circleci.com/docs/docker-layer-caching/), but I'm not sure if it can be used for our iOS dependencies. (I have an incredibly limited understanding of what Docker does, so I could be wrong about that!)

calvincestari:
> > All Package.resolve and Podfile.lock files were deleted so that the latest allowed versions of dependencies are used to ensure they continue to work with built projects.
> 
> Do we need to gitignore these files as well? Or I guess if we are running `pod update` not just `pod install` (and respective commands for SPM), then it shouldn't matter anyways. I'm not entirely clear why we need to delete them at all.

I want to know ahead of time if a dependency update is going to cause issues with projects and we may need to narrow the scope of allowed dependency versions until we can figure out what's going on. The alternative is waiting for an issue to be created by the community letting us know.

calvincestari:
> > All test projects have been configured to treat warnings as errors. This was done to catch Swift warnings such as 'public' modifier is redundant for type alias declared in a public extension which we've seen in the past.

This could come back to bite us if we add `#warnings` in to remind us to come back to something later. Although we generally only do that in long lived feature branches or work in progress, so nothing that should end up in `main`.

AnthonyMDev:
> > > All Package.resolve and Podfile.lock files were deleted so that the latest allowed versions of dependencies are used to ensure they continue to work with built projects.
> > 
> > 
> > Do we need to gitignore these files as well? Or I guess if we are running `pod update` not just `pod install` (and respective commands for SPM), then it shouldn't matter anyways. I'm not entirely clear why we need to delete them at all.
> 
> I want to know ahead of time if a dependency update is going to cause issues with projects and we may need to narrow the scope of allowed dependency versions until we can figure out what's going on. The alternative is waiting for an issue to be created by the community letting us know.

Definitely! I understand the objective you're trying to achieve here.

What I was saying is if we run `pod update` it's going to ignore the `Podfile.lock` and update to the newest versions of dependencies anyways. So I'm not sure what value we get out of deleting those files. 

The first time I run these test code gen configurations locally, the file is going to get readded anyways, so unless we are adding them to the `.gitignore` also, this doesn't seem to make sense.

calvincestari:
> The first time I run these test code gen configurations locally, the file is going to get readded anyways, so unless we are adding them to the `.gitignore` also, this doesn't seem to make sense.

Yes, both `Package.resolved` and `Podfile.lock` were added to `.gitignore`.

AnthonyMDev:
> > All test projects have been configured to treat warnings as errors. This was done to catch Swift warnings such as 'public' modifier is redundant for type alias declared in a public extension which we've seen in the past.
> 
> This could come back to bite us if we add `#warnings` in to remind us to come back to something later. Although we generally only do that in long lived feature branches or work in progress, so nothing that should end up in `main`.

This leads me to 2 thoughts!

1. Can we set it up to only treat warnings as errors for the warnings in the actual example project's sources? I think there is a way to suppress all warnings in your dependencies without suppressing warnings in your project's source code.

2. Since this is such a long running job and we may have unresolved warnings during work in progress, should/can we configure this job to only run in certain circumstances?
  - I don't think it needs to run every time I push a commit to a branch that I haven't even made a PR for yet. It should probably only run on PRs (and maybe not draft PRs?)
  - Is there some way for those situations we can set some option or flag to make it run this job on other commits in the mean time, if we are working on something that heavily affects codegen and we want to run it for sanity checks while doing work in progress? (Though I guess in that situation we can just run the tests locally anyways?)

BobaFetters (approved):

.gitignore
calvincestari: This is done in https://github.com/apollographql/apollo-ios/pull/2991/commits/eaad7d6845a91dc6cec5cc351bf2aeedfd3aceb8.

PR #2996, opened 2023-05-04
Updating casing of variable names and keys

### LINKED ISSUE #2858: InputObjects generated with incorrect getter/setter key
### Summary

I'm having a issue with a custom input that results in different string keys being generated for `InputObjects` in the `init` and the associated getter/setter. For example:

```swift
// @generated

import ApolloAPI

public extension API {
  struct EquipmentFilterInput: InputObject {
    public private(set) var __data: InputDict

    public init(_ data: InputDict) {
      __data = data
    }

    public init(
      iD: GraphQLNullable<String> = nil
    ) {
      __data = InputDict([
        "ID": iD
      ])
    }

    public var iD: GraphQLNullable<String> {
      get { __data["iD"] }
      set { __data["iD"] = newValue }
    }
}
```

Note that the `__data` in the `init` uses the string key "ID" while the `var iD` getter/setter uses `iD`.

### Version

1.0.7 (I also saw it on 1.0.5)

The above issue was seen using the following operation definitions:

```graphql
fragment EquipmentFields on Equipment {
    __typename,
    ID,
}

query ReadEquipments($where: EquipmentFilterInput) {
    equipments(where: $where) {
        ...EquipmentFields
    }
}
```

Where `EquipmentFilterInput` is defined as the following in the schema:
```graphql
input EquipmentFilterInput {
  ID: String
}
```

### Logs

The above bug results in the query being malformed and returning errors like

```
[Variable "$where" got invalid value "{"iD":{"isEqualTo":"E4600CC4-0FAC-40F7-8EC7-5CE65AF5B05F"}}".
In field "iD": Unknown field.]
```

If I manually change `__data["iD"]` to `__data["ID"]` then queries are successful.

### Anything else?

_No response_

BobaFetters:
-Updated the handling of all uppercase variable names & dictionary keys to ensure keys always match what is provided in the schema, and variables are converted to lowercase.

Closes #2858

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
@@ -24,7 +24,7 @@ extension OperationTemplateRenderer {
     _ variables: [CompilationResult.VariableDefinition]
   ) -> TemplateString {
     """
-    \(variables.map { "public var \($0.name.asInputParameterName): \($0.type.rendered(as: .inputValue, config: config.config))"}, separator: "\n")
+    \(variables.map { "public var \($0.name.asFieldAccessorPropertyName): \($0.type.rendered(as: .inputValue, config: config.config))"}, separator: "\n")
     """
   }
 
@@ -46,7 +46,7 @@ extension OperationTemplateRenderer {
     }
 
     return """
-      public var __variables: \(if: !graphQLOperation, "GraphQLOperation.")Variables? { [\(list: variables.map { "\"\($0.name)\": \($0.name.asInputParameterName)"})] }
+      public var __variables: \(if: !graphQLOperation, "GraphQLOperation.")Variables? { [\(list: variables.map { "\"\($0.name)\": \($0.name.asFieldAccessorPropertyName)"})] }
       """
   }
 
--- a/Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
+++ b/Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
@@ -13,7 +13,7 @@ extension OperationTemplateRenderer {
     return """
     \(`init`)(\(list: variables.map(VariableParameter))) {
       \(variables.map {
-        let name = $0.name.asInputParameterName
+        let name = $0.name.asFieldPropertyName
         return "self.\(name) = \(name)"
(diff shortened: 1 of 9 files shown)


Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
```diff
@@ -24,7 +24,7 @@ extension OperationTemplateRenderer {
     _ variables: [CompilationResult.VariableDefinition]
   ) -> TemplateString {
     """
-    \(variables.map { "public var \($0.name.asInputParameterName): \($0.type.rendered(as: .inputValue, config: config.config))"}, separator: "\n")
+    \(variables.map { "public var \($0.name.asFieldAccessorPropertyName): \($0.type.rendered(as: .inputValue, config: config.config))"}, separator: "\n")
```
calvincestari, line 27: If I read this right we could have a situation where we're creating the property name here with `asFieldAccessorPropertyName` but referring to it with a name generated by `asInputParameterName` up on [line 17](https://github.com/apollographql/apollo-ios/pull/2996/files#diff-905e0defd35e8857aa7177892f45520af989a81895465ffb4f447e53e36dfed2R17). I think it works for now because they actually seem to escape the same list of strings but having them be separate is leaving room for error. Is it worth reducing this into a single function and corresponding list-to-be-escaped?

calvincestari (commented):
Looks good, with one comment on reducing room for error.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
BobaFetters: Oh good catch, I was using `asInputParameterName` there because it was for the input parameter of the `init` but didn't catch that it would be accessing the generated `var` name. Right now `asInputParameterName` and `asFieldAccessorPropertyName` are both functioning exactly the same, so we could definitely combine them into one function. I can't think of a need to have input parameters for functions formatted differently than variables but if that case did come up we could always break this back out in the future.

calvincestari: > I can't think of a need to have input parameters for functions formatted differently than variables but if that case did come up we could always break this back out in the future.

Me neither, I think it'd be a good decision to combine them.

calvincestari (approved):
Nice work, thanks for the fix @BobaFetters!

PR #3009, opened 2023-05-12
Check type name conflicts in SelectionSet

### LINKED ISSUE #2850: Generates conflicting types for fields of singular and plural names
### Summary

When querying different properties of similar name — one being singular, one being plural — with a different subset of properties , Apollo generates conflicting type names for the query response.

We have a query with two properties on a parent type. One is a single instance, the other is an array. Both are of the same type in the schema. When we query different subsets of properties on each, Apollo generated separate types (as expected) but names them the same. This is a conflict because they are defined on the same level.

### Version

1.0.7

Query example: 
```graphql
value {
  propertyA
  propertyB
  propertyC
  propertyD
}

values {
  propertyA
  propertyC
}
```

Both resulting types are called `Value` in the Swift project. This makes any reference ambiguous and fails the build.

### Logs

_No response_

### Anything else?

Apollo on Android seems to notice the conflicting type names, as it generates one `Value` and one `Value1` type.

BobaFetters:
Adding validation to check for type name conflicts in selection set fields based on pluralization.

Closes #2850


Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -624,7 +624,7 @@ fileprivate extension IR.SelectionSet {
 
 }
 
-fileprivate extension IR.EntityField {
+extension IR.EntityField {
```
BobaFetters, line 627: Considered moving this out to another file since it is being used outside of `SelectionSetTemplate` now however it uses another fileprivate extension within `SelectionSetTemplate` so opted to keep it here for now.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -212,6 +218,24 @@ public class ApolloCodegen {
       throw Error.schemaNameConflict(name: context.schemaNamespace)
     }
   }
+  
+  /// Validates that there are no type conflicts within a SelectionSet
+  static private func validateTypeConflicts(for selectionSet: IR.SelectionSet, with context: ConfigurationContext) throws {
+    // Check for type conflicts resulting from singularization/pluralization of fields
+    var typeNames = [String: String]()
```
AnthonyMDev, line 225: This is a little hard to understand. It's not clear what the key vs value represent here. Maybe changing this to `fieldNamesByFormattedTypeName` or at least adding an inline documentation comment to clarify?

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -212,6 +218,24 @@ public class ApolloCodegen {
       throw Error.schemaNameConflict(name: context.schemaNamespace)
     }
   }
+  
+  /// Validates that there are no type conflicts within a SelectionSet
+  static private func validateTypeConflicts(for selectionSet: IR.SelectionSet, with context: ConfigurationContext) throws {
```
AnthonyMDev, line 223: Great work on this algorithm! Looks great! Just a couple nits.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -212,6 +218,24 @@ public class ApolloCodegen {
...
+    // Check for type conflicts resulting from singularization/pluralization of fields
+    var typeNames = [String: String]()
+    var fields: [IR.EntityField] = selectionSet.selections.direct?.fields.values.compactMap({ $0 as? IR.EntityField }) ?? []
+    fields.append(contentsOf: selectionSet.selections.merged.fields.values.compactMap({ $0 as? IR.EntityField }))
+
+    try fields.forEach({ field in
```
AnthonyMDev, line 229: I'm not opposed to using `forEach` but I'm not sure what the value of that over `for field in fields` is here.

Nit: To better follow swift standard style guidelines, you can omit the `()` here. `try fields.forEach { ... }`

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -2334,6 +2334,78 @@ class ApolloCodegenTests: XCTestCase {
     // then
     expect(try ApolloCodegen._validate(config: config)).notTo(throwError())
   }
+  
+  func test__validation__selectionSet_typeConflicts_shouldThrowError() throws {
```
AnthonyMDev, line 2338: I'd love an additional test here that tests this for merged fields.

That test would need `values` to be in an inline fragment. So in another test:
schema:
```graphql
      type Query {
        user: User
      }
      type User {
        containers: [ContainerInterface]
      }
      interface ContainerInterface {
        value: Value
      }
      type Container implements ContainerInterface{
        value: Value
        values: [Value]
      }
      type Value {
        propertyA: String!
        propertyB: String!
        propertyC: String!
        propertyD: String!
      }
```
operation:
"""graphql
query ConflictingQuery {
          user {
              containers {
                  value {
                      propertyA
                      propertyB
                      propertyC
                      propertyD
                  }
                 ... on Container {
                    values {
                        propertyA
                        propertyC
                    }
                  }
              }
          }
      }
"""

Remove the line that validates the `merged` fields, this test should fail. Add the line back in, and it should then throw the validation error on the `Data.User.Container.AsContainer` selection set instead of `Data.User.Container`.

AnthonyMDev (changes requested):
This is awesome! I'd just like to add the operation/fragment name to the validation error and add a test for merged selections.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -53,8 +54,13 @@ public class ApolloCodegen {
...
       case let .targetNameConflict(name):
         return """
-          Target name '\(name)' conflicts with a reserved library name. Please choose a different \
-          target name.
+        Target name '\(name)' conflicts with a reserved library name. Please choose a different \
+        target name.
+        """
+      case let .typeNameConflict(name, conflictingName):
+        return """
+        Field '\(conflictingName)' conflicts with field '\(name)' in selection set. Recommend \
```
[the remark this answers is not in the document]
BobaFetters: Yea it shouldn't be difficult to add operation/fragment name here, line numbers is probably more work than the value it gives but I can check on that as well.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -304,6 +328,7 @@ public class ApolloCodegen {
     for fragment in compilationResult.fragments {
       try autoreleasepool {
         let irFragment = ir.build(fragment: fragment)
+        try validateTypeConflicts(for: irFragment.rootField.selectionSet, with: config)
```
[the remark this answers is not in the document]
BobaFetters: Agreed, was thinking about that when writing this, also why I tried to name this new function fairly generically at least to allow for other check to be added in it. But like you said refactoring all of this into a separate object so we can group validations together better would be preferable.

Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: Honestly was just a style preference, happy to change it to `for field in fields` if we want to enforce that style for this type of loop.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
AnthonyMDev: If they are being used outside of this file, they should probably be moved into the `Utilities` folder. `fileprivate` is good for it being scope to just the `SelectionSetTemplate`, but if other places are dependent on the behavior of this, we should make that clear by pulling it out and documenting what it's doing more.

Sources/ApolloCodegenLib/ApolloCodegen.swift
AnthonyMDev: Really doesn't matter much to me. I have always preferred `for ... in ...` but it's just a personal preference! It was just odd for me to see that! Go ahead and leave it. :)

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
BobaFetters: Pulled these out into a `IR+Formatting.swift` file in the `Extensions` folder since there wasn't a `Utilities` folder in `ApolloCodegenLib`

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -243,8 +245,6 @@ public class ApolloCodegen {
     // gather nested fragments to loop through and check as well
     var nestedSelectionSets: [IR.SelectionSet] = selectionSet.selections.direct?.inlineFragments.values.elements ?? []
     nestedSelectionSets.append(contentsOf: selectionSet.selections.merged.inlineFragments.values)
-    nestedSelectionSets.append(contentsOf: selectionSet.selections.direct?.fragments.values.map { $0.fragment.rootField.selectionSet } ?? [])
```
calvincestari: @BobaFetters - is it correct that these two lines were removed?

Issue #2919, opened 2023-03-31
`makefile` improvement - CLI


asbhat:
### Summary

After upgrading to Xcode v14.3 the `InstallCLI` Command Plugin seems to be missing.

I created a new project and installed `apollo-ios` but the plugin is still missing. Am I doing something wrong?

<img width="394" alt="image" src="https://user-images.githubusercontent.com/3885798/229012643-b0a6441a-2ae8-433a-974e-3c48eaf78920.png">

### Version

v1.0.7

### Steps to reproduce the behavior

1. Install Xcode v14.3
2. Create a new project
3. Add package `apollo-ios`
4. Right click on the project in the navigator
5. The command plugin is missing

calvincestari:
Hi @asbhat - we've taken a look and don't have a clear idea on what could be causing it to be missing; [our implementation of `XcodeCommandPlugin`](https://github.com/apollographql/apollo-ios/blob/main/Plugins/InstallCLI/InstallCLIPluginCommand.swift) hasn't changed. We're also concerned it could be a regression in Xcode 14.3 support for SPM plugins since the plugin is visible when opening a `package.swift` based project (vs. an Xcode project that uses SPM dependencies).

I realize it is inconvenient not having it in the menu and until we can figure out the root cause the workaround would be to grab the pre-built CLI binary we attach to each [GitHub release](https://github.com/apollographql/apollo-ios/releases) and place that in the root of your project directory; the effect will be the same and the CLI has a version checker so if you ever update the dependency but forget to manually download the pre-built CLI it will warn you.

asbhat:
Thanks for the response Calvin!

I did end up just grabbing the pre-built CLI binary ([docs here](https://www.apollographql.com/docs/ios/code-generation/codegen-cli#installation)) that works just fine.

I took a quick look at the plugin too and couldn't figure it out :(
but it does seem other projects are having similar issues (e.g., https://github.com/nicklockwood/SwiftFormat/issues/1404), so you're probably right that it's a regression and something for Apple to fix.

Should I close the issue?

calvincestari:
I think leave it open for a bit longer and let's monitor if it's showing up in other libraries too.

PawanSinghatia:
Hi @calvincestari , @asbhat  please provide the exact steps for the pre-built CLI binary solution for this issue

PawanSinghatia:
> Thanks for the response Calvin!
> 
[...]

PR #3015, opened 2023-05-15
Plug in for CLI

amgundeti:
The plug in for the xcode CLI doesn't seem to be working - however I can take care of code generation by manually inserting the cli binary in the directory. Any way I can get the plug in back up within xcode? Thanks!


BobaFetters:
@amgundeti This is a known bug with Xcode 14.3 that is affecting many projects including Apollo, there is an issue related to this you can see [here](https://github.com/apollographql/apollo-ios/issues/2919). For future reference please post questions like this as an Issue in the repo with the label `question` vs opening a PR

amgundeti:
Thank you!

PR #3020, opened 2023-05-16
chore: Verify version format in set script

calvincestari:
@AnthonyMDev - here's a change to the version setting script to avoid the versioning mistake from the `1.2` release yesterday. All three digits should now be required when using it.


AnthonyMDev (approved):
lgtm thanks!

PR #3027, opened 2023-05-18
fix: flaky urlsessionclient tests

### LINKED ISSUE #2902: test: Fix flaky `URLSessionClientTests`
Been failing recently in multiple PRs and on `main` with the 1.1 release.

BobaFetters:
After evaluating the existing `URLSessionClientTests` discovered the tests were flaky due to their reliance on an external live server (httpbin.org), after further investigation/discussion we determined these tests do not need to and should not be hitting live services.

These tests have been redone to use a `MockURLProtocol` to handle simulating responses for the various requests while still allowing the necessary `URLSessionClient` code to be called and tested without reliance on a live server.

These tests have also been moved from the Integration test plan to the Unit test plan as a result of this work.

Closes #2902


Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
BobaFetters, line 1: Tried to keep this class as simple as possible for current use case while leaving room to expand upon it in the future if we find more use cases etc for our various tests.

Tests/ApolloTests/Network/URLSessionClientTests.swift
```diff
@@ -153,26 +181,29 @@ class URLSessionClientLiveTests: XCTestCase {
...
+
     self.client.cancel(task: task)
-    
+
     self.wait(for: [expectation], timeout: 5)
-    
+
   }
   
   func testMultipleSimultaneousRequests() {
```
BobaFetters, line 209: Ended up not even needing to do some sort of random response delay for this, they complete in a different order each time the test is run.

Apollo.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,9 @@
 		5AC6CA4322AAF7B200B7C94D /* GraphQLHTTPMethod.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5AC6CA4222AAF7B200B7C94D /* GraphQLHTTPMethod.swift */; };
 		5BB2C0232380836100774170 /* VersionNumberTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BB2C0222380836100774170 /* VersionNumberTests.swift */; };
 		66321AE72A126C4400CC35CB /* IR+Formatting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 66321AE62A126C4400CC35CB /* IR+Formatting.swift */; };
+		66B18E872A15367300525DFB /* URLSessionClientTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B4F4542244A2AD300C2CF7D /* URLSessionClientTests.swift */; };
+		66B18E8C2A1674FB00525DFB /* MockURLProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 66B18E8B2A1674FB00525DFB /* MockURLProtocol.swift */; };
+		66B18E8D2A1674FB00525DFB /* MockURLProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 66B18E8B2A1674FB00525DFB /* MockURLProtocol.swift */; };
```
BobaFetters, line 20: Need to check why this new file has 2 references in the xcodeproj

Tests/ApolloTests/Network/URLSessionClientTests.swift
AnthonyMDev: Even if that's the case, it's due to some implementation detail of `URLSession` that is liable to change in the future. I feel like we should be adding some random delay in here anyways, just to ensure we are in control of the behavior of the tests _always_ doing what we expect in the future.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+public class MockURLProtocol: URLProtocol {
+  
+  public typealias MockRequestHandler = ((URLRequest) throws -> Result<(HTTPURLResponse, Data?), Error>)
+  
+  // Dictionary of mock request handlers where the `key` is the URL of the request.
+  public static var requestHandlers = [URL: MockRequestHandler]()
```
AnthonyMDev, line 8: When unit tests are mutating static properties, we should use a `TestObserver` to ensure that state is reset between tests. Especially for tests that run async code, you can easily break test isolation.

Check out the `MockSchemaMetadata` to see an example of using `TestObserver`.

Actually, since we run unit tests in parallel, this could still be an issue even with the `TestObserver`. Is there a way to indicate that certain tests must not be run in parallel I wonder?

Tests/ApolloTests/Network/URLSessionClientTests.swift
```diff
@@ -1,14 +1,18 @@
...
+class URLSessionClientTests: XCTestCase {
   
   var client: URLSessionClient!
+  var sessionConfiguration: URLSessionConfiguration!
 
   override func setUp() {
     super.setUp()
-
-    client = URLSessionClient()
+    Self.testObserver.start()
```
BobaFetters, line 12: Running the `testObserver.start()` in the setup for every test rather that on value change for `requestHandlers` like what is done in `MockShemaData` due to some tests applying more than 1 request handler, in testing this on top of the `MockURLProtocol` clearing out request handlers as it uses them ensured that there were no left over request handlers between tests.

Tests/ApolloTests/Network/URLSessionClientTests.swift
BobaFetters: Added a random 0-1 seconds execution delay in `MockURLProtocol` to give some randomness to the test responses.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
```diff
@@ -0,0 +1,57 @@
...
+          let handler = RequestProvider.requestHandlers[url] else {
+      fatalError("No URL available for URLRequest.")
+    }
+    
+    defer {
+      RequestProvider.requestHandlers.removeValue(forKey: url)
+    }
+    
+    // Provide random response delay between 0 and 1 seconds
+    Thread.sleep(forTimeInterval: Double.random(in: 0.0...1.0))
```
BobaFetters, line 24: need to re-evaluate handling the response delay since this seems to be causing some issues when run on the CI server.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
```diff
@@ -0,0 +1,56 @@
...
+    return request
+  }
+  
+  override public func startLoading() {
+    guard let url = self.request.url,
+          let handler = RequestProvider.requestHandlers[url] else {
+      fatalError("No URL available for URLRequest.")
+    }
+    
+    DispatchQueue.main.asyncAfter(deadline: .now() + Double.random(in: 0.0...0.5)) {
```
BobaFetters, line 19: Thoughts on the time delay range? Is 0.5 seconds to high in the range for testing purposes? Wondering if I should close this range down even further

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
```diff
@@ -0,0 +1,56 @@
...
+  }
+  
+  override class public func canonicalRequest(for request: URLRequest) -> URLRequest {
+    return request
+  }
+  
+  override public func startLoading() {
+    guard let url = self.request.url,
+          let handler = RequestProvider.requestHandlers[url] else {
+      fatalError("No URL available for URLRequest.")
```
calvincestari, line 16: Isn't it the handler rather than the URL that's likely to be missing?

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
calvincestari: I _think_ it's fine. Since it was using a service before we have no idea what the real delay was but this seems like a wide enough variance.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
BobaFetters: Ah yea nice catch I need to re-word that error message

Tests/ApolloTests/Network/URLSessionClientTests.swift
```diff
@@ -181,61 +213,67 @@ class URLSessionClientLiveTests: XCTestCase {
     expectation.expectedFulfillmentCount = iterations
     @Atomic var taskIDs: [Int] = []
     
-    DispatchQueue.concurrentPerform(iterations: iterations, execute: { index in
-      let request = self.request(for: .getWithIndex(index: index))
+    var responseStrings = [Int: String]()
+    var requests = [Int: URLRequest]()
+    for i in 0..<20 {
```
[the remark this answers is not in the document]
BobaFetters: Yea good call I overlooked that

calvincestari (approved):
I'm happy with this but we should wait for @AnthonyMDev opinion before merging I think.l

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
AnthonyMDev: I do think that 0.5 seconds is pretty long. I'd probably do `0.0...0.2` personally. But it's definitely not an issue either way.

AnthonyMDev: My one concern here is if we should be dispatching this on the `main` queue. Does `URLSessionClient` traditionally dispatch to the main thread? My assumption is that it would call on a background thread. And we should be trying to mimic the real life behaviors in these tests as closely as possible.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
```diff
@@ -0,0 +1,57 @@
...
+      client?.urlProtocol(self, didFailWithError: error)
+    }
+  }
+  
+  override public func stopLoading() {
+  }
+  
+}
+
+public protocol MockRequestProvider {
```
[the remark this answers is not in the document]
AnthonyMDev: Using a generic type based on the unit tests class itself was really clever. I don't think I would have thought of that approach and I love it.

Tests/ApolloTests/Network/URLSessionClientTests.swift
AnthonyMDev: Yupp that's perfect!

Tests/ApolloTests/Network/URLSessionClientTests.swift
```diff
@@ -17,14 +21,38 @@ class URLSessionClientLiveTests: XCTestCase {
     super.tearDown()
```
AnthonyMDev, line 21: We need to add `sessionConfiguration = nil` here.

The way XCTest works is that an instance of the `XCTestCase` subclass is instantiated for each individual test. These instances are not deallocated after their test finishes. Because they contain all of the necessary information for constructing the test report. So every single instance variable will be retained in memory _multiplied by the number of tests_ until the entire test suite finishes. This is why we `nil` things out in `tearDown`.

Usually this doesn't matter much. The amount of objects you would need to have in order to actually cause memory issues is absurdly high. But it's still best practice. And it does end up mattering when you have things in async tests that can be doing things on background threads and keeping those threads alive past the test's completion.

AnthonyMDev (changes requested):
Very happy with this! Thanks so so much @BobaFetters! Just one suggested change and a question about the `DispatchQueue` to use and this is good to go!

AnthonyMDev (changes requested):
Very happy with this! Thanks so so much @BobaFetters! Just one suggested change and a question about the `DispatchQueue` to use and this is good to go!

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
calvincestari: We [default to the main queue](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/URLSessionClient.swift#L65) but that can be overridden.

calvincestari:
Good suggestions @AnthonyMDev. Thanks @BobaFetters, I hope this is the turning point for these tests being stable.

Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
BobaFetters: Ahhh yea looks like the delegates and completion handlers are default to the main queue. So what do we think about having this changed to the global queue vs staying on main?

calvincestari: It will definitely change during 2.0 but I'm skeptical whether we should make the change now. It's arguably better behaviour but we have no idea what calls people making out of the callback. It could be disastrous for the less informed.

AnthonyMDev: You should use `DispatchQueue(label: "MockURLProtocol", qos: .background)`.

PR #3028, opened 2023-05-18
Update docs for SelectionSet internal initializer usage

AnthonyMDev:
Just clearing up usage of this function and pointing people to the right place in the docs.


calvincestari (approved):
Thanks for this.

BobaFetters (approved):
Good call adding the extra clarification for this

PR #3019, opened 2023-05-16
Adding new iOS Tutorial

BobaFetters:
Created new iOS tutorial using version 1.2.0 and SwiftUI

Closes https://github.com/apollographql/team-prometheus/issues/72

PR #3032, opened 2023-05-19
Suggested tutorial revisions

bignimbus:
Here are some follow-ups after stepping through our awesome new Tutorial: https://github.com/apollographql/apollo-ios/pull/3019

See also: https://github.com/apollographql/iOSTutorial/pull/7

### DIFF
--- a/docs/source/tutorial/tutorial-execute-first-query.mdx
+++ b/docs/source/tutorial/tutorial-execute-first-query.mdx
@@ -20,7 +20,7 @@ class Network {
 
 ## Implement the query
 
-To make sure your `ApolloClient` instance is communicating correctly with the server, go to `LaunchListViewModel`, add the following imports and update the `init()` method to contain the following code:
+To make sure your `ApolloClient` instance is communicating correctly with the server, go to `LaunchListViewModel`, add these imports, then add the following code to the `init()` method just below the TODO:
 
 ```swift title="LaunchListViewModel.swift"
 import Apollo
@@ -29,6 +29,7 @@ import RocketReserverAPI
 ...
 
 init() {
+    // TODO (Section 13 - https://www.apollographql.com/docs/ios/tutorial/tutorial-subscriptions#use-your-subscription)
     Network.shared.apollo.fetch(query: LaunchListQuery()) { result in
         switch result {
         case .success(let graphQLResult):
--- a/docs/source/tutorial/tutorial-execute-first-query.mdx
+++ b/docs/source/tutorial/tutorial-execute-first-query.mdx
@@ -48,6 +48,13 @@ Build and run your application. The web host might take a few seconds to spin up
 
 This means the request was correctly executed and you now have a list of launch sites 🚀🚀🚀.
 
-Go ahead and remove the code added to the `init()` method.
+Go ahead and remove the code added to the `init()` method and replace it with a TODO for later:
+
+
+```swift title="LaunchListViewModel.swift"
+init() {
+    // TODO (Section 13 - https://www.apollographql.com/docs/ios/tutorial/tutorial-subscriptions#use-your-subscription)
+}
+```
 
 Next, let's [connect this data to your UI](tutorial-connect-queries-to-ui)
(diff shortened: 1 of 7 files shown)


docs/source/tutorial/tutorial-execute-first-query.mdx
```diff
@@ -6,7 +6,7 @@ To use the generated operations in `RocketReserverAPI`, you first create an inst
 
 ## Create an ApolloClient
 
-1. Create a new Swift file called `Network.swift` and add `import Apollo` to the top of the file.
+1. Create a new Swift file within `RocketReserver` called `Network.swift`.  Set the target to `RocketServer` and add `import Apollo` to the top of the file.
```
BobaFetters, line 9:
```suggestion
1. Create a new Swift file within `RocketReserver` called `Network.swift`.  Set the target to `RocketReserver` and add `import Apollo` to the top of the file.
```

docs/source/tutorial/tutorial-paginate-results.mdx
```diff
@@ -94,6 +94,14 @@ func loadMoreLaunchesIfTheyExist() {
 }
 ```
 
+Lastly, update the `init()` method to pass `cursor: nil` to `LaunchListQuery`:
+
+```swift title="LaunchListViewModel.swift"
+init() {
+    Network.shared.apollo.fetch(query: LaunchListQuery(cursor: nil)) { result in// highlight-line
```
BobaFetters, line 101: There is some guidance in the section for executing your first query after adding this code to the `init()` method that says to remove it since its only meant to be an initial illustration of running a query. If it's not clear I think we should update that section rather than continuing to reference this code.

docs/source/tutorial/tutorial-subscriptions.mdx
```diff
@@ -146,7 +146,9 @@ Next, update the `init()` method to the following:
 
 ```swift title="LaunchListViewModel.swift"
 init() {
-    startSubscription()
+    Network.shared.apollo.fetch(query: LaunchListQuery(cursor: nil)) { result in
```
BobaFetters, line 149: Same as other comment, this query code shouldn't be here after the section for executing your first query.

calvincestari (approved):
These are all good suggestions/corrections. I agree with @BobaFetters that we should revise the prior instruction to remove code rather than reference and update it.

docs/source/tutorial/tutorial-execute-first-query.mdx
```diff
@@ -44,10 +44,17 @@ init() {
 
 Build and run your application. The web host might take a few seconds to spin up your GraphQL server if nobody's been using it recently, but once it's up, you should see a response that resembles the following:
 
-<img src="images/query_console_log_sample.png" class="screenshot" alt="Query console log sample"/>
+<img src="images/query_console_log_sample.png" class="screenshot" alt="Query console log output, beginning with 'Success! Result: GraphQLResult...'"/>
 
 This means the request was correctly executed and you now have a list of launch sites 🚀🚀🚀.
 
-Go ahead and remove the code added to the `init()` method.
+Go ahead and remove the code added to the `init()` method and replace it with a TODO for later:
```
bignimbus, line 51: @calvincestari @BobaFetters I totally blew by this section in my run-through.  I'm happy to throw this change away if it's a little "too much" but maybe an explicit step to replace the fetch code with a TODO will call more attention to it?  I might just be imposing my user error on the tutorial but let me know if this is worth doing.

BobaFetters: Do you think we should have the user replace the code with a `TODO` pointing to section 13, or have a `TODO` for section 13 already there in the starter project and have the user add this temporary code below it, and then instruct to remove it to end up back with the provided code snippet?

calvincestari: I think having a TODO already in the starter project, adding code below. The removal at the end should then be backed up by the still-existing TODO marker which is clear they still need make some change (remove the code).

calvincestari (approved):

docs/source/tutorial/tutorial-execute-first-query.mdx
BobaFetters, line 51:
```suggestion
Go ahead and remove the code added to the `init()` method so there is just the TODO for later:
```

BobaFetters (approved):

PR #3037, opened 2023-05-22
fix: ScalarType casting for Int values

### LINKED ISSUE #3034: NetworkResponseExecutionSource.DataTransformer Returns `nil` When It Should Not
### Summary

Using the subscript for the new `ObjectData`, which uses the `NetworkResponseExecutionSource.DataTransformer`, returns `nil` for integer id fields.

```swift
let value = object._rawData["id"]  // value is Optional(AnyHashable(<<Int>>))

switch value {
case let scalar as ScalarType:
  print(scalar)
case let customScalar as CustomScalarType:
  print(customScalar._jsonValue as? ScalarType)
default: print("nil!")  // prints "nil!"
}

switch value {
case let intID as Int: print(intID)  // this properly prints the int
default:
    print("\(type) could not translate \(value) from \(object)!!")
    return nil
}

```

### Version

v1.2

My project is closed, but the API I'm using (which I have no control over) is open.

URL: https://graphql.anilist.co

Example query
```graphql
query MediaDetails($id: Int!) {
  Media(id: $id) {
    id
    updatedAt
  }
}
```

Schema Configuration
```swift
public enum SchemaConfiguration: ApolloAPI.SchemaConfiguration {
    public static func cacheKeyInfo(for type: ApolloAPI.Object, object: ApolloAPI.ObjectData) -> ApolloAPI.CacheKeyInfo? {
        switch type {
        case Objects.Media:
            let value = object._rawData["id"]  // value is Optional(AnyHashable(<<Int>>))

            switch value {
            case let scalar as ScalarType:
                print(scalar)
            case let customScalar as CustomScalarType:
                print(customScalar._jsonValue as? ScalarType)
            default: print("subscript returns nil")  // this line executes
            }

            switch value {
            case let intID as Int: 
                print(intID)  // this properly prints the int
                return CacheKeyInfo(id: String(intID))
            default:
                print("\(type) could not translate \(value) from \(object)!!")
                return nil
            }

        default: return nil
        }
    }
}
```

### Logs

_No response_

### Anything else?

https://anilist.github.io/ApiV2-GraphQL-Docs/

Let me know if there's anything else I can add to help!

BobaFetters:
`Int32` and `Int64` values were failing to cast to `ScalarType` because only `Int` conformed to `ScalarType`, adding code to the `subscript` functions of `ObjectData` and `ListData` to handle casting `AnyHashable`  representations of `Int32` and `Int64` to just `Int` before the attempted casting to `ScalarType`

Closes #3034

### DIFF
--- a/Sources/ApolloAPI/JSONStandardTypeConversions.swift
+++ b/Sources/ApolloAPI/JSONStandardTypeConversions.swift
@@ -42,6 +42,32 @@ extension Int: JSONDecodable, JSONEncodable {
   }
 }
 
+extension Int32: JSONDecodable, JSONEncodable {
+  @inlinable public init(_jsonValue value: JSONValue) throws {
+    guard let number = value as? NSNumber else {
+      throw JSONDecodingError.couldNotConvert(value: value, to: Int32.self)
+    }
+    self = number.int32Value
+  }
+
+  @inlinable public var _jsonValue: JSONValue {
+    return self
+  }
+}
+
+extension Int64: JSONDecodable, JSONEncodable {
+  @inlinable public init(_jsonValue value: JSONValue) throws {
+    guard let number = value as? NSNumber else {
+      throw JSONDecodingError.couldNotConvert(value: value, to: Int64.self)
+    }
+    self = number.int64Value
+  }
+
+  @inlinable public var _jsonValue: JSONValue {
+    return self
+  }
+}
+
 extension Float: JSONDecodable, JSONEncodable {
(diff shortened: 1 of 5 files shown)


calvincestari (approved):
This looks good, thanks @BobaFetters. My only nitpick is updating the PR description to accurately reflect the diff.

calvincestari:
@BobaFetters - I was thinking this morning that none of the execution sources perform JSON deserialization. We should probably add another test that actually uses `JSONSerialization` which would match the execution pattern that led us to this bug.

PR #3045, opened 2023-05-25
Fix RootEntityType and FulfilledFragment name generation

### LINKED ISSUE #2962: Generated code is sometimes incomplete when using @include directive
### Summary

Codegen seems to omit some type when using @include directive on top level fragment.

### Version

1.1.2

Codegen seems to leave this line incomplete when using @include directive on top level fragment (in the definition of `IfFetchParticularEntity`):

`public typealias RootEntityType = `

```
query SomeQuery(
  $id: ID!
  $fetchParticularEntity: Boolean!
) {
  ...ParticularEntityFragment @include(if: $fetchParticularEntity)
  foo(id: $id) {
    ...SomeFragment
  }
}

```

Codegen doesn't crash or anything but the resulting code can't compile.

### Logs

_No response_

### Anything else?

I had no issue in version 1.0.7

AnthonyMDev:
Fully qualified names should be generated at all times for `FulfilledFragments` and `RootEntityType`. Due to the issue indicated in #2949 

Fully qualified names were not being calculated properly for fragments on the root of a query. Simplified the calculation of these by adding the fully qualified root type names to the root of the `FieldPathComponents`. This fixes #2962.

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
+++ b/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
@@ -23,7 +23,7 @@ public class CompilationResult: JavaScriptObject {
     lazy var subscriptionType: GraphQLNamedType? = self["subscriptionType"]
   }
   
-  public class OperationDefinition: JavaScriptObject, Equatable {
+  public class OperationDefinition: JavaScriptObject, Hashable {
     lazy var name: String = self["name"]
     
     lazy var operationType: OperationType = self["operationType"]
@@ -44,6 +44,10 @@ public class CompilationResult: JavaScriptObject {
       "\(name) on \(rootType.debugDescription)"
     }
 
+    public func hash(into hasher: inout Hasher) {
+      hasher.combine(name)
+    }
+    
     public static func ==(lhs: OperationDefinition, rhs: OperationDefinition) -> Bool {
       return lhs.name == rhs.name
     }
--- a/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
+++ b/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
@@ -11,8 +11,7 @@ extension IR {
     )
 
     let rootEntity = Entity(
-      source: .operation(operationDefinition),
-      rootTypePath: LinkedList(operationDefinition.rootType)
+      source: .operation(operationDefinition)      
     )
 
     let result = RootFieldBuilder.buildRootEntityField(
--- a/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
+++ b/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
@@ -11,11 +11,8 @@ extension IR {
     )
 
     let rootEntity = Entity(
-      rootTypePath: LinkedList(operationDefinition.rootType),
-      fieldPath: [.init(
-        name: "\(operationDefinition.generatedDefinitionName).Data",
-        type: rootField.type
-      )]
+      source: .operation(operationDefinition),
+      rootTypePath: LinkedList(operationDefinition.rootType)
     )
 
     let result = RootFieldBuilder.buildRootEntityField(
--- a/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
+++ b/Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
@@ -12,7 +12,10 @@ extension IR {
 
     let rootEntity = Entity(
       rootTypePath: LinkedList(operationDefinition.rootType),
-      fieldPath: [.init(name: rootField.name, type: rootField.type)]
+      fieldPath: [.init(
+        name: "\(operationDefinition.generatedDefinitionName).Data",
+        type: rootField.type
+      )]
     )
 
     let result = RootFieldBuilder.buildRootEntityField(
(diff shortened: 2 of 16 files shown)


Sources/ApolloCodegenLib/IR/IR+OperationBuilder.swift
```diff
@@ -12,7 +12,10 @@ extension IR {
 
     let rootEntity = Entity(
       rootTypePath: LinkedList(operationDefinition.rootType),
-      fieldPath: [.init(name: rootField.name, type: rootField.type)]
+      fieldPath: [.init(
+        name: "\(operationDefinition.generatedDefinitionName).Data",
```
AnthonyMDev, line 16: This is really hacky and a violation of separation of responsibilities. I'm working on a refactor that provides the information about the IR needed in order for the templates to format the name properly.

I'll be putting this PR back into draft.

Sources/ApolloCodegenLib/IR/IR.swift
```diff
@@ -62,30 +62,87 @@ class IR {
   /// Multiple `SelectionSet`s may select fields on the same `Entity`. All `SelectionSet`s that will
   /// be selected on the same object share the same `Entity`.
   class Entity {
-    struct FieldPathComponent: Hashable {
-      let name: String
-      let type: GraphQLType
+
+    /// Represents the location within a GraphQL definition (operation or fragment) of an `Entity`.
+    struct Location: Hashable {
```
AnthonyMDev, line 67: In order to generate selection set names properly, an entity needs to be able to indicate where it is located in its definition. This includes needing to know _what_ definition (operation or fragment) it is defined in.

Previously, we were using a hacky solution of adding an initial `FieldPath` component that included the name of the fragment, or the operation type. This was being used to help properly generate the selection set names, but it was hacky for 2 reasons: 

1. It was making the "field path" not really accurate. Field path included a first component for the source definition that any code that was consuming the field path just needed to understand was there. Causing tight coupling of implementation code.
2. This partially put the responsibility of how we generate names in the templates into the IR. The IR should just represent the shape of the definitions, and the templates consume them to understand how to generate the correct code.

Representing the `Location` with a `SourceDefinition` and a `FieldPath` allows us to better represent the shape of the operations in the IR and allows the templates to do all the logic for formatting names of the Selection Sets.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
```diff
@@ -44,6 +44,10 @@ public class CompilationResult: JavaScriptObject {
       "\(name) on \(rootType.debugDescription)"
     }
 
+    public func hash(into hasher: inout Hasher) {
+      hasher.combine(name)
```
AnthonyMDev, line 48: There can never be 2 operations with the same name, as graphql-js would fail validation. So this is safe

Sources/ApolloCodegenLib/LinkedList.swift
```diff
@@ -71,24 +75,25 @@ public struct LinkedList<T>: ExpressibleByArrayLiteral {
...
+  @_disfavoredOverload
   public init(_ headValue: T) {    
     self.init(head: HeadNode(value: headValue))
   }
 
-  public init(array: [T]) {
-    precondition(!array.isEmpty, "Cannot initialize LinkedList with an empty array. LinkedList must have at least one element.")
-    var segments = array
-    let headNode = HeadNode(value: segments.removeFirst())
+  public init<C: Collection<T>>(_ collection: C) {
```
AnthonyMDev, line 83: Made this function take in a generic collection instead of only an array. This allows us to pass in another linked list or an ordered set.

Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -707,28 +706,116 @@ fileprivate extension IR.MergedSelections.MergedSource {
...
+    pluralizer: Pluralizer
+  ) -> String {
+    generatedSelectionSetName(
+      for: mergedSource.typeInfo,
+      format: format,
+      pluralizer: pluralizer
+    )
+  }
+
+  private static func generatedSelectionSetName(
```
AnthonyMDev, line 744: This is the core of the changes in the template. The rest is mostly updating call sites to use these new functions.

This refactors the selection set name computation to account for the fact that the `fieldPath` no longer includes a component for the root definition.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -6037,6 +6037,153 @@ class SelectionSetTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected, atLine: 1, ignoringExtraLines: true))
   }
 
+  func test__render_conditionalFragmentOnQueryRoot__rendersRootEntityType() throws {
```
AnthonyMDev, line 6040: These are the tests for the actual bug fix. The rest of the test changes are just refactoring for other minor API changes.

Sources/ApolloCodegenLib/IR/IR.swift
BobaFetters: Like this change, especially helping separate responsibilities between the IR and Templates

BobaFetters (approved):
I think this looks good, nice work!

calvincestari (approved):
LGTM!

PR #3048, opened 2023-05-26
Also look for Package.resolved in .xcworkspace when performing version check

alexdeem:
When using an xcworkspace the Package.resolved file is stored at `my.xcworkspace/xcshareddata/swiftpm/Package.resolved`

This change prioritises Package.resolved located in .xcworkspace over those in .xcodeproj, but below those located in the root.

### DIFF
--- a/Sources/CodegenCLI/Extensions/VersionChecker.swift
+++ b/Sources/CodegenCLI/Extensions/VersionChecker.swift
@@ -44,24 +44,22 @@ enum VersionChecker {
     // `.xcworkspace` or the `.xcproject` package. Since we don't know the name of your project,
     // we just look for the first workspace or project that depends on Apollo, prioritising
     // workspaces. This may not be 100% fool-proof, but it should be accurate in almost all cases.
-    func findInXcodeWorkspaces() -> PackageResolvedModel? {
+    func findInXcode(
+      fileSuffix: String,
+      packagePath: String,
+      excludeFilePathComponents: [String] = []
+    ) -> PackageResolvedModel? {
       let projectEnumerator = fileManager.base.enumerator(atPath: projectRootURL?.path ?? ".")
-      while let file = projectEnumerator?.nextObject() as? String {
-        if file.hasSuffix(".xcworkspace") {
-          let projectPackagePath = "\(file)/xcshareddata/swiftpm/\(Package_resolved)"
-          if let package = apolloDependantPackage(atPath: projectPackagePath) {
-            return package
+      enumeratorLoop: while let file = projectEnumerator?.nextObject() as? String {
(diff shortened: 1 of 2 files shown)


Sources/CodegenCLI/Extensions/VersionChecker.swift
```diff
@@ -40,42 +40,63 @@ enum VersionChecker {
...
-    // 100% fool-proof, but it should be accurate in almost all cases.
-    func findInXcodeProject(named projectFileName: String) -> Data? {
-      let projectPackagePath =
-      "\(projectFileName)/project.xcworkspace/xcshareddata/swiftpm/\(Package_resolved)"
-
+    // When using SPM via Xcode, the `Package.resolved` file is nested inside either the
+    // `.xcworkspace` or the `.xcproject` package. Since we don't know the name of your project,
+    // we just look for the first workspace or project that depends on Apollo, prioritising
+    // workspaces. This may not be 100% fool-proof, but it should be accurate in almost all cases.
+    func findInXcodeWorkspaces() -> PackageResolvedModel? {
```
calvincestari, line 47: These two functions appear identical except for the file suffix and package path. I think it would be a good improvement to extract the bulk of the logic into a single function that has parameters for those two pieces that are different.

Tests/CodegenCLITests/VersionCheckerTests.swift
```diff
@@ -186,6 +186,41 @@ class VersionCheckerTests: XCTestCase {
...
+
+    // when
+    let result = try VersionChecker.matchCLIVersionToApolloVersion(
+      projectRootURL: fileManager.directoryURL
+    )
+
+    // then
+    expect(result).to(equal(.versionMismatch(cliVersion: Constants.CLIVersion, apolloVersion: apolloVersion)))
+  }
+
```
calvincestari, line 223: There should also be a test that creates both an `xcworkspace` and `xcproject` file to ensure that the workspace is prioritized over the project file.

calvincestari (changes requested):
Thanks for the contribution @alexdeem. It looks good and I've made a couple suggestions for improvements.

calvincestari:
The test failures also look suspect. I recommend rebasing on `main` because we made some improvements to our CI jobs recently.

Tests/CodegenCLITests/VersionCheckerTests.swift
calvincestari: Resolved in https://github.com/apollographql/apollo-ios/pull/3048/commits/9a7a1a09e1d708e1dbe0af8a0be5d695ffbaac1c.

calvincestari (approved):
LGTM!

PR #3051, opened 2023-05-30
Add linker settings for iOSApplicationExtension

Iron-Ham:
This package doesn't use any APIs disallowed by `iOSApplicationExtension`.  In some instances, linking against this package can still trigger the warning: `ld: warning: linking against a dylib which is not safe for use in application extensions`.  This commit adds linker settings to suppress this warning, while simultaneously adding a build setting that treats these warnings as errors.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -44,13 +44,25 @@ let package = Package(
       ],
       exclude: [
         "Info.plist"
+      ],
+      swiftSettings: [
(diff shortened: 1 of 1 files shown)


Iron-Ham:
Closing bases on this limitation https://github.com/apple/swift-evolution/blob/main/proposals/0238-package-manager-build-settings.md

PR #3058, opened 2023-06-02
fix: Adding type name suffixing for reserved keywords

### LINKED ISSUE #3006: Using reserved keyword `Type` as in selection fields does not compile
### Summary

Using reserved keywords for fields in a type results in the failure to compile the project. I have a type called `QueueDto` which has several fields, one of which is called `Type`. When using version 1.0.5 the `Type` will be generated as `GraphQLEnum<Type>` which is fine, however, when using any version higher than 1.0.5 (even the latest, 1.1.3), it will append it with the namespace (in my case `MainApi`) which tries to use `GraphQLEnum<MainType.Type>`, where `.Type` is probably some reserved Swift keyword or property. Is there any way to get around this problem, I would not like to rename the field at the server-side, is there any way to transform it for code generation or any other possible option I could try?

Part of the generated query with version 1.0.5
```swift
public var trackId: Long { __data["trackId"] }
public var type: GraphQLEnum<Type> { __data["type"] }
```

Part of the generated query with version 1.1.3
```swift
public var trackId: MainApi.Long { __data["trackId"] }
public var type: GraphQLEnum<MainApi.Type> { __data["type"] }
```

### Version

1.1.3

In order to reproduce, you need to have a schema which has a type that has a field called `type`. When generating the code, a swift file will be generated (in this case GetQueueQuery.graphql.swift) which will contain a reference to `YourNamespace.Type`. I have attached all possible files I could think of to reproduce, please let me know if there is something missing. 

Schema.graphqls file
```graphqls
"""
A directive used by the Apollo iOS client to annotate operations or fragments that should be used exclusively for generating local cache mutations instead of as standard operations.
"""
directive @apollo_client_ios_localCacheMutation on QUERY | MUTATION | SUBSCRIPTION | FRAGMENT_DEFINITION

"""A type representing a DateTime"""
scalar DateTime

"""Long type"""
scalar Long

input PagingDtoInput {
  limit: Int!
  offset: Int!
}

type Query {
  getQueue(paging: PagingDtoInput!): [QueueDto!]!
}

type QueueDto {
  trackId: Long!
  type: Type!
}

enum Type {
  FALLBACK
  ADDED_BY_VOTE
  VOTED_FALLBACK
}
```

GetQueue.graphql
```graphql
query GetQueue($offset: Int!, $limit: Int!) {
    getQueue(paging: { offset: $offset, limit: $limit }) {
        trackId
        type
    }
}
```

apollo-codegen-config-main.json
```json
{
  "schemaName": "MainApi",
  "options": {
    "deprecatedEnumCases": "include",
    "schemaDocumentation": "include",
    "conversionStrategies": {
      "enumCases": "camelCase"
    }
  },
  "input": {
    "operationSearchPaths": [
      "graphql/main/**/*.graphql"
    ],
    "schemaSearchPaths": [
      "graphql/main/**/*.graphqls"
    ]
  },
  "output": {
    "testMocks": {
      "none": {
      }
    },
    "schemaTypes": {
      "path": "./graphql/main/graphql-main-api",
      "moduleType": {
        "swiftPackageManager": {
        }
      }
    },
    "operations": {
      "inSchemaModule": {
      }
    }
  }
}
```

### Logs

_No response_

### Anything else?

_No response_

BobaFetters:
- Added `formattedName` value to `GraphlQLNamedType` and overrides in subclasses for specific behavior
- Added checks for reserved keyword usage to type name to apply appropriate suffixes to avoid conflicts

Closes #3006

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
+++ b/Sources/ApolloCodegenLib/Frontend/GraphQLSchema.swift
@@ -31,6 +31,10 @@ public class GraphQLSchema: JavaScriptObject {
 
 public class GraphQLNamedType: JavaScriptObject, Hashable {
   lazy var name: String = self["name"]
+  
+  var formattedName: String {
+    swiftName
+  }
 
   lazy var documentation: String? = self["description"]
 
@@ -66,10 +70,26 @@ public class GraphQLScalarType: GraphQLNamedType {
       return false
     }
   }
+  
+  override var formattedName: String {
+    if !isCustomScalar {
+      return swiftName
+    }
+    
+    let uppercasedName = swiftName.firstUppercased
+    return SwiftKeywords.TypeNamesToSuffix.contains(uppercasedName) ?
+            "\(uppercasedName)_Scalar" : uppercasedName
+  }
 }
 
 public class GraphQLEnumType: GraphQLNamedType {
   lazy var values: [GraphQLEnumValue] = try! invokeMethod("getValues")
+  
+  override var formattedName: String {
+    let uppercasedName = swiftName.firstUppercased
+    return SwiftKeywords.TypeNamesToSuffix.contains(uppercasedName) ?
+            "\(uppercasedName)_Enum" : uppercasedName
+  }
 }
 
 public class GraphQLEnumValue: JavaScriptObject {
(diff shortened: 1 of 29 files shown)


calvincestari (approved):
This looks great Zach - thanks!

This might need to go out in a minor release since it could change the naming of existing structs/classes in generated code. So it's a breaking change, not a very serious one, but we can't take advantage of Xcode renaming annotation which is the annoying bit for users.

BobaFetters:
@calvincestari Agreed, definitely a minor breaking change and should go out as part of 1.3.

AnthonyMDev (approved):

PR #3065, opened 2023-06-08
fix: `RequestChain` `managedSelf` cleanup

### LINKED ISSUE #3057: Memory leak from InterceptorRequestChain when ending the chain with returnValueAsync
### Summary

I'm reporting a memory leak in the iOS Apollo library version 1.2.0 where the request chain does not get deallocated when there is a cache hit.

When a request chain finishes with `returnValueAsync` and does not call `proceedAsync`, the `releaseManagedSelf` never gets called. The retain cycle never gets broken, and the chain leaks. This typically happens in the success case of `CacheReadInterceptor.returnCacheDataElseFetch`.

I confirmed that adding a call to `self.releaseManagedSelf` from `InterceptorRequestChain.returnValueAsync` solves the memory leak: all of the chain instances get deallocated properly. 

However, this solution would break the `CacheReadInterceptor.returnCacheDataAndFetch` case, which calls both `returnValueAsync` and `proceedAsync`. It wouldn't be right to call `releaseManagedSelf` prematurely before the chain has finished. There's no way to know when `returnValueAsync` is called whether that is the end or not.

There's another leak in the `handleErrorAsync` function because it does not call `self.releaseManagedSelf`, either. But that is always the end of the chain so it should be safe to release self there.

### Version

1.2.0

I discovered the problem using the memory graph debugger in Xcode after executing many queries in the app. As long as some of them return cached data, a large multi-node memory cycle will appear in the debugger.

### Logs

```shell
▿ 68 elements
  - 0 : "0   ???                                 0x000000010d6ed4cc 0x0 + 4520334540"
  - 1 : "1   ???                                 0x000000010d6ed5a7 0x0 + 4520334759"
  - 2 : "2   AppName                                0x0000000100dc9050 main + 0"
  - 3 : "3   AppName                                0x0000000101ca6cd7 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1335"
  - 4 : "4   AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 5 : "5   AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 6 : "6   AppName                                0x00000001013e0a38 $s18AppNameServiceClients25GraphQLMetricsInterceptorC14interceptAsync5chain7request8response10completiony6Apollo12RequestChain_p_AI11HTTPRequestCyxGAI12HTTPResponseCyxGSgys6ResultOyAI0D8QLResultVy4DataQzGs5Error_pGct0M3API0D11QLOperationRzlF6$deferL_yyA_A0_RzlF + 184"
  - 7 : "7   AppName                                0x00000001013e094e $s18AppNameServiceClients25GraphQLMetricsInterceptorC14interceptAsync5chain7request8response10completiony6Apollo12RequestChain_p_AI11HTTPRequestCyxGAI12HTTPResponseCyxGSgys6ResultOyAI0D8QLResultVy4DataQzGs5Error_pGct0M3API0D11QLOperationRzlF + 350"
  - 8 : "8   AppName                                0x00000001013e57d5 $s18AppNameServiceClients25GraphQLMetricsInterceptorC6Apollo0gF0AadEP14interceptAsync5chain7request8response10completionyAD12RequestChain_p_AD11HTTPRequestCyqd__GAD12HTTPResponseCyqd__GSgys6ResultOyAD0D8QLResultVy4DataQyd__Gs5Error_pGct0G3API0D11QLOperationRd__lFTW + 21"
  - 9 : "9   AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 10 : "10  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 11 : "11  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 12 : "12  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 13 : "13  AppName                                0x0000000101c6e5b1 $s6Apollo21CacheWriteInterceptorV14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 1249"
  - 14 : "14  AppName                                0x0000000101c6e8e7 $s6Apollo21CacheWriteInterceptorVAA0aD0A2aDP14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyqd__GAA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0P11QLOperationRd__lFTW + 23"
  - 15 : "15  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 16 : "16  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 17 : "17  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 18 : "18  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 19 : "19  AppName                                0x00000001013e76e2 $s18AppNameServiceClients32GraphQLResponseHeaderInterceptorV14interceptAsync5chain7request8response10completiony6Apollo12RequestChain_p_AI11HTTPRequestCyxGAI12HTTPResponseCyxGSgys6ResultOyAI0D8QLResultVy4DataQzGs5Error_pGct0N3API0D11QLOperationRzlF + 1794"
  - 20 : "20  AppName                                0x00000001013e7df9 $s18AppNameServiceClients32GraphQLResponseHeaderInterceptorV6Apollo0hG0AadEP14interceptAsync5chain7request8response10completionyAD12RequestChain_p_AD11HTTPRequestCyqd__GAD12HTTPResponseCyqd__GSgys6ResultOyAD0D8QLResultVy4DataQyd__Gs5Error_pGct0H3API0D11QLOperationRd__lFTW + 9"
  - 21 : "21  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 22 : "22  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 23 : "23  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 24 : "24  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 25 : "25  AppName                                0x0000000101c6b972 $s6Apollo34AutomaticPersistedQueryInterceptorV14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0Q11QLOperationRzlF + 2402"
  - 26 : "26  AppName                                0x0000000101c6c0d9 $s6Apollo34AutomaticPersistedQueryInterceptorVAA0aE0A2aDP14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyqd__GAA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0Q11QLOperationRd__lFTW + 9"
  - 27 : "27  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 28 : "28  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 29 : "29  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 30 : "30  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 31 : "31  AppName                                0x0000000101cad3ca $s6Apollo30JSONResponseParsingInterceptorV14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 1370"
  - 32 : "32  AppName                                0x0000000101cad919 $s6Apollo30JSONResponseParsingInterceptorVAA0aD0A2aDP14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyqd__GAA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0P11QLOperationRd__lFTW + 9"
  - 33 : "33  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 34 : "34  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 35 : "35  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 36 : "36  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 37 : "37  AppName                                0x0000000101cc3723 $s6Apollo23ResponseCodeInterceptorV14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 611"
  - 38 : "38  AppName                                0x0000000101cc39d9 $s6Apollo23ResponseCodeInterceptorVAA0aD0A2aDP14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyqd__GAA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0P11QLOperationRd__lFTW + 9"
  - 39 : "39  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 40 : "40  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 41 : "41  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 42 : "42  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 43 : "43  AppName                                0x00000001013e5c79 $s18AppNameServiceClients25GraphQLLatencyTimerFinishC14interceptAsync5chain7request8response10completiony6Apollo12RequestChain_p_AI11HTTPRequestCyxGAI12HTTPResponseCyxGSgys6ResultOyAI0D8QLResultVy4DataQzGs5Error_pGct0N3API0D11QLOperationRzlF + 249"
  - 44 : "44  AppName                                0x00000001013e6075 $s18AppNameServiceClients25GraphQLLatencyTimerFinishC6Apollo0H11InterceptorAadEP14interceptAsync5chain7request8response10completionyAD12RequestChain_p_AD11HTTPRequestCyqd__GAD12HTTPResponseCyqd__GSgys6ResultOyAD0D8QLResultVy4DataQyd__Gs5Error_pGct0H3API0D11QLOperationRd__lFTW + 21"
  - 45 : "45  AppName                                0x0000000101c6350a $s6Apollo0A27InterceptorReentrantWrapperC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlF + 314"
  - 46 : "46  AppName                                0x0000000101ca6bd8 $s6Apollo23InterceptorRequestChainC12proceedAsync7request8response10completion11interceptoryAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGcAA0aB16ReentrantWrapperCt0A3API0N11QLOperationRzlF + 1080"
  - 47 : "47  AppName                                0x0000000101c62e6e $s6Apollo0A27InterceptorReentrantWrapperC12proceedAsync7request8response10completionyAA11HTTPRequestCyxG_AA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0M11QLOperationRzlF + 174"
  - 48 : "48  AppName                                0x0000000101c63331 $s6Apollo0A27InterceptorReentrantWrapperCAA12RequestChainA2aDP12proceedAsync7request8response10completionyAA11HTTPRequestCyqd__G_AA12HTTPResponseCyqd__GSgys6ResultOyAA13GraphQLResultVy4DataQyd__Gs5Error_pGct0A3API0O11QLOperationRd__lFTW + 17"
  - 49 : "49  AppName                                0x0000000101cb98f3 $s6Apollo23NetworkFetchInterceptorC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlFyARy10FoundationAUV_So17NSHTTPURLResponseCtsAX_pGcfU_ + 1315"
  - 50 : "50  AppName                                0x0000000101cb9a3c $s6Apollo23NetworkFetchInterceptorC14interceptAsync5chain7request8response10completionyAA12RequestChain_p_AA11HTTPRequestCyxGAA12HTTPResponseCyxGSgys6ResultOyAA13GraphQLResultVy4DataQzGs5Error_pGct0A3API0P11QLOperationRzlFyARy10FoundationAUV_So17NSHTTPURLResponseCtsAX_pGcfU_TA + 76"
  - 51 : "51  AppName                                0x0000000101cc9318 $s6Apollo16URLSessionClientC10urlSession_4task20didCompleteWithErrorySo12NSURLSessionC_So0K4TaskCs0J0_pSgtF + 1240"
  - 52 : "52  AppName                                0x0000000101cc9666 $s6Apollo16URLSessionClientC10urlSession_4task20didCompleteWithErrorySo12NSURLSessionC_So0K4TaskCs0J0_pSgtFTo + 134"
  - 53 : "53  CFNetwork                           0x00007ff804105dbd CFNetwork + 56765"
  - 54 : "54  Foundation                          0x00007ff800bd3d1a __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 7"
  - 55 : "55  Foundation                          0x00007ff800bd3c12 -[NSBlockOperation main] + 98"
  - 56 : "56  Foundation                          0x00007ff800bd6c17 __NSOPERATION_IS_INVOKING_MAIN__ + 17"
  - 57 : "57  Foundation                          0x00007ff800bd2e7f -[NSOperation start] + 782"
  - 58 : "58  Foundation                          0x00007ff800bd755d __NSOPERATIONQUEUE_IS_STARTING_AN_OPERATION__ + 17"
  - 59 : "59  Foundation                          0x00007ff800bd70ab __NSOQSchedule_f + 182"
  - 60 : "60  libdispatch.dylib                   0x000000010c07b0fd _dispatch_block_async_invoke2 + 83"
  - 61 : "61  libdispatch.dylib                   0x000000010c0697ec _dispatch_client_callout + 8"
  - 62 : "62  libdispatch.dylib                   0x000000010c06ca44 _dispatch_continuation_pop + 836"
  - 63 : "63  libdispatch.dylib                   0x000000010c06b806 _dispatch_async_redirect_invoke + 997"
  - 64 : "64  libdispatch.dylib                   0x000000010c07e8d7 _dispatch_root_queue_drain + 414"
  - 65 : "65  libdispatch.dylib                   0x000000010c07f57c _dispatch_worker_thread2 + 278"
  - 66 : "66  libsystem_pthread.dylib             0x00007ff837749c0f _pthread_wqthread + 257"
  - 67 : "67  libsystem_pthread.dylib             0x00007ff837748bbf start_wqthread + 15"
```

### Anything else?

_No response_

calvincestari:
Fixes #3057 

Resolves the leak of `managedSelf` by providing a `terminate` function on `RequestChain` that allows an interceptor to signal whether it is ending the request chain. This allows interceptors such as the `CacheReadInterceptor` to signal an early exit of the request, such as in the case of using the `returnCacheDataDontFetch` cache policy with a cache hit.

### DIFF
--- a/Sources/Apollo/InterceptorRequestChain.swift
+++ b/Sources/Apollo/InterceptorRequestChain.swift
@@ -174,6 +174,20 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
     _ = self.releaseManagedSelf
   }
 
+  /// Ends execution of the request chain.
+  ///
+  /// Call this function when there is an early exit of the request chain, such as a cache read
+  /// interceptor that handles a request to only fetch from the cache and not send the request to
+  /// the server.
+  public func terminate() {
+    guard !self.isCancelled else {
+      // Do not proceed, this chain has been cancelled.
+      return
+    }
+
+    _ = self.releaseManagedSelf
+  }
+
   /// Restarts the request starting from the first interceptor.
   ///
   /// - Parameters:
--- a/Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
+++ b/Sources/Apollo/ApolloInterceptorReentrantWrapper.swift
@@ -56,6 +56,15 @@ class ApolloInterceptorReentrantWrapper: RequestChain {
     }
   }
 
+  func terminate() {
+    guard !self.isCancelled else {
+      // Do not proceed, this chain has been cancelled.
+      return
+    }
+
+    requestChain.takeUnretainedValue().terminate()
+  }
+
   func retry<Operation>(
     request: HTTPRequest<Operation>,
     completion: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void
--- a/Sources/Apollo/CacheReadInterceptor.swift
+++ b/Sources/Apollo/CacheReadInterceptor.swift
@@ -65,6 +65,7 @@ public struct CacheReadInterceptor: ApolloInterceptor {
               chain.returnValueAsync(for: request,
                                      value: graphQLResult,
                                      completion: completion)
+              chain.terminate()
             }
           }
         case .returnCacheDataDontFetch:
@@ -81,6 +82,8 @@ public struct CacheReadInterceptor: ApolloInterceptor {
                                      value: result,
                                      completion: completion)
             }
+
+            chain.terminate()
           }
         }
       }
--- a/Sources/Apollo/RequestChain.swift
+++ b/Sources/Apollo/RequestChain.swift
@@ -3,6 +3,22 @@ import Foundation
 import ApolloAPI
 #endif
 
+/// A request chain defines a sequence of interceptors that handle the lifecycle of a particular
+/// GraphQL operation's execution. One interceptor might add custom HTTP headers to a request,
+/// while the next might be responsible for actually sending the request to a GraphQL server over
+/// HTTP. A third interceptor might then write the operation's result to the cache.
+///
+/// When an operation is executed an object called an `InterceptorProvider` generates a RequestChain
+/// for the operation. Then `kickoff` is called on the request chain, which runs the first
+/// interceptor in the chain.
+///
+/// An interceptor can perform arbitrary, asynchronous logic on any thread. When an interceptor
+/// finishes running, it calls `proceedAsync` on its `RequestChain`, which advances to the next
+/// interceptor.
+///
+/// By default when the last interceptor in the chain finishes, if a parsed operation result is
+/// available, that result is returned to the operation's original caller. Otherwise, error-handling
+/// logic is called.
 public protocol RequestChain: Cancellable {
   func kickoff<Operation>(
     request: HTTPRequest<Operation>,
@@ -16,6 +32,8 @@ public protocol RequestChain: Cancellable {
   ) where Operation : GraphQLOperation
 
   func cancel()
+
+  func terminate()
 
   func retry<Operation>(
     request: HTTPRequest<Operation>,
--- a/Tests/ApolloInternalTestHelpers/InterceptorTester.swift
+++ b/Tests/ApolloInternalTestHelpers/InterceptorTester.swift
@@ -49,6 +49,8 @@ fileprivate class ResponseCaptureRequestChain: RequestChain {
 
   func cancel() {}
 
+  func terminate() {}
+
   func retry<Operation>(
     request: Apollo.HTTPRequest<Operation>,
     completion: @escaping (Result<Apollo.GraphQLResult<Operation.Data>, Error>) -> Void
--- a/Tests/ApolloTests/RequestChainTests.swift
+++ b/Tests/ApolloTests/RequestChainTests.swift
@@ -19,7 +19,7 @@ class RequestChainTests: XCTestCase {
       defer {
         expectation.fulfill()
       }
-      
+
       switch result {
       case .success:
         XCTFail("This should not have succeeded")
@@ -33,11 +33,11 @@ class RequestChainTests: XCTestCase {
         }
       }
     }
-    
-    
+
+
     self.wait(for: [expectation], timeout: 1)
   }
-  
+
   func testCancellingChainCallsCancelOnInterceptorsWhichImplementCancellableAndNotOnOnesThatDont() {
     class TestProvider: InterceptorProvider {
       let cancellationInterceptor = CancellationHandlingInterceptor()
@@ -60,29 +60,29 @@ class RequestChainTests: XCTestCase {
       XCTFail("This should not have gone through")
       expectation.fulfill()
     }
-    
+
     cancellable.cancel()
     XCTAssertTrue(provider.cancellationInterceptor.hasBeenCancelled)
     XCTAssertFalse(provider.retryInterceptor.hasBeenCancelled)
     self.wait(for: [expectation], timeout: 2)
   }
-  
+
   func test__send__ErrorInterceptorGetsCalledAfterAnErrorIsReceived() {
     class ErrorInterceptor: ApolloErrorInterceptor {
       var error: Error? = nil
-      
+
       func handleErrorAsync<Operation: GraphQLOperation>(
           error: Error,
           chain: RequestChain,
           request: HTTPRequest<Operation>,
           response: HTTPResponse<Operation>?,
           completion: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void) {
-      
+
         self.error = error
         completion(.failure(error))
       }
     }
-    
+
     class TestProvider: InterceptorProvider {
       let errorInterceptor = ErrorInterceptor()
       func interceptors<Operation: GraphQLOperation>(for operation: Operation) -> [ApolloInterceptor] {
@@ -91,7 +91,7 @@ class RequestChainTests: XCTestCase {
           AutomaticPersistedQueryInterceptor()
         ]
       }
-      
+
       func additionalErrorInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> ApolloErrorInterceptor? {
         return self.errorInterceptor
       }
@@ -101,7 +101,7 @@ class RequestChainTests: XCTestCase {
     let transport = RequestChainNetworkTransport(interceptorProvider: provider,
                                                  endpointURL: TestURL.mockServer.url,
                                                  autoPersistQueries: true)
-    
+
     let expectation = self.expectation(description: "Hero name query complete")
     _ = transport.send(operation: MockQuery.mock()) { result in
       defer {
@@ -120,9 +120,9 @@ class RequestChainTests: XCTestCase {
         }
       }
     }
-    
+
     self.wait(for: [expectation], timeout: 1)
-    
+
     switch provider.errorInterceptor.error {
     case .some(let error):
       switch error {
@@ -213,33 +213,33 @@ class RequestChainTests: XCTestCase {
       XCTFail("Error interceptor did not receive an error!")
     }
   }
-  
+
   func testErrorInterceptorGetsCalledInDefaultInterceptorProviderSubclass() {
     class ErrorInterceptor: ApolloErrorInterceptor {
       var error: Error? = nil
-      
+
       func handleErrorAsync<Operation: GraphQLOperation>(
         error: Error,
         chain: RequestChain,
         request: HTTPRequest<Operation>,
         response: HTTPResponse<Operation>?,
         completion: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void) {
-        
+
         self.error = error
         completion(.failure(error))
       }
     }
-    
+
     class TestProvider: DefaultInterceptorProvider {
       let errorInterceptor = ErrorInterceptor()
-      
+
       override func interceptors<Operation: GraphQLOperation>(for operation: Operation) -> [ApolloInterceptor] {
         return [
           // An interceptor which will error without a response
           AutomaticPersistedQueryInterceptor()
         ]
       }
-      
+
       override func additionalErrorInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> ApolloErrorInterceptor? {
         return self.errorInterceptor
       }
@@ -249,7 +249,7 @@ class RequestChainTests: XCTestCase {
     let transport = RequestChainNetworkTransport(interceptorProvider: provider,
                                                  endpointURL: TestURL.mockServer.url,
                                                  autoPersistQueries: true)
-    
+
     let expectation = self.expectation(description: "Hero name query complete")
     _ = transport.send(operation: MockQuery.mock()) { result in
       defer {
@@ -268,9 +268,9 @@ class RequestChainTests: XCTestCase {
         }
       }
     }
-    
+
     self.wait(for: [expectation], timeout: 1)
-    
+
     switch provider.errorInterceptor.error {
     case .some(let error):
       switch error {
@@ -286,7 +286,7 @@ class RequestChainTests: XCTestCase {
   }
 
   // MARK: Multipart subscription tests
-  
+
   struct RequestTrapInterceptor: ApolloInterceptor {
     let callback: (URLRequest) -> (Void)
 
@@ -418,7 +418,7 @@ class RequestChainTests: XCTestCase {
     var name: String { __data["name"] }
   }
 
-  func test__retain_release__givenQuery_shouldNotHaveRetainCycle() throws {
+  func test__managedSelf__givenQuery_whenCompleted_shouldNotHaveRetainCycle() throws {
     // given
     let client = MockURLSessionClient(
       response: .mock(
@@ -479,7 +479,7 @@ class RequestChainTests: XCTestCase {
     XCTAssertNil(weakRequestChain)
   }
 
-  func test__retain_release__givenSubscription_whenCancelled_shouldNotHaveRetainCycle() throws {
+  func test__managedSelf__givenSubscription_whenCancelled_shouldNotHaveRetainCycle() throws {
     // given
     let client = MockURLSessionClient(
       response: .mock(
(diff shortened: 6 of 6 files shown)


Sources/Apollo/InterceptorRequestChain.swift
```diff
@@ -25,11 +25,6 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
   @Atomic public var isCancelled: Bool = false
 
   private var managedSelf: Unmanaged<InterceptorRequestChain>!
-  // This is to fix #2932, caused by overreleasing managedSelf on cancellation of a query or
-  // mutation. It can now only be released once.
-  private lazy var releaseManagedSelf: Void = {
```
calvincestari, line 30: Removed this singleton 'management' of the `Unmanaged` object so we can retain against on retries.

calvincestari:
Note that this PR only updates the `CacheReadInterceptor` with the required calls to `terminate()`. The other Apollo-built interceptors will need to be updated to prevent memory leaks when they have early-exit logic paths.

calvincestari:
OK, I think I've finally got this PR ready to go with all Apollo interceptors updated with calls to `terminate()`.

Sources/Apollo/CacheWriteInterceptor.swift
```diff
@@ -69,6 +71,15 @@ public struct CacheWriteInterceptor: ApolloInterceptor {
...
                              response: response,
                              completion: completion)
+
+      terminate(chain: chain, ifNeededFor: Operation.operationType)
+    }
+  }
+
+  private func terminate(chain: RequestChain, ifNeededFor operationType: GraphQLOperationType) {
+    // HTTP-based subscriptions are re-entrant so this may not be terminal if not a subscription
+    if operationType != .subscription {
```
calvincestari, line 81: This is an example of the implementation detail that interceptor authors would be required to know for this solution; because this interceptor is after the `MultipartResponseParsingInterceptor` and HTTP-based subscriptions are re-entrant it means that `terminate` should only be called for non-subscription operations or in cases where it's known that a subscription event could never be received again.

Sources/Apollo/InterceptorRequestChain.swift
```diff
@@ -23,13 +23,9 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
   private var interceptors: [ApolloInterceptorReentrantWrapper]
   private var callbackQueue: DispatchQueue
   @Atomic public var isCancelled: Bool = false
+  @Atomic private var isFinished: Bool = false
```
calvincestari, line 26: I added this because `cancel` could still be called after an operation had completed which would result in `managedSelf` having unbalanced `release` calls.

Sources/Apollo/AutomaticPersistedQueryInterceptor.swift
```diff
@@ -48,6 +48,8 @@ public struct AutomaticPersistedQueryInterceptor: ApolloInterceptor {
                                request: request,
                                response: response,
                                completion: completion)
+
+        chain.terminate()
```
AnthonyMDev, line 52: If we proceed with this solution, would it be reasonable to just terminate the chain in `handleErrorAsync()`? Or are there situations in which we want to handle and error without termination?

Sources/Apollo/InterceptorRequestChain.swift
AnthonyMDev: Concerned that this will regress 2932 though. How are we ensuring we don't overrelease?

Sources/Apollo/MultipartResponseParsingInterceptor.swift
```diff
@@ -112,6 +112,7 @@ public struct MultipartResponseParsingInterceptor: ApolloInterceptor {
             )
 
             // These are fatal-level transport errors, don't process anything else.
+            chain.terminate()
```
AnthonyMDev, line 115: Do we also need a `terminate` on line 99 here?

AnthonyMDev (commented):
I see what you are saying about this approach feeling wrong. It feels like as we continue to try to work around this issue, we keep making the workaround more and more confusing, hacky, and fragile.

I think the entire team agrees that the `RequestChain` is not a good way forward, and I'm looking forward to the opportunity to completely re-write this.

The other approach here doesn't on its surface feel all that great either. It sounds like you have a vision for it that you feel will work better @calvincestari. But it seems to me that branching into 2 separate implementations of the request chain that we are on the hook to support until 2.0 is going to make things more complicated and make our maintenance (and testing) burden higher. But if you think that approach will be cleaner and more direct, I'm open to it.

My main priority is trying not to spend a disproportionate amount of time building and maintaining solutions that we plan on replacing in the coming months. I wonder, if we do take the second approach you are proposing, how much of the interceptor code do you think could be shared between the two APIs? If we are only branching out in a few minor areas, but can re-use most of the same interceptors, that makes it a lot more reasonable to me.

Sources/Apollo/AutomaticPersistedQueryInterceptor.swift
calvincestari: `handleErrorAsync` doesn't mean the request chain has ended unfortunately; that's in the interceptor implementation and there is nothing in the request chain logic that ends a request with a call to `handleErrorAsync`. It's simply a way for you to get additional information about an error that has been encountered.

Sources/Apollo/MultipartResponseParsingInterceptor.swift
calvincestari: No because there could be more chunks arriving after all other errors. It's only fatal-level transport errors that we can assume the router/server will close the connection.

calvincestari:
> The other approach here doesn't on its surface feel all that great either. It sounds like you have a vision for it that you feel will work better @calvincestari. But it seems to me that branching into 2 separate implementations of the request chain that we are on the hook to support until 2.0 is going to make things more complicated and make our maintenance (and testing) burden higher. But if you think that approach will be cleaner and more direct, I'm open to it.
> 
> My main priority is trying not to spend a disproportionate amount of time building and maintaining solutions that we plan on replacing in the coming months. I wonder, if we do take the second approach you are proposing, how much of the interceptor code do you think could be shared between the two APIs? If we are only branching out in a few minor areas, but can re-use most of the same interceptors, that makes it a lot more reasonable to me.

My gut tells me the other solution is more viable long-term and less maintenance for us. Simply put it's another function in the `RequestChain` protocol and we deprecate the existing `proceedAsync` function. They should be able to share the bulk of the logic once an interceptor has been identified; that identification _should_ be the only difference between the two functions.

We didn't explore this when first building HTTP-based subscriptions because we wanted it to be 100% backwards compatible. It's only now that we're getting more comfortable with the idea of a minor breaking change that this is a potential solution.

calvincestari:
Closed in favour of #3070

PR #3070, opened 2023-06-13
fix: Remove `managedSelf` from `InterceptorRequestChain`

### LINKED ISSUE #3057: Memory leak from InterceptorRequestChain when ending the chain with returnValueAsync

calvincestari:
Fixes #3057 

This PR removes `managedSelf` from `InterceptorRequestChain` going back to the memory management model that was used before the introduction of HTTP-based subscriptions.

The gist of this solution is to deprecate the current `proceedAsync` function, adding a new one that requires the interceptor to identify itself. It's the interceptor identifier that allows us to get rid of `ApolloInterceptorReentrantWrapper` and `managedSelf`.


Sources/Apollo/ApolloInterceptor.swift
```diff
@@ -4,6 +4,8 @@ import ApolloAPI
 
 /// A protocol to set up a chainable unit of networking work.
 public protocol ApolloInterceptor {
+
+  var id: String { get set }
```
calvincestari, line 8: `Identifiable` is only available on iOS 13+ and macOS 15+, we support lower versions of both. This imposes the same 'identifiable' requirement for interceptors.

Sources/Apollo/AutomaticPersistedQueryInterceptor.swift
```diff
@@ -36,35 +38,46 @@ public struct AutomaticPersistedQueryInterceptor: ApolloInterceptor {
...
         return
       }
 
       guard let result = response?.parsedResponse else {
         // This is in the wrong order - this needs to be parsed before we can check it.
-        chain.handleErrorAsync(APQError.noParsedResponse,
-                               request: request,
-                               response: response,
-                               completion: completion)
+        chain.handleErrorAsync(
```
calvincestari, line 52: The changes to these other functions are purely formatting.

Sources/Apollo/InterceptorRequestChain.swift
```diff
@@ -9,28 +9,26 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
...
         return "`proceedAsync` was called for index \(index), which is out of bounds of the receiver for this chain. Double-check the order of your interceptors."
+      case let .unknownInterceptor(id):
+        return "`proceedAsync` was called by unknown interceptor \(id)."
       }
     }
   }
 
-  private var interceptors: [ApolloInterceptorReentrantWrapper]
+  private var interceptors: [any ApolloInterceptor]
+  private var interceptorIndexes: [AnyHashable: Int] = [:]
```
calvincestari, line 27: A new lookup table that gives us a faster way to find an interceptors index.

Sources/Apollo/InterceptorRequestChain.swift
```diff
@@ -97,40 +92,73 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
...
+
+    proceedAsync(
+      interceptorIndex: nextIndex,
+      request: request,
+      response: response,
+      completion: completion
+    )
+  }
+
+  private func proceedAsync<Operation: GraphQLOperation>(
```
calvincestari, line 140: The shared logic of `proceedAsync` so the other two public functions are purely for interceptor lookup.

Sources/Apollo/InterceptorRequestChain.swift
AnthonyMDev: Good call on using a hash map to improve performance here!

If the interceptor id is always a `String` can't this just be a `String: Int` dictionary?

AnthonyMDev: Why are `interceptors` and `interceptorIndexes` variables? Could they be `let` properties, since they are created during initialization?

AnthonyMDev (approved):
Awesome work! This definitely feels way better. Sorry for pushing us in the wrong direction previously.

Sources/Apollo/InterceptorRequestChain.swift
calvincestari: > If the interceptor id is always a String can't this just be a String: Int dictionary?

Yes, it can be. I thought I'd updated that when I moved away from an `associatedtype`. I'll update it.

> Why are interceptors and interceptorIndexes variables? Could they be let properties, since they are created during initialization?

`interceptors` and `callbackQueue` can become `let` but `interceptorIndexes` needs to be `var` because of the way we enumerate through `interceptors` to assign the `index` to the `id`.

Tests/ApolloTests/RequestChainTests.swift
```diff
@@ -963,4 +963,136 @@ class RequestChainTests: XCTestCase {
...
+
+    let request = JSONRequest(
+      operation: MockQuery<Hero>(),
+      graphQLEndpoint: TestURL.mockServer.url,
+      clientName: "test-client",
+      clientVersion: "test-client-version"
+    )
+
+    // when
+    requestChain.kickoff(request: request) { result in }
```
calvincestari, line 1029: We don't care about the response in these tests, there are other tests for that. These purely ensure that the interceptors are executed in the correct order.

BobaFetters (approved):

Sources/Apollo/ApolloInterceptor.swift
lhunath: 1. It would be really helpful to document this method and explain what the purpose and scope of this identifier is, so that custom implementers can decide what the best identifier would be for their specific use case.
2. In order to avoid the breaking change introduced by this, have you considered a default implementation of this requirement using a protocol extension?

calvincestari: > It would be really helpful to document this method and explain what the purpose and scope of this identifier is, so that custom implementers can decide what the best identifier would be for their specific use case.

Documentation can be added to clarify it's purpose, I'll get that done. As long as it's unique to other identifiers and a `String` it shouldn't matter. This was required as a workaround to the interceptor architecture while fixing the memory leak and maintaining the ability to support HTTP-based subscriptions.

> In order to avoid the breaking change introduced by this, have you considered a default implementation of this requirement using a protocol extension?

Yes I did, but Swift extensions cannot hold stored properties and the identifier needs to be constant during the lifetime of the interceptor. Static constants don't work here either.

lhunath: > Documentation can be added to clarify it's purpose, I'll get that done.

Thanks! Perhaps the scope is the interceptor instance? What does "other identifiers" mean concretely? Unique within the set of installed interceptors active during the processing of one request?

> Swift extensions cannot hold stored properties and the identifier needs to be constant during the lifetime of the interceptor.

Ahh, makes sense. Perhaps an alternative might've been an `id { ObjectIdentifier(self) }`, but it would incur a protocol requirement for class-semantics by the implementer.

calvincestari: > What does "other identifiers" mean concretely? Unique within the set of installed interceptors active during the processing of one request?

Correct. The interceptor provider passed into `RequestChainNetworkTransport` during initialization is used to build the [request chain for each request](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/RequestChainNetworkTransport.swift#L130). They need to be unique amongst that array of interceptors.

PR #3073, opened 2023-06-14
Fix merged source name generation when shared root is operation root

### LINKED ISSUE #3071: Code generation throwing an error in 1.2.1 only
### Summary

I'm getting this error during generation of mocks. in only 1.2.1. This works well in 1.2.0.

```
Range requires lowerBound <= upperBound
```

This is what the log looks like when it happens. For reference I'm calling `try ApolloCodegen.build(with: config)` from a swift tools package myself by importing `ApolloCodegenLib`.

```
.
.
.
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
Swift/arm64e-apple-macos.swiftinterface:15840: Fatal error: Range requires lowerBound <= upperBound
```

### Version

1.2.1

I'm still working on isolating a project where I can reproduce it.

### Logs

```shell
.
.
.
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
[DEBUG - ApolloCodegenLib:Glob.swift:225] - Matched /<file-path>/Mocks/<type-name>+Mock.graphql.swift
Swift/arm64e-apple-macos.swiftinterface:15840: Fatal error: Range requires lowerBound <= upperBound
```

### Anything else?

The generated mock from 1.2.0 looks like this

```swift
import Apollo
import ModuleName

public class TypeName: MockObject {
  public static let objectType: Object = ModuleName.Objects.TypeName
  public typealias MockValueCollectionType = Array<Mock<TypeName>>

  public struct MockFields {
  }
}

// The original Object 

public extension Objects {
  static let TypeName = Object(
    typename: "TypeName",
    implementedInterfaces: [Interfaces.InterfaceName.self]
  )
}
```

Another thing to add is that this type does not add anything on top of the fields from the interface itself. All the fields that are part of this type are from the interface.

AnthonyMDev:
This fixes #3071 

When the shared root of a merged source is the direct parent or a direct sibling, we remove the parent from the namespacing of the referenced selection set. If the shared root is the root of the operation/fragment, that component of the namespacing should be removed.

Since the change to the way we indicate the location of a `SelectionSet` in the `IR` changed in #3045, in this case, the shared root is now not part of the `fieldPath`, causing the "shared root" to be at an index of `-1` (because its represented by the `location.source` now). This was causing a crash with an invalid range bounds.

This PR fixes this by using `max(0, sharedRootIndex)` to ensure we don't use `-1`, and then _not_ removing the first component in that case (because that component would be the `location.source` which is already not included.

### DIFF
--- a/Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Fragments/ClassroomPetDetails.graphql.swift
+++ b/Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Fragments/ClassroomPetDetails.graphql.swift
@@ -58,7 +58,7 @@ public struct ClassroomPetDetails: AnimalKingdomAPI.SelectionSet, Fragment {
         "__typename": __typename,
       ],
       fulfilledFragments: [
-        ObjectIdentifier(Self.self)
+        ObjectIdentifier(ClassroomPetDetails.self)
       ]
     ))
   }
@@ -88,8 +88,8 @@ public struct ClassroomPetDetails: AnimalKingdomAPI.SelectionSet, Fragment {
           "species": species,
         ],
         fulfilledFragments: [
-          ObjectIdentifier(Self.self),
-          ObjectIdentifier(ClassroomPetDetails.self)
+          ObjectIdentifier(ClassroomPetDetails.self),
+          ObjectIdentifier(ClassroomPetDetails.AsAnimal.self)
         ]
       ))
     }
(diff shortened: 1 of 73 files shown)


calvincestari (approved):
Thanks @AnthonyMDev 🎉

PR #3074, opened 2023-06-14
Fix parsing of input objects as default values for input params

### LINKED ISSUE #2978: Default values specified in the schema on an input object are causing a crash during code generation
### Summary

I am using a 16-inch, 2019 Laptop ( Intel ) , XCode 14.3 and when I try to run `./Pods/Apollo/apollo-ios-cli generate` I always get the error `illegal hardware instruction`

Is this a known issue? Is it something I am missing to pass ? 

### Version

1.1.2

Just created a new config and run the command

### Logs

```shell
How can I get logs for this operation?
```

### Anything else?

I would really appreciate if anyone could help. Thanks!

Emanuel

AnthonyMDev:
Fixes #2978

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/GraphQLValue.swift
+++ b/Sources/ApolloCodegenLib/Frontend/GraphQLValue.swift
@@ -42,7 +42,22 @@ extension GraphQLValue: JavaScriptValueDecodable {
       }
       self = .list(.fromJSValue(value, bridge: bridge))
     case "ObjectValue":
-      self = .object(.fromJSValue(jsValue["value"], bridge: bridge))
+      let value = jsValue["value"]
+
+      /// The JS frontend does not do value conversions of the default values for input objects,
+      /// because no other compliation is needed, these are passed through as is from `graphql-js`.
+      /// We need to handle both converted object values and default values and represented by
+      /// `graphql-js`.
+      if !value.isUndefined {
+        self = .object(.fromJSValue(value, bridge: bridge))
+
+      } else {
+        let fields = jsValue["fields"].toOrderedDictionary { field in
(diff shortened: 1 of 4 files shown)


Sources/ApolloCodegenLib/Frontend/GraphQLValue.swift
```diff
@@ -42,7 +42,22 @@ extension GraphQLValue: JavaScriptValueDecodable {
       }
       self = .list(.fromJSValue(value, bridge: bridge))
     case "ObjectValue":
-      self = .object(.fromJSValue(jsValue["value"], bridge: bridge))
+      let value = jsValue["value"]
+
+      /// The JS frontend does not do value conversions of the default values for input objects,
+      /// because no other compliation is needed, these are passed through as is from `graphql-js`.
```
BobaFetters, line 48:
```suggestion
      /// because no other compilation is needed, these are passed through as is from `graphql-js`.
```
Assuming this should say compilation?

BobaFetters (approved):
This looks good, nice work!

Sources/ApolloCodegenLib/Frontend/JavaScriptBridge.swift
```diff
@@ -394,6 +394,25 @@ extension JSValue {
 
     return dictionary
   }
+
+  // The regular `toDictionary()` creates an `NSDictionary` that while it preserves the order of
+  // `keys` from JavaScript during initialization, there is no order afterwards. `OrderedDictionary`
+  // provides for the preservation and subsequent use of ordering in the collection.
```
calvincestari, line 400: 👍🏻

calvincestari (approved):
Thanks for fixing this so quickly @AnthonyMDev!

PR #3076, opened 2023-06-15
Make SOCKS proxy support public

### LINKED ISSUE #2788: Restore ability to use SOCKS proxies for debugging
### Use case

I'm trying to debug our web socket usage over the wire, and am mostly unable to use tools like Proxyman or Charles. SOCKS proxying was previously enabled but seems to have been enabled due to a Starscream upgrade (#1104) that was [reverted in 0.47.0](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md#v0470). 

Re-expose the `useSOCKSProxy` variable.

tahirmt:
Closes #2788 

This was originally added in #1108 but removed when `Starscream` was removed but for debugging tools like Charles, it is very useful to be able to see the web socket traffic. 

I made the implementation non-breaking by using the sam approach of `SOCKSProxyable` protocol from #1108.

Added tests for `WebSocketTransport` but I couldn't find create tests of `WebSocket` due to how it is written.

### DIFF
--- a/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
+++ b/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
@@ -40,7 +40,7 @@ public struct SSLSettings {
 
 //WebSocket implementation
 
-public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSocketStreamDelegate {
+public final class WebSocket: NSObject, WebSocketClient, StreamDelegate, WebSocketStreamDelegate, SOCKSProxyable {
 
   public enum OpCode : UInt8 {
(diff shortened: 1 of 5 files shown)


calvincestari (approved):
This looks good, thanks for the contribution @tahirmt.

> Added tests for WebSocketTransport but I couldn't find create tests of WebSocket due to how it is written.

Correct - some parts of the networking stack are difficult to work with at the moment, part of the motivation for the work to be done in 2.0.

calvincestari:
Merging - ignore the test failure, I understand what's going wrong and will fix it.

_tl;dr - it's a CI/test configuration, not incorrect logic._

PR #3083, opened 2023-06-15
fix: Exclude script from target source

calvincestari:
Excludes the internal script from the target source to fix the build warning:
```
❯ swift build
warning: 'apollo-ios': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/calvincestari/Projects/apollo-ios/Sources/ApolloCodegenLib/Frontend/auto_rollup.sh
```


BobaFetters (approved):
Good catch!

PR #3089, opened 2023-06-22
fix: Updating test mock set method definitions

### LINKED ISSUE #3023: Mock that contains an array of Union type crashes on access
### Summary

Hey folks,

I'm unsure if this is a problem of us "holding it wrong" or something that is an issue, so we appreciate any comments or advice.

We have an issue where after migrating from version 1.1.2 to 1.2.0 of Apollo, we see crashes when accessing specific values on mocks. 

These crashes always occur when accessing properties that have a type that is an array of a union type. We do _not_ see similar crashes when accessing array properties that use a solid type, only those that are part of a union.

### Version

1.2.0

In our schema we have a union type called UILine, defined as the following in our generated code:
```
public extension Unions {
  static let UILine = Union(
    name: "UILine",
    possibleTypes: [
      Objects.UITitleLine.self,
      Objects.UITextLine.self
    ]
  )
}
```

Multiple types utilise this union type, often as an array. An example of this is a UICard type.  
I don't know if this matters, but the UICard type is also part of a Union.

For the UICard type the codegen step creates this Mock object (I've removed all other properties for brevity):
```
public class UICard: MockObject {
  public static let objectType: Object = MenuGQL.Objects.UICard
  public typealias MockValueCollectionType = Array<Mock<UICard>>

  public struct MockFields {
    @Field<[UILine]>("ui_lines") public var ui_lines
  }
}

public extension Mock where O == UICard {
  convenience init(
    ui_lines: [AnyMock]? = nil
  ) {
    self.init()
    _set(ui_lines, for: \.ui_lines)
  }
}
```

Also generated are these Mock objects:
```
public extension MockObject {
  typealias UILine = Union
}

public class UITitleLine: MockObject {
  public static let objectType: Object = MenuGQL.Objects.UITitleLine
  public typealias MockValueCollectionType = Array<Mock<UITitleLine>>

  public struct MockFields {
    @Field<String>("text") public var text
  }
}

public extension Mock where O == UITitleLine {
  convenience init(
    text: String? = nil
  ) {
    self.init()
    _set(text, for: \.text)
  }
}
```

These are used to attempt to create a Mock and then instantiate our `UiBlockFields` fragment.
```
let uiCard = Mock<UICard>.init(
  ui_lines: [Mock<UITitleLine>.init(text: "text")]
)
let block = Blocks.CardBlock(uiCard: UiBlockFields.from(uiCard).asUICard!)
```

This leads to a crash on this final line, with the actual error occurring here:
Screenshot 2023-05-17 at 7 12 03 pm

Something I did find when attempting to debug this was that in the Mock init, when the `_set(_:,for:)` function is called, the `(value as? AnyHashable)` always resolves to nil. 
```
public func _set<T: MockFieldValue>(
  _ value: T.MockValueCollectionType.Element?,
  for keyPath: KeyPath<O.MockFields, Field<T>>
) {
  let field = O._mockFields[keyPath: keyPath]
  _data[field.key.description] = (value as? AnyHashable)
}
```

### Logs

```shell
#0	0x000000018bfa78e0 in _swift_runtime_on_report ()
#1	0x000000018c034580 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x000000018bce32a8 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x000000018bce216c in _assertionFailure(_:_:file:line:flags:) ()
#4	0x000000015883d228 in Array<τ_0_0>.init(_fieldData:) at /SourcePackages/checkouts/apollo-ios/Sources/ApolloAPI/DataDict.swift:181
#5	0x000000015883d77c in protocol witness for SelectionSetEntityValue.init(_fieldData:) in conformance <τ_0_0> [τ_0_0] ()
#6	0x000000015883a878 in DataDict.subscript.getter at /SourcePackages/checkouts/apollo-ios/Sources/ApolloAPI/DataDict.swift:71
#7	0x0000000158f27eec in UiBlockFields.AsUICard.ui_lines.getter at /Sources/MenuGQL/Fragments/UiBlockFields.graphql.swift:2084
#8	0x0000000159383cd0 in Blocks.CardBlock.init(uiCard:) at /Sources/Menu/GraphQL/Adapters/BlockAdapter.swift:175
#9	0x0000000158124860 in BlockAdapterTests.testCardCard() at Tests/MenuTests/GraphQL/Adapters/BlockAdapterTests.swift:267
```

### Anything else?

_No response_

### LINKED ISSUE #3043: In Version 1.2, arrays of Union Types don't get set properly in mocks
### Summary

In version 1.1, we had code like
```
let itemOne = Mock<TypeOne>(propertyOne: "Hello World")
let itemTwo = Mock<TypeTwo>(propertyTwo: "Goodbye World")
let mock = Mock<OuterItemType>(
     description: "This is my outer type",
     items: [ itemOne, itemTwo]
)
```
Note that in this example, itemOne and itemTwo are two different types that are of the same Union type. So for example, our mock would have it defined as as something like

`@Field<[MyUnionType]>("items") public var items` but the initializer would take
`items: [AnyMock]? = nil,`

In 1.1.1 this worked fine. However in `1.2` after initializing the mock with `items`, the `items` property is still nil. And it is not set in the internal dictionary of the object. i.e if we print `itemCollection._data` it will only show `__typename` and `description`.  I'm not 100% sure if the union type is relevant, but i believe it is. Because ultimately it is calling

```
  public func _set<T: MockFieldValue>(
    _ value: T.MockValueCollectionType.Element?,
    for keyPath: KeyPath<O.MockFields, Field<T>>
  ) {
    let field = O._mockFields[keyPath: keyPath]
    _data[field.key.description] = (value as? AnyHashable)
  }
```

but in this case `value as? AnyHashable`  is nil. ( `value as? [AnyHashable]` however would have worked fine ). But i do see there are `_set` functions here for arrays, but i don't think they work because the typing is based on generics whereas we're passing `[AnyMock]`. 

### Version

1.2.0

1. Create a schema that has a type with a property that is a nested union type of at least two different types. 
2. Generate the code for these types. 
3. Attempt to initialize a mock for the other type with an array of two different union types. 
4. Print the `items` property, note it is nil even though it was passed to the initializer. 

### Logs

_No response_

### Anything else?

_No response_

BobaFetters:
-Updated the definitions of the test mock _set method definitions 
-Updated MockObjectTemplate to generate mocks with new definitions

Closes #3023
Closes #3043


Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -21,16 +21,19 @@ struct MockObjectTemplate: TemplateRenderer {
 
   var template: TemplateString {
     let objectName = graphqlObject.formattedName
-    let fields: [TemplateField] = ir.fieldCollector
+    let fields: [(TemplateField, GraphQLType)] = ir.fieldCollector
```
BobaFetters, line 24: Turned this into an array of tuples to keep the `GraphQLType` together with the `TemplateField` to check the type to determine the appropriate `_set` function to call.

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -56,21 +59,38 @@ struct MockObjectTemplate: TemplateRenderer {
...
+          }, separator: "\n")
         }
       }
       """) : TemplateString(stringLiteral: "")
     )
     
     """
   }
+  
+  private func getMockSetType(_ graphQLType: GraphQLType) -> String {
```
BobaFetters, line 80: Not sure I like the name of this function, open to other suggestions here.

calvincestari: Maybe just remove the `get` - `mockSetType` or something completely different in `mockFunctionDescriptor`.

calvincestari (commented):
I think this is the right approach. Didn't look like we were going to get it right through typing alone.

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
AnthonyMDev: Since `TemplateField` is already a tuple should we just change the `type` field of that tuple to the `GraphQLType` and do the formatting of the `rendered(as: .testMockField)` in the rendering on like 52?

Sources/AnimalKingdomAPI/AnimalKingdomAPI/TestMocks/Bird+Mock.graphql.swift
```diff
@@ -47,7 +47,7 @@ public extension Mock where O == Bird {
     _setScalar(laysEggs, for: \.laysEggs)
     _setEntity(owner, for: \.owner)
     _setList(predators, for: \.predators)
-    _setEntity(skinCovering, for: \.skinCovering)
+    _setScalar(skinCovering, for: \.skinCovering)
```
calvincestari, line 50: What's causing these to change? `GraphQLEnum` is simply a custom scalar but how did we end up with `_setEntity` to begin with?

calvincestari (approved):
Looks good, just one question on the changes to `AnimalKingdomAPI`.

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -56,21 +59,38 @@ struct MockObjectTemplate: TemplateRenderer {
...
+  private func mockFunctionDescriptor(_ graphQLType: GraphQLType) -> String {
+    switch graphQLType {
+      case .list(_):
+        return "List"
+      case .scalar(_), .enum(_):
+        return "Scalar"
+      case .entity(_):
+        return "Entity"
+      case .inputObject(_):
+        return ""
```
AnthonyMDev, line 89: This should probably be a `preconditionFailure` as if we return an empty string here, you're just going to get code that doesn't compile. Better to fail during generation with an error message that people can see and put in a bug report for us.

Sources/AnimalKingdomAPI/AnimalKingdomAPI/TestMocks/Bird+Mock.graphql.swift
BobaFetters: I originally had it mistakenly grouped with `entity` instead of `scalar` in the switch statement when determining the function descriptor, I believe I fixed that before the initial PR creation, but hadn't regenerated the code for AnimalKingdomAPI

AnthonyMDev (approved):
Two minor suggestions, but this loos great! Thanks so much @BobaFetters

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -63,14 +63,31 @@ struct MockObjectTemplate: TemplateRenderer {
...
+  private func mockFunctionDescriptor(_ graphQLType: GraphQLType) -> String {
+    switch graphQLType {
+      case .list(_):
+        return "List"
+      case .scalar(_), .enum(_):
+        return "Scalar"
+      case .entity(_):
+        return "Entity"
+      case .inputObject(_):
+        preconditionFailure()
```
AnthonyMDev, line 86: Nit: might want to also add a message to this failure to help us identify the issue if a user ever encounters this and files a bug report.

AnthonyMDev (approved):

PR #3093, opened 2023-06-27
RFC: GraphQL `@defer`

calvincestari:
Read the formatted RFC [here](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md).

```[tasklist]
### Codegen
- [ ] https://github.com/apollographql/apollo-ios/issues/3113
- [ ] https://github.com/apollographql/apollo-ios/issues/3114
- [ ] https://github.com/apollographql/apollo-ios/issues/3139
- [ ] https://github.com/apollographql/apollo-ios/issues/3140
- [ ] https://github.com/apollographql/apollo-ios/issues/3141
- [ ] https://github.com/apollographql/apollo-ios/issues/3142
```

```[tasklist]
### Execution
- [ ] https://github.com/apollographql/apollo-ios/issues/3144
- [ ] https://github.com/apollographql/apollo-ios/issues/3145
```

```[tasklist]
### Networking
- [ ] https://github.com/apollographql/apollo-ios/issues/3118
- [ ] https://github.com/apollographql/apollo-ios/issues/3143
- [ ] https://github.com/apollographql/apollo-ios/issues/3146
- [ ] https://github.com/apollographql/apollo-ios/issues/3147
- [ ] https://github.com/apollographql/apollo-ios/issues/3148
```

```[tasklist]
### Caching
- [ ] https://github.com/apollographql/apollo-ios/issues/3149
```

### DIFF
--- a/Design/3093-graphql-defer.md
+++ b/Design/3093-graphql-defer.md
@@ -14,12 +14,14 @@ Based on the progress of `@defer`/`@stream` through the approval process there m
 
 Apollo iOS uses [graphql-js](https://github.com/graphql/graphql-js) for validation of the GraphQL schema and operation documents as the first step in the code generation workflow. The version of this [dependency](https://github.com/apollographql/apollo-ios/blob/spike/defer/Sources/ApolloCodegenLib/Frontend/JavaScript/package.json#L16) is fixed at [`16.3.0-canary.pr.3510.5099f4491dc2a35a3e4a0270a55e2a228c15f13b`](https://www.npmjs.com/package/graphql/v/16.3.0-canary.pr.3510.5099f4491dc2a35a3e4a0270a55e2a228c15f13b?activeTab=versions). This is a version of graphql-js that supports the experimental [Client Controlled Nullability](https://github.com/graphql/graphql-wg/blob/main/rfcs/ClientControlledNullability.md) feature but does not support the `@defer` directive.
 
-The latest `16.x` release of graphql-js with support for the `@defer` directive is [`16.1.0-experimental-stream-defer.6`](https://www.npmjs.com/package/graphql/v/16.1.0-experimental-stream-defer.6) but it looks like the 'experimental' named releases for `@defer` have been discontinued and the recommendation is to use [`17.0.0-alpha.2`](https://www.npmjs.com/package/graphql/v/17.0.0-alpha.2). This is further validated by the fact that [`16.7.0` does not](https://github.com/graphql/graphql-js/blob/v16.7.0/src/type/directives.ts#L167) include the @defer directive whereas [`17.0.0-alpha.2` does](https://github.com/graphql/graphql-js/blob/v17.0.0-alpha.2/src/type/directives.ts#L159).
+The latest `16.x` release of graphql-js with support for the `@defer` directive is [`16.1.0-experimental-stream-defer.6`](https://www.npmjs.com/package/graphql/v/16.1.0-experimental-stream-defer.6) but it looks like the 'experimental' named releases for `@defer` have been discontinued and the recommendation is to use [`17.0.0-alpha.2`](https://www.npmjs.com/package/graphql/v/17.0.0-alpha.2). This is further validated by the fact that [`16.7.0` does not](https://github.com/graphql/graphql-js/blob/v16.7.0/src/type/directives.ts#L167) include the `@defer` directive whereas [`17.0.0-alpha.2` does](https://github.com/graphql/graphql-js/blob/v17.0.0-alpha.2/src/type/directives.ts#L159).
 
 Potential solutions:
 1. Add support for Client Controlled Nullability to `17.0.0-alpha.2`, or the latest 17.0.0 alpha release, and publish that to NPM. The level of effort for this is unknown but it would allow us to maintain support for CCN.
 2. Use `17.0.0-alpha.2`, or the latest 17.0.0 alpha release, as-is and remove the experimental Client Controlled Nullability feature. We do not know how many users rely on the CCN functionality so this may be a controversial decision. This path doesn’t necessarily imply an easier dependency update because there will be changes needed to our frontend javascript to adapt to the changes in graphql-js.
 3. Another option is a staggered approach where we adopt `17.0.0-alpha.2`, or the latest 17.0.0 alpha release, limiting the changes to our frontend javascript only and at a later stage bring the CCN changes from [PR `#3510`](https://github.com/graphql/graphql-js/pull/3510) to the `17.x` release path and reintroduce support for CCN to Apollo iOS. This would also require the experiemental CCN feature to be removed, with no committment to when it would be reintroduced.
+
+**Decision:** Option 3 is the preferred solution. The work to port the CCN PRs to `17.0.0-alpha.2` is being done externally as part of the renewed interest in the CCN proposal.
 
 ## Rename `PossiblyDeferred` types/functions
 
@@ -86,6 +88,8 @@ client.fetch(query: ExampleQuery()) { result in
 }
 
 ```
+
+**Decision:** Option 4 is the preferred solution.
 
 Regardless of the fragment/field solution chosen all deferred fragment definitions in generated models `__selections` will get an additional property to indicate they are deferred. This helps to understand the models when reading them as well as being used by internal code.
 
@@ -247,7 +251,7 @@ struct DeferResponseParser: MultipartResponseSpecificationParser {
 
 The initial response data and data received in each incremental response will need to be retained and combined so that each incremental response can insert the latest received incremental response data at the correct path and return an up-to-date response to the request callback.
 
-The data being retained and combined should not require another pass through the GraphQL executor though.
+The data being retained and combined will be passed through the GraphQL executor on each response, initial and incremental.
 
 ### Completion handler
 
@@ -314,11 +318,13 @@ client.fetch(query: ExampleQuery()) { result in
 }
 ```
 
+**Decision:** Option 1 is the preferred solution.
+
 ## GraphQL execution
 
-The executor currently executes on an entire operation selection set. It will need to be adapted to be able to execute on a partial response when deferred fragments have not been received and on an isolated fragment selection set so that incremental responses can be parsed alone instead of needing to execute on the whole operation’s selection set.
+The executor currently executes on an entire operation selection set. It will need to be adapted to be able to execute on a partial response when deferred fragments have not been received, this may Just Work™ once the generated models have been updated for optional fragments.
 
-An alternative to parsing isolated fragment selection sets would be to execute on all the data currently received. The inefficiency with this is executing on data that would have already been executed on during prior responses.
+Each response will be passed to the GraphQL executor.
 
 ## Caching
 
--- a/Design/3093-graphql-defer.md
+++ b/Design/3093-graphql-defer.md
@@ -77,18 +77,6 @@ _Example usage in a generated model_
   ] }
 ```
 
-### Deferred fragments
-
-_In progress_
-
-### Merged fields
-
-_In progress_
-
-### Selection set initializers
-
-_In progress_
-
 ## Networking 
 
 ### Request header
--- a/Design/3093-graphql-defer.md
+++ b/Design/3093-graphql-defer.md
@@ -77,6 +77,18 @@ _Example usage in a generated model_
   ] }
 ```
 
+### Deferred fragments
+
+_In progress_
+
+### Merged fields
+
+_In progress_
+
+### Selection set initializers
+
+_In progress_
+
 ## Networking 
 
 ### Request header
--- a/Design/3093-graphql-defer.md
+++ b/Design/3093-graphql-defer.md
@@ -33,10 +33,11 @@ Generated models will definitely be affected by `@defer` statements in the opera
 
 Potential fragment/field solutions:
 1. Property wrappers - I explored Swift's property wrappers but they suffer from the limitation of not being able to be applied to a computed property. All GraphQL fields in the generated models are computed properties because they simply route access to the value in the underlying data dictionary storage. It would be nice to be able to simply annotate fragments and fields with something like `@Deferred` but unfortunately that is not possible.
-4. `Enum` wrapper - an idea that was suggested by [`@Iron-Ham`](https://github.com/apollographql/apollo-ios/issues/2395#issuecomment-1433628466) is to wrap the type in a Swift enum that can expose the deferred state as well as the underlying value once it has been received.
+2. Optional types - this solution would change the deferred property type to an optional version of that type. This may not seem necessary when considering that only fragments can be marked as deferred but it would be required to cater for the way that Apollo iOS does field merging in the generated model fragments. Field merging is non-optional at the moment but there is an issue ([#2560](https://github.com/apollographql/apollo-ios/issues/2560)) that would make this a configuration option. This solution hides detail though because you wouldn't be able to tell whether the field value is `nil` because the response data hasn't been received yet (i.e.: deferred) or whether the data was returned and it was explicitly `null`. It also gets more complicated when a field type is already optional; would that result in a Swift double-optional type? As we learnt with the legacy implementation of GraphQL nullability, double-optionals are difficult to interpret and easily lead to mistakes.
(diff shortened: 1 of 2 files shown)


calvincestari:
This is ready for review and comment. The section on the generated models is the weakest of the RFC right now, I will put more time into that when I get back. If you have comments/ideas for that though please comment on it and I will address it.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,326 @@
...
+```swift
+  public static var __selections: [ApolloAPI.Selection] { [
+    .inlineFragment(AsEntity.self, deferred: true),
+  ] }
+
+  public static var __selections: [ApolloAPI.Selection] { [
+    .fragment(EntityFragment.self, deferred: true),
+  ] }
+```
+
```
BoD, line 79: Just for a bit of context, here's how it works on Apollo Kotlin:

- fragments spread/inline fragments generate a dedicate field in the generated model
- before `@defer`, this field could be nullable if:
  - its type condition means the fields won't always be returned
  - or it has a `@skip` or `@include` (unless it's trivial)
- after `@defer`: now the field is always nullable. If it's `null` it means you haven't received this fragment yet.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,326 @@
...
+The most simple solution is to change the deferred property type to an optional version of that type. This hides detail though because you wouldn't be able to tell whether the value is `nil` because the response data has been received yet (i.e.: deferred) or whether the data was returned and it was explicitly `null`. It also gets more complicated when a type is already optional; would that result in a Swift double-optional type? As we learnt with the legacy implementation of GraphQL nullability, double-optionals are difficult to interpret and easy lead to errors.
+
+I explored Swift's property wrappers but they suffer from the limitation of not being able to be applied to a computed property. All model properties are computed properties because they simply route access the value in the underlying dictionary data storage. It would be nice to be able to simply annotate fragments and fields with some like `@Deferred` but it doesn't look like that is possible.
```
Iron-Ham, line 36: Not pressing: Investigate macros for this purpose?

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,326 @@
...
+      switch (data.response) {
+      case .complete:
+      case .partial:
+      }
+    case let .failure(error):
+    }
+  }
+```
+
+Another way which may be a bit more intuitive is to make the `server` case on `Source` have an associated value since `cache` sources will always be complete. The cache could return partial responses for deferred operations but for the initial implementation we will probably only write the cache record once all deferred fragments have been received.
```
Iron-Ham, line 254: > The cache could return partial responses for deferred operations but for the initial implementation we will probably only write the cache record once all deferred fragments have been received.

Just thinking on edge cases here. Let's imagine this scenario:
* User triggers a network request.
* One of the response fields is `@deferred` as it takes, let's say 5 seconds to resolve.
* The user navigates away from the view, ending all ongoing network requests for that view. 

In this case, there would be no cache write at all, despite the majority of data being retrieved (and valid!). This perhaps could lead to the unintended effect wherein a partial fetch that should have updated the values of various entity types across the cache don't.

Design/3093-graphql-defer.md
calvincestari: Macros might be a way to obscure the implementation in the generated code so the generated models wouldn't need to be regenerated if the underlying implementation changed but I don't think macros can provide a mechanism to expose the 'state' and values of deferred fragments?

calvincestari:
I've [updated](https://github.com/apollographql/apollo-ios/pull/3093/commits/323838cacc7d740283c39fac185297f62bfe56b5) the section on the generated models adding "Optional fragments" which aligns with @BoD's [comment](https://github.com/apollographql/apollo-ios/pull/3093#discussion_r1250992850) earlier.

calvincestari:
I'm going to be creating issues/tasks for the parts of the RFC that are less contentious and need to happen to be able to effectively support the `@defer` work.

calvincestari:
@AnthonyMDev / @Iron-Ham - do you have any feedback on the [completion handler](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#completion-handler) changes?

Working on the defer protocol parser today I realized there is the [`label` property](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#payload-format) which can be part of the deferred response and I wonder if we need to feed that back to the user too or just swallow it? Is there value in passing it back to the user through the completion data?

calvincestari:
> Working on the defer protocol parser today I realized there is the [label property](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#payload-format) which can be part of the deferred response and I wonder if we need to feed that back to the user too or just swallow it? Is there value in passing it back to the user through the completion data?

Thinking about this more today I don't believe there is any value in passing back the label to the user. Users work with models and models are typed which is a much stronger 'reference' than a label string. If anything it should maybe be a property on deferred fragment instances.

calvincestari:
@Iron-Ham, I've updated the [Generated Models](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#generated-models) section with an option (4) that would make fragments optional. I'd love to get your thoughts on it when you get a moment - thanks.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+4. Optional fragments (disabling field merging) - optional types are only needed when fragment fields are merged into entity selection sets. If field merging were disabled automatically for deferred fragments then the solution is simplified and we only need to alter the deferred fragments to be optional. Consuming the result data is intuitive too where a `nil` fragment value would indicate that the fragment data has not yet been received (i.e.: deferred) and when the complete response is received the fragment value is populated and the result sent to the client. This seems a more elegant and ergonimic way to indicate the status of deferred data but complicates the understanding of field merging.
+
+```swift
+// Sample usage in a generated model
+public class ExampleQuery: GraphQLQuery {
+  // other properties and types not shown
+
+  public struct Data: ExampleSchema.SelectionSet {
+    public static var __selections: [ApolloAPI.Selection] { [
+      .fragment(EntityFragment?.self)
```
AnthonyMDev, line 63: Below, you are showing something else. Should this be modified to:
```swift
.fragment(EntityFragment.self, deferred: true)
```

Even if we are making the fragment accessor on the model optional to indicate deferred fragments, the actual selection itself (and it's consumption in the execution layer) is going to be more clear if we specifically indicate `deferred`.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+
+// Sample model with a deferred property
+public struct ModelSelectionSet: GraphAPI.SelectionSet {
+  // other properties not shown
+
+  public var name: DeferredValue<String?> { __data["name"] }
+}
+```
+
+4. Optional fragments (disabling field merging) - optional types are only needed when fragment fields are merged into entity selection sets. If field merging were disabled automatically for deferred fragments then the solution is simplified and we only need to alter the deferred fragments to be optional. Consuming the result data is intuitive too where a `nil` fragment value would indicate that the fragment data has not yet been received (i.e.: deferred) and when the complete response is received the fragment value is populated and the result sent to the client. This seems a more elegant and ergonimic way to indicate the status of deferred data but complicates the understanding of field merging.
```
AnthonyMDev, line 54: I think this is probably the best approach we have. I do believe there is one specific situation where we already allow optional fragments. I think it had to do with having a conditionally included fragment with a `@include/skip` directive. We should double check that and make sure that we maintain consistency and clarity here.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+  }
+}
+
+// Sample usage in an app completion block
+client.fetch(query: ExampleQuery()) { result in
+  switch (result) {
+  case let .success(data):
+    client.fetch(query: ExampleQuery()) { result in
+      switch (result) {
+      case let .success(data):
```
AnthonyMDev, line 74: I do think we should add a property to the `GraphQLResult` to indicate if the response is complete/partial as well.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+      // call request chain error handler
+
+      return
+    }
+
+    let dataHandler: ((Data) -> Void) = { data in
+      // proceed ahead on the request chain
+    }
+
+    let errorHandler: (() -> Void) = {
```
AnthonyMDev, line 192: We'll just need to make sure that we have adequate testing around this to ensure these closures don't cause retain cycles. Easy to capture `self` here and cause memory leaks.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+    // new code to parse the defer specification
+  }
+}
+```
+
+#### Response data
+
+The initial response data and data received in each incremental response will need to be retained and combined so that each incremental response can insert the latest received incremental response data at the correct path and return an up-to-date response to the request callback.
+
+The data being retained and combined should not require another pass through the GraphQL executor though.
```
AnthonyMDev, line 250: Really? I would assume it would have to go back through the executor. How do you plan on merging it without execution?

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+// New Response type and property
+public struct GraphQLResult<Data: RootSelectionSet> {
+  // other properties and types not shown
+
+  public enum Response {
+    case partial
+    case complete
+  }
+
+  public let response: Response
```
AnthonyMDev, line 269: Oh, this is what I was referring to above! :) 

I don't like calling this `Response` though. It's not actually the response data, it's just indicating if the response is partial/complete. Think we just need some consideration on naming here.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+    switch (data.response) {
+    case .complete:
+    case .partial:
+    }
+  case let .failure(error):
+  }
+}
+```
+
+2. Another way which may be a bit more intuitive is to make the `server` case on `Source` have an associated value since `cache` sources will always be complete. The cache could return partial responses for deferred operations but for the initial implementation we will probably only write the cache record once all deferred fragments have been received. This solution becomes invalid though once the cache can return partial responses, with that in mind maybe option 1 is better.
```
AnthonyMDev, line 285: Interesting thought!  But I agree for the reason you stated that we should stick with option 1.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+  case let .failure(error):
+  }
+}
+```
+
+## GraphQL execution
+
+The executor currently executes on an entire operation selection set. It will need to be adapted to be able to execute on a partial response when deferred fragments have not been received and on an isolated fragment selection set so that incremental responses can be parsed alone instead of needing to execute on the whole operation’s selection set.
+
+An alternative to parsing isolated fragment selection sets would be to execute on all the data currently received. The inefficiency with this is executing on data that would have already been executed on during prior responses.
```
AnthonyMDev, line 321: Yeah, I'm skeptical that we're going to be able to get partial execution to work without a whole lot of new bugs in a reasonable amount of time. I'm sure it's possible, but I'm not sure if it's worth attempting for this 1st iteration on defer.

Design/3093-graphql-defer.md
calvincestari: Noted - thanks.

Design/3093-graphql-defer.md
calvincestari: Yes it would have the `deferred` property, these two code pieces were simply done at different times. I'll update for consistency.

calvincestari, line 63:
```suggestion
      .fragment(EntityFragment?.self, deferred: true)
```

Design/3093-graphql-defer.md
calvincestari: Noted

Design/3093-graphql-defer.md
calvincestari: This was mentioned purely from an efficiency point of view. If the implementation of doing that is overly complicated then yes each response will have to be executed on.

Design/3093-graphql-defer.md
```diff
@@ -0,0 +1,356 @@
...
+    switch (data.response) {
+    case .complete:
+    case .partial:
+    }
+  case let .failure(error):
+  }
+}
+```
+
+2. Another way which may be a bit more intuitive is to make the `server` case on `Source` have an associated value since `cache` sources will always be complete. The cache could return partial responses for deferred operations but for the initial implementation we will probably only write the cache record once all deferred fragments have been received. This solution becomes invalid though once the cache can return partial responses, with that in mind maybe option 1 is better.
```
BobaFetters, line 285: Agreed, ultimately would be great to be able to get partial responses from cache but for this implementation I think option 1 is fine.

Design/3093-graphql-defer.md
BobaFetters: I do like the idea of providing this data to the user through an enum. As for naming, yea just `Response` maybe isn't the best, thinking about what we are trying to represent which is the status of the result, maybe something like `ResultState`? `ResponseResult` doesn't really feel right, and `ResultData` would make me think it is data and not a state representation.

Design/3093-graphql-defer.md
Iron-Ham: +1 -- I think this is a brilliant approach.

Design/3093-graphql-defer.md
calvincestari: The naming we can settle on in Slack or the PR. It seems like we're aligned on the state being available at the root of `GraphQLResult` and not only on the `server` source case.

Design/3093-graphql-defer.md
AnthonyMDev: Yeah, I thought about this a lot. The problem is there really isn't any amount of generated code that could really make this work. We actually want to be able to have a "metadata" accessor on a property. The only way to do this is through the `projectedValue` of a property wrapper.

We could wrap the value in some struct that had the metadata and the value, but the 99% of the time you just wanted to value and not the "state" metadata, you'd need to call `myProperty.value` which is definitely not what we want here.

We just wanted to be able to use that sweet swift syntactic sugar that only exists with property wrappers, but they have another limitation that prevents us from using them right now.

Design/3093-graphql-defer.md
AnthonyMDev: Great call out @Iron-Ham. I think we're all aligned this is something we want to implement, just not for the MVP. And yes, we should clearly document this behavior.

AnthonyMDev:
> Thinking about this more today I don't believe there is any value in passing back the label to the user. Users work with models and models are typed which is a much stronger 'reference' than a label string. If anything it should maybe be a property on deferred fragment instances.

I'm not sure about this. There may be some valuable functionality that this could unlock, but I don't think it needs to be part of the MVP either way. We can see if we get any user feedback requesting us to expose this later.

calvincestari:
> > Thinking about this more today I don't believe there is any value in passing back the label to the user. Users work with models and models are typed which is a much stronger 'reference' than a label string. If anything it should maybe be a property on deferred fragment instances.
> 
> I'm not sure about this. There may be some valuable functionality that this could unlock, but I don't think it needs to be part of the MVP either way. We can see if we get any user feedback requesting us to expose this later.

I initially thought there might be value until I went back and read the RFCs description/purpose of the `label` property. To me it reads that it's helpful for data patching and not so much once you have a complete response.
> `label`: `String`
> - A unique label across all `@defer` and `@stream` directives in an operation.
> - This label should be used by GraphQL clients to identify the data from patch responses and associate it with the correct fragment.
> - If provided, the GraphQL Server must add it to the payload.

calvincestari:
I've updated the RFC with decisions on the preferred option where there were multiple and cleaned up some of the text based on review comments.

In summary:
1. [Update graphql-js dependency](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#update-graphql-js-dependency) - option 3 is the preferred solution.
2. [Generated models](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#generated-models) - option 4 is the preferred solution.
3. [Completion handler](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#completion-handler) - option 1 is the preferred solution.

Issue #3114, opened 2023-07-11
Maybe: Refactor CCN PR on top of graphql-js latest 17.x.x release


calvincestari:
This is step 2 in an implementation of Option 3 from ['Update graphql-js dependency'](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#update-graphql-js-dependency).

**It is undecided whether this will be done ahead of a release for @defer or afterwards.**

> Another option is a staggered approach where we adopt 17.0.0-alpha.2, or the latest 17.0.0 alpha release, limiting the changes to our frontend javascript only and at a later stage bring the CCN changes from https://github.com/graphql/graphql-js/pull/3510 to the 17.x release path and reintroduce support for CCN to Apollo iOS. This would also require the experiemental CCN feature to be removed, with no committment to when it would be reintroduced.

calvincestari:
I'm pushing to get this done by an external contributor as part of the renewed CCN work.

calvincestari:
Closing this issue as CCN has been removed in the defer work. Once the future of CCN is figured out we can make a decision on it but it can be separate from the defer work.

_For context: CCN was removed because the defer work required moving to graphql-js v17 which does not support CCN as the original changes to graphql-js was done in v16._

PR #3115, opened 2023-07-11
dependency: update graphql-js

### LINKED ISSUE #3113: Update JS frontend to use graphql-js (latest 17.x.x) without support for CCN
This is step 1 in an implementation of Option 3 from ['Update graphql-js dependency'](https://github.com/apollographql/apollo-ios/blob/spike/defer/Design/3093-graphql-defer.md#update-graphql-js-dependency). Next steps to handle the Apollo iOS configuration setting and/or reintroduce the CCN changes to graphql-js will be in subsequent PRs.

> Another option is a staggered approach where we adopt 17.0.0-alpha.2, or the latest 17.0.0 alpha release, limiting the changes to our frontend javascript only and at a later stage bring the CCN changes from https://github.com/graphql/graphql-js/pull/3510 to the 17.x release path and reintroduce support for CCN to Apollo iOS. This would also require the experiemental CCN feature to be removed, with no committment to when it would be reintroduced.

calvincestari:
Closes #3113 

* Updates the graphql-js dependency to `17.0.0-alpha.2` and updates our frontend to be compatible
* Effectively removes support for CCN because `17.0.0-alpha.2` does not support it
* Skips all CCN tests
* Added a warning to remind me to go back to the CCN configuration option depending on whether CCN is added to graphql-js 17.0.0

_This PR generates warnings that are dependent on the outcome of #3114._


AnthonyMDev (commented):
This looks good, but I don't want to merge it into `main` with the warning there if we can avoid it. Would like to try to do the work for #3114 and determine if we can get that done before merging.

Let's not spend too much time on #3114. If it's pretty easy, fine. But if you start running into complexities, just punt it for now.

calvincestari:
> This looks good, but I don't want to merge it into `main` with the warning there if we can avoid it. Would like to try to do the work for #3114 and determine if we can get that done before merging.

No it won't go into `main`, this PR will get merged into the `feature/defer` branch. It's just there as a reminder to me to either clean it up or do the work before finalizing the feature.

> Let's not spend too much time on #3114. If it's pretty easy, fine. But if you start running into complexities, just punt it for now.

Exactly. I've prefixed the name with "Maybe:" as an indication that it's left for the end of delivery and only if we determine it's beneficial to the release. I won't be starting on it until most of the other work is done and timing looks good.

PR #3119, opened 2023-07-13
refactor: multipart response parsers

### LINKED ISSUE #3118: Support multiple response parsers
This is the work described in the [Multipart parsing protocol](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#multipart-parsing-protocol) section of the `@defer` RFC.

Refactor `MultipartResponseParsingInterceptor` to be able to support multiple response parsers based on the response specification.

calvincestari:
Closes #3118

* Introduced a new protocol for multipart response parsers
* Refactored the subscription parsing logic into `MultipartResponseSubscriptionParser`
* Refactored errors for the context of where they are raised from


BobaFetters (approved):
I think this looks good

calvincestari:
Thanks for the review @BobaFetters but I think I'm going to put this PR into draft because I'm not happy with the way each multipart parser needs to implement errors and a full parser. I think those should be centralized and each spec parser should only be unique in the actual data format.

calvincestari:
I've changed my mind and am going to merge this in because it sets up the framework for what I want to achieve. There is a note in https://github.com/apollographql/apollo-ios/issues/3146 with guidance on the desired improvement.

PR #3123, opened 2023-07-15
Issue 2426 improve response code error api

dfperry5:
Addressing Issue 2426:

- Adding in a computed property on ResponseCodeError for GraphQL Error. It converts the data into a JSONObject and uses that to create a new GraphQL Error. If there is not a valid JSON response, graphQLError computed property return nil.


dfperry5:
@AnthonyMDev

AnthonyMDev (approved):
Thanks so much for the contribution! There just seems to be some odd whitespace that got committed in your unit tests. Would you mine cleaning up the added whitespace before we merge this?

dfperry5:
Apologies for my delay! Looks like @BobaFetters  knocked it out (Thanks a ton!) :) 

Happy to contribute in other ways too!

Issue #2942, opened 2023-04-07
Generate error when codegen configuration keys are misplaced


calvincestari:
_Context in https://github.com/apollographql/apollo-ios/issues/2940#issuecomment-1499768938_

PR #3125, opened 2023-07-17
Throw error when an invalid key is present in a `JSON` Apollo configuration

Iron-Ham:
### What are you trying to accomplish?

closes https://github.com/apollographql/apollo-ios/issues/2942

Notify users when they create an invalid `apollo-codegen-config.json`. 

### What approach did you choose and why?

I created an `AnyCodingKey` struct which cannot fail. I create sets of all the keys present in the input, and then make sure that the input keys are a subset of the `CodingKeys` defined for that object.

### Anything you want to highlight for special attention from reviewers?

Is the error text here sufficient to explain to users that something is wrong?

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -202,7 +202,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
 
     // MARK: Codable
 
-    enum CodingKeys: CodingKey {
+    enum CodingKeys: CodingKey, CaseIterable {
       case schemaTypes
       case operations
       case testMocks
@@ -214,6 +214,15 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     /// specified defaults when not present.
     public init(from decoder: Decoder) throws {
       let values = try decoder.container(keyedBy: CodingKeys.self)
+
+      let allKeysContainer = try decoder.container(keyedBy: AnyCodingKey.self)
+      guard Set(allKeysContainer.allKeys.map(\.stringValue)).isSubset(of: Set(CodingKeys.allCases.map(\.stringValue))) else {
+        throw DecodingError.typeMismatch(Self.self, DecodingError.Context.init(
+          codingPath: values.codingPath,
+          debugDescription: "Unrecognized key found",
+          underlyingError: nil
+        ))
+      }
 
       schemaTypes = try values.decode(
         SchemaTypesFileOutput.self,
@@ -617,7 +626,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
 
     // MARK: Codable
 
-    enum CodingKeys: CodingKey {
+    enum CodingKeys: CodingKey, CaseIterable {
       case additionalInflectionRules
       case queryStringLiteralFormat
       case deprecatedEnumCases
@@ -633,6 +642,15 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
 
     public init(from decoder: Decoder) throws {
       let values = try decoder.container(keyedBy: CodingKeys.self)
+
+      let allKeysContainer = try decoder.container(keyedBy: AnyCodingKey.self)
+      guard Set(allKeysContainer.allKeys.map(\.stringValue)).isSubset(of: Set(CodingKeys.allCases.map(\.stringValue))) else {
+        throw DecodingError.typeMismatch(Self.self, DecodingError.Context.init(
+          codingPath: values.codingPath,
+          debugDescription: "Unrecognized key found",
+          underlyingError: nil
+        ))
+      }
 
       additionalInflectionRules = try values.decodeIfPresent(
         [InflectionRule].self,
@@ -757,6 +775,13 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
 
     public init(from decoder: Decoder) throws {
       let values = try decoder.container(keyedBy: CodingKeys.self)
+      guard values.allKeys.first != nil else {
+        throw DecodingError.typeMismatch(Self.self, DecodingError.Context.init(
+          codingPath: values.codingPath,
+          debugDescription: "Invalid number of keys found, expected one.",
+          underlyingError: nil
+        ))
+      }
 
       enumCases = try values.decodeIfPresent(
         CaseConversionStrategy.self,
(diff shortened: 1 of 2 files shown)


calvincestari:
Thanks @Iron-Ham! 

It looks like the key validation has been implemented in:
* `FileOutput`
* `OutputOptions`
* `ExperimentalFeatures`
* base (`ApolloCodegenConfiguration`)

It's not implemented in these structs though:
* `FileInput`
* `SchemaTypesFileOutput`
* `ConversionStrategies`
* `OperationDocumentFormat`
* `SelectionSetInitializers`

We should decide if we're going to enforce a 100% key-correct configuration file or only perform key validation on the most common places we've noticed incorrect configurations.
* If we're going to enforce 100% key-correctness then some of the structs will need to get `decode` implementations because at the moment we've only done that where custom behaviour was needed.
* If we're only going to validate at the most error-prone structs then we can likely remove validation of `ExperimentalFeatures`.

I'm not sure it's worth the effort to go the 100% key-correctness route and I think I'd instead choose to just validate the most error-prone. Thoughts?

Iron-Ham:
Agreed that we should really just tackle the most error prone ones. I'll remove experimental features and adjust the error String.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -1370,3 +1381,36 @@ extension ApolloCodegenConfiguration.OutputOptions {
...
+private struct AnyCodingKey: CodingKey {
+    var stringValue: String
+
+    init?(stringValue: String) {
+        self.stringValue = stringValue
+    }
+
+    var intValue: Int?
+
+    init?(intValue: Int) {
```
AnthonyMDev, line 1394: The `intValue` stuff here is throwing me off. This seems like it can't actually be used properly for validating  the coding keys. Am I missing the reason for this?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -1370,3 +1381,36 @@ extension ApolloCodegenConfiguration.OutputOptions {
...
+  decoder: Decoder
+) throws {
+  // Map all keys from the input object
+  let allKeys = Set(try decoder.container(keyedBy: AnyCodingKey.self).allKeys.map(\.stringValue))
+  // Map all valid keys from the given `CodingKey` enum
+  let validKeys = Set(C.allCases.map(\.stringValue))
+  guard allKeys.isSubset(of: validKeys) else {
+    throw DecodingError.typeMismatch(type, DecodingError.Context.init(
+      codingPath: container.codingPath,
+      debugDescription: "Unrecognized key found",
```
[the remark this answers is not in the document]
AnthonyMDev: I'm wondering if this could be put into a protocol that automatically calls this function so we don't have to remember to call it in each of the decoder functions? We could call it `StrictlyKeyedDecodable`.

I'm not sure if that would cause more complexity with the `CodingKeys`?

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
```diff
@@ -923,4 +923,49 @@ class ApolloCodegenConfigurationCodableTests: XCTestCase {
       ))
     )
   }
+
+  func test__decodeApolloCodegenConfiguration__withInvalidOptions() throws {
```
AnthonyMDev, line 927: To prevent regressions and ensure that we are actually calling the validation function in each of the decoder methods (or if we move to a protocol, that all of the proper objects conform to and use the `StrictlyKeyedDecodable` protocol), we should probably have multiple tests here, one where there is an invalid key in each of the objects.

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
```diff
@@ -923,4 +923,49 @@ class ApolloCodegenConfigurationCodableTests: XCTestCase {
...
+        }
+      }
+    }
+    """.asData
+
+    func decodeConfiguration(subject: Data) throws -> ApolloCodegenConfiguration {
+      try JSONDecoder().decode(ApolloCodegenConfiguration.self, from: subject)
+    }
+    XCTAssertThrowsError(try decodeConfiguration(subject: subject)) { error in
+      XCTAssert(error is DecodingError)
```
AnthonyMDev, line 968: Can we be more specific on testing that the error is the correct error?
```swift
guard case let DecodingError.typeMismatch(type, context) = error else {
  fail("Incorrect error type")
}

XCTAssert(type is $ExpectedType)
XCTAssertEqual(context.codingPath, $expectedPath)
```

AnthonyMDev (approved):
Thanks so much for putting this together @Iron-Ham . I think this could be improved by turning it into a protocol, and adding additional test cases, but this is great!

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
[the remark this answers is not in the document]
Iron-Ham: > I'm wondering if this could be put into a protocol that automatically calls this function so we don't have to remember to call it in each of the decoder functions?

That was my first thought, but I shied away from it when I realized that `Codable` (`Encodable` + `Decodable`) doesn't have any reference to `CodingKeys`. Adds a lot of complexity for a function that we'd otherwise call a handful of times in one file.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
Iron-Ham: Truthfully, I borrowed the `AnyCodingKey` from elsewhere; and it's implemented for int-keyed dictionaries. However, that doesn't really apply to decoding from JSON, so I'll remove it.

Iron-Ham: It's a requirement of the protocol
Screenshot 2023-07-18 at 8 44 43 PM

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
Iron-Ham: Added more tests -- and an additioanl one to handle mutliple failures in a single object.

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
Iron-Ham: Addressed in [02f4072](https://github.com/apollographql/apollo-ios/pull/3125/commits/02f40727da07e760569353fb1d66142bda2d0e40)

calvincestari (approved):
LGTM - @AnthonyMDev, thoughts on the protocol route?

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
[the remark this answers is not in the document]
AnthonyMDev: Yupp, that makes sense. If we were using this all over might be worth it, but for this one place, totally fine.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
AnthonyMDev: Ahhh odd. Okay, thanks for explaining :)

AnthonyMDev (approved):

PR #3128, opened 2023-07-19
Operation Manifest Generation Updates

BobaFetters:
- Updated the Operation Manifest generation to be able to do either an absolute path as provided (which is the current functionality) or to take relative paths that begin with './' and have the pathing be relative to the config rootURL 
-Updated tests to reflect these changes

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
@@ -57,8 +57,22 @@ struct OperationManifestFileGenerator {
   func generate(fileManager: ApolloFileManager = .default) throws {
     let rendered: String = try template.render(operations: operationManifest)
 
+    var manifestPath = config.output.operationManifest.unsafelyUnwrapped.path
+    let relativePrefix = "./"
+      
+    // if path begins with './' the path should be relative to the config.rootURL
+    if manifestPath.hasPrefix(relativePrefix) {
+      let fileURL = URL(fileURLWithPath: String(manifestPath.dropFirst(relativePrefix.count)), relativeTo: config.rootURL)
+      manifestPath = fileURL
+          .resolvingSymlinksInPath()
+          .appendingPathExtension("json")
+          .path
+    } else {
+      manifestPath.append(".json")
+    }
+      
     try fileManager.createFile(
-      atPath: config.output.operationManifest.unsafelyUnwrapped.path,
+      atPath: manifestPath,
       data: rendered.data(using: .utf8),
       overwrite: true
     )
--- a/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationManifestFileGeneratorTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationManifestFileGeneratorTests.swift
@@ -71,7 +71,7 @@ class OperationManifestFileGeneratorTests: XCTestCase {
 
   // MARK: Generate Tests
 
-  func test__generate__givenOperation_shouldWriteToPath() throws {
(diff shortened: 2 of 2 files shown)


Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
```diff
@@ -57,8 +57,22 @@ struct OperationManifestFileGenerator {
...
+      
+    // if path begins with './' the path should be relative to the config.rootURL
+    if manifestPath.hasPrefix(relativePrefix) {
+      let fileURL = URL(fileURLWithPath: String(manifestPath.dropFirst(relativePrefix.count)), relativeTo: config.rootURL)
+      manifestPath = fileURL
+          .resolvingSymlinksInPath()
+          .appendingPathExtension("json")
+          .path
+    } else {
+      manifestPath.append(".json")
```
calvincestari, line 71: Do we care whether they've already added `.json` to their `path` value? I suppose it's not going to cause this to stop working, it'll just result in a file ending in `.json.json`.

calvincestari (approved):
Looks good, just one question on whether we care to check for an existing file extension.

Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
BobaFetters: Good point, I will add a check for an existing suffix of `.json` to not duplicate it, won't cover other potential suffixes but we also can't predicate what someone may be using.

Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
```diff
@@ -57,8 +57,22 @@ struct OperationManifestFileGenerator {
...
 
+    var manifestPath = config.output.operationManifest.unsafelyUnwrapped.path
+    let relativePrefix = "./"
+      
+    // if path begins with './' the path should be relative to the config.rootURL
+    if manifestPath.hasPrefix(relativePrefix) {
+      let fileURL = URL(fileURLWithPath: String(manifestPath.dropFirst(relativePrefix.count)), relativeTo: config.rootURL)
+      manifestPath = fileURL
+          .resolvingSymlinksInPath()
+          .appendingPathExtension("json")
```
AnthonyMDev, line 68: Are we forgetting to check for the existing path extension here? Glad you added it below due to Calvin's comment, but it seems like in the case of a relative path, we aren't checking for the `.json` already existing?

Unless `Swift.URL.appendingPathExtension()` already handles checking for an existing path extension for you?

Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationManifestFileGeneratorTests.swift
```diff
@@ -71,7 +71,7 @@ class OperationManifestFileGeneratorTests: XCTestCase {
```
AnthonyMDev, line 71: I think we should also add tests for to ensure that we never get a doubled `.json.json` extension for both relative and absolute paths.

AnthonyMDev (approved):
Awesome! Just want to make sure we add tests for the `.json.json` issue and this is good to go!

Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
BobaFetters: I thought the `appendingPathExtension` would check for that, but adding the tests for this can verify that for sure.

BobaFetters: Looks like it didn't handle that so thats fixed and tests added

AnthonyMDev (approved):
Awesome! Thank you LGTM!

PR #3132, opened 2023-07-20
fix: Update graphql-js error handling

### LINKED ISSUE #3126: Update graphql-js error handling in Swift
By default graphql-js has a max error limit of 100, when this is surpassed it will return 101 errors with the last one being an "abort" error that you have exceeded the error limit. The Swift code is expecting all errors that return to have source location information and when the "abort" error comes through it causes a precondition failure and the user does not get the useful error logs they should.

To resolve this we need to update the error handling in Swift to properly handle this "abort" error to ensure the user is always receiving the proper error logs to aid in debugging errors.

BobaFetters:
-Updating the handling of graphql-js errors to check if source locations exist to determine how to process and display the error 
-This fixes the issue of reaching the max error limit of graphql-js

Closes #3126

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -309,7 +309,7 @@ public class ApolloCodegen {
         if let logLines = $0.logLines {
           return logLines
         } else {
-          return ["[Error] \($0.name ?? "unknown"): \($0.message ?? "")"]
+          return ["\($0.name ?? "unknown"): \($0.message ?? "")"]
         }
       })
       CodegenLogger.log(errorlines.joined(separator: "\n"), logLevel: .error)
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -305,8 +305,14 @@ public class ApolloCodegen {
     )
 
     guard graphqlErrors.isEmpty else {
-      let errorlines = graphqlErrors.flatMap({ $0.logLines })
-      CodegenLogger.log(String(describing: errorlines), logLevel: .error)
+      let errorlines = graphqlErrors.flatMap({
+        if let logLines = $0.logLines {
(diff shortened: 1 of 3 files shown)


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -305,8 +305,14 @@ public class ApolloCodegen {
     )
 
     guard graphqlErrors.isEmpty else {
-      let errorlines = graphqlErrors.flatMap({ $0.logLines })
-      CodegenLogger.log(String(describing: errorlines), logLevel: .error)
+      let errorlines = graphqlErrors.flatMap({
```
BobaFetters, line 308: Thinking here is by making the `sourceLocations` optional and therefore the `logLines` optional, we can determine how to display a given error, so errors with `sourceLocations` get displayed as they always have, but errors without `sourceLocations` no longer cause a precondition failure and instead just have their `name` and `message` displayed to ensure the user always gets usable error information

AnthonyMDev: This is an awesome solution!

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -305,8 +305,14 @@ public class ApolloCodegen {
...
+        if let logLines = $0.logLines {
+          return logLines
+        } else {
+          return ["[Error] \($0.name ?? "unknown"): \($0.message ?? "")"]
```
AnthonyMDev, line 312: I believe the `[Error]` label is already prepended to the log by the `CodegenLogger`. So putting it here might make it display it twice in the actual log.

AnthonyMDev (approved):
Nice job! Just one thing to verify and this is good to go.

Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: Good callout I'll remove that

PR #3135, opened 2023-07-21
fix: Fragment definitions support escaped characters

### LINKED ISSUE #3134: apollo-ios-cli code generation fails for fragment containing double quotes.
### Summary

If the fragment contains double quotes, a compile error will occur because the double quotes are not escaped in the fragmentDefinition of the generated code.

### Version

v1.3.2

Example fragment
```
fragment ExampleFragment on Example {
  field(parameter: "value")
}
```
Generated:
```
public struct ExampleFragment: SelectionSet, Example {
  public static var fragmentDefinition: StaticString {
    "fragment ExampleFragment on Example { __typename field( parameter: "value" ) }"
  }
```

### Logs

_No response_

### Anything else?

_No response_

BobaFetters:
-Updated the fragment definition generation to be a raw string like the operation definition to be able to support containing characters like quotes 
-Updated run-test-codegen-configurations.sh script to fail if code generation gives an error 
-Added a new field to Crocodile in AnimalKingdomAPI that accepts a String parameter 
-Added a new fragment to AnimalKingdomAPI that uses the field with string parameter to validate fragment definition generation

Closes #3134

### DIFF
--- a/Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Fragments/ClassroomPetDetails.graphql.swift
+++ b/Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Fragments/ClassroomPetDetails.graphql.swift
@@ -4,30 +4,9 @@
 @_exported import ApolloAPI
 
 public struct ClassroomPetDetails: AnimalKingdomAPI.SelectionSet, Fragment {
-  public static var fragmentDefinition: StaticString { """
-    fragment ClassroomPetDetails on ClassroomPet {
-      __typename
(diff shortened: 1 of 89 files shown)


AnthonyMDev (approved):
LGTM

calvincestari (approved):
The multiline to single line changes to the generated code - are those included here simply because it wasn't regenerated in that earlier PR that made that change?

BobaFetters:
> The multiline to single line changes to the generated code - are those included here simply because it wasn't regenerated in that earlier PR that made that change?

Yea

PR #3150, opened 2023-07-24
fix: Updated mock function descriptor handling

### LINKED ISSUE #3130: apollo-ios-cli mock code generation fails for string arrays
### Summary

I tried out the code generation for mock objects and stubled upon an issue related to the `_setList` invocation of the initializer of the generated `MockObjects`. 

For all of those invocations I'm getting the compiler error 
```
Type of expression is ambiguous without more context
```

I alreay read through this related issue: [https://github.com/apollographql/apollo-ios/issues/3023](#3023). However in my case it even fails for string arrays.

Any help / workarounds for the meantime are appreciated.

### Version

v1.3.1

Example fragment
```graphql
fragment HumanName on HumanName {
    id
    ...
    maidenNames # [String]
}
```

Generated:
```swift
public extension Mock where O == HumanName {
  convenience init(
    id: String? = nil,
    ...
    maidenNames: [String]? = nil,
  ) {
    self.init()
    _setScalar(id, for: \.id)
    ...
    _setList(maidenNames, for: \.maidenNames) // <- Causing the compiler error
  }
}

```

iOS Version: 15
Xcode Version: 14.3.1

### Logs

_No response_

### Anything else?

_No response_

BobaFetters:
-Updated mock function descriptor handling to check for entity vs scalars (scalar, custom scalar, enum) lists to check for proper function descriptor 
-Also checks nested lists
-Updated test to include cases for these

Closes #3130

### DIFF
--- a/Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
+++ b/Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
@@ -76,10 +76,18 @@ struct MockObjectTemplate: TemplateRenderer {
   
   private func mockFunctionDescriptor(_ graphQLType: GraphQLType) -> String {
     switch graphQLType {
-      case .list(.nonNull(.scalar(_))):
-        return "ScalarList"
-      case .list(_):
-        return "List"
+      case .list(let type):
+        switch type {
+        case .nonNull(.list(_)),
+             .list(_):
+          return mockFunctionDescriptor(type)
+        case .nonNull(.entity(_)),
+             .entity(_):
+          return "List"
+        default:
+          break
+        }
+      return "ScalarList"
       case .scalar(_), .enum(_):
         return "Scalar"
       case .entity(_):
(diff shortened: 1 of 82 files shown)


BobaFetters:
Failing tests in this PR are fixed by #3135

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
```diff
@@ -76,10 +76,18 @@ struct MockObjectTemplate: TemplateRenderer {
...
+      case .list(let type):
+        switch type {
+        case .nonNull(.list(_)),
+             .list(_):
+          return mockFunctionDescriptor(type)
+        case .nonNull(.entity(_)),
+             .entity(_):
+          return "List"
+        default:
+          break
```
AnthonyMDev, line 88: Should the `return "ScalarList"` just be here in the `default` case instead of the `break`?

AnthonyMDev (approved):
Minor nit but thanks for the fix!

Sources/ApolloCodegenLib/Templates/MockObjectTemplate.swift
BobaFetters: It could go there there as well, but it would still need to be where it is now so it would be in two places because of the nested switch statement so I just had it break out of the inner switch and have a single return for `ScalarList`

AnthonyMDev: ooooh i see! Nevermind this is fine!

calvincestari (approved):

PR #3152, opened 2023-07-25
feature: Adding CLI command for Operation Manifest

BobaFetters:
- Added new CLI command for generating operation manifest
- Added tests for new CLI command
- Updated CLI documentation to contain new command


Sources/CodegenCLI/Commands/GenerateOperationManifest.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    var configData: Data?
+    switch (inputs.string, inputs.path) {
+    case let (.some(string), _):
+      configData = try string.asData()
+    case let (nil, path):
+      configData = try fileManager.unwrappedContents(atPath: path)
+    }
+    
+    guard let data = configData else {
```
calvincestari, line 49: It seems like the only reason we need the `guard` here is because of `configData` being defined as optional when both `asData` and `unwrappedContents(atPath:)` return a non-optional type. I'm not sure there is any value in the error message with those return types and command validation should ensure that one of `--path` or `--string` must be supplied with the command.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -124,6 +124,37 @@ public class ApolloCodegen {
...
+    let compilationResult = try compileGraphQLResult(
+      configContext,
+      experimentalFeatures: configuration.experimentalFeatures
+    )
+    
+    try validate(configContext, with: compilationResult)
+    
+    let ir = IR(compilationResult: compilationResult)
+    
+    var operationIDsFileGenerator = OperationManifestFileGenerator(config: configContext)
```
calvincestari, line 149: Could this be a `let` constant too?

calvincestari (commented):
There is also a [README file for the CLI](https://github.com/apollographql/apollo-ios/blob/main/Sources/apollo-ios-cli/README.md) that needs to be updated. It's a manual edit and I usually just copy-paste the help output directly from what the CLI spits out.

Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: So `collectOperationIdentifier` is a mutating function so this does still need to be a var.

PR #3157, opened 2023-07-27
docs: Update Persisted Queries documentation

BobaFetters:
Converted the existing APQs page into a Persisted Queries page that covers both Persisted Queries and APQs and how to use them.


docs/source/config.json
```diff
@@ -78,7 +78,7 @@
         "Error Handling":  "/fetching/error-handling",
         "Type Conditions": "/fetching/type-conditions",
         "Custom Scalars": "/custom-scalars",
-        "Automatic Persisted Queries": "/fetching/apqs"
+        "Persisted Queries": "/fetching/persisted-queries"
```
zionts, line 81: Do you need to set up a redirect from the old endpoint?

docs/source/fetching/persisted-queries.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: Persisted Queries
+---
+
+Apollo supports features called "Persisted Queries" and "Automatic Persisted Queries" (APQs), to avoid sending the same large query documents over and over, and to allow safelisting of queries.
+
+Each query or mutation is identified by the SHA256 hash of its contents that can be sent to the server in lieu of sending a full query document. For APQs specifically, if the hash can't be found by the server, it sends back an error indicating that it needs the full query. If it receives this specific error, Apollo iOS will automatically retry the operation with the full query document without you having to do anything.
```
zionts, line 7: nit: Do we want to say "each operation" since operations now include subscriptions as well?

Also PQs does allow for user-customizable identifiers that are _not_ the SHA256 hash of the document contents, but we could specify that in a note rather than avoiding the nice simplicity you have here. Maybe a block comment below this could call it out?

docs/source/config.json
BobaFetters: Yea that's a good idea!

docs/source/fetching/persisted-queries.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+### Support
+
+| | **Persisted Queries** | **Automatic Persisted Queries** |
+| - | :---------------: | :-------------------------: |
+| **GraphOS** | Yes | No |
```
zionts, line 13: Not sure what it means to say Automatic Persisted Queries isn't supported in GraphOS? Would it be better to specify Gateway vs. Router vs. Server here?

docs/source/fetching/persisted-queries.mdx
```diff
@@ -0,0 +1,98 @@
...
+| **Apollo Server** | Yes?, version ?? | Yes, version ?? |
+
+### Feature
+
+| | **Persisted Queries** | **Automatic Persisted Queries** |
+| - | :---------------: | :-------------------------: |
+| **Pre-registered queries** | Yes | Yes |
+| **Automatically persist queries from client** | No | Yes |
+| **Allow only safelisted persisted queries** | Yes | Yes |
```
zionts, line 22: To me this is a little confusing - would it be better to say that persisted queries' default implementation is compatible with APQs and then additional safelisting requires APQs? Though yes, I suppose you can have safelisting of _documents_ and still use APQs. Is that what you're going for in this table?

docs/source/fetching/persisted-queries.mdx
```diff
@@ -0,0 +1,98 @@
...
+      "name": "GetItem",
+      "type": "query"
+    }
+  ]
+}
+```
+
+> **Note:** This example is using the `persistedQueries` version of the operation manifest, the `legacyAPQs` version will look slightly different.
+
+### 2. Enable persisted queries on your `ApolloClient`
```
zionts, line 74: Should this say "automatic" persisted queries? I figure if folks have done the above manifest input then they don't need to use `autoPersistQueries`, right?

docs/source/fetching/persisted-queries.mdx
BobaFetters: So saying instead of "Each query or mutation" say "Each operation"?

Also we don't currently have support for supplying custom operation identifiers implemented in iOS.

docs/source/fetching/persisted-queries.mdx
BobaFetters: This is definitely one of the areas I'm looking for input for what the messaging should be here. We could definitely specify Gateway vs Server vs Router versions, which would give a more detail view of the support.

docs/source/fetching/persisted-queries.mdx
BobaFetters: So right now in iOS the `autoPersistQueries` flag is used to enable both PQs and APQs for the network code, eventually we will add a separate flag specifically for PQs though.

docs/source/fetching/persisted-queries.mdx
BobaFetters: So this table I was just trying to see if I could break down some of the key features or callouts between the two, definitely open to suggestions for changes or another way to represent the differences especially if this is confusing.

docs/source/fetching/persisted-queries.mdx
Meschreiber: Agree with replacing "Each query or mutation" with "Each operation". I'm trying to consistently use "operation" instead of "query" in the [main PQ docs](https://github.com/apollographql/docs/pull/496).

docs/source/fetching/persisted-queries.mdx
Meschreiber: Generally, I'm a fan of tables, but wondering if just listing out requirements might be clearer here.

I believe for PQ it's just:
- GraphOS (Enterprise plan)
- Router >= v1.25.0

For APQ, is it just:
- Apollo Server (not sure of version)

I might be misunderstanding why we would call it support though.

docs/source/fetching/persisted-queries.mdx
Meschreiber: I would either use [this table](https://deploy-preview-496--apollo-monodocs.netlify.app/graphos/operations/persisted-queries#differences-from-automatic-persisted-queries), or link to it with a slight introduction.

(BTW this is entire doc, table included, is still WIP and I'm happy to change it.)

docs/source/fetching/persisted-queries.mdx
```diff
@@ -0,0 +1,98 @@
...
+let networkTransport = RequestChainNetworkTransport(
+    interceptorProvider: interceptorProvider,
+    endpointURL: URL(string: "http://localhost:4000/graphql")!
+)
+
+let client = ApolloClient(networkTransport: networkTransport, store: store)
+```
+For more information on configuring your `ApolloClient`, `NetworkTransport`, `InterceptorProvider`, and the request chain, see the [Request Pipeline](./../networking/request-pipeline) documentation.
+
+> **Note:** Sending APQs as `Get` requests
```
Meschreiber, line 92: Curious as to why this was made a note rather than [a separate section](https://www.apollographql.com/docs/ios/fetching/apqs/#send-apqs-as-get-requests) as in the current docs.

Meschreiber (commented):
Thanks so much for this! Left a couple of questions, but will also stack a PR on this one with some suggestions.

Meschreiber:
Hmm, also wondering why the @apollographql/docs team didn't get pinged on this one, since it's listed in the [codeowners file](https://github.com/apollographql/apollo-ios/blob/aaeed597fbb6c9382f69b521ecd6fe352cfe8e2c/.github/CODEOWNERS.md?plain=1#L2)...

docs/source/config.json
Meschreiber: Handled in https://github.com/apollographql/apollo-ios/pull/3160

BobaFetters:
Closing this, work moved to #3160

PR #3161, opened 2023-07-28
refactor: `Selection` enum and `Condition` type

### LINKED ISSUE #3140: Modify fragment protocols to support deferred properties
This is the more general fragment work mentioned in the [Generated models](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#generated-models) section of the RFC. This will probably have a dependency on #3139 being completed first.

> Regardless of the fragment/field solution chosen all deferred fragment definitions in generated models `__selections` will get an additional property to indicate they are deferred. This helps to understand the models when reading them as well as being used by internal code.
> ```
> // Updated Selection enum
> public enum Selection {
>   // other cases not shown
>   case fragment(any Fragment.Type, deferred: Bool)
>   case inlineFragment(any InlineFragment.Type, deferred: Bool)
> 
>   // other properties and types not shown
> }
> 
> // Sample usage in a generated model
> public class ExampleQuery: GraphQLQuery {
>   // other properties and types not shown
> 
>   public struct Data: ExampleSchema.SelectionSet {
>     public static var __selections: [ApolloAPI.Selection] { [
>       .fragment(EntityFragment.self, deferred: true),
>       .inlineFragment(AsEntity.self, deferred: true),
>     ] }
>   }
> }
> ```

calvincestari:
Closes #3140 

This covers the work to refactor the `Selection` enum allowed for the `.inlineFragment` and `.fragment` cases to be generated with support for `@defer`.

### DIFF
--- a/Sources/Apollo/FieldSelectionCollector.swift
+++ b/Sources/Apollo/FieldSelectionCollector.swift
@@ -77,14 +77,16 @@ struct DefaultFieldSelectionCollector: FieldSelectionCollector {
                             info: info)
         }
 
-      case let .fragment(fragment):
+      // TODO: _ is fine for now but will need to be handled in #3145
+      case let .fragment(fragment, _):
         groupedFields.addFulfilledFragment(fragment)
         try collectFields(from: fragment.__selections,
                           into: &groupedFields,
                           for: object,
                           info: info)
 
-      case let .inlineFragment(typeCase):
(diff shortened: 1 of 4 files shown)


BobaFetters (approved):

AnthonyMDev (approved):
LGTM

PR #3163, opened 2023-07-31
fix: Removing newline characters from operation id hashes

### LINKED ISSUE #3162: 1.3.2 - Wrong \n hash with persistent queries
### Summary

It looks like the hash for persistent requests hashes `\n` as two characters. But by default on the backend, this character is hashed as a single character. Due to a hash mismatch, the backend rejects any request

### Version

1.3.2

1. Turn on persistent queries
2. Send any query with fragment

### Logs

```shell
Query:
{
  "extensions": {
    "persistedQuery": {
      "sha256Hash": "536acf1071fce858f0ccb1def84d13ddc4d60f0ceb7a72b711f7f1c3bfc0392f",
      "version": 1
    }
  },
  "operationName": "PhoneCountriesList",
  "query": "query PhoneCountriesList { system { __typename phoneCountries { __typename ...UPhoneCountry } } }\nfragment UPhoneCountry on PhoneCountry { __typename code phoneCode englishName nativeName }"
}

Responce:
{
  "errors": [
    {
      "message": "Wrong hash was passed"
    }
  ]
}
Right hash:
64a17fdeb09cbed60e07b1d7ac4f46da17ab0d98e22fa6adbf9e2d18c57fc58d
```

### Anything else?

_No response_

BobaFetters:
- Updating operation/fragment definition handling to remove newline characters

Closes #3162

### DIFF
--- a/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
@@ -14,7 +14,7 @@ struct OperationManifestItem {
 
     var source = operation.definition.source.convertedToSingleLine()
     for fragment in operation.referencedFragments {
-      source += #"\n\#(fragment.definition.source.convertedToSingleLine())"#
+      source += #" \#(fragment.definition.source.convertedToSingleLine())"#
     }
     self.source = source
   }
--- a/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
+++ b/Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
@@ -14,7 +14,7 @@ struct OperationManifestItem {
 
     var source = operation.definition.source.convertedToSingleLine()
     for fragment in operation.referencedFragments {
-      source += #" \#(fragment.definition.source.convertedToSingleLine())"#
+      source += #"\n\#(fragment.definition.source.convertedToSingleLine())"#
     }
     self.source = source
   }
--- a/Sources/ApolloCodegenLib/IR/IR.swift
+++ b/Sources/ApolloCodegenLib/IR/IR.swift
@@ -169,7 +169,7 @@ class IR {
 
         var newline: String
         for fragment in referencedFragments {
-          newline = #"\n"#
+          newline = "\n"
           updateHash(with: &newline)
           var fragmentSource = fragment.definition.source.convertedToSingleLine()
           updateHash(with: &fragmentSource)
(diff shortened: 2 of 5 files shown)


Sources/ApolloCodegenLib/FileGenerators/OperationManifestFileGenerator.swift
```diff
@@ -14,7 +14,7 @@ struct OperationManifestItem {
 
     var source = operation.definition.source.convertedToSingleLine()
     for fragment in operation.referencedFragments {
-      source += #"\n\#(fragment.definition.source.convertedToSingleLine())"#
+      source += #" \#(fragment.definition.source.convertedToSingleLine())"#
```
calvincestari, line 17: The recent change here was to make this a raw string with the `#` character for Swift. I think the root issue here is that because it is now a raw string the hash computation sees the raw two characters instead of when it was a regular string it would have been escaped down to an actual newline (single) character.

I don't know if it's 100% correct here to be replacing this with a string unless that's what we'll actually be sending - given that we're modifying the source it looks like it will be. Question then - does this break any backwards compatibility with older generated fragments or do we not need to care about that?

BobaFetters: That's a good point, the raw string could be contributing to the issue here. I would lean towards still removing the newline character as I don't remember there being any specific reason to have it other than we thought we should keep it before each fragment. This route would bring us inline with how Kotlin is handling this as well.

As for backwards compatibility I don't think this should be an issue because the original change was very recent for persisted queries and with this bug operations containing fragments aren't working correctly to begin with.

@AnthonyMDev thoughts?

AnthonyMDev (changes requested):
If we are changing this, we definitely need to also change the way we send the document over the wire. We may need to chat about this further.

Sources/ApolloCodegenLib/IR/IR.swift
```diff
@@ -169,7 +169,7 @@ class IR {
 
         var newline: String
         for fragment in referencedFragments {
-          newline = #"\n"#
+          newline = "\n"
```
BobaFetters, line 172: Ok so did a little bit of testing (thanks @calvincestari for calling out the raw string potentially being an issue) and believe the issue is the operation string in the json that is getting processed with the `\n` treated as an actual newline, but because we switched this piece of code in hash generation to a raw string it was being processed as the actual characters. After discussion with @AnthonyMDev this should fix the actual issue while maintaining backward compatibility for legacy apqs

AnthonyMDev (approved):

Issue #3146, opened 2023-07-24
Implement `deferSpec` response parser


calvincestari:
This is part 2 to #3118 and is an implementation of the [parser for the `deferSpec` protocol](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#multipart-parsing-protocol) specification.

I have a local branch with an implementation but I'm not satisfied with it and I think it should change direction as I [commented on in #3118](https://github.com/apollographql/apollo-ios/pull/3119#issuecomment-1648207874)

> .. I'm not happy with the way each multipart parser needs to implement errors and a full parser. I think those should be centralized and each spec parser should only be unique in the actual data format.

calvincestari:
Improve on the protocol implementation per [this comment](https://github.com/apollographql/apollo-ios/pull/3119#issuecomment-1648207874).

calvincestari:
The `MultipartResponseDeferParser` was introduced in https://github.com/apollographql/apollo-ios/pull/3167. This work should build on that.

PR #3167, opened 2023-07-31
feature: Adds request `Accept` header

### LINKED ISSUE #3143: Automatically add `deferSpec` to `Accept` header for deferred operations
This work is part of the [Networking](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#request-header) section of the RFC.

~_This work will probably have a dependency on #3139 or #3141.~

calvincestari:
Closes #3143 
Related to #3146 (introduces the `MultipartResponseDeferParser`)

This PR ensures that when an operation is recognized as having deferred fragments the request automatically has the correct `Accept` HTTP header set to specify a multipart response and one that matches the currently configured defer spec parser.

### DIFF
--- a/Sources/Apollo/RequestChainNetworkTransport.swift
+++ b/Sources/Apollo/RequestChainNetworkTransport.swift
@@ -83,7 +83,7 @@ open class RequestChainNetworkTransport: NetworkTransport {
     cachePolicy: CachePolicy,
     contextIdentifier: UUID? = nil
   ) -> HTTPRequest<Operation> {
-    var request = JSONRequest(
+    let request = JSONRequest(
       operation: operation,
       graphQLEndpoint: self.endpointURL,
       contextIdentifier: contextIdentifier,
--- a/Sources/Apollo/RequestChainNetworkTransport.swift
+++ b/Sources/Apollo/RequestChainNetworkTransport.swift
@@ -69,7 +69,7 @@ open class RequestChainNetworkTransport: NetworkTransport {
     self.useGETForPersistedQueryRetry = useGETForPersistedQueryRetry
   }
   
-  /// Constructs a default (ie, non-multipart) GraphQL request.
+  /// Constructs a GraphQL request for the given operation.
   ///
   /// Override this method if you need to use a custom subclass of `HTTPRequest`.
   ///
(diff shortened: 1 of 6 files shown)


Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -70,6 +71,10 @@ public extension GraphQLOperation {
     return nil
   }
 
+  static var hasDeferredFragments: Bool {
+    false
```
calvincestari, line 75: This will get set for each operation that has deferred fragments in it as part of the work for https://github.com/apollographql/apollo-ios/issues/3114.

Sources/Apollo/RequestChainNetworkTransport.swift
```diff
@@ -81,18 +81,37 @@ open class RequestChainNetworkTransport: NetworkTransport {
...
+    }
+
+    if Operation.hasDeferredFragments {
+      request.addHeader(
+        name: "Accept",
+        value: "multipart/mixed;boundary=\"graphql\";\(MultipartResponseDeferParser.protocolSpec),application/json"
+      )
+    }
+
+    return request
```
AnthonyMDev, line 114: Moving these here (instead of in `send`) is definitely an improvement. :)

Sources/Apollo/RequestChainNetworkTransport.swift
```diff
@@ -81,18 +81,37 @@ open class RequestChainNetworkTransport: NetworkTransport {
...
+    )
+
+    if Operation.operationType == .subscription {
+      request.addHeader(
+        name: "Accept",
+        value: "multipart/mixed;boundary=\"graphql\";\(MultipartResponseSubscriptionParser.protocolSpec),application/json"
+      )
+    }
+
+    if Operation.hasDeferredFragments {
```
AnthonyMDev, line 107: Is there a chance that we will have a subscription that also has deferred fragments? This would override that protocol spec in that case.

AnthonyMDev (commented):
Looks good, just have a question about deferred fragments _in_ subscriptions.

Sources/Apollo/RequestChainNetworkTransport.swift
calvincestari: No, I don't think defer is supported in subscriptions. In that case we'd be indirectly relying on graphql-js to invalidate the operation.

BobaFetters: Is graphql-js already handling that case?

calvincestari: I went looking for validation of my assumption and it was not as easy to find as I thought it was. I did find reference to not being allowed on a Subscription in the [spec edits PR](https://github.com/graphql/graphql-spec/pull/742/files#diff-607ee7e6b71821eecadde7d92451b978e8a75e23d596150950799dc5f8afa43eR1559) though.

AnthonyMDev: As long as it's definitely not allowed in the specification, I'm fine with merging this then. If there is a bug in the future, that is a bug in GraphQL-js not Apollo-iOS. :)

calvincestari: I think we're fine. IMO defer on subscriptions is pointless.

calvincestari: OK, more confirmation in [Apollo Client's httpLink](https://github.com/apollographql/apollo-client/blob/main/src/link/http/createHttpLink.ts#L149-L150).

Issue #3164, opened 2023-07-31
Make Codegen Namespacing Optional


Mordil:
### Use case

The Codegen API tries to support many architecture use cases, and in some it's possible to have name collisions if the generated code happens to be in the same module.

This is resolved by wrapping the code in a namespace enum, which is configurable.

However, this is completely unnecessary in some architectures, where all the generated code is encapsulated in a single Swift module - the Swift module acts as the namespace.

This is particularly painful for projects that are upgrading from the previous codegen to the new, where thousands of lines of code are needing to be updated to append the namespace, with no other changes to the type reference.

### Describe the solution you'd like

In the Codegen API, `namespace` should be entirely optional.

calvincestari:
Hi @Mordil, this is already available with the `other` module type. I've attached a sample project that demonstrates the configuration.

[Archive.zip](https://github.com/apollographql/apollo-ios/files/12232291/Archive.zip)

Mordil:
@calvincestari Then documentation needs to be updated as this is not clearly explained and actually seems counter-intuitive from the symbol docs on the enum options.

The `other` type mentions that `schemaNamespace` is required and that the option is more for things like Cocoapods (when I'm using a SwiftPM target to host this code)

and `embeddedInTarget` mentions manually including files into the project

Mordil:
What about for the use case where I want all of my schema types to be defined in one module, but all of my operations are defined in downstream modules, but I don't want to namespace the schema types in the sole module?

What configuration combination should I use?

```
- schema module
  - operationSubsetModule
  - operationSubsetModule
```

Mordil:
For context in this other use case,

we want a top level schema module that has all of the "currency" types of the schema, as public, and then we want each submodule to encapsulate the operations as internal, only exposing the currency types to our app that imports these operation modules.

What way can we setup the configuration to have the submodules import the schema module, and control over their access modifiers?

calvincestari:
I think you should still be able to use a schema module type of `other`, eliminating the enum namespace. Combined with an operations output of `relative(subpath:accessModifier:)`, which would require you to place the operation definitions in each of the downstream modules.

> we want a top level schema module that has all of the "currency" types of the schema, as public, and then we want each submodule to encapsulate the operations as internal, only exposing the currency types to our app that imports these operation modules.

I think the above configuration should satisfy that.

> What way can we setup the configuration to have the submodules import the schema module, and control over their access modifiers?

[...]

PR #3170, opened 2023-08-01
docs: Clarify `other` schema module type

calvincestari:
I've tried to add clarity to the `other` schema module type as a result of feedback in https://github.com/apollographql/apollo-ios/issues/3164. Let me know if you think we're going in the wrong direction here.


AnthonyMDev (approved):
LGTM

PR #3171, opened 2023-08-02
feature: Add support for field casing strategy to code gen

### LINKED ISSUE #2738: Add casing strategies for property names
### Use case

I created a new issue for this feature request since we closed #2641.

The gist is to provide casing strategies for other parts of generated code, such as fields, similar to the current option for enums.

_No response_

BobaFetters:
- Added new codegen configuration option under `ConversionStrategies` for `fieldCasing` to allow customization of how field names are generated.
- This will default to the new `default` `CaseConversionStrategy` which will function the same way field generation has previously.
- The `camelCase` option will apply camel case conversion to the field names
- Also deprecated the `none` case and replaced it with the `default` case for enum casing

Closes #2738

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -668,20 +668,31 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     /// cases on the generated Swift enums.
     /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
     public let enumCases: CaseConversionStrategy
+    
+    /// Determines how the names of fields in the GraphQL schema will be converted into
+    /// properties in the generated Swift code.
+    /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
+    public let fieldCasing: CaseConversionStrategy
 
     /// Default property values
     public struct Default {
       public static let enumCases: CaseConversionStrategy = .camelCase
+      public static let fieldCasing: CaseConversionStrategy = .camelCase
     }
 
-    public init(enumCases: CaseConversionStrategy = Default.enumCases) {
+    public init(
+      enumCases: CaseConversionStrategy = Default.enumCases,
+      fieldCasing: CaseConversionStrategy = Default.fieldCasing
+    ) {
       self.enumCases = enumCases
+      self.fieldCasing = fieldCasing
     }
 
     // MARK: Codable
 
     public enum CodingKeys: CodingKey {
       case enumCases
+      case fieldCasing
     }
 
     public init(from decoder: Decoder) throws {
@@ -698,6 +709,11 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
         CaseConversionStrategy.self,
         forKey: .enumCases
       ) ?? Default.enumCases
+      
+      fieldCasing = try values.decodeIfPresent(
+        CaseConversionStrategy.self,
+        forKey: .fieldCasing
+      ) ?? Default.fieldCasing
     }
   }
   
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -651,24 +651,13 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     case exclude
   }
 
-  /// ``CaseConversionStrategy`` is used to specify the strategy used to convert the casing of
-  /// GraphQL schema values into generated Swift code.
-  @available(*, deprecated, message: "Use EnumCaseConversionStrategy instead.")
-    public enum CaseConversionStrategy: String, Codable, Equatable {
-      /// Generates swift code using the exact name provided in the GraphQL schema
-      /// performing no conversion.
-      case none
-      /// Convert to lower camel case from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
-      case camelCase
-  }
-
   /// ``ConversionStrategies`` configures rules for how to convert the names of values from the
   /// schema in generated code.
   public struct ConversionStrategies: Codable, Equatable {
 
     /// ``ApolloCodegenConfiguration/ConversionStrategies/EnumCase`` is used to specify the strategy
     /// used to convert the casing of enum cases in a GraphQL schema into generated Swift code.
-    public enum EnumCase: String, Codable, Equatable {
+    public enum EnumCases: String, Codable, Equatable {
       /// Generates swift code using the exact name provided in the GraphQL schema
       /// performing no conversion.
       case none
@@ -676,10 +665,10 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       case camelCase
     }
 
-    /// ``ApolloCodegenConfiguration/ConversionStrategies/FieldAccessor`` is used to specify the
+    /// ``ApolloCodegenConfiguration/ConversionStrategies/FieldAccessors`` is used to specify the
     /// strategy used to convert the casing of fields on GraphQL selection sets into field accessors
     /// on the response models in generated Swift code.
-    public enum FieldAccessor: String, Codable, Equatable {
+    public enum FieldAccessors: String, Codable, Equatable {
       /// This conversion strategy will:
       /// - Lowercase the first letter of all fields.
       /// - Convert field names that are all `UPPERCASE` to all `lowercase`.
@@ -693,45 +682,32 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     /// Determines how the names of enum cases in the GraphQL schema will be converted into
     /// cases on the generated Swift enums.
     /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
-    public let enumCases: EnumCase
+    public let enumCases: EnumCases
     
     /// Determines how the names of fields in the GraphQL schema will be converted into
     /// properties in the generated Swift code.
     /// Defaults to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
-    public let fieldAccessor: FieldAccessor
+    public let fieldAccessors: FieldAccessors
 
     /// Default property values
     public struct Default {
-      public static let enumCases: EnumCase = .camelCase
-      public static let fieldAccessor: FieldAccessor = .idiomatic
-    }
-
-    @available(*, deprecated, renamed: "init(enumCases:fieldAccessor:)")
-    public init(
-      enumCases: CaseConversionStrategy
-    ) {
-      switch enumCases {
-      case .none:
-        self.enumCases = .none
-      case .camelCase:
-        self.enumCases = .camelCase
-      }
-      self.fieldAccessor = Default.fieldAccessor
+      public static let enumCases: EnumCases = .camelCase
+      public static let fieldAccessors: FieldAccessors = .idiomatic
     }
       
     public init(
-      enumCases: EnumCase = Default.enumCases,
-      fieldAccessor: FieldAccessor = Default.fieldAccessor
+      enumCases: EnumCases = Default.enumCases,
+      fieldAccessors: FieldAccessors = Default.fieldAccessors
     ) {
       self.enumCases = enumCases
-      self.fieldAccessor = fieldAccessor
+      self.fieldAccessors = fieldAccessors
     }
 
     // MARK: Codable
 
     public enum CodingKeys: CodingKey {
       case enumCases
-      case fieldAccessor
+      case fieldAccessors
     }
 
     public init(from decoder: Decoder) throws {
@@ -756,15 +732,15 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
         }
       } else {
         enumCases = try values.decodeIfPresent(
-          EnumCase.self,
+          EnumCases.self,
           forKey: .enumCases
         ) ?? Default.enumCases
       }
       
-      fieldAccessor = try values.decodeIfPresent(
-        FieldAccessor.self,
-        forKey: .fieldAccessor
-      ) ?? Default.fieldAccessor
+      fieldAccessors = try values.decodeIfPresent(
+        FieldAccessors.self,
+        forKey: .fieldAccessors
+      ) ?? Default.fieldAccessors
     }
   }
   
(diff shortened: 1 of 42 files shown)


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -718,20 +735,31 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     /// cases on the generated Swift enums.
     /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
     public let enumCases: CaseConversionStrategy
+    
+    /// Determines how the names of fields in the GraphQL schema will be converted into
+    /// properties in the generated Swift code.
+    /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
```
AnthonyMDev, line 741:
```suggestion
    /// Defaults to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -704,11 +704,28 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
   /// GraphQL schema values into generated Swift code.
   public enum CaseConversionStrategy: String, Codable, Equatable {
-    /// Generates swift code using the exact name provided in the GraphQL schema
-    /// performing no conversion.
+    @available(*, deprecated, message: "`none` casing strategy has been renamed to `default`")
     case none
+    /// Generates swift code using a default casing strategy determined based on the type of code
+    /// being generated.
+    ///
+    /// For enums, this will render as provided in the GraphQL schema and will escape certina reserved keywords.
```
AnthonyMDev, line 712:
```suggestion
    /// For enums, this will render as provided in the GraphQL schema and will escape certain reserved keywords.
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -718,20 +735,31 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
     public let enumCases: CaseConversionStrategy
+    
+    /// Determines how the names of fields in the GraphQL schema will be converted into
+    /// properties in the generated Swift code.
+    /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
+    public let fieldCasing: CaseConversionStrategy
 
     /// Default property values
     public struct Default {
       public static let enumCases: CaseConversionStrategy = .camelCase
```
AnthonyMDev, line 746: I don't think it's right to have an enum case called `default` if it's not actually the default value for all the configurations.

Along with the fact that the `default` behavior is different for `enumCases` and `fieldCasing`, it really feels like a bad design choice to me. Which is leading me back to thinking we should bite the bullet and just go with two separate enums.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -718,20 +735,31 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+      public static let fieldCasing: CaseConversionStrategy = .default
```
AnthonyMDev, line 747: Can we rename this to `fieldAccessors`. `enumCases` is referring to the cases of an enum, not to upper"case"/lower"case". This is the `CaseConversionStrategy` for `fieldAccessors`.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -718,20 +735,31 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
     }
 
     // MARK: Codable
 
     public enum CodingKeys: CodingKey {
       case enumCases
+      case fieldCasing
     }
 
     public init(from decoder: Decoder) throws {
```
AnthonyMDev, line 765: I think the validation that throws the `DecodingError` in this initializer is incorrect now, we should change the `debugDescription` to just "Invalid value found."

AnthonyMDev (dismissed):
After looking at the way we need to define the `default` case to do different things for different types, and the awkwardness of having the `.default` case not actually be the default value for `enumCases`, I'm strongly in favor of using two different enums for the two types.

@calvincestari what's your opinion?

calvincestari:
> After looking at the way we need to define the `default` case to do different things for different types, and the awkwardness of having the `.default` case not actually be the default value for `enumCases`, I'm strongly in favor of using two different enums for the two types.
> 
> @calvincestari what's your opinion?

`default` is not descriptive of what the behaviour is, it simply allows us to give the case a name that fits the current behaviour. `none` is a more descriptive name IMO because it indicates that there will be no conversion.

If we used a second enum what would those cases be?

AnthonyMDev:
> > After looking at the way we need to define the `default` case to do different things for different types, and the awkwardness of having the `.default` case not actually be the default value for `enumCases`, I'm strongly in favor of using two different enums for the two types.
> > @calvincestari what's your opinion?
> 
> `default` is not descriptive of what the behaviour is, it simply allows us to give the case a name that fits the current behaviour. `none` is a more descriptive name IMO because it indicates that there will be no conversion.
> 
> If we used a second enum what would those cases be?

`none` only makes sense for `enumCases` where we said it's okay if the first letter is either upper or lower cased. But for `fieldAccessor` which does _not_ allow uppercased first letters, `default` makes more sense to me.

So my proposal is:
- Deprecate `CaseConversionStrategy`
- Add two new enums for enum and field accessor casing.
  - I would put these _inside_ of `ConversionStrategies`

@BobaFetters here is a code snippet of my suggestion. I even updated the documentation for you!

```swift
/// ``CaseConversionStrategy`` is used to specify the strategy used to convert the casing of
  /// GraphQL schema values into generated Swift code.
  @available(*, deprecated, message: "Use EnumCaseConversionStrategy instead.")
  public enum CaseConversionStrategy: String, Codable, Equatable {
    /// Generates swift code using the exact name provided in the GraphQL schema
    /// performing no conversion.
    case none
    /// Convert to lower camel case from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
    case camelCase
  }

  /// ``ConversionStrategies`` configures rules for how to convert the names of values from the
  /// schema in generated code.
  public struct ConversionStrategies: Codable, Equatable {

    /// ``ApolloCodegenConfiguration/ConversionStrategies/EnumCase`` is used to specify the strategy
    /// used to convert the casing of enum cases in a GraphQL schema into generated Swift code.
    public enum EnumCase: String, Codable, Equatable {
      /// Generates swift code using the exact name provided in the GraphQL schema
      /// performing no conversion.
      case none
      /// Convert to lower camel case from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
      case camelCase
    }

    /// ``ApolloCodegenConfiguration/ConversionStrategies/FieldAccessor`` is used to specify the
    /// strategy used to convert the casing of fields on GraphQL selection sets into field accessors
    /// on the response models in generated Swift code.
    public enum FieldAccessor: String, Codable, Equatable {
      /// This conversion strategy will:
      /// - Lowercase the first letter of all fields.
      /// - Convert field names that are all `UPPERCASE` to all `lowercase`.
      case `default`
      /// This conversion strategy will:
      /// - Convert to `lowerCamelCase` from `snake_case`, or `UpperCamelCase`.
      /// - Convert field names that are all `UPPERCASE` to all `lowercase`.
      case camelCase
    }

  ...
```

BobaFetters:
Yea I think I would still be in favor of keeping this as `none` instead of `default`, I think it's reasonable that casing strategies function differently for different types of code. If we don't want to do that then im curious how we would handle casing for different items, still have `none` and `camelCase` for enums, and then `default` and `camelCase` for field accessors?

calvincestari:
`swift` as a case name seems more descriptive than `default` considering we're changing casing to match what we'd expect to see in idiomatic Swift.

AnthonyMDev:
> `swift` as a case name seems more descriptive than `default` considering we're changing casing to match what we'd expect to see in idiomatic Swift.

I'm not that opinionated on the naming, but `swift` seems really confusing to me. If I didn't have context on this, I would think, "Isn't it all Swift code? Is there an option to generate code for other languages?"

Maybe something like `idiomatic` might be more expressive? I definitely see why `default` seems odd, but `swift` in my opinion is worse due to being unclear.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
   /// ``CaseConversionStrategy`` is used to specify the strategy used to convert the casing of
   /// GraphQL schema values into generated Swift code.
-  public enum CaseConversionStrategy: String, Codable, Equatable {
-    /// Generates swift code using the exact name provided in the GraphQL schema
-    /// performing no conversion.
-    case none
-    /// Convert to lower camel case from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
-    case camelCase
+  @available(*, deprecated, message: "Use EnumCaseConversionStrategy instead.")
+    public enum CaseConversionStrategy: String, Codable, Equatable {
```
AnthonyMDev, line 707: We should move this to the bottom of the file where the other deprecated stuff lives.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+      case idiomatic
+      /// This conversion strategy will:
+      /// - Convert to `lowerCamelCase` from `snake_case`, or `UpperCamelCase`.
+      /// - Convert field names that are all `UPPERCASE` to all `lowercase`.
+      case camelCase
+    }
+    
     /// Determines how the names of enum cases in the GraphQL schema will be converted into
     /// cases on the generated Swift enums.
     /// Defaultss to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
```
AnthonyMDev, line 745:
```suggestion
    /// Defaults to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
-    public let enumCases: CaseConversionStrategy
+    public let enumCases: EnumCase
+    
+    /// Determines how the names of fields in the GraphQL schema will be converted into
+    /// properties in the generated Swift code.
+    /// Defaults to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
+    public let fieldAccessor: FieldAccessor
```
AnthonyMDev, line 751:
```suggestion
    public let fieldAccessors: FieldAccessors
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+      /// performing no conversion.
+      case none
+      /// Convert to lower camel case from `snake_case`, `UpperCamelCase`, or `UPPERCASE`.
+      case camelCase
+    }
+
+    /// ``ApolloCodegenConfiguration/ConversionStrategies/FieldAccessor`` is used to specify the
+    /// strategy used to convert the casing of fields on GraphQL selection sets into field accessors
+    /// on the response models in generated Swift code.
+    public enum FieldAccessor: String, Codable, Equatable {
```
AnthonyMDev, line 732:
```suggestion
    public enum FieldAccessors: String, Codable, Equatable {
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+      case camelCase
   }
 
   /// ``ConversionStrategies`` configures rules for how to convert the names of values from the
   /// schema in generated code.
   public struct ConversionStrategies: Codable, Equatable {
+
+    /// ``ApolloCodegenConfiguration/ConversionStrategies/EnumCase`` is used to specify the strategy
+    /// used to convert the casing of enum cases in a GraphQL schema into generated Swift code.
+    public enum EnumCase: String, Codable, Equatable {
```
AnthonyMDev, line 721: I think all of these should be plural.
```suggestion
    public enum EnumCases: String, Codable, Equatable {
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+    /// Determines how the names of fields in the GraphQL schema will be converted into
+    /// properties in the generated Swift code.
+    /// Defaults to ``ApolloCodegenConfiguration/CaseConversionStrategy/camelCase``
+    public let fieldAccessor: FieldAccessor
 
     /// Default property values
     public struct Default {
-      public static let enumCases: CaseConversionStrategy = .camelCase
+      public static let enumCases: EnumCase = .camelCase
+      public static let fieldAccessor: FieldAccessor = .idiomatic
```
AnthonyMDev, line 756:
```suggestion
      public static let fieldAccessors: FieldAccessors = .idiomatic
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+    ) {
       self.enumCases = enumCases
+      self.fieldAccessor = fieldAccessor
     }
 
     // MARK: Codable
 
     public enum CodingKeys: CodingKey {
       case enumCases
+      case fieldAccessor
```
AnthonyMDev, line 784:
```suggestion
      case fieldAccessors
```

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -703,51 +703,118 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+        enumCases = try values.decodeIfPresent(
+          EnumCase.self,
+          forKey: .enumCases
+        ) ?? Default.enumCases
+      }
+      
+      fieldAccessor = try values.decodeIfPresent(
+        FieldAccessor.self,
+        forKey: .fieldAccessor
+      ) ?? Default.fieldAccessor
```
AnthonyMDev, line 817:
```suggestion
      fieldAccessors = try values.decodeIfPresent(
        FieldAccessors.self,
        forKey: .fieldAccessors
      ) ?? Default.fieldAccessors
```

AnthonyMDev (commented):
Haven't gotten time to review the actual code that does the case conversions. But here are some comments on the API for the configuration.

AnthonyMDev (approved):
Lets ship it!

PR #3172, opened 2023-08-03
Add output-path argument to Generate Operation Manifest 

AnthonyMDev:
Adds `--output-path` and `--manifest-version` arguments to generate-operation-manifest CLI command.

When used, these override any operation manifest configuration in the codegen configuration. This allows us to use the command even when there is no operation manifest in the config.


AnthonyMDev:
Closing in favor of #3175

PR #3174, opened 2023-08-03
refactor: Fragment types in IR

calvincestari:
This brings clarity to parts of the IR with regards to when work is being done on an inline fragment spread vs. a named fragment spread.

What was previously `IR.SelectionSet` when dealing with inline fragment spreads is now `IR.InlineFragmentSpread` which in turn now contains a selection set, deferred property, type info and inclusion conditions. Similarly `IR.FragmentSpread` is now `IR.NamedFragmentSpread`.


calvincestari:
@AnthonyMDev as we discussed earlier I'm going to merge this in without explicit approval as we paired on these changes so are both familiar with them.

PR #3175, opened 2023-08-04
Updating Persisted Query Configuration

BobaFetters:
-Moved all codegen config options related to persisted queries to a new top level `OperationManifestConfiguration` object
-Added an `autoGenerate` option to determine if the operation manifest should be generated when using the `generate` cli command
-Cleaned up/combined some code

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -96,14 +96,14 @@ public class ApolloCodegen {
     withRootURL rootURL: URL? = nil,
     buildOptions: CodeGenerationBuildOptions = [.code]
   ) throws {
-    try build(with: configuration, rootURL: rootURL)
+    try build(with: configuration, rootURL: rootURL, buildOptions: buildOptions)
   }
 
   internal static func build(
     with configuration: ApolloCodegenConfiguration,
     rootURL: URL? = nil,
     fileManager: ApolloFileManager = .default,
-    buildOptions: CodeGenerationBuildOptions = [.code]
+    buildOptions: CodeGenerationBuildOptions
   ) throws {
 
     let configContext = ConfigurationContext(
@@ -126,8 +126,10 @@ public class ApolloCodegen {
       var operationIDsFileGenerator = OperationManifestFileGenerator(config: configContext)
       
       for operation in compilationResult.operations {
-        let irOperation = ir.build(operation: operation)
-        operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+        autoreleasepool {
+          let irOperation = ir.build(operation: operation)
+          operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+        }
       }
       
       try operationIDsFileGenerator?.generate(fileManager: fileManager)
@@ -144,7 +146,8 @@ public class ApolloCodegen {
         compilationResult: compilationResult,
         ir: ir,
         config: configContext,
-        fileManager: fileManager
+        fileManager: fileManager,
+        buildOptions: buildOptions
       )
 
       if configuration.options.pruneGeneratedFiles {
@@ -411,7 +414,8 @@ public class ApolloCodegen {
     compilationResult: CompilationResult,
     ir: IR,
     config: ConfigurationContext,
-    fileManager: ApolloFileManager = .default
+    fileManager: ApolloFileManager = .default,
+    buildOptions: CodeGenerationBuildOptions
   ) throws {
     for fragment in compilationResult.fragments {
       try autoreleasepool {
@@ -431,11 +435,15 @@ public class ApolloCodegen {
         try OperationFileGenerator(irOperation: irOperation, config: config)
           .generate(forConfig: config, fileManager: fileManager)
 
-        operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+        if buildOptions.contains(.operationManifest) {
+          operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+        }
       }
     }
 
-    try operationIDsFileGenerator?.generate(fileManager: fileManager)
+    if buildOptions.contains(.operationManifest) {
+      try operationIDsFileGenerator?.generate(fileManager: fileManager)
+    }
     operationIDsFileGenerator = nil
 
     for graphQLObject in ir.schema.referencedTypes.objects {
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -67,6 +67,22 @@ public class ApolloCodegen {
       }
     }
   }
+  
+  public struct CodeGenerationBuildOptions: OptionSet {
+    public var rawValue: Int
+    
+    public static let code = CodeGenerationBuildOptions(rawValue: 1 << 0)
+    public static let operationManifest = CodeGenerationBuildOptions(rawValue: 1 << 1)
+    public static let all = [
+      CodeGenerationBuildOptions.code,
+      CodeGenerationBuildOptions.operationManifest
+    ]
+    
+    public init(rawValue: Int) {
+      self.rawValue = rawValue
+    }
+    
+  }
 
   /// Executes the code generation engine with a specified configuration.
   ///
@@ -77,7 +93,8 @@ public class ApolloCodegen {
   ///     If `nil`, the current working directory of the executing process will be used.
   public static func build(
     with configuration: ApolloCodegenConfiguration,
-    withRootURL rootURL: URL? = nil
+    withRootURL rootURL: URL? = nil,
+    buildOptions: CodeGenerationBuildOptions = [.code]
   ) throws {
     try build(with: configuration, rootURL: rootURL)
   }
@@ -85,7 +102,8 @@ public class ApolloCodegen {
   internal static func build(
     with configuration: ApolloCodegenConfiguration,
     rootURL: URL? = nil,
-    fileManager: ApolloFileManager = .default
+    fileManager: ApolloFileManager = .default,
+    buildOptions: CodeGenerationBuildOptions = [.code]
   ) throws {
 
     let configContext = ConfigurationContext(
@@ -103,57 +121,39 @@ public class ApolloCodegen {
     try validate(configContext, with: compilationResult)
 
     let ir = IR(compilationResult: compilationResult)
+    
+    if buildOptions.contains(.operationManifest) {
+      var operationIDsFileGenerator = OperationManifestFileGenerator(config: configContext)
+      
+      for operation in compilationResult.operations {
+        let irOperation = ir.build(operation: operation)
+        operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+      }
+      
+      try operationIDsFileGenerator?.generate(fileManager: fileManager)
+    }
 
-    var existingGeneratedFilePaths = configuration.options.pruneGeneratedFiles ?
-    try findExistingGeneratedFilePaths(
-      config: configContext,
-      fileManager: fileManager
-    ) : []
+    if buildOptions.contains(.code) {
+      var existingGeneratedFilePaths = configuration.options.pruneGeneratedFiles ?
+      try findExistingGeneratedFilePaths(
+        config: configContext,
+        fileManager: fileManager
+      ) : []
 
-    try generateFiles(
-      compilationResult: compilationResult,
-      ir: ir,
-      config: configContext,
-      fileManager: fileManager
-    )
-
-    if configuration.options.pruneGeneratedFiles {
-      try deleteExtraneousGeneratedFiles(
-        from: &existingGeneratedFilePaths,
-        afterCodeGenerationUsing: fileManager
+      try generateFiles(
+        compilationResult: compilationResult,
+        ir: ir,
+        config: configContext,
+        fileManager: fileManager
       )
-    }
-  }
 
-  public static func generateOperationManifest(
-    with configuration: ApolloCodegenConfiguration,
-    withRootURL rootURL: URL? = nil,
-    fileManager: ApolloFileManager = .default
-  ) throws {
-    let configContext = ConfigurationContext(
-      config: configuration,
-      rootURL: rootURL
-    )
-    
-    try validate(configContext)
-    
-    let compilationResult = try compileGraphQLResult(
-      configContext,
-      experimentalFeatures: configuration.experimentalFeatures
-    )
-    
-    try validate(configContext, with: compilationResult)
-    
-    let ir = IR(compilationResult: compilationResult)
-    
-    var operationIDsFileGenerator = OperationManifestFileGenerator(config: configContext)
-    
-    for operation in compilationResult.operations {
-      let irOperation = ir.build(operation: operation)
-      operationIDsFileGenerator?.collectOperationIdentifier(irOperation)
+      if configuration.options.pruneGeneratedFiles {
+        try deleteExtraneousGeneratedFiles(
+          from: &existingGeneratedFilePaths,
+          afterCodeGenerationUsing: fileManager
+        )
+      }
     }
-    
-    try operationIDsFileGenerator?.generate(fileManager: fileManager)
   }
 
   // MARK: Internal
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -73,9 +73,9 @@ public class ApolloCodegen {
     
     public static let code = ItemsToGenerate(rawValue: 1 << 0)
     public static let operationManifest = ItemsToGenerate(rawValue: 1 << 1)
-    public static let all = [
-      ItemsToGenerate.code,
-      ItemsToGenerate.operationManifest
+    public static let all: ItemsToGenerate = [
+      .code,
+      .operationManifest
     ]
     
     public init(rawValue: Int) {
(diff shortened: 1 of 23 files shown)


Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -164,16 +164,15 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
     /// The local path structure for the test mock operation object files.
     public let testMocks: TestMockFileOutput
-    /// Configures the generation of an operation manifest JSON file for use with persisted queries
-    /// or [Automatic Persisted Queries (APQs)](https://www.apollographql.com/docs/apollo-server/performance/apq).
-    /// Defaults to `nil`.
-    public let operationManifest: OperationManifestFileOutput?
+    
+    /// This var helps maintain backwards compatibility with legacy APQ config with the new
+    /// `OperationManifestConfiguration` and will be fully removed in v2.0
+    fileprivate let operationIDsPath: String?
```
BobaFetters, line 170: Added this and a private `apqConfig` var because moving the operation manifest options out of their existing container structs meant the current way of decoding them to handle backward compatibility didn't work so these variables allow that backward compatibility to happen in the main decoding of the over codegen configuration file.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -77,15 +93,17 @@ public class ApolloCodegen {
...
   ) throws {
     try build(with: configuration, rootURL: rootURL)
   }
 
   internal static func build(
     with configuration: ApolloCodegenConfiguration,
     rootURL: URL? = nil,
-    fileManager: ApolloFileManager = .default
+    fileManager: ApolloFileManager = .default,
+    buildOptions: CodeGenerationBuildOptions = [.code]
```
AnthonyMDev, line 106: I usually only add the default value to the public initializers in situations like this.

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -103,58 +121,40 @@ public class ApolloCodegen {
     try validate(configContext, with: compilationResult)
 
     let ir = IR(compilationResult: compilationResult)
+    
+    if buildOptions.contains(.operationManifest) {
+      var operationIDsFileGenerator = OperationManifestFileGenerator(config: configContext)
+      
+      for operation in compilationResult.operations {
+        let irOperation = ir.build(operation: operation)
```
AnthonyMDev, line 129: `ir.build(operation:)` does a _LOT_ of computing and each operation hold a lot in memory (thats why they are in an `autoreleasepool` in the code generation below. We should be careful to only do this once per operation.

But looking at the `OperationManifestFileGenerator`, we actually don't need the built operation to do this. We should be able to change `collectOperationIdentifier()` to take in the `CompilationResult.OperationDefinition` instead of the `IR. Operation`.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -228,21 +226,10 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
         forKey: .testMocks
       )
 
-      if let operationManifest = try values.decodeIfPresent(
```
AnthonyMDev, line 231: Need to also remove the `operationManifest` coding key on line 207.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -750,60 +695,6 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       ) ?? Default.enumCases
     }
   }
-
-  public struct OperationDocumentFormat: OptionSet, Codable, Equatable {
```
AnthonyMDev, line 754: I don't think `OperationDocumentFormat` should be moved to the manifest config. This is something that actually affects the way _code_ is generated and is not actually even used when generating the operation manifest itself.

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
```diff
@@ -0,0 +1,179 @@
...
+    
+    /// Configures the generation of an operation manifest JSON file for use with persisted queries
+    /// or [Automatic Persisted Queries (APQs)](https://www.apollographql.com/docs/apollo-server/performance/apq).
+    ///
+    /// The operation manifest is a JSON file that maps all generated GraphQL operations to an
+    /// operation identifier. This manifest can be used to register operations with a server utilizing
+    /// persisted queries
+    /// or [Automatic Persisted Queries (APQs)](https://www.apollographql.com/docs/apollo-server/performance/apq).
+    /// Defaults to `nil`.
+    public struct OperationManifestFileOutput: Codable, Equatable {
```
AnthonyMDev, line 94: When we move `operationDocumentFormat` back into it's original spot, we can probably just have `path` and `version` be properties on the top level `OperationManifestConfiguration` along with `autoGenerate`.

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
```diff
@@ -0,0 +1,179 @@
...
+    /// Configures the generation of an operation manifest JSON file for use with persisted queries
+    /// or [Automatic Persisted Queries (APQs)](https://www.apollographql.com/docs/apollo-server/performance/apq).
+    /// Defaults to `nil`.
+    public let operationManifest: OperationManifestFileOutput?
+    
+    /// How to generate the operation documents for your generated operations.
+    public let operationDocumentFormat: OperationDocumentFormat
+    
+    /// If set to `true` will generate the operation manfiest every time code generation is run. Defaults to `false`
+    public let autoGenerate: Bool
```
AnthonyMDev, line 18: Can we change the name of this to `generateManifestOnCodeGeneration` or something to that end?

Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -67,6 +67,22 @@ public class ApolloCodegen {
       }
     }
   }
+  
+  public struct CodeGenerationBuildOptions: OptionSet {
```
AnthonyMDev, line 71: These are more "Types of things to generate" rather than "build options". I do like the approach, but I think we should change the name maybe.

I could see us in the future altering this to allow you to say only re-generate fragments, or even only regenerate models in a specific sub-folder or something also.

AnthonyMDev (changes requested):

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
BobaFetters: My thinking was there was no use case for this outside of persisted queries, although that may be an incorrect assumption? Technically I guess you could do APQs with no manifest and still need this document for option yea?

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
BobaFetters: Yea that would definitely be more descriptive

Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: Yea I was having a hard time coming up with a good name for this, but figured the overall approach could have some benefits so I am open to different names for this.

Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: Looks like we do need the `operationIdentifier` and `referencedFragments` which are on the `IR.Operation` and not on the `CompilationResult.OperationDefinition`. Is there something I am missing here since it looks like those are calculated as part of the IR process?

AnthonyMDev: Oooof you're right. Okay well, then in that case, we should try to build the operations once and use it for both things. We also need to make sure that we _release_ the operation after we're done using it, too much memory usage on those objects.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -1281,16 +1257,12 @@ extension ApolloCodegenConfiguration.FileOutput {
     self.schemaTypes = schemaTypes
     self.operations = operations
     self.testMocks = testMocks
-    if let operationIdentifiersPath {
```
AnthonyMDev, line 1284: Need to change the deprecation message on this function, as it doesn't have the `operationManifest` param anymore.

I don't think we need to deprecate that function, as it existed for a short time and was really never used by anyone other than those in the preview (if at all).

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -954,11 +906,14 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
   public let experimentalFeatures: ExperimentalFeatures
   /// Schema download configuration.
   public let schemaDownloadConfiguration: ApolloSchemaDownloadConfiguration?
+  /// `OperationManifestConfiguration` for persisted queries or legacy APQs
```
AnthonyMDev, line 909:
```suggestion
  /// Configuration for generating an operation manifest for use with persisted queries.
```

AnthonyMDev (commented):
Can you also do a quick pass over the inline docs here, finding any reference to APQ and making sure it still is accurate. We now know that APQ and legacy safelisting are **not** the same feature. And we should just replace any references to the legacy safelisting with the new "persisted queries" only.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
BobaFetters: Shouldn't we still deprecate this because we want users to move from using this `operationIdentifiersPath` to using the new `operationManifestConfiguration`?

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+    public let path: String
+    /// The version format to use when generating the operation manifest. Defaults to `.persistedQueries`.
+    public let version: Version
+
+    public enum Version: String, Codable, Equatable {
+      /// Generates an operation manifest for use with persisted queries.
+      case persistedQueries
+      /// Generates an operation manifest for pre-registering operations with the legacy
+      /// [Automatic Persisted Queries (APQs)](https://www.apollographql.com/docs/apollo-server/performance/apq).
+      /// functionality of Apollo Server/Router.
```
AnthonyMDev: Need to update this documentation, as we discussed. (Re: legacy safelisting !== APQs)

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+      case legacy
+    }
+    
+    /// If set to `true` will generate the operation manfiest every time code generation is run. Defaults to `false`
```
AnthonyMDev, line 22:
```suggestion
    /// If set to `true` will generate the operation manifest every time code generation is run. Defaults to `false`
```

Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration+OperationManifestConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+      public static let generateManifestOnCodeGeneration: Bool = false
+    }
+    
+    // MARK: - Initializers
+    
+    /// Designated initializer
+    ///
+    /// - Parameters:
+    ///   - operationManifes: The `OperationManifestFileOutput` used to determine where and how to output the operation manifest JSON
+    ///   - operationDocumentFormat: The `OperationDocumentFormat` used to determine how to output operations in the generated code files.
```
AnthonyMDev, line 37: Documentation is way out of date here.

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
AnthonyMDev: Yes, but the deprecation message points to an initializer name that has changed. `operationManifest` is no longer a param there.

BobaFetters: Ah yea good catch

Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
```diff
@@ -1002,6 +959,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     case options
     case experimentalFeatures
     case schemaDownloadConfiguration
+    case operationManifestConfiguration
```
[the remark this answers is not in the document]
BobaFetters: I'm open to doing that, I named it `operationManifestConfiguration` to keep with the existing naming format from `schemaDownloadConfiguration` and since the struct is called `OperationManifestConfiguration`

Sources/CodegenCLI/Commands/Generate.swift
```diff
@@ -79,8 +66,13 @@ public struct Generate: ParsableCommand {
         schemaDownloadProvider: schemaDownloadProvider
       )
     }
+    let itemsToGenerate: ApolloCodegen.ItemsToGenerate = (configuration.operationManifestConfiguration?.generateManifestOnCodeGeneration ?? false) ? [.code, .operationManifest] : [.code]
```
AnthonyMDev, line 69:
```suggestion
    var itemsToGenerate: ApolloCodegen.ItemsToGenerate = .code
    
    if configuration.operationManifestConfiguration?.generateManifestOnCodeGeneration {
      itemsToGenerate += .operationManifest
    }
```

Sources/CodegenCLI/Commands/GenerateOperationManifest.swift
```diff
@@ -26,36 +26,39 @@ public struct GenerateOperationManifest: ParsableCommand {
...
   }
+
+  // MARK: - Validation
+
+  func validate(configuration: ApolloCodegenConfiguration) throws {
+    try checkForCLIVersionMismatch(with: inputs)
+
+    guard configuration.operationManifestConfiguration != nil else {
+      throw ValidationError("""
+          `operationManifestConfiguration` section must be set in the codegen configuration JSON in order
```
AnthonyMDev, line 58: If we are changing this JSON key, then:

```suggestion
          `operationManifest` section must be set in the codegen configuration JSON in order
```

Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationManifestFileGeneratorTests.swift
```diff
@@ -269,7 +275,7 @@ class OperationManifestFileGeneratorTests: XCTestCase {
 
   func test__template__givenOperationManifestVersion_apqLegacy__isLegacyAPQTemplate() throws {
```
AnthonyMDev, line 276:
```suggestion
  func test__template__givenOperationManifestVersion_legacy__isLegacyTemplate() throws {
```

docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -472,7 +435,7 @@ The top-level properties are:
 			"MyFragment"
 		]
 	},
-	"operationDocumentFormat" : ["definition", "operationId"],
+    "operationDocumentFormat" : ["definition", "operationId"],
```
AnthonyMDev, line 438: I think the indentation got messed up here.

docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -667,3 +630,72 @@ let configuration = ApolloCodegenConfiguration(
 </MultiCodeBlock>
 
 For more details, see the section on [downloading a schema](./downloading-schema).
+
+## Operation Manifest Configuration
+
+Optional settings used to configure the usage of [Persisted Queries or Automatic Persisted Queries](./../fetching/persisted-queries) in order to generate the operation identfier manifest and determine how operation files are generated.
```
AnthonyMDev, line 636:
```suggestion
Optional settings used to configure generation of the operation identifier manifest for use with [Persisted Queries](./../fetching/persisted-queries).
```

docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -667,3 +630,72 @@ let configuration = ApolloCodegenConfiguration(
...
+    output: ApolloCodegenConfiguration.FileOutput(
+        operationManifest: .init(
+            path: "./generated/operationIdentifiers.json",
+            version: .persistedQueries
+        )
+    )
+)
+```
+
+</MultiCodeBlock>
```
AnthonyMDev, line 701: I think this section needs to be deleted? It's an older version of whats directly above it.

AnthonyMDev (approved):
One minor typo, but otherwise good! Let's ship this!

PR #3176, opened 2023-08-05
refactor: Selection set generation for `@defer`

### LINKED ISSUE #3141: Modify selection set template to generate deferred fragments
This work is the implementation of the [generated models](https://github.com/apollographql/apollo-ios/blob/feature/defer/Design/3093-graphql-defer.md#generated-models) section of the RFC.

This work is linked to #3142 but can be done separately.

calvincestari:
Closes #3141

Introduces `IsDeferred` expression logic which is used in the selection set template to:
* Add Operation-level `hasDeferredFragments` property
* Add `deferred` property to inline and named fragment selections
* Mark deferred fragment accessors as Optional

_If there are other tests you think would be beneficial, I'm sure there are, please let me know and get them added._

### DIFF
--- a/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
+++ b/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
@@ -2,6 +2,7 @@ import JavaScriptCore
 
 /// The output of the frontend compiler.
 public class CompilationResult: JavaScriptObject {
+  /// String constants used to match JavaScriptObject instances.
   private enum Constants {
     enum DirectiveNames {
       static let LocalCacheMutation = "apollo_client_ios_localCacheMutation"
--- a/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
+++ b/Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
@@ -3,7 +3,7 @@ import JavaScriptCore
 /// The output of the frontend compiler.
 public class CompilationResult: JavaScriptObject {
   /// String constants used to match JavaScriptObject instances.
-  private enum Constants {
+  fileprivate enum Constants {
     enum DirectiveNames {
       static let LocalCacheMutation = "apollo_client_ios_localCacheMutation"
       static let Defer = "defer"
@@ -174,33 +174,14 @@ public class CompilationResult: JavaScriptObject {
     }
   }
 
-  public class InlineFragment: JavaScriptObject, Hashable {
+  public class InlineFragment: JavaScriptObject, Hashable, Deferrable {
     lazy var selectionSet: SelectionSet = self["selectionSet"]
 
     lazy var inclusionConditions: [InclusionCondition]? = self["inclusionConditions"]
 
     lazy var directives: [Directive]? = self["directives"]
 
-    lazy var isDeferred: IsDeferred = {
-      guard let directive = directives?.first(where: { $0.name == Constants.DirectiveNames.Defer }) else {
-        return false
-      }
-
-      guard let argument = directive.arguments?.first(where: { $0.name == Constants.ArgumentLabels.If }) else {
-        return true
-      }
-
-      switch (argument.value) {
-      case let .boolean(value):
-        return .value(value)
-
-      case let .variable(variable):
-        return .variable(variable)
-
-      default:
-        preconditionFailure("Incompatible argument value. Expected Boolean or Variable, got \(argument.value).")
-      }
-    }()
+    lazy var isDeferred: IsDeferred = getIsDeferred()
 
     public override var debugDescription: String {
       selectionSet.debugDescription
@@ -219,33 +200,14 @@ public class CompilationResult: JavaScriptObject {
 
   /// Represents an individual selection that includes a named fragment in a selection set.
   /// (ie. `...FragmentName`)
-  public class FragmentSpread: JavaScriptObject, Hashable {
+  public class FragmentSpread: JavaScriptObject, Hashable, Deferrable {
     lazy var fragment: FragmentDefinition = self["fragment"]
 
     lazy var inclusionConditions: [InclusionCondition]? = self["inclusionConditions"]
 
     lazy var directives: [Directive]? = self["directives"]
 
-    lazy var isDeferred: IsDeferred = {
-      guard let directive = directives?.first(where: { $0.name == Constants.DirectiveNames.Defer }) else {
-        return false
-      }
-
-      guard let argument = directive.arguments?.first(where: { $0.name == Constants.ArgumentLabels.If }) else {
-        return true
-      }
-
-      switch (argument.value) {
-      case let .boolean(value):
-        return .value(value)
-
-      case let .variable(variable):
-        return .variable(variable)
-
-      default:
-        preconditionFailure("Incompatible argument value. Expected Boolean or Variable, got \(argument.value).")
-      }
-    }()
+    lazy var isDeferred: IsDeferred = getIsDeferred()
 
     var parentType: GraphQLCompositeType { fragment.type }
 
@@ -475,4 +437,35 @@ public class CompilationResult: JavaScriptObject {
     }
   }
 
+}
+
+fileprivate protocol Deferrable {
+  var directives: [CompilationResult.Directive]? { get }
+}
+
+fileprivate extension Deferrable where Self: JavaScriptObject {
+  func getIsDeferred() -> CompilationResult.IsDeferred {
+    guard let directive = directives?.first(
+      where: { $0.name == CompilationResult.Constants.DirectiveNames.Defer }
+    ) else {
+      return false
+    }
+
+    guard let argument = directive.arguments?.first(
+      where: { $0.name == CompilationResult.Constants.ArgumentLabels.If }
+    ) else {
+      return true
+    }
+
+    switch (argument.value) {
+    case let .boolean(value):
+      return .value(value)
+
+    case let .variable(variable):
+      return .variable(variable)
+
+    default:
+      preconditionFailure("Incompatible argument value. Expected Boolean or Variable, got \(argument.value).")
+    }
+  }
 }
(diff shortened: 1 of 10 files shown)


Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
```diff
@@ -173,6 +181,27 @@ public class CompilationResult: JavaScriptObject {
 
     lazy var directives: [Directive]? = self["directives"]
 
+    lazy var isDeferred: IsDeferred = {
```
calvincestari, line 184: This is duplicated for both `InlineFragment` and `FragmentSpread`, I'm not sure if there is a better way to do it; they're both just simple `JavaScriptObject` types.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
```diff
@@ -412,4 +462,17 @@ public class CompilationResult: JavaScriptObject {
     }
   }
 
+  public enum IsDeferred: ExpressibleByBooleanLiteral {
```
calvincestari, line 465: This type is basically the same as `IR.IsDeferred`. I like having them defined in both places even if they're similar. I'm unsure how I feel about sharing a type between the JS frontend (`CompilationResult`) and `IR`. I'm open to suggestions.

Sources/ApolloCodegenLib/IR/IR+InclusionConditions.swift
```diff
@@ -54,6 +54,21 @@ extension IR {
       self.conditions = [condition]
     }
 
+    /// A failable initializer to convert from an array of `CompilationResult.InclusionCondition`.
+    init?(_ conditions: [CompilationResult.InclusionCondition]?) {
```
calvincestari, line 58: @AnthonyMDev - I've added this since we determined that `ScopeCondition` should match the inclusion conditions as well as the parent type. Problem is that the named fragment passed into the scope condition is from `CompilationResult` and the two `InclusionCondition` types can't be compared. This is a simple conversion similar to how we treat conversions for the `IR.InclusionCondition.Result` type.

calvincestari: I wonder if we could instead have a custom `==` operator that will compare `CompilationResult.InclusionConditions` to `IR.InclusionConditions`. That might be a better solution than an initializer that should never be used?

Sources/ApolloCodegenLib/IR/ScopeDescriptor.swift
```diff
@@ -3,9 +3,6 @@ import OrderedCollections
 
 extension IR {
 
-  // TODO: Write tests that two inline fragments with same type and inclusion conditions,
```
calvincestari, line 6: Done in https://github.com/apollographql/apollo-ios/pull/3176/files#diff-75a0a416f524d6139314ef8b30651767a446ba7f6a1dbc0e00a4c64f130251d1R2232.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
```diff
@@ -50,4 +50,40 @@ extension OperationTemplateRenderer {
       """
   }
 
+  func DeferredProperties(
+    _ definition: IR.Definition
+  ) -> TemplateString {
+    func isDeferred(_ selectionSet: IR.SelectionSet) -> Bool {
```
calvincestari, line 56: @AnthonyMDev - needing your feedback here whether this is the best way to be determining at an operation-level whether any fragments are deferred; I think it should be since we're only doing direct selections.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
AnthonyMDev:
```swift
protocol Deferrable {
  var directives: [Directive]? { get }
}

fileprivate extension Deferrable where Self: JavaScriptObject {
  func getIsDeferred() -> IsDeferred {
    ...
  }
}

public class InlineFragment: Deferrable {
  ...
  lazy var isDeferred: IsDeferred = getIsDeferred()
}
```

Wish we could easily namespace this `Deferrable` protocol in `CompilationResult`. Until we break `CompilationResult` into it's own package, we might want to call the protocol `CompilationResult_Deferrable`

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
AnthonyMDev: Yeah, let's leave these separate for now. I don't know if this one needs to be `ExpressibleByBooleanLiteral`, or have the `static func if` though. We're never initializing them directly via literals, are we?

Sources/ApolloCodegenLib/IR/IR+InclusionConditions.swift
AnthonyMDev: The problem with including this initializer is that it doesn't have an accurate 1:1 way to map. For `@skip` with no variable, we would be omitting the entire fragment. The `preconditionFailure` here seems way too easy to accidentally make occur and cause a crash.

I think that a custom matching function that is `fileprivate` to the file that is using it is a better approach. It's not an actual `==`, so I would make it a function with a more descriptive name.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
calvincestari: The protocol and extension can have the `fileprivate` scope which is effectively the same as putting it into it's own package for this particular case.

Sources/ApolloCodegenLib/Frontend/CompilationResult.swift
calvincestari: Both the `ExpressibleByBooleanLiteral` and the `.if` shortcuts are used in the evaluation function as conveniences. I don't think it's all that bad to include them, they sometimes make for easier reading.

Sources/ApolloCodegenLib/IR/IR+RootFieldBuilder.swift
```diff
@@ -230,8 +230,29 @@ extension IR {
...
+      let scopedIsDeferred = matchedParentTypes && matchedInclusionConditions
+      ? namedFragment.isDeferred : false
+
+      return scopeCondition(for: namedFragment, in: parentTypePath, isDeferred: scopedIsDeferred)
+    }
+
+    private func scopeCondition(
+      for conditionalSelectionSet: ConditionallyIncludable,
+      in parentTypePath: SelectionSet.TypeInfo,
+      isDeferred: CompilationResult.IsDeferred
```
AnthonyMDev, line 255: We should change the type of this parameter to `IR.IsDeferred` and do ``IsDeferred(compilationResult: namedFragment.isDeferred)` on line 247.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -1969,6 +1969,806 @@ class SelectionSetTemplateTests: XCTestCase {
...
+      }
+    }
+    """
+
+    let expected = """
+      public static var __selections: [ApolloAPI.Selection] { [
+        .field("__typename", String.self),
+        .include(if: "filter", [
+          .inlineFragment(AsDogIfFilter.self, deferred: true),
+          .inlineFragment(AsDogIfFilter.self, deferred: .if("other")),
```
AnthonyMDev, line 2268: This looks like it's going to generate two separate selection sets both named `AsDogIfFilter`, which is not going to compile. We're going to need to handle this more carefully.

Sources/ApolloCodegenLib/IR/ScopeDescriptor.swift
AnthonyMDev: I'm thinking we should also be adding tests for this behavior to the IR building tests, not just the `SelectionSetTemplateTests`. Making sure the actual IR is structured properly is important and having those tests makes it a LOT easier to debug and figure out the cause of bugs than just the template tests.

Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
```diff
@@ -50,4 +50,40 @@ extension OperationTemplateRenderer {
       """
   }
 
+  func DeferredProperties(
```
AnthonyMDev, line 53: The naming `DeferredProperties` confused me here. Can we just rename this to `HasDeferredFragmentsTemplate`

Sources/ApolloCodegenLib/Templates/RenderingHelpers/OperationTemplateRenderer.swift
AnthonyMDev: I think this code should be valid, but I'm wondering if we could prevent another iteration through the entire selection set here.

What if we added `hasDeferredFragments` to the `RootFieldBuilder.Result`. We could keep track of this as a `var` on the `RootFieldBuilder`, and while building, if we encounter a deferred fragment, we just set that to `true`.

Sources/ApolloCodegenLib/IR/IR+RootFieldBuilder.swift
```diff
@@ -241,7 +262,11 @@ extension IR {
       let type = parentTypePath.parentType == conditionalSelectionSet.parentType ?
       nil : conditionalSelectionSet.parentType
 
-      return ScopeCondition(type: type, conditions: inclusionResult.conditions)
+      return ScopeCondition(
+        type: type,
+        conditions: inclusionResult.conditions,
+        isDeferred: IsDeferred(compilationResult: isDeferred)
+      )
```
AnthonyMDev, line 269: These changes also should really be tested in tests on the `RootFieldBuilder`, not just the `SelectionSetTemplateTests`.

PR #3180, opened 2023-08-10
refactor: Update `CompilationResult` for deferred

calvincestari:
This incorporates the changes and suggestions from review of #3176.


calvincestari:
@AnthonyMDev, these are the isolated changes from #3176 for `CompilationResult`. This diff should still be valid despite the RFC changes we went through today.

AnthonyMDev (approved):

Issue #3187, opened 2023-08-10
`@defer` label validation


calvincestari:
We will require `@defer` directives to have an accompanying `label:` argument. This is needed to provide naming of the generated fragment and for identifying incremental response data.

calvincestari:
Done with the merging of https://github.com/apollographql/apollo-ios-dev/pull/28.

PR #3188, opened 2023-08-10
feature: `@defer` generated models

calvincestari:
This sets in place some of the supporting infrastructure needed to complete #3141 and #3187. There are a bunch of tests that are forced to fail and will be built to pass as the generated models are adapted to match the updated RFC edits in #3181.


calvincestari:
Closing this since moving to work in the [`apollo-ios-dev`](https://github.com/apollographql/apollo-ios-dev) repo.

PR #3189, opened 2023-08-12
feature: final class definitions for graphql codegen

### LINKED ISSUE #3183: Why were GraphQLOperations changed from final to non-final?
### Question

During our migration from 0.36 to 1.3, we noticed that some of our code no longer compiles due to the `final` modifier being dropped from the generated GraphQL operation classes.

```swift
protocol SupportedOfflineMutation {
  init?(variables: GraphQLOperation.Variables)
}

extension UpdateBookingPrimaryGuestMutation: SupportedOfflineMutation {
  public var bookingID: String { self.inputParams.id }
  public var requestType: OfflineRequestType { .standard(self) }
  public var variables: GraphQLOperation.Variables? { self.__variables }

  public convenience init?(variables: GraphQLOperation.Variables) {
     ^^^^^^^^^^^^^^^
        Initializer requirement 'init(variables:)' can only be satisfied by a 'required' initializer in the definition of non-final class 'UpdateBookingPrimaryGuestMutation'
  }
```

Was this intentional, or a miss in parity during the re-write of the engine?

Would it be reasonable to at least offer this as a codegen option?

Mordil:
Re-adds generated GraphQL operations to be marked as `final` through a new `OutputOptions` property: `markClassesAsFinal`.

This is to support fixing this in a backwards-compatible way, and still allowing people control over how they use the generated operation models.

This resolves #3183

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegenConfiguration.swift
@@ -503,6 +503,8 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///
     ///  Defaults to `true`.
     public let pruneGeneratedFiles: Bool
+    /// Whether generated GraphQL operation and local cache mutation class types will be marked as `final`.
+    public let markOperationDefinitionsAsFinal: Bool
 
     /// Default property values
     public struct Default {
@@ -515,6 +517,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       public static let warningsOnDeprecatedUsage: Composition = .include
       public static let conversionStrategies: ConversionStrategies = .init()
       public static let pruneGeneratedFiles: Bool = true
+      public static let markOperationDefinitionsAsFinal: Bool = false
     }
 
     /// Designated initializer.
@@ -535,6 +538,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///   - conversionStrategies: Rules for how to convert the names of values from the schema in
     ///     generated code.
     ///   - pruneGeneratedFiles: Whether unused generated files will be automatically deleted.
+    ///   - markOperationDefinitionsAsFinal: Whether generated GraphQL operation and local cache mutation class types will be marked as `final`.
     public init(
       additionalInflectionRules: [InflectionRule] = Default.additionalInflectionRules,
       deprecatedEnumCases: Composition = Default.deprecatedEnumCases,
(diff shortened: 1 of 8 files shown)


AnthonyMDev (changes requested):
Thanks so much for working on this! Implementation looks spot on to me! Just a couple things to consider.

- We need to add the new option to the documentation of the CLI

- I think we should also add the `final` keyword to local cache mutations if this option is `true`. That would be done in the `LocalCacheMutationDefinitionTemplate`.

- @BobaFetters what do you think of the naming of this configuration option? I'm not coming up with a better name off the top of my head right now, but I feel like it's not clear what `finalOperationDefinitions` means without having to read the docs.

Mordil:
I just came up with `markClassesAsFinal`, but now I'm wondering if that name now pushes is so we should mark `Mock` definitions as `final` as well?

It'd probably be easier to make it so that it applies to _all_ classes to start with, and if people say they want more granular control, later the options are split into further sub-options

Mordil:
> Overall I think this looks great, thanks for taking this on!
> 
> As for naming, since the `final` keyword is a modifier thoughts on naming the option something like `applyFinalModifier`?

I don't have a strong preference. It's what y'all think is best.

BobaFetters:
> > Overall I think this looks great, thanks for taking this on!
> > As for naming, since the `final` keyword is a modifier thoughts on naming the option something like `applyFinalModifier`?
> 
> I don't have a strong preference. It's what y'all think is best.

@AnthonyMDev thoughts?

calvincestari:
> I just came up with `markClassesAsFinal`, but now I'm wondering if that name now pushes is so we should mark `Mock` definitions as `final` as well?

I'm less inclined to want `final` mock classes. I think the fact that they're intended to be test classes should allow for more complicated, do-as-you-want code.

> It'd probably be easier to make it so that it applies to _all_ classes to start with, and if people say they want more granular control, later the options are split into further sub-options

Yes I agree.

I'm happy with the config option named `markClassesAsFinal`, I think it's descriptive enough.

Mordil:
> I felt like `finalOperationDefinitions` wasn't completely clear about what `final` meant in the context. But now `markClassesAsFinal` makes me confused about what `Classes` means. I don't think people using the codegen really know exactly what it is we are generating as classes vs structs.
> 
> What about `markOperationDefinitionsAsFinal`? I think that covers all our bases.
> 
> And I agree with @calvincestari that mocks should not be `final`.

Again, I have zero preferences.

I'll update the PR with this new proposed name, at the same time that I'll squash the commits and update with the latest version of the dev branch.

Appreciate the feedback!

AnthonyMDev (approved):

AnthonyMDev:
Thanks so much for this @Mordil! This looks great! Merging this in now, we'll get it out in the next release!

PR #3198, opened 2023-08-17
Adds RequestContext to the RequestChain/Interceptor pipeline

danieltiger:
There is sometimes a need to do things in an interceptor that relies on information not contained in the operation that is passed in. An example of this might be wanting to do custom headers on a per-request or pre-request-type basis. 

As a solution, this PR adds a new marker protocol called RequestContext, that can be used to pass anything that a consumer might want/need through the request chain, making it available in any custom interceptor.

This was intentionally left as only a marker protocol, not any kind of specific context implementation. That's to allow the maximum flexibility for users of the chain, and to not try to solve for the probably endless needs they might have.

### DIFF
--- a/Sources/Apollo/ApolloClientProtocol.swift
+++ b/Sources/Apollo/ApolloClientProtocol.swift
@@ -29,6 +29,7 @@ public protocol ApolloClientProtocol: AnyObject {
   func fetch<Query: GraphQLQuery>(query: Query,
                                   cachePolicy: CachePolicy,
                                   contextIdentifier: UUID?,
+                                  context: RequestContext?,
                                   queue: DispatchQueue,
                                   resultHandler: GraphQLResultHandler<Query.Data>?) -> Cancellable
 
@@ -42,6 +43,7 @@ public protocol ApolloClientProtocol: AnyObject {
   /// - Returns: A query watcher object that can be used to control the watching behavior.
   func watch<Query: GraphQLQuery>(query: Query,
                                   cachePolicy: CachePolicy,
+                                  context: RequestContext?,
                                   callbackQueue: DispatchQueue,
                                   resultHandler: @escaping GraphQLResultHandler<Query.Data>) -> GraphQLQueryWatcher<Query>
 
@@ -55,6 +57,7 @@ public protocol ApolloClientProtocol: AnyObject {
   /// - Returns: An object that can be used to cancel an in progress mutation.
   func perform<Mutation: GraphQLMutation>(mutation: Mutation,
                                           publishResultToStore: Bool,
+                                          context: RequestContext?,
                                           queue: DispatchQueue,
                                           resultHandler: GraphQLResultHandler<Mutation.Data>?) -> Cancellable
 
@@ -68,6 +71,7 @@ public protocol ApolloClientProtocol: AnyObject {
   /// - Returns: An object that can be used to cancel an in progress request.
   func upload<Operation: GraphQLOperation>(operation: Operation,
                                            files: [GraphQLFile],
+                                           context: RequestContext?,
                                            queue: DispatchQueue,
                                            resultHandler: GraphQLResultHandler<Operation.Data>?) -> Cancellable
 
@@ -80,6 +84,7 @@ public protocol ApolloClientProtocol: AnyObject {
   ///   - resultHandler: An optional closure that is called when mutation results are available or when an error occurs.
   /// - Returns: An object that can be used to cancel an in progress subscription.
   func subscribe<Subscription: GraphQLSubscription>(subscription: Subscription,
+                                                    context: RequestContext?,
                                                     queue: DispatchQueue,
                                                     resultHandler: @escaping GraphQLResultHandler<Subscription.Data>) -> Cancellable
 }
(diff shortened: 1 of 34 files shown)


Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -24,11 +24,13 @@ public protocol ApolloClientProtocol: AnyObject {
   ///   - cachePolicy: A cache policy that specifies when results should be fetched from the server and when data should be loaded from the local cache.
   ///   - queue: A dispatch queue on which the result handler will be called. Should default to the main queue.
   ///   - contextIdentifier: [optional] A unique identifier for this request, to help with deduping cache hits for watchers. Should default to `nil`.
+  ///   - context: [optional] A context that is being passed through the request chain. Should default to `nil`.
   ///   - resultHandler: [optional] A closure that is called when query results are available or when an error occurs.
   /// - Returns: An object that can be used to cancel an in progress fetch.
   func fetch<Query: GraphQLQuery>(query: Query,
                                   cachePolicy: CachePolicy,
                                   contextIdentifier: UUID?,
+                                  context: RequestContext?,
```
AnthonyMDev, line 33: The only concern I have is that this is technically a minor breaking change. Anyone who is implementing the `ApolloClientProtocol` (which I would think is usually just for mock objects in unit tests), will have to add the new parameter. It's not a huge change, but it will not be backwards compatible. 

I'm not really sure that there is a great method for deprecation of the old signatures though. So I think this falls under acceptable minor breaking changes.

AnthonyMDev (approved):
Thanks so much for the contribution! I don't have a problem with this approach at all, though I would like to get input from @calvincestari on this before merging in. He works on the networking layer a lot more than I do. He is out until the 27th, so I'll get his input when he's back.

Sources/Apollo/ApolloClientProtocol.swift
danieltiger: Yeah, I couldn't think of a way to do it without creating a breaking change. Glad to hear you're likely ok with it!

danieltiger:
Added the tests that were requested :)

Tests/ApolloTests/RequestContextTests.swift
```diff
@@ -0,0 +1,82 @@
+import XCTest
```
AnthonyMDev, line 1: These tests are awesome! Nice work.

calvincestari:
What about if we make the request context part of `HTTPRequest` instead? Interceptors already have access to that through the `request` property that is passed along in all the `ApolloInterceptor` functions. That would bind the context to the actual request instead of having to link them through the additional `ApolloInterceptor` function parameter.

My original thought was that two requests with different contexts should be recognized as different requests, i.e.: not equatable. _I don't know yet where that comparison would be useful but I feel it's still technically correct._

I'm unable to create a branch on your fork so I've put the changes into a [patch file](https://github.com/apollographql/apollo-ios/files/12470766/0001-Moved-request-context-to-HTTPRequest.patch) which you can apply and see the difference in implementation. I think it creates less of a breaking change too.

calvincestari:
> My original thought was that two requests with different contexts should be recognized as different requests, i.e.: not equatable. I don't know yet where that comparison would be useful but I feel it's still technically correct.

Unfortunately my original goal of the difference in equality is still not possible due to `RequestContext` being a protocol whereas a concrete type is needed for conformance to `Hashable`.

AnthonyMDev:
I actually really like that idea @calvincestari! The `RequestContext` would be a parameter in the `ApolloClient` functions and the `NetworkTransport`, but once the `RequestChainNetworkTransport` constructs the request (in this case a `JSONRequest` or `UploadRequest` which are both subclasses of `HTTPRequest`, then the `RequestContext` would be passed into it, making it easier to access throughout the request chain without having to continue to pass it as a parameter.

This also feels like a logical place to store it and ensures it is retained until the request is complete and deallocated.

@danieltiger Do you mind making that change?

calvincestari:
> The RequestContext would be a parameter in the ApolloClient functions and the NetworkTransport, but once the RequestChainNetworkTransport constructs the request (in this case a JSONRequest or UploadRequest which are both subclasses of HTTPRequest, then the RequestContext would be passed into it, making it easier to access throughout the request chain without having to continue to pass it as a parameter.

Correct.

calvincestari:
> @danieltiger Do you mind making that change?

Simply applying the patch should give you exactly that.

danieltiger:
Sorry, I haven't had a chance to take a look yet, but I will tomorrow morning!

danieltiger:
Ok. PR is updated. This approach is great. I had briefly considered it when I started, but pivoted for reasons I no longer remember ;)

Anyway, I updated all the comments everywhere and made sure the context is being passed into the upload request as well. We should be good to go now!

calvincestari (approved):
Looks good, thanks for the contribution @danieltiger.

AnthonyMDev (approved):

AnthonyMDev:
Merged! Thanks so much @danieltiger!

We're going to get a minor release out to include this today.

danieltiger:
Amazing! Thanks so much for helping this get done!

petarbelokonski:
I have achieved the same by subclassing query classes. Then in the interceptor I'd downcast to check whether I need what I am after. Since we are forced to use downcasting the RequestContext as well I wonder if this approach has more merits ? 

Also worth mentioning that I am using query subclasses to differentiate operations in the `InterceptorProvider` level  as described here https://github.com/apollographql/apollo-ios/issues/3213. We have no way to use the RequestContext in the `InterceptorProvider` as of now.

PR #3217, opened 2023-09-12
chore: Changed to use isEmpty instead of count == 0

tokizuoh:
I ran grep with `count == 0` on the whole code and got only one hit.
In this case, the amount of calculation is O(1) with or without `isEmpty`, but since isEmpty is used throughout, would you like to replace?

- https://developer.apple.com/documentation/swift/array/isempty
- https://developer.apple.com/documentation/swift/collection/count-4l4qk
  - Array conforms to RandomAccessCollection, so O(1).


calvincestari:
Thanks @tokizuoh. This particular code is vendor code for StarScream version 3 that we pulled into the codebase instead of linking out to the package as a dependency. Long-story-short is that is why it's the only place you find `count == 0` vs `isEmpty`.

I don't think we'll get a large performance improvement from this change but given that it is used each time an input message is processed it can still make a difference. Thanks for the contribution.

calvincestari (approved):

PR #3218, opened 2023-09-13
Remove: Error inheritance

tokizuoh:
Since LocalizedError inherits from Error, there is no need to inherit both Error and LocalizedError. 

https://developer.apple.com/documentation/foundation/localizederror > Inherits From Error


tokizuoh:
Sorry, I noticed you have a custom Error defined.

https://github.com/apollographql/apollo-ios/blob/main/Sources/CodegenCLI/Error.swift

PR #3219, opened 2023-09-13
Provide a way to pass an input operation manifest file

jimisaacs:
TL;DR - This provides a preferred way to provide input operation IDs.

Details: We need a way to input our own operation ID during the codegen process, and this supports the following workflow:
1. Run Apollo Codegen outputting only the operationManifest.json file (with itemsToGenerate)
2. Process the previously operationManifest.json file, (using the `body` field) regenerating all the `id` fields, and resave the operationManifest.json file
3. Run Apollo Codegen **again**, but this time feeding the operationManifest.json file as input, into the source generation process

Ideally we wouldn't need to run Apollo Codegen twice for this workflow. Though I'm struggling to find a reasonable way to do it otherwise. Thoughts are appreciated.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -94,6 +94,28 @@ public class ApolloCodegen {
     
   }
 
+  /// This is exposed for external use cases that need to read an operation manifest file and decode it.
+  public struct OperationManifestV1: Equatable, Codable {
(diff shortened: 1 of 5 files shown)


AnthonyMDev:
Thanks for providing this as an example. As per conversations offline, we are going to be going ahead with a solution that is purely in Swift code. This means it will only work when using `ApolloCodegenLib` directly, but avoids all of these shenanigans that need to happen for the CLI to support this.

PR #3221, opened 2023-09-15
allow the injection of a NetworkSession in the schema download fetch

jimisaacs:
The title is pretty self explanatory, but I'll go into a little bit about our and use case can reasoning.

### Use Case

We require an enterprise level of security / authentitication, that is just easier to inject at the NetworkSession level than to try another workaround.

### Reasoning

aka, why this way?

I simply thought this was the least intrusive set of changes to achieve the use case. `NetworkSession` already existed, and was injectable in the internal `URLDownloader` initializer. I just passed it through all the way from the public static `fetch` method.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
+++ b/Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
@@ -46,7 +46,7 @@ public struct ApolloSchemaDownloader {
   ///   - configuration: The `ApolloSchemaDownloadConfiguration` used to download the schema.
   ///   - rootURL: The root `URL` to resolve relative `URL`s in the configuration's paths against.
   ///     If `nil`, the current working directory of the executing process will be used.
-  ///   - session: The network session to use for the download
+  ///   - session: The network session to use for the download. If `nil` the `URLSession.Shared` will be used by default.
   /// - Returns: Output from a successful fetch or throws an error.
   /// - Throws: Any error which occurs during the fetch.
   public static func fetch(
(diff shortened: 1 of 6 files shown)


Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
```diff
@@ -46,11 +46,13 @@ public struct ApolloSchemaDownloader {
   ///   - configuration: The `ApolloSchemaDownloadConfiguration` used to download the schema.
   ///   - rootURL: The root `URL` to resolve relative `URL`s in the configuration's paths against.
   ///     If `nil`, the current working directory of the executing process will be used.
+  ///   - session: The network session to use for the download
```
AnthonyMDev, line 49:
```suggestion
  ///   - session: The network session to use for the download. If `nil` the `URLSession.Shared` will be used by default.
```

AnthonyMDev (approved):
Thanks for this jimi!

AnthonyMDev:
It looks like the `SchemaDownloadProvider` also needs to be updated to make this build properly and  pass CI. Can you make those changes and I'll merge it in!

jimisaacs:
> It looks like the `SchemaDownloadProvider` also needs to be updated to make this build properly and pass CI. Can you make those changes and I'll merge it in!

Thanks for the catch! Should be fixed 🤞

jimisaacs:
@AnthonyMDev just a friendly nudge on this to see if your refactor might also touch this

AnthonyMDev:
Yeah, this code looks good, but I'm going to hold off and rebase it on top of my changes before merging it in!

jimisaacs:
Closed in favor of https://github.com/apollographql/apollo-ios-codegen/pull/1

Issue #3226, opened 2023-09-18
Why is there no use of async/await in ApolloCodegenLib?


jimisaacs:
### Question

I understand not wanting to use `async/await` in public client APIs yet, but I'm surprised it hasn't trickled in yet internally in ApolloCodegenLib. 

This question is basically around whether contributions I might make to ApolloCodegenLib using `async/await` would be rejected?

jimisaacs:
I think I might have answered my own question in that `Task` is only available in macOS 10.15, but it seems ApolloCodegenLib is already meant to support only that version. It might be cumbersome to use `if #available(macOS 10.15, *)` everywhere, but I'm thinking, would if the entire process was wrapped in a `Task` and forced to be synchronous right at the build function.

calvincestari:
> Why is there no use of async/await in ApolloCodegenLib?

It's on the roadmap for 2.0 which we simply haven't gotten to yet.

> This question is basically around whether contributions I might make to ApolloCodegenLib using async/await would be rejected?

Unfortunately we won't be including/accepting contributions for Swift Concurrency in any of the 1.x versions. Our stance is that anything like that would simply be band-aid whereas 2.0 is a complete rethink of the networking architecture/stack to have first-class support for modern networking standards.

jimisaacs:
@calvincestari Did you see that this issue is about ApolloCodegenLib? Because what you are mentioning I though was all referring to client code.

calvincestari:
No I didn't notice that I was too quick to answer, apologies. I see Anthony got you the right answer in the PR.

PR #3227, opened 2023-09-18
support async/await in ApolloCodegenLib

jimisaacs:
Related to this question: https://github.com/apollographql/apollo-ios/issues/3226

This seemed like a good solution to start to using async/await everywhere below ApolloCodegen.build

### DIFF
--- a/Apollo.xcodeproj/project.pbxproj
+++ b/Apollo.xcodeproj/project.pbxproj
@@ -6746,7 +6746,7 @@
 			repositoryURL = "https://github.com/Quick/Nimble.git";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 10.0.0;
+				minimumVersion = 12.3.0;
 			};
 		};
 		DEAFB784270652AF00BE02F3 /* XCRemoteSwiftPackageReference "swift-collections" */ = {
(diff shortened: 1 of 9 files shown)


jimisaacs:
The latest commit pushed this to a rather larger change, but I still like the changes.

- AsyncParsableCommand makes using async/await in a CLI easy
- XCTest already has great async/await support
- Though upgraded Nimble to 12, which needed its own support for async/await

jimisaacs:
Hmm, rather confused that there are codegen tests run on macOS 13.2, when there is clearly a macOS 15 check in IR.Operation with a fatalError. @BobaFetters @AnthonyMDev, could you provide some insight?

jimisaacs:
Yeah, it might be wait to cumbersome until the build official changes 10.15 as the minimum.

calvincestari:
> Hmm, rather confused that there are codegen tests run on macOS 13.2, when there is clearly a macOS 15 check in IR.Operation with a fatalError. @BobaFetters @AnthonyMDev, could you provide some insight?

I think this is because ApolloCodegenLib requires macOS 15 whereas the other modules support much lower versions of macOS and corresponding APIs in iOS.

AnthonyMDev:
@jimisaacs, thanks for starting this PR. I'm working on a refactoring of a lot of the codegen code right now. When it's finished, I expect that we will raise the minimum deployment target to macOS 15. I also hope to add parallelization to the codegen lib at that time. This seems like a good start, but there are some complications around race conditions when building fragment dependencies that need a bit more work than this.

I'm not going to be accepting a PR for async/await additions at the moment, but once I've got this refactoring work done, I'll be looking into this further.

PR #3228, opened 2023-09-18
loop through operation definitions once

jimisaacs:
### Background (the why)
I'm trying to implement a *clean'ish* way to to inject a closure to compute operation identifiers, while providing the proper source of said operation to the injection closure, and while doing so with as much parallelization as possible, i.e. I do not want to make the computation closure synchronous.

### The what
So... all this being said, it was getting rather cumbersome to always account for multiple places of looping through the operation definitions and generating an `IR.Operation`. So this makes a single loop more complex, but also hopefully simplifies by removing a loop, and ensuring all known files, including the manifest, are generated from the same generateFiles method.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -131,36 +131,28 @@ public class ApolloCodegen {
 
     let ir = IR(compilationResult: compilationResult)
 
-    if itemsToGenerate.contains(.code) {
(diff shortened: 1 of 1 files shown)


jimisaacs:
I don' think `testCancellingTaskThroughClientDoesNotCallCompletion` had anything to do with these changes.

calvincestari:
@AnthonyMDev will you pick this one up post-refactoring of codegen lib?

AnthonyMDev:
Yeah I'm looking into this!

PR #3243, opened 2023-09-28
Adding codegen deprecation message

BobaFetters:
Adding deprecation message to `ApolloCodegen` to alert users in advance of the upcoming 1.6.0 changes.

### DIFF
--- a/Sources/ApolloCodegenLib/ApolloCodegen.swift
+++ b/Sources/ApolloCodegenLib/ApolloCodegen.swift
@@ -5,7 +5,7 @@ import OrderedCollections
 #if os(macOS)
 
 /// A class to facilitate running code generation
-@available(*, deprecated, message: "Starting with the upcoming 1.6.0 release, the code generation library will be its own SPM package which you will need to add as a dependency. For more details see here: https://github.com/apollographql/apollo-ios/issues/3240")
+@available(*, deprecated, message: "Starting with the 1.6.0 release, the code generation library will be its own SPM package which you will need to add as a dependency. For more details see here: https://github.com/apollographql/apollo-ios/issues/3240")
 public class ApolloCodegen {
 
   // MARK: Public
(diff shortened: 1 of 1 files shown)


Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -5,6 +5,7 @@ import OrderedCollections
 #if os(macOS)
 
 /// A class to facilitate running code generation
+@available(*, deprecated, message: "Starting with the upcoming 1.6.0 release, the code generation library will be its own SPM package which you will need to add as a dependency. For more details see here: https://github.com/apollographql/apollo-ios/issues/3240")
```
calvincestari, line 8: We can probably remove the word "upcoming" given that we want to release these in tandem? So that folks who are pedantic about no warnings can do the change to `1.6.0` immediately.

calvincestari (approved):

Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: Or maybe not if the patch release will only be this deprecation. They they have no motivation to stay on it and can rollback to the previous (current) release prior to the warning.

bignimbus: If we do end up shipping a subsequent patch to `1.5.x` then maybe it would be prudent for the user to be able to hide this message with an environment variable

BobaFetters: I think the idea is we push this patch ahead of time? Maybe I misunderstood the purpose though

BobaFetters: @bignimbus @AnthonyMDev thoughts?

calvincestari: Ahead of `1.6.0` is fine but then the patch should _only_ contain the deprecation otherwise people who want the other changes but no warnings couldn't stay on the patch release. If there are no other changes then they can see the message, decide they don't want the build warning and go back to the previous version. Once `1.6.0` is out they can catch-up and go from there.

BobaFetters: I'm not aware of any changes we have planned to ship in a patch release before 1.6.0 at this point.

PR #3391, opened 2024-06-10
Fix building ApolloSQLite with Xcode 16

rastersize:
In Xcode 16, or rather Foundation in the iOS 18/etc. SDKs, there’s a new `Expression` type. The SQLite library also expose this type. This caused an issue compiling the `SQLiteDotSwiftDatabase.swift` file, as it imports both libraries. This PR disambiguates it for the compiler.

With this change the project once again builds with Xcode 16.

Screenshot 2024-06-10 at 3 59 48 PM


BrentMifsud:
any word on when this is going out in a release?

BobaFetters:
@BrentMifsud This will go out in a release this week

BobaFetters:
Quick update on this, we will be doing a release containing this on Monday.

PR #3400, opened 2024-06-30
[Legacy] Remove nil values from `variables` field in `requestBody`

yramocan:
### Summary

Fixes an issue in legacy versions of the Apollo SDK. In my testing, this only occurred in my project when working with Xcode 16 Beta. This issue did not occur with previous versions of Xcode.

**Apollo version**: 0.53.0 (legacy)

### Issue Description

I have a GQL query which takes in an optional value as part of the request body. During serialization, I'd hit a `fatalError` in the Apollo library which reads `Dictionary is only JSONEncodable if Value is (and if Key is String)` ([permalink](https://github.com/apollographql/apollo-ios/blob/5c231061afab1929073a3f4f2789498cfd6d41f5/Sources/Apollo/JSONStandardTypeConversions.swift#L133)). This occurred because the `variables` dictionary contained a key-value pair where the key was present and the value was `nil`. In this version of Xcode, casting an `Optional<Wrapped>` value to `Wrapped` does not succeed.

I don't believe this is an issue in v1.0+ versions of the library, however, we cannot yet upgrade the Apollo version due to [this issue](https://github.com/apollographql/apollo-ios/issues/2560).

### Proposed Solution

In this PR, I've changed the way that values in the `variables` dictionary are included in the `requestBody`. Now, nil values in `variables` are removed before serialization. Additionally, if all values in `variables` are `nil`, `variables` won't be included in the `requestBody`.

### DIFF
--- a/Sources/Apollo/RequestBodyCreator.swift
+++ b/Sources/Apollo/RequestBodyCreator.swift
@@ -27,9 +27,12 @@ extension RequestBodyCreator {
                                                        sendQueryDocument: Bool,
                                                        autoPersistQuery: Bool) -> GraphQLMap {
     var body: GraphQLMap = [
-      "variables": operation.variables,
       "operationName": operation.operationName,
     ]
+
+    if let variables = operation.variables?.compactMapValues({ $0 }) {
+      body["variables"] = variables
+    }
 
     if sendOperationIdentifiers {
       guard let operationIdentifier = operation.operationIdentifier else {
--- a/Tests/ApolloTests/RequestBodyCreatorTests.swift
+++ b/Tests/ApolloTests/RequestBodyCreatorTests.swift
@@ -37,4 +37,19 @@ class RequestBodyCreatorTests: XCTestCase {
 
     XCTAssertEqual(query.queryDocument, req["test_query"] as? String)
   }
+
+  func testRequestBodyOmitsVariablesFieldWhenValuesAreNil() {
+    let query = HeroNameQuery(episode: nil)
+    let req = self.create(with: apolloRequestBodyCreator, for: query)
+
+    XCTAssertFalse(req.keys.contains(where: { $0 == "variables" }))
+  }
+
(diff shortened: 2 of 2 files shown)


calvincestari:
Hi @yramocan - I think there might be unintended side-effects to this change, and if this works for your case that might not be the case for all server implementations.

I'm not 100% familiar with the legacy version's implementation of GraphQL nullability but I believe `nil` would have been [converted](https://github.com/apollographql/apollo-ios/blob/5c231061afab1929073a3f4f2789498cfd6d41f5/Sources/Apollo/JSONStandardTypeConversions.swift#L105) to `null` in the request sent to the server and there is a [semantic difference in GraphQL](https://spec.graphql.org/draft/#sec-Null-Value) between `nil` (not present) and `null`.

As you can see in the GraphQL spec example there are two ways to send an optional variable and your proposed change would lock everyone into the second way. Right now I don't think we can accept this change as it is, it's going to need more debugging to try understand what is causing the issue with Xcode 16.

yramocan:
> Hi @yramocan - I think there might be unintended side-effects to this change, and if this works for your case that might not be the case for all server implementations.
> 
> I'm not 100% familiar with the legacy version's implementation of GraphQL nullability but I believe `nil` would have been [converted](https://github.com/apollographql/apollo-ios/blob/5c231061afab1929073a3f4f2789498cfd6d41f5/Sources/Apollo/JSONStandardTypeConversions.swift#L105) to `null` in the request sent to the server and there is a [semantic difference in GraphQL](https://spec.graphql.org/draft/#sec-Null-Value) between `nil` (not present) and `null`.
> 
> As you can see in the GraphQL spec example there are two ways to send an optional variable and your proposed change would lock everyone into the second way. Right now I don't think we can accept this change as it is, it's going to need more debugging to try understand what is causing the issue with Xcode 16.

Sounds good. I'll close this PR and investigate further.

PR #3655, opened 2026-09-01
Report a peer-initiated close instead of swallowing it

reljai:
## Summary
`didCloseWith` discarded the close code and reason, reporting a peer-initiated close as a clean shutdown. Subscribers were never told their subscription had ended, so they kept believing they were subscribed while the transport silently rebuilt the socket beneath them.

## Why
Against a `graphql-transport-ws` server closing with 4408 (connection_init timeout), the client reconnected and the UI stayed live, but messages sent during the gap never arrived — the app had to be killed and reopened to see them.

Reconnecting restores the SOCKET, not the subscriptions on it. The transport replays the subscribe frames it first sent, so each carries the state from then and the server sends nothing for the gap. Only the subscriber can re-establish against its own current position, and it never learns it needs to.

## What changed
`cleanupSession` now receives a `WebSocketClosedByPeerError` carrying the code and reason, which routes `websocketDidDisconnect` down the error branch and reaches `notifyErrorAllHandlers`.

Our own disconnects are unaffected: `disconnect()` clears `state.session` before its close handshake, and `cleanupSession` ignores a session it has already replaced — so only a close for the CURRENT session reports.

## Two judgement calls worth reviewing
**Every peer close reports, including `.normalClosure`.** A close is not a failure of the socket layer, but it ends every subscription on it, and a subscriber handler is `(Result<JSONObject, any Error>) -> Void` — there is no way to signal termination other than `.failure`. Excluding graceful codes would drop the backfill exactly during a server deploy, when a whole fleet reconnects at once.

**Reconnection is unchanged.** `handleDisconnection(with:)` falls through to `handleDisconnection()`, so the socket comes back as before. The one delta: it sets `socketConnectionState = .failed` first, and a close arriving while already `.failed` takes the `break` and skips reconnection. `cleanupSession`'s session guard appears to make that unreachable — a second close for the same socket is filtered — but I have reasoned about it rather than exercised it, and it is the place to look if a socket ever stops reconnecting.

## Test plan
- `swift build --target ApolloWebSocket` passes.
- Not exercised against a live server. The consuming app (portolans/apple) has a matching change and is where the end-to-end behaviour shows up: a message sent while the socket is down should now arrive without a force-quit.


reljai:
Opened against the wrong remote by mistake — this targets our internal fork, not upstream. Closing.

